PR #11, opened 2023-07-13
Add Rollup config for UMD bundles.

alantreadway:
Adds `.umd.js` bundle variants, allowing direct use via a `<script>` tag in a browser.


taktran (approved):
Simple enough 😄

PR #16, opened 2023-07-14
Add Example Runner

taktran:
## Changes

* Add example runner from AG Grid
* Persist framework/internal framework selections to localstorage
* Add example plunkr endpoint for the `index.html` file required for plunkr (will probably change how this works in future)

## Known issues

* Example generation only works for JavaScript and Typescript
* Plunkr only works for JavaScript


packages/ag-charts-website/src/utils/pages.ts
```diff
@@ -49,6 +49,14 @@ export const DEV_FILE_PATH_MAP: Record<string, string> = {
     'ag-charts-enterprise/dist/ag-charts-enterprise.umd.js': 'packages/ag-charts-enterprise/dist/main.umd.cjs',
 };
 
+export const getChartScriptPath = () => {
+    return `https://testing.ag-grid.com/ag-charts/dev/ag-charts-community/dist/ag-charts-community.umd.js`;
```
alantreadway, line 53: These should probably not be hardcoded to the testing host?

packages/ag-charts-website/src/utils/pages.ts
```diff
@@ -49,6 +49,14 @@ export const DEV_FILE_PATH_MAP: Record<string, string> = {
...
+};
+
+export const getChartEnterpriseScriptPath = () => {
+    return `https://testing.ag-grid.com/ag-charts/dev/ag-charts-enterprise/dist/ag-charts-enterprise.umd.js`;
```
alantreadway, line 57: Ditto

packages/ag-charts-website/src/utils/pages.ts
alantreadway: Ok to follow-up on a later PR on this point as discussed offline.

packages/ag-charts-website/src/utils/pages.ts
alantreadway: Ok to follow-up on a later PR on this point as discussed offline.

alantreadway (approved):

PR #20, opened 2023-07-17
Add dynamic site url support for JavaScript example runner

taktran:
## Changes

* Remove hardcoded testing url for charts and enterprise charts urls and allow for a custom site configuration
* Use `SITE_URL` env var for determining URLs in different environments, and set in Astro config
  * Set in `.env` files for dev, staging and production
* Fix `pathJoin` to work for URL objects


packages/ag-charts-website/.env.build
```diff
@@ -0,0 +1,2 @@
+# Production site
+SITE_URL=https://charts.ag-grid.com
```
alantreadway: We should avoid pointing to production by default - could we point to the local development URL by default, then use `.env.production` for `--configuration production` to pick it up?

alantreadway (commented):
Looks good, other than I have a strong preference for not defaulting to production config.

packages/ag-charts-website/.env.build
taktran: Good point 👍🏼  - I'll make this change later

taktran: @alantreadway I've updated the default to be dev, and added the `production` configuration

I tried using `.staging.env` and `.env.staging` ([nx env docs](https://nx.dev/recipes/environment-variables/define-environment-variables#setting-environment-variables)) to replace `.env.build.staging` and `.env.preview.staging`, but it doesn't apply the env to the tasks eg, if I run `nx preview ag-charts-website -c staging`, it doesn't apply the `staging` env for `nx build` and `nx preview`. Might need some more investigation as to why that is, but having the expanded form of `.env.build.staging` and `.env.preview.staging` works.

PR #39, opened 2023-07-19
Add vitest for ag-charts-website

taktran:
* `jest` had issues with ESM modules which are used by astro/vite
* Fix snippetTransfomer.test.ts file

---

To run

```
nx test ag-charts-website
```


alantreadway (approved):
Nice, LGTM!

PR #40, opened 2023-07-20
More CI caching improvements

alantreadway:
Experimenting with parallelisation and caching.


alantreadway:
Redundant, Jest sharding of just the Jest test execution step gives better bang-for-buck ratio.

PR #53, opened 2023-07-25
AG-8577 Simplify Radar animations

alexanderby:


lsjroberts (approved):
LGTM. If we do a third radar series type we could share that maths, but all good for now.

PR #54, opened 2023-07-26
AG-8469 Refactor agChartOptions types into files by topic

lsjroberts:
PR for https://ag-grid.atlassian.net/browse/AG-8469

This should help with maintainability as we add more series types.

To minimise impact I've re-exported all the options back out of `agChartOptions.ts` though I'm not sure that's necessarily the best approach as for example a line series should be the only series that deals with line series options. So could just import from `lineOptions.ts` directly.


alexanderby (approved):
Superb!

PR #59, opened 2023-07-27
AG-8395 Fix node highlight outside chart pending crosshair hover option

lsjroberts:
PR for https://ag-grid.atlassian.net/browse/AG-8395

This was added for https://ag-grid.atlassian.net/browse/AG-4371 to enable the crosshair to be triggered when hovering slightly outside the chart series area.

A safer solution would be to have it as an option specific to hovering with crosshairs. Though perhaps still generic such that different features can trigger at different paddings? (Ticket added https://ag-grid.atlassian.net/browse/AG-8983)

### DIFF
--- a/packages/ag-charts-community/src/chart/cartesianChart.ts
+++ b/packages/ag-charts-community/src/chart/cartesianChart.ts
@@ -42,15 +42,12 @@ export class CartesianChart extends Chart {
             left: seriesAreaPadding.left,
         });
 
-        const hoverRectPadding = 20;
-        const hoverRect = seriesPaddedRect.clone().grow(hoverRectPadding);
-
-        this.hoverRect = hoverRect;
+        this.hoverRect = seriesPaddedRect;
 
         this.layoutService.dispatchLayoutComplete({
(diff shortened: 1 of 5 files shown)


alexanderby (approved):

packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1040,7 +1038,7 @@ export abstract class Chart extends Observable implements AgChartInstance {
         this.lastInteractionEvent = undefined;
     });
     protected handlePointer(event: InteractionEvent<'hover'>) {
-        const { lastPick, hoverRect } = this;
+        const { lastPick, seriesRect } = this;
```
manapeirov: should this be using `seriesPaddedRect`? because `hoverRect` was growing from that rather than `seriesRect`

lsjroberts: Good spot, I've updated it to use the padded rect throughout instead.

manapeirov (approved):
LGTM

PR #60, opened 2023-07-27
AG-8868 Fix chart update to mark series as dirty in scheduled update

lsjroberts:
PR for https://ag-grid.atlassian.net/browse/AG-8868

Fixes when an update is called during another update with `forceNodeDataRefresh: true`, the series were marked clean by the current update before the new scheduled update was processed.

This may cause performance issues, if the `forceNodeDataRefresh` value lingers. And may also cause bad side effects. Requires further discussion.


lsjroberts:
As discussed, should look into short circuiting the `update()` fn instead

lsjroberts:
Closing in favour of https://github.com/ag-grid/ag-charts/pull/95

PR #62, opened 2023-07-27
AG-8972 Nightingale animation

alexanderby:


packages/ag-charts-enterprise/src/polar-series/nightingale/nightingaleSeries.ts
```diff
@@ -478,6 +478,80 @@ export class NightingaleSeries extends _ModuleSupport.PolarSeries<NightingaleNod
...
+    }
+
+    protected animateEmptyUpdateReady() {
+        if (!this.visible) {
+            return;
+        }
+
+        const { sectorSelection, labelSelection } = this;
+
+        const duration = 1000;
```
manapeirov: const duration = this.ctx.animationManager?.defaultOptions.duration ?? 1000;

manapeirov (approved):
Perfect

PR #66, opened 2023-07-28
AG-8997 Polar chart axis hole

alexanderby:


packages/ag-charts-community/src/chart/polarChart.ts
```diff
@@ -40,25 +40,31 @@ export class PolarChart extends Chart {
...
-            }
+        const angleAxis = this.axes.find((axis) => axis.direction === ChartAxisDirection.X);
+        const radiusAxis = this.axes.find((axis) => axis.direction === ChartAxisDirection.Y);
+        if (!(angleAxis instanceof PolarAxis) || !(radiusAxis instanceof PolarAxis)) {
+            return;
+        }
+
+        const angleScale = angleAxis.scale;
+        const angles = angleScale.ticks?.().map((value: string) => angleScale.convert(value));
+        const innerRadiusOffset = radiusAxis.innerRadiusOffset;
```
manapeirov, line 51: Why do we need consts for this? can't we just do
`angleAxis.innerRadiusOffset = radiusAxis.innerRadiusOffset;`

manapeirov (approved):
LGTM

packages/ag-charts-community/src/chart/polarChart.ts
alexanderby: The variable is reused in several places and makes the code look shorter.

manapeirov: Sure it does

PR #73, opened 2023-08-03
AG-8490 Add scatter series add/update/remove animation

lsjroberts:
PR for https://ag-grid.atlassian.net/browse/AG-8490

https://github.com/ag-grid/ag-charts/assets/3817697/431258b9-5e6d-4294-add4-605b686452bc

### DIFF
--- a/packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
+++ b/packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
@@ -1,6 +1,6 @@
 import type { Selection } from '../../../scene/selection';
 import type { SeriesNodeDataContext } from '../series';
-import { SeriesTooltip, SeriesNodePickMode, valueProperty } from '../series';
+import { SeriesTooltip, SeriesNodePickMode, valueProperty, keyProperty } from '../series';
 import type { ChartLegendDatum, CategoryLegendDatum } from '../../legendDatum';
 import { ColorScale } from '../../../scale/colorScale';
 import { LinearScale } from '../../../scale/linearScale';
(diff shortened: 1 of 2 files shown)


packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
```diff
@@ -149,11 +160,15 @@ export class ScatterSeries extends CartesianSeries<SeriesNodeDataContext<Scatter
 
         const { dataModel, processedData } = await dataController.request<any, any, true>(this.id, data ?? [], {
             props: [
+                keyProperty(this, xKey, isContinuousX, { id: 'xKey-raw' }),
```
manapeirov, line 163: Aren't these just value properties?

packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
```diff
@@ -314,7 +333,11 @@ export class ScatterSeries extends CartesianSeries<SeriesNodeDataContext<Scatter
         }
 
         const data = enabled ? nodeData : [];
-        return markerSelection.update(data);
+        return markerSelection.update(
+            data,
+            undefined,
```
manapeirov, line 338: This was throwing some error in the line series when running tests, so I think Alex removed it. Not sure if it's been resolved now

manapeirov (approved):
Nice

packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
lsjroberts: They are, but they are needed as keys for the diff processor to assign a unique id to each datum. So it becomes a composite key.

See here https://github.com/ag-grid/ag-charts/blob/latest/packages/ag-charts-community/src/chart/data/processors.ts#L269-L270

Admittedly, there may be a better way.

packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
lsjroberts: This line is still there in line series (and is very much needed). Alex's fix over here seems to have done the job - https://github.com/ag-grid/ag-charts/commit/eaa63bb907a0397f6b46882cda1711b19df6c443

I think it's a race condition of some kind, or my logic is wrong haha. But this should be fine now.

PR #77, opened 2023-08-04
AG-8261 Radial column shape

alexanderby:
Connect the bottom of a radial column with the inner axis circle (but not when the connection is going to be too large).


manapeirov (approved):
Wonderful

PR #81, opened 2023-08-07
Move thumbnail generation to postinstall command

taktran:
Also only install chromium, so it's faster


taktran:
Will try and replace playwright with node-canvas, and work on this then

PR #86, opened 2023-08-08
Ag 8889/range bar axes

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-8889


packages/ag-charts-community/src/scale/continuousScale.ts
```diff
@@ -34,7 +34,7 @@ export abstract class ContinuousScale<D extends number | Date, I = number> imple
 
     abstract toDomain(d: number): D;
 
-    protected getDomain() {
+    public getDomain() {
```
alexanderby: `public` is unnecessary.

alexanderby (approved):
LBTM

PR #89, opened 2023-08-09
AG-9057 - Fix nx dev server full restart on example edit.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-9057

Reduces the severity of reload when an example is edited, from full server reload (before) to page reload (after).


packages/ag-charts-website/src/astro/integrations/.gitkeep
taktran, line 1: We probably don't need this `.gitkeep` and the `astro/integrations` folder if we don't have any astro integrations

taktran (approved):
Nice one 👍🏼

packages/ag-charts-website/src/astro/integrations/.gitkeep
alantreadway: Done

PR #91, opened 2023-08-09
AG-9057 - Fix redundant Astro startup errors.

alantreadway:
Fixes type and import errors on Astro startup - however embedded TS in .mdoc files importing `ag-charts-community` will stop working after this change (this doesn't affect examples, only the placeholder example we placed on the first landing page).


taktran (approved):
Cool, do we need to delete the homepage example then?

alantreadway:
> Cool, do we need to delete the homepage example then?

Fixed

taktran (approved):
Double 👍🏼 👍🏼

PR #92, opened 2023-08-10
AG-8489 Add pie series add/remove/update animation

lsjroberts:
PR for https://ag-grid.atlassian.net/browse/AG-8489

Will require some back and forth with QA as the spec didn't detail what should happen with colours (sequential colours need to change to match legend changing) or the center moving about due to repositioning callout labels.


alantreadway (approved):
LGTM

PR #93, opened 2023-08-10
Remove animation state machine actions

lsjroberts:
Removes the animation state machine actions on transitioning between states, since we were not using them and the extra nesting from the `on` property was annoying.


alantreadway (approved):
LGTM, nice cleanup - FYI there is a merge conflict.

PR #95, opened 2023-08-10
AG-8868 Add chart update shortcuts when triggered within another update

lsjroberts:
PR for https://github.com/ag-grid/ag-charts/pull/60

Fixes when an update is called within another update, to ensure the rendered state matches the latest update attempt. This is limited to a maximum of 4 updates to prevent infinite loops and excessive re-runs.

Before:
Screenshot 2023-08-10 at 13 12 47

After:
Screenshot 2023-08-10 at 13 58 00


alantreadway (approved):
Nice and neat, thanks! LGTM

PR #96, opened 2023-08-10
AG-9072 - Add docs left hand nav

taktran:
## Changes

* Add docs left hand nav
* Add Astro validation for menu data contents
* Add placeholder doc pages for missing content

## Notes

Just the first pass of side menu. In later passes, we'll work on

* Mobile view
* Filtering out javascript only pages

### DIFF
--- a/packages/ag-charts-website/src/content/menu/data.json
+++ b/packages/ag-charts-website/src/content/menu/data.json
@@ -1,0 +1,352 @@
+{
+    "main": {
+        "items": [
+            {
+                "title": "Getting Started",
+                "path": "getting-started"
+            },
+            {
+                "title": "Building & Tooling",
+                "items": [
+                    { "title": "Install with NPM", "path": "npm", "frameworks": ["javascript"] },
+                    { "title": "Download Bundles", "path": "download", "frameworks": ["javascript"] },
+                    { "title": "Community and Enterprise", "path": "licensing" },
+                    { "title": "Packages vs Modules", "path": "packages-modules" },
+                    { "title": "Packages", "path": "packages", "frameworks": ["javascript"] },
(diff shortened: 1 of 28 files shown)


packages/ag-charts-website/src/content/menu/data.json
```diff
@@ -0,0 +1,352 @@
...
+        ]
+    },
+    "charts": {
+        "items": [
+            {
+                "title": "Line",
+                "icon": "chartsLine",
+                "items": [
+                    {
+                        "title": "Overview",
```
alantreadway: `"title": "Overview"` is repeated across all 2nd-level items, why is that?

packages/ag-charts-website/src/content/menu/data.json
```diff
@@ -0,0 +1,352 @@
...
+                ]
+            },
+            {
+                "title": "AG Grid Integrated Charts",
+                "url": "https://ag-grid.com/javascript-data-grid/integrated-charts/",
+                "isEnterprise": true
+            }
+        ]
+    },
+    "charts": {
```
alantreadway, line 93: Is this for the gallery or showcase? (It's not clear to me from the name)

packages/ag-charts-website/src/content/menu/data.json
taktran: They are placeholders for now, and may be generated in future. It's easier to change as items in the menu data json for now

packages/ag-charts-website/src/content/menu/data.json
taktran: Neither. It's just for the left hand nav in docs as it's in `menu/data.json`, but it's just for the charts part of the menu (ie, with icons), which is separate from the main part.

alantreadway (approved):

PR #100, opened 2023-08-11
AG-9075 - Add prism syntax highlighting

taktran:
## Changes

* Add prism syntax highlighting
* Extend default astro prism plugin to add the `code` class for design system styles to come through


alantreadway (approved):
LGTM

PR #103, opened 2023-08-14
AG-8932 - Move enterprise typings into community.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-8932

Moves us towards the desired end-state of all typings being part of `ag-charts-community`. with runtime messages to highlight use of enterprise features outside the scope of `ag-charts-community`.

This change attempts to move typings wholesale, with the minimal changes to enable `nx build`/`nx lint` and `nx test` to pass.

### DIFF
--- a/packages/ag-charts-community/src/chart/options/chartOptions.ts
+++ b/packages/ag-charts-community/src/chart/options/chartOptions.ts
@@ -1,0 +1,152 @@
+import type { AgChartBackgroundImage } from '../background/backgroundOptions';
+import type { AgBaseChartListeners } from './eventOptions';
+import type { AgChartTheme } from './themeOptions';
+import type { AgChartTooltipOptions } from './tooltipOptions';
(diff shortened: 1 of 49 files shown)


packages/ag-charts-community/src/chart/options/chartOptions.ts
lsjroberts, line 1: Nice, yeah, I couldn't think of anywhere to put these previously, but `chartOptions` works for sure.

lsjroberts (approved):
Awesome, LGTM.

packages/ag-charts-community/src/chart/options/chartOptions.ts
alantreadway: Yeah - I figured this was ok for a first pass, we can revise the location if we come up with a better scheme going forward 🤞

lsjroberts (approved):

PR #105, opened 2023-08-14
AG-9058 - Fix mdoc prettier formatting via nx format

alantreadway:
https://ag-grid.atlassian.net/browse/AG-9058

Since `nx` uses the `prettier` JS API, the addition of `*.mdoc` overrides in `.prettierrc` doesn't work. The only solution from inspecting the `nx` and `prettier` sourcecode is to patch `prettier` to include the `.mdoc` extension in it's list of know extensions.

This PR adds `patch-package` to handle NPM package patching responsibility (including the previous adhoc `postinstall` patch of `nx` itself).


taktran (approved):
Noice 👍🏼

Are `patches` a nx thing, or is there a script that needs to run to apply them?

alantreadway:
> Noice 👍🏼
> 
> 
> 
> Are `patches` a nx thing, or is there a script that needs to run to apply them?

No, they are managed by a NPM package called [patch-package](https://www.npmjs.com/package/patch-package) - there is a post install script to invoke it (see [here](https://github.com/ag-grid/ag-charts/pull/105/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R8)).

PR #110, opened 2023-08-15
AG-8826 - Improve high series count performance

alantreadway:
https://ag-grid.atlassian.net/browse/AG-8826

Significantly improves performance for high series-count charts by counter-intuitively reducing the number of layers. This effectively batches up more canvas operations per layer rasterisation, which is why it is faster for low per-series data sizes.

Concrete changes:
- Reduces the 1:1 mapping for series<->layer if the series count is >= 30
- Also now collapses highlight groups into shared layers to reduce layer count overhead at minimal performance change.
- Fixes double render on initial load due to later resize (I increase the back-off count, which seems to have fixed the issue although I'm not sure why).

Bonus:
- Adds emojis to `window.agChartsSceneRenderCharts` outputs to better distinguish timing measurements.

### DIFF
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -611,7 +611,7 @@ export abstract class Chart extends Observable implements AgChartInstance {
         const end = performance.now();
         this.log('Chart.performUpdate() - end', {
             chart: this,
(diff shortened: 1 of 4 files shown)


manapeirov (approved):
LGTM :)

PR #113, opened 2023-08-16
AG-8826 - Ignore incorrect resize event before DOMContentLoaded.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-8826

Attempt to fix render-resize-render glitch on initial load.

From observation, this Plnkr had the following behaviour before this change:
- ResizeObserver is setup and monitors the chart element.
- Resize event fires with non-stable sizes.
- Chart.update() processing runs and renders a chart.
- DOMContentLoaded event fires.
- Resize event fires with stable sizes.
- Chart.update() processing runs and renders a chart at a different size.

This change attempt to prevent the first resize event from triggering a resize by waiting for `DOMContentLoaded` to fire if the DOM is still being loaded at chart setup.

### DIFF
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -146,14 +146,14 @@ export abstract class Chart extends Observable implements AgChartInstance {
 
     @ActionOnSet<Chart>({
         newValue(value) {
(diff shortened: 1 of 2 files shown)


manapeirov (approved):
LGTM!

PR #115, opened 2023-08-17
AG-8490 Fix scatter snapshots

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-8490

Seems it was being overly eager on cleaning up markers. I'm not sure of the root cause of this, as in theory when no animation is happening (or has completed) it should be fine to run `selection.cleanup()` as much as one likes.

But removing the two animation states that just reset things fixes it. Other animation bugs may be related so they may drive out a deeper issue, if there is one.


alantreadway (approved):
LGTM

PR #116, opened 2023-08-17
AG-8463 Fix line series markers to use both axes for datum id

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-8463

Fixes the case where a line series has multiple nodes at the same x-value. Previously it would only show the one marker that shared that x-value. Now it uses both x and y-values for the datum id.


lsjroberts:
Closing as I remembered why this doesn't work. You can't use the y-value as part of the id as it is the value, and you need the line to transition between y-values when it changes. Not think it is a new node and instantly pop to it.

PR #117, opened 2023-08-17
Enable enterprise gallery examples

alantreadway:
- Enables use of enterprise package for gallery examples by default.
- Enables `animation`, `contextMenu` and `zoom` features by default for enterprise.


packages/ag-charts-enterprise/src/navigator/chartNavigatorModule.ts
```diff
@@ -8,50 +8,3 @@ export const CHART_NAVIGATOR_MODULE: _ModuleSupport.Module = {
     chartTypes: ['cartesian'],
     instanceConstructor: Navigator,
 };
-
-export interface AgNavigatorMaskOptions {
```
lsjroberts, line 12: good spot

lsjroberts (approved):
LGTM, I like this idea

PR #119, opened 2023-08-17
AG-8889 - Fix series grouped and stacked options processing

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-8889


packages/ag-charts-community/src/chart/mapping/prepareSeries.test.ts
```diff
@@ -144,258 +186,560 @@ describe('transform series options', () => {
...
+                    const groupable = seriesTypes[sType].groupable;
+                    const stackable = seriesTypes[sType].stackable;
+                    const stacked = stackable;
+                    const grouped = !stacked && groupable;
+                    result.forEach((s) => {
+                        if (stackable) {
+                            expect(s.stacked).toBe(stacked);
+                            expect(s.grouped).toBe(grouped);
+                        } else {
+                            expect(console.warn).toBeCalled();
```
alantreadway: Nit: Can we also verify that the console warning is the expected warning message?

I think this can be achieved with something like:
```
expect(console.warn).toHaveBeenCalledWith(/some regexp matching a console message/);
```

packages/ag-charts-community/src/chart/mapping/prepareSeries.test.ts
```diff
@@ -144,258 +186,560 @@ describe('transform series options', () => {
...
+                    );
+                    const groupable = seriesTypes[sType].groupable;
+                    const stackable = seriesTypes[sType].stackable;
+                    const stacked = false;
+                    const grouped = !stacked && groupable;
+                    result.forEach((s) => {
+                        if (stackable) {
+                            expect(s.stacked).toBe(stacked);
+                            expect(s.grouped).toBe(grouped);
+                        }
```
alantreadway, line 275: Should we make an assertion in the negative case as a sanity check?

packages/ag-charts-community/src/chart/mapping/prepareSeries.test.ts
```diff
@@ -144,258 +186,560 @@ describe('transform series options', () => {
...
+                    const stackable = seriesTypes[sType].stackable;
+                    const groupable = seriesTypes[sType].groupable;
+                    const stacked = seriesTypes[sType].stackedByDefault;
+                    const grouped = !stacked && groupable;
+
+                    result.forEach((s) => {
+                        if (groupable || stackable) {
+                            expect(s.stacked).toBe(stacked);
+                            expect(s.grouped).toBe(grouped);
+                        }
```
alantreadway, line 303: Ditto negative case.

packages/ag-charts-community/src/chart/mapping/prepareSeries.test.ts
```diff
@@ -144,258 +186,560 @@ describe('transform series options', () => {
...
+                    );
+                    const groupable = seriesTypes[sType].groupable;
+                    const stacked = !groupable;
+                    const grouped = groupable;
+                    result.forEach((s) => {
+                        if (groupable) {
+                            expect(s.stacked).toBe(stacked);
+                            expect(s.grouped).toBe(grouped);
+                        } else {
+                            expect(console.warn).toBeCalled();
```
alantreadway: Ditto

packages/ag-charts-community/src/chart/mapping/prepareSeries.test.ts
```diff
@@ -144,258 +186,560 @@ describe('transform series options', () => {
...
+                    );
+                    const groupable = seriesTypes[sType].groupable;
+                    const stackable = seriesTypes[sType].stackable;
+                    const stacked = stackable && seriesTypes[sType].stackedByDefault;
+                    const grouped = false;
+                    result.forEach((s) => {
+                        if (groupable) {
+                            expect(s.stacked).toBe(stacked);
+                            expect(s.grouped).toBe(grouped);
+                        }
```
alantreadway, line 363: Ditto negative case.

packages/ag-charts-community/src/chart/mapping/prepareSeries.test.ts
manapeirov: So at the moment we don't do a waring if the options set `stacked: false` and the series is not stackable, the warning only appears if it's been set to true and the series isn't stackable

packages/ag-charts-community/src/chart/mapping/prepareSeries.test.ts
manapeirov: Sure

alantreadway (approved):
LGTM - have you verified that the code without your fixes fails the tests as you'd expect?

PR #120, opened 2023-08-18
AG -9077 (subtask: AG-9156)- First pass of interface documentation

taktran:
## Changes

* Add `ApiDocumentation` and `InterfaceDocumentation` components from grid
* Add `interfaceDocumentation` markdoc tag
* Update all docs to use `interfaceDocumentation` markdoc tag
* Add prism types

## Known issues

* `ApiDocumentation` component not done yet - `charts-api-create-update` and `charts-api-download`
* Code details does not recursively show all interfaces
* Defaults values are not included
* Required fields are not highlighted and not sorted to the top

PR #123, opened 2023-08-18
AG-9077 - API Documentation Round 2

taktran:
Follow up to #120

## Changes

* Fix details code not showing all interfaces, by adding `getTypeLink` back in
* Enable `apiDocumentation` to work
  * Update `charts-api-create-update` and `charts-api-download` pages to use `apiDocumentation` tag

Bonus

* Add support for framework specific docs by using markdoc `if` tag and custom functions `isFramework` and `isNotJavascriptFramework`
* Clean up html tags in markdoc
* Add `externalLink` markdoc tag for links that open in a new tab

## Known issues

(to be dealt with in another ticket)

* Defaults values are not included
* Required fields are not highlighted and not sorted to the top


packages/ag-charts-website/src/features/api-documentation/utils/getTypeLinks.ts
```diff
@@ -0,0 +1,85 @@
...
+    AgPolarChartOptions: '/integrated-charts-customisation/#example-piedoughnut-chart-overrides',
+    AgPolarAxesTheme: '/integrated-charts-customisation/#example-piedoughnut-chart-overrides',
+    AgPolarSeriesTheme: '/integrated-charts-customisation/#example-piedoughnut-chart-overrides',
+    AgPieSeriesOptions: '/integrated-charts-customisation/#example-piedoughnut-chart-overrides',
+    Blob: 'https://developer.mozilla.org/en-US/docs/Web/API/Blob',
+    CellPosition: '/keyboard-navigation/#cellposition',
+    CellRange: '/range-selection/#range-selection-api',
+    ICellRendererParams: '/component-cell-renderer/#reference-ICellRendererParams',
+    ICellEditorParams: '/component-cell-editor/#reference-ICellEditorParams',
+    ChartModel: '/integrated-charts-api-save-restore-charts/#api-reference',
```
alantreadway, line 26: Some of these types are Grid types - maybe we could drop them, or break them into a distinct definition to make it clearer?

alantreadway (approved):
LGTM other than one comment (I think we can defer the fix though)

packages/ag-charts-website/src/features/api-documentation/utils/getTypeLinks.ts
taktran: I've got a todo to come back to this. We'll need to see how it's being used in Charts (or whether there is a better way)

PR #129, opened 2023-08-23
AG-9040 - Make separate docs pages for Range Bar and Column

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-9040


packages/ag-charts-website/src/content/docs/charts-range-column-series/index.mdoc
alantreadway, line 1: Is there any overlap with the `range-bar` documentation, and if so could you explorer how to extract a fragment that could be re-used?

alantreadway:
Conflicts - could you resolve before re-review?

alantreadway:
Should we close this?

PR #133, opened 2023-08-24
Attempt to improve Github Actions cache artifact sizes and overlap.

alantreadway:
Tweaks Nx + `node_modules` cache rules to reduce cache sizes and speed-up cache hits.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,6 +65,8 @@ jobs:
                   path: |
                       node_modules/
                       packages/*/node_modules/
+                      !node_modules/.cache/
+                      !packages/*/node_modules/.cache/
                   key: node_modules-${{hashFiles('yarn.lock')}}
                   restore-keys: node_modules- # Take any latest cache if failed to find it for current yarn.lock
             - name: Cache Nx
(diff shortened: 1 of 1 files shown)


taktran (approved):
Noice ✨

.github/workflows/ci.yml
```diff
@@ -81,6 +83,9 @@ jobs:
                   node-version: '18'
                   cache: ${{ steps.cache.outputs.cache-hit != 'true' && 'yarn' || '' }}
 
+            - name: Purge Nx cache of old entries
+              run: find node_modules/.cache/nx -maxdepth 1 -mtime +7 -exec rm -rf '{}' \;
```
lsjroberts, line 87: I must be missing something, but if you are skipping `node_modules/.cache` up on line 68, then there would be nothing to remove here?

alantreadway: Ah, so the naunce is that the `Cache Nx` step is specifically managing save/restore of `node_modules/.cache/nx` - the additions on line 68 is to ensure that the `Cache node_modules` step doesn't doubly cache the Nx cache.

PR #139, opened 2023-08-25
AG-8890 - Remove both markers if either low or high value are invalid…

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-8890


alantreadway (approved):
LGTM

PR #147, opened 2023-08-30
AG-8489 Change pie animation method and clear on complex animations

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-9236

- Removed pie sectors are taken from the end and others are replaced by the updated sectors
- If an animation includes added and removed sectors, it fades out the whole pie and redraws the initial animation

https://github.com/ag-grid/ag-charts/assets/3817697/13bf86e2-9c09-41e2-8c94-9c4a60457c11


alantreadway (approved):
LGTM

PR #148, opened 2023-08-31
Series tooltip implementation improvement/suggestion

iMoses:
Switched `SeriesTooltip` to use generics, so we won't have to extend this class on every type of chart

### DIFF
--- a/packages/ag-charts-community/src/chart/series/series.ts
+++ b/packages/ag-charts-community/src/chart/series/series.ts
@@ -24,7 +24,7 @@ import { ChartAxisDirection } from '../chartAxisDirection';
 import type { AgSeriesTooltipRendererParams, AgTooltipRendererResult, InteractionRange } from '../agChartOptions';
 import type { DatumPropertyDefinition, ScopeProvider } from '../data/dataModel';
 import { fixNumericExtent } from '../data/dataModel';
-import { TooltipPosition } from '../tooltip/tooltip';
+import { TooltipPosition, toTooltipHtml } from '../tooltip/tooltip';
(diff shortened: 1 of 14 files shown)


packages/ag-charts-community/src/chart/series/series.ts
```diff
@@ -249,13 +251,19 @@ export class HighlightStyle {
...
     enabled: boolean = true;
 
     @Validate(OPT_BOOLEAN)
     showArrow?: boolean = undefined;
 
+    @Validate(OPT_FUNCTION)
+    renderer?: (params: P) => string | AgTooltipRendererResult = undefined;
+
+    @Validate(OPT_STRING)
+    format?: string = undefined;
```
alantreadway, line 266: So I'm just wondering if in the cases where we don't support `format` we could ensure that trying to set it will still a console warn?

The hacky way to do this would be to NOT assign `undefined `here so that the field isn't present at runtime by default, and then in the constructor passing an option to control an initial assignment to `undefined` if we're expecting users to be able to set it?

alantreadway (commented):
I like the improvement, thanks for this! One thing I think we should try to address to ensure we give feedback to users still if they try and set a `formatter` for a case where we don't support it yet?

packages/ag-charts-community/src/chart/series/series.ts
alantreadway: Discussed offline: We'll add `formatter` support universally.

alantreadway (approved):
LGTM, thanks for the cleanup!

alantreadway (approved):
LGTM

PR #150, opened 2023-08-31
AG-9234 - Add devFiles wildcard support.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-9234

Adds `DEV_FILE_PATH_MAP` support for wildcards, which opens the door to chunked bundles for `ag-charts-community` and `ag-charts-enterprise` in the future if we decide to make that change.

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -24,7 +24,7 @@
       "tsDeclarations",
       "sharedGlobals"
     ],
-    "sharedGlobals": ["{workspaceRoot}/rollup.*.js", "{workspaceRoot}/tsconfig.*.json"]
+    "sharedGlobals": ["{workspaceRoot}/rollup.*.cjs", "{workspaceRoot}/tsconfig.*.json"]
   },
   "targetDefaults": {
     "build": {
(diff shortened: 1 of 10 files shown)


nx.json
```diff
@@ -24,7 +24,7 @@
       "tsDeclarations",
       "sharedGlobals"
     ],
-    "sharedGlobals": ["{workspaceRoot}/rollup.*.js", "{workspaceRoot}/tsconfig.*.json"]
+    "sharedGlobals": ["{workspaceRoot}/rollup.*.cjs", "{workspaceRoot}/tsconfig.*.json"]
```
alantreadway, line 27: NOTE: This is an unrelated fix to the dependencies for many targets.

packages/ag-charts-website/src/utils/pages.ts
```diff
@@ -169,8 +164,7 @@ export const getBoilerPlateUrl = ({
  * Get dev files for local development
  */
 export function getDevFiles(): DevFileRoute[] {
-    const files = Object.keys(DEV_FILE_PATH_MAP).map((filePath) => {
-        const sourceFilePath = DEV_FILE_PATH_MAP[filePath];
+    const files = Object.entries(DEV_FILE_PATH_MAP).map(([filePath, sourceFilePath]) => {
```
taktran, line 167: 👍🏼 I'm sure I've not used `Object.entries` in other places where it would've been useful too

taktran (approved):
Nice addition 👍🏼 Good stuff!

PR #154, opened 2023-08-31
Enable madge lint target.

alantreadway:
Follow up to #153 to ensure we don't accidentally introduce further circular deps.


iMoses (approved):
LGTM 👍

PR #155, opened 2023-09-01
AG-8890 - Remove data model id aliases, use [scope, id] tuples instead

manapeirov:
https://ag-grid.atlassian.net/browse/AG-9040

### DIFF
--- a/packages/ag-charts-community/src/chart/data/dataModel.ts
+++ b/packages/ag-charts-community/src/chart/data/dataModel.ts
@@ -134,9 +134,9 @@ type PropertyIdentifiers = {
     /** Scope(s) a property definition belongs to (typically the defining entities unique identifier). */
     scopes?: string[];
     /** Unique id for a property definition within the scope(s) provided. */
+    /** Tuples of [scope, id] that match this definition. */
+    ids?: [string, string][];
     id?: string;
-    /** Alternative unique ids for the property. */
-    idAliases?: string[];
     /** Optional group a property belongs to, for cross-scope combination. */
     groupId?: string;
 };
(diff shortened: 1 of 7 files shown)


packages/ag-charts-community/src/chart/data/dataModel.ts
```diff
@@ -322,7 +312,7 @@ export class DataModel<
         searchId: string,
         type: PropertyDefinition<any>['type'] = 'value'
     ): { type: typeof type; index: number; def: PropertyDefinition<any> } | never {
-        const { index, def } = this.resolveProcessedDataDefById(scope, searchId, type) ?? {};
```
alantreadway, line 325: Can we drop `type` as a parameter to this method? If it was being set to anything other than `'value'` is this change introducing a bug?

(My recollection was we discussed this offline, so perhaps we just need to remove the `type` parameter above?)

packages/ag-charts-community/src/chart/data/dataModel.ts
```diff
@@ -331,40 +321,32 @@ export class DataModel<
         searchId: string | RegExp,
         type: PropertyDefinition<any>['type'] = 'value'
     ): { type: typeof type; index: number; def: PropertyDefinition<any> }[] | never {
-        return this.resolveProcessedDataDefsById(scope, searchId, type).map(({ index, def }) => ({ type, index, def }));
```
alantreadway, line 334: Ditto

packages/ag-charts-community/src/chart/data/dataModel.ts
```diff
@@ -487,24 +469,25 @@ export class DataModel<
...
-                    return false;
-                }
-                return !(matchScopes && (def.scopes == null || !matchScopes.some((s) => def.scopes?.includes(s))));
+                return !(
+                    matchIds &&
+                    (def.ids == null ||
+                        !matchIds.some(([matchScope, matchId]) =>
+                            def.ids?.some(([defScope, defId]) => defScope === matchScope && defId === matchId)
+                        ))
+                );
```
alantreadway: I'd imagine this could be broken into a few simpler steps?

e.g.:
```
if (!matchIds) return true;
if (def.ids == null) return false;

return matchIds.some(([matchScope, matchId]) =>
  def.ids?.some(([defScope, defId]) => defScope === matchScope && defId === matchId)
));
```

alantreadway (approved):
LGTM! Thanks for this

PR #156, opened 2023-09-01
AG-9234 - Improve private entry points

alantreadway:
https://ag-grid.atlassian.net/browse/AG-9234


alantreadway:
Closing for now, I don't think we can bite this off in one step so we need to rethink the approach.

PR #159, opened 2023-09-04
AG-8463 Fix static line series with datums sharing x-values

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-8463

This can be improved later by adding some kind of `datumId` getter to the chart options. 

Will also add a note to the animation docs regarding this issue.


alantreadway (approved):
LGTM

PR #160, opened 2023-09-04
AG-9024 - Add basic box-plot-bar series

iMoses:
https://ag-grid.atlassian.net/browse/AG-9024


packages/ag-charts-community/src/chart/series/cartesian/boxPlotOptions.ts
```diff
@@ -0,0 +1,45 @@
...
+    minKey?: string;
+    minName?: string;
+    q1Key?: string;
+    q1Name?: string;
+    medianKey?: string;
+    medianName?: string;
+    q3Key?: string;
+    q3Name?: string;
+    maxKey?: string;
+    maxName?: string;
```
alantreadway, line 34: We will need to document these properly in a future iteration.

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.test.ts
```diff
@@ -0,0 +1,57 @@
...
+            minKey: 'min',
+            q1Key: 'q1',
+            medianKey: 'median',
+            q3Key: 'q3',
+            maxKey: 'max',
+        },
+    ],
+};
+
+describe.skip('Chart', () => {
```
alantreadway: Do we need to skip, I'm assuming this should work now?

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
```diff
@@ -0,0 +1,319 @@
...
+    readonly q3Value: number;
+    readonly maxValue: number;
+
+    readonly fill: string;
+    readonly stroke: string;
+    readonly strokeWidth: number;
+}
+
+export class BoxPlotSeries extends CartesianSeries<_ModuleSupport.SeriesNodeDataContext<BoxPlotNodeDatum>> {
+    constructor(moduleCtx: _ModuleSupport.ModuleContext) {
```
alantreadway: Nit: Move to after the property definitions, it feels wrong here?

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
```diff
@@ -0,0 +1,319 @@
...
+                valueProperty(this, maxKey, true, { id: `maxValue` }),
+            ],
+            dataVisible: this.visible,
+        });
+
+        this.dataModel = dataModel;
+        this.processedData = processedData;
+    }
+
+    // @ts-ignore
```
alantreadway: Needed?

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
```diff
@@ -0,0 +1,319 @@
...
+                maxValue: Math.round(xAxis.scale.convert(maxValue, { strict: false })),
+            };
+
+            context.nodeData.push(nodeData);
+        });
+
+        return [context];
+    }
+
+    async processData(dataController: _ModuleSupport.DataController): Promise<void> {
```
alantreadway, line 98: Nit: Could you move this and `getDomain` to appear above `createNodeData` - this is the order we use in all other series implementations, and it logically follows the processing order?

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
```diff
@@ -0,0 +1,319 @@
...
+            ],
+            dataVisible: this.visible,
+        });
+
+        this.dataModel = dataModel;
+        this.processedData = processedData;
+    }
+
+    // @ts-ignore
+    getDomain(direction: _ModuleSupport.ChartAxisDirection): any[] {
```
alantreadway: Maybe try and drop the explicit return type to remove the `any`?

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
```diff
@@ -0,0 +1,319 @@
...
+            const yValueIndex = dataModel.resolveProcessedDataIndexById(this, `yValue`).index;
+            return processedData.domain.keys[yValueIndex];
+        }
+    }
+
+    getLegendData(): _ModuleSupport.ChartLegendDatum[] {
+        return [];
+    }
+
+    // @ts-ignore
```
alantreadway: Ditto?

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
```diff
@@ -0,0 +1,319 @@
...
+        const { nodeData, datumSelection } = opts;
+        const data = nodeData ?? [];
+        return datumSelection.update(
+            data.filter(
+                (datum) =>
+                    datum.minValue <= datum.q1Value &&
+                    datum.q1Value <= datum.medianValue &&
+                    datum.medianValue <= datum.q3Value &&
+                    datum.q3Value <= datum.maxValue
+            )
```
alantreadway: I would do this filtering earlier on, in the `processData()` step (you might be able to configure a data validator for this purpose).

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
```diff
@@ -0,0 +1,319 @@
...
+
+            boxes[1].x = datum.medianValue;
+            boxes[1].y = datum.yValue;
+            boxes[1].width = datum.q3Value - datum.medianValue;
+            boxes[1].height = datum.bandwidth;
+
+            caps[0].x1 = caps[0].x2 = datum.minValue;
+            caps[1].x1 = caps[1].x2 = datum.maxValue;
+            caps[0].y1 = caps[1].y1 = datum.yValue + datum.bandwidth * 0.25;
+            caps[0].y2 = caps[1].y2 = datum.yValue + datum.bandwidth * 0.75;
```
alantreadway: I guess this is going to become more configurable in later JIRAs?

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
```diff
@@ -0,0 +1,319 @@
...
+
+            whiskers[0].x1 = datum.minValue;
+            whiskers[0].x2 = datum.q1Value;
+            whiskers[1].x1 = datum.q3Value;
+            whiskers[1].x2 = datum.maxValue;
+            whiskers[0].y1 = whiskers[0].y2 = whiskers[1].y1 = whiskers[1].y2 = datum.yValue + datum.bandwidth * 0.5;
```
alantreadway: Should the `0.5` be tied to the `0.25` and `0.75` above?

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
```diff
@@ -0,0 +1,319 @@
...
+                caps[0].stroke =
+                caps[1].stroke =
+                    datum.stroke;
+            boxes[0].strokeWidth =
+                boxes[1].strokeWidth =
+                whiskers[0].strokeWidth =
+                whiskers[1].strokeWidth =
+                caps[0].strokeWidth =
+                caps[1].strokeWidth =
+                    datum.strokeWidth;
```
alantreadway: Nit: I would avoid these chained assignments, it doesn't follow our typical coding pattern.

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
alantreadway: Ditto, could we avoid the chained assignments?

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
alantreadway: Ditto, chained assignments.

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
```diff
@@ -0,0 +1,319 @@
...
+        });
+    }
+
+    // @ts-ignore
```
alantreadway: Needed?

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
iMoses: this is needed because `getTooltipHtml` still isn't implemented and `seriesDatum` isn't used

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
iMoses: I moved it to `processData`, if you think I should add a validator then let's talk about it

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
iMoses: yes, i already extracted the `capSize` value into a const for future proofing

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
iMoses: yes and no :)
see newer iteration

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
iMoses: yes, because of unimplemented method and unused arguments

packages/ag-charts-community/src/chart/series/cartesian/boxPlotOptions.ts
iMoses: will be part of AG-9023, unless you mean within the code?

iMoses:
note that I removed dead code from the `Shape` class
if you think I shouldn't have or it's better I do it in a separate PR let me know

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
alantreadway: I'd prefer we didn't use this, we don't have any other examples in library code - for unused variables use the more specific `_` prefix to ignore for now?

packages/ag-charts-community/src/scene/shape/line.ts
```diff
@@ -28,6 +28,16 @@ export class Line extends Shape {
...
 
+    set x(value: number) {
+        this.x1 = value;
+        this.x2 = value;
+    }
+
+    set y(value: number) {
+        this.y1 = value;
+        this.y2 = value;
+    }
```
alantreadway, line 39: This addition to the `Line` interface seems confusing (as it already has `x1`, `x2`, `y1` and `y2` properties)?

packages/ag-charts-community/src/scene/shape/shape.ts
```diff
@@ -61,6 +37,13 @@ export abstract class Shape extends Node {
         }
     );
 
+    setStyles<T>(this: T, styles: { [K in keyof T]?: T[K] }) {
```
alantreadway: Nit: Maybe `setProperties` would be a better name, as you can set non-style-related properties with this method?

packages/ag-charts-community/src/scene/shape/shape.ts
```diff
@@ -14,36 +14,12 @@ type CanvasContext = CanvasFillStrokeStyles &
 const LINEAR_GRADIENT_REGEXP = /^linear-gradient\((.*?)deg,\s*(.*?)\s*\)$/i;
 
 export abstract class Shape extends Node {
```
alantreadway, line 16: Thanks for the code cleanup here :tada:

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
```diff
@@ -0,0 +1,383 @@
...
+
+            median.setStyles({
+                x: medianValue,
+                y1: yValue + strokeWidth,
+                y2: yValue + bandwidth - strokeWidth,
+            });
+
+            const capSize = 0.5;
+            const capY1 = yValue + bandwidth * capSize * 0.5;
+            const capY2 = yValue + bandwidth * capSize * 1.5;
```
alantreadway: Nit: Could you extract the `0.5` and `1.5` numbers to constants to try and make them a little less magic here? Are these offsets in proportion of `capSize` (that's how I'm interpreting it)? I'm also presuming that the implied offset of `1` between the two is so that `capSize` is the actual width of the cap?

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
```diff
@@ -0,0 +1,383 @@
...
+                x2: Math.round(maxValue - strokeWidth / 2),
+                y: yValue + bandwidth * 0.5,
+            });
+
+            // fill only elements
+            for (const element of boxes) {
+                element.fill = fill;
+                element.fillOpacity = fillOpacity;
+                element.strokeWidth = strokeWidth * 2;
+                element.strokeOpacity = 0;
```
alantreadway: Any reason not to use `setStyles()` here too?

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
```diff
@@ -0,0 +1,383 @@
...
+            }
+
+            // stroke only elements
+            for (const element of [outline, median, ...whiskers, ...caps]) {
+                element.fillOpacity = 0;
+                element.stroke = stroke;
+                element.strokeWidth = strokeWidth;
+                element.strokeOpacity = strokeOpacity;
+                element.lineDash = lineDash;
+                element.lineDashOffset = lineDashOffset;
```
alantreadway: Ditto

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
```diff
@@ -0,0 +1,383 @@
...
+        labelData: BoxPlotNodeDatum[];
+        labelSelection: _Scene.Selection<_Scene.Text, BoxPlotNodeDatum>;
+        seriesIdx: number;
+    }) {
+        const { labelData, labelSelection } = opts;
+        return labelSelection.update(labelData);
+    }
+
+    protected nodeFactory() {
+        const group = new _Scene.Group({ layer: true });
```
alantreadway: Please don't enable `layer: true` here - this means we're adding an entier `OffscreenCanvas` per datum with the current configuration, which is excessive 😉 .

alantreadway, line 366: Please don't enable `layer: true` here - this means we're adding an entier `OffscreenCanvas` per datum with the current configuration, which is excessive 😉 .

alantreadway (commented):
Thanks - some nice cleanup here too, much appreciated!

This looks to be close now.

packages/ag-charts-community/src/scene/shape/line.ts
iMoses: this is actually following d3 convention
it's a utility setter to use in case both points on the same axis are of identical values
it helps turn this:
```
element.x1 = 123;
element.x2 = 123;
```
into:
```
element.x = 123;
```
I think it helps cleanup the code and it acts as an alternative to chaining, such as `element.x1 = element.x2 = 123;`

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
iMoses: I changed the implementation to not use constants

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
```diff
@@ -0,0 +1,276 @@
...
+            lineDashOffset,
+        } = this;
+
+        const context: _ModuleSupport.SeriesNodeDataContext<BoxPlotNodeDatum> = {
+            itemId: yKey,
+            nodeData: [],
+            labelData: [],
+        };
+
+        const defs = dataModel.resolveProcessedDataDefsByIds(this, [
```
alantreadway, line 165: Nice, tidies up the call-site here quite cleanly!

packages/ag-charts-community/src/scene/shape/line.ts
alantreadway: I'm not 100% convinced (especially if we also have `setProperties()` to improve readability), but let's roll with it.

alantreadway (approved):
LGTM

PR #171, opened 2023-09-06
AG-8489 Fix pie animation interrupts and reset

lsjroberts:

This commit is the actual fix https://github.com/ag-grid/ag-charts/pull/171/commits/deceadb46a4789d443b541f737e1abb0ed534a97

Bonus - cleaned up the animation manager, removed some unused functions for tweening, added a `defaultDuration()` function for that common use-case rather than having a magic number everywhere and added comments.


packages/ag-charts-community/src/chart/series/polar/pieSeries.ts
```diff
@@ -1823,7 +1826,9 @@ export class PieSeries extends PolarSeries<PieNodeDatum> {
                     sector.fillOpacity = fillOpacity;
                 },
                 onComplete: () => {
-                    if (cleanup) this.resetSectors();
+                    // TODO: Calling reset here causes a discrepancy between the data and the sectors, such that
+                    // sectors that were previously removed then shown do not get rendered.
+                    // if (cleanup) this.resetSectors();
```
lsjroberts, line 1831: Theoretically, this should be fine. I've put these "reset" functions at the end of animations mostly to cover my back in case the animation goes wrong. However in this situation, it was causing the issue. It's something I should delve deeper into at some point, but this fixes it for now.

alantreadway (approved):
LGTM

PR #176, opened 2023-09-06
AG 9153: Add basic configuration for box plot bar series

iMoses:
In addition to AG 9153 this also fixes some leftovers from AG 9024, such as flipping the axes and removing `yKey`


packages/ag-charts-community/src/scene/shape/line.ts
```diff
@@ -42,8 +42,9 @@ export class Line extends Shape {
         return new BBox(this.x1, this.y1, this.x2 - this.x1, this.y2 - this.y1);
     }
 
-    isPointInPath(_x: number, _y: number): boolean {
-        return false;
+    isPointInPath(px: number, py: number): boolean {
+        const { x, y } = this.transformPoint(px, py);
+        return this.computeBBox().grow(this.strokeWidth).containsPoint(x, y);
```
alantreadway: As discussed offline: This is incorrect for non-vertical/non-horizontal lines (diagonal line being the most pessimistic case) - but I think it's an ok compromise unless there is a straightforward solution.

We might be able to use something related to https://en.wikipedia.org/wiki/Distance_from_a_point_to_a_line#Line_defined_by_two_points (although I don't think this will tell you if the distance calculated is to a point past either end of the line)

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
```diff
@@ -79,11 +104,18 @@ export class BoxPlotSeries extends CartesianSeries<
...
     lineDashOffset: number = 0;
 
+    @Validate(OPT_FUNCTION)
+    formatter?: (params: AgBoxPlotSeriesFormatterParams<BoxPlotNodeDatum>) => AgBoxPlotSeriesFormat = undefined;
+
+    cap = new BoxPlotSeriesCap();
+
+    whisker = new BoxPlotSeriesWhisker();
+
     tooltip = new SeriesTooltip<AgCartesianSeriesTooltipRendererParams>();
```
alantreadway: Nit: Just wondering, could/should these properties be `readonly` if the reference never changes?

packages/ag-charts-community/src/chart/series/cartesian/boxPlotOptions.ts
```diff
@@ -32,14 +22,59 @@ export interface AgBoxPlotSeriesOptions<DatumType = any>
...
+export interface AgBoxPlotCapOptions {
+    lengthRatio?: Ratio;
+}
 
-    /** Configuration for the labels shown on columns. */
-    // label?: AgColumnSeriesLabelOptions;
+export type AgBoxPlotWhiskerOptions = StrokeOptions & LineDashOptions;
+
+export interface BoxPlotNodeDatum
+    extends Omit<CartesianSeriesNodeDatum, 'yKey'>,
```
alantreadway: Nit: Maybe move the `Omit<...>` to the end, as at first glance I was trying to parse whether the following extends clauses were parameters to that `Omit`?

alantreadway (commented):
Very nice, just a few minor points.

packages/ag-charts-community/src/scene/shape/line.ts
iMoses: oh, didn't think about that..
I initially tried this solution, but it didn't work because I didn't know yet about `transformPoint`..
I can give it another try, but I'm wondering, would it be more efficient to use the existing logic for straight lines?
we could do a simple check if the x's or y's are identical

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
iMoses: np, I was merely following existing code examples
so shouldn't this also be the case for `tooltip`

packages/ag-charts-community/src/chart/series/cartesian/boxPlotOptions.ts
iMoses: sure, if you think it helps readability

packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
alantreadway: Yes, SGTM for `tooltip` :-)

alantreadway (approved):
LGTM

packages/ag-charts-community/src/scene/shape/line.ts
alantreadway: LGTM now - thanks!

PR #177, opened 2023-09-06
AG-8646 Fix warning when no axes

alexanderby:


packages/ag-charts-community/src/chart/series/series.ts
```diff
@@ -311,6 +311,8 @@ export abstract class Series<C extends SeriesNodeDataContext = SeriesNodeDataCon
     @Validate(STRING)
     readonly id = createId(this);
 
+    readonly canHaveAxes: boolean = false;
```
alantreadway: Having property overrides on sub-classes is a bit of a smell - let's assign this in the constructor (the field can still be `readonly` if there is exactly one assignment in the constructor) and pass it in as an options argument?

alantreadway (approved):
LGTM

PR #179, opened 2023-09-07
AG-9206 - Options reference round 3 - Add scroll to property

taktran:
## Changes

* Show top level items on no selection view ie, default when landing on options reference page
  * Make `options` at the top clickable and show the no selection view
  * For `axes` and `series` only show 1 level deep
* Instead of filtering out selection, display all properties and scroll to value
* Show `type` in the navigation when expanded

## TODO

* [ ] Theme heading references don't work - nav uses `#reference-theme--1--baseTheme` and heading id is `reference--1--baseTheme` (missing theme)

## Screenshots

No selection or `options` clicked view shows top level properties

Screenshot 2023-09-08 at 5 26 53 pm

Show type in nav
Screenshot 2023-09-08 at 5 24 19 pm

Only show 1 level deep for `axes` and `series`
Screenshot 2023-09-08 at 5 24 43 pm


markdurrant (approved):
LGTM 👍 👍 👍

PR #180, opened 2023-09-07
AG-9212 - Improve theme typings

alantreadway:

Removes:
- `theme.overrides.cartesian`
- `theme.overrides.polar`
- `theme.overrides.heirarchy`

Adjusts:
- `theme.overrides.[series-type].series.[series-type]` (removes repetition of series types).
- Aligns label colours across series/legend/axes.

Compromises:
- Unfortunately out TS metadata generator can't handle much beyond basic `interface` types, and some simple `Omit`, `Required`, `Partial` cases - so there is still a lot more manual repetition than I would like, but fixing that is a much more involved task.

Reduces generated theme config output by removing the repeat of series configs under series-specific sections (e.g. `bar.series.bar` is now `bar.series`, and `bar.series.area` no longer exists - instead `area.series` should be used).

NOTE: Some niggles are still present (seeing some console warnings on options application), I will try to work these through a bit more before merge.


packages/ag-charts-community/src/chart/options/themeOptions.ts
```diff
@@ -48,30 +39,24 @@ export interface AgBaseChartTheme<T> extends AgBaseChartThemeOptions<T> {
...
-    treemap?: AgBaseHierarchyThemeOptions<AgHierarchySeriesOptions> & T;
-
-    /** Specifies defaults for all chart types. Be careful to only use properties that apply to all chart types here. For example, don't specify `navigator` configuration here as navigators are only available in cartesian charts. */
-    common?: any;
-}
+type SeriesThemeSkipOptions = 'id' | 'xKey' | 'yKey' | 'type';
+type SeriesThemeOptions<Series, Type> = Omit<Partial<Extract<Series, { type?: Type }>>, SeriesThemeSkipOptions>;
+
+export type AgBaseChartThemeOverrides<T> = {
+    [K in NonNullable<AgCartesianSeriesOptions['type']>]?: AgBaseCartesianThemeOptions<
```
manapeirov: what if there are overlaps with the  series options?

manapeirov (commented):
Looks like the default font colors for the axis labels and legend labels have changed

alantreadway:
> Looks like the default font colors for the axis labels and legend labels have changed

Yes, so there is now better consistency across all labels - I think that makes sense?

Element | Before | After
-------- | ------- | ------
`axes[].label.color` | `rgb(87, 87, 87)` | `rgb(70, 70, 70)` (standard label colour)
`legend.item.label.color` | `black` |  `rgb(70, 70, 70)` (standard label colour)

olegat (commented):
LGTM. No concerns on my end but someone more familiar with the code should approve.

alexanderby (commented):
The only issue seems to be the legend text colour.

packages/ag-charts-community/src/chart/series/cartesian/rangeBarOptions.ts
```diff
@@ -119,6 +98,32 @@ export interface AgRangeBarSeriesOptions<DatumType = any> extends AgBaseSeriesOp
...
+    /** A human-readable description of the x-values. If supplied, this will be shown in the default tooltip and passed to the tooltip renderer as one of the parameters. */
+    xName?: string;
+    /** A human-readable description of the y-low-values. If supplied, this will be shown in the default tooltip and passed to the tooltip renderer as one of the parameters. */
+    yLowName?: string;
+    /** A human-readable description of the y-high-values. If supplied, this will be shown in the default tooltip and passed to the tooltip renderer as one of the parameters. */
+    yHighName?: string;
+    /** A human-readable description of the y-values. If supplied, this will be shown in the default tooltip and passed to the tooltip renderer as one of the parameters. */
+    yName?: string;
+    /** The title to use for the series. Defaults to `yName` if it exists, or `yKey` if not. */
+    title?: string;
```
manapeirov: I think title can be removed it's not used

packages/ag-charts-community/src/chart/series/seriesOptions.ts
```diff
@@ -27,7 +27,18 @@ export interface AgSeriesHighlightStyle {
     series?: AgSeriesHighlightSeriesStyle;
 }
 
-export interface AgBaseSeriesOptions<DatumType> {
+export interface AgBaseSeriesThemeableOptions {
+    /** Whether to include the series in the legend. */
+    showInLegend?: boolean;
+    /** The cursor to use for hovered area markers. This config is identical to the CSS `cursor` property. */
```
manapeirov: Looks good but now some of these comments don't apply to all series types so need to be made generic

manapeirov (approved):

alantreadway:
> The only issue seems to be the legend text colour.

See https://github.com/ag-grid/ag-charts/pull/180#issuecomment-1710193113 - as there is parallel work with Mark and Mana to rework theme colours completely, this is acceptable (I've ran it past Rob).

packages/ag-charts-community/src/chart/series/seriesOptions.ts
alantreadway: Fixed

packages/ag-charts-community/src/chart/options/themeOptions.ts
alantreadway: Discussed offline - there's no overlap.

packages/ag-charts-community/src/chart/series/cartesian/rangeBarOptions.ts
alantreadway: Fixed

PR #186, opened 2023-09-08
AG-9252 Polar cross lines docs

alexanderby:


packages/ag-charts-website/src/content/docs/polar-axes/_examples/polar-axes-crosslines-angle/main.ts
```diff
@@ -0,0 +1,48 @@
...
+    axes: [
+        {
+            type: 'angle-category',
+            shape: 'circle',
+            crossLines: [
+                {
+                    type: 'line',
+                    value: 'Meeting Deadlines',
+                    label: {
+                        text: 'Useless Skill',
```
alantreadway, line 29: I think this might be somewhat controversial, some would argue being able to manage time, expectations and scope is a vital life skill :D

alantreadway (commented):
LGTM

PR #187, opened 2023-09-08
AG-9015 - Add basic 'Hello World' implementation error bars

olegat:
https://ag-grid.atlassian.net/browse/AG-9015

This is the basic initial 'Hello World' implementation of error bars. This initial implementation is built with the intent of not relying on `CartesianSeries` too much to facilitate migration into the enterprise module later on.

This implementation has many limitations:

- A hardcoded placeholder styling is used.
- No caps are drawn (just the whiskers).
- Only the y-axis supports error bars.
- The y-axis domain is not extended to fit error bar data values. Values outside the domain are converted into `NaN` pixel coordinates and are therefore not drawn.
- All error bars are part of one group (no subgrouping).

### DIFF
--- a/packages/ag-charts-community/src/chart/errorBar.ts
+++ b/packages/ag-charts-community/src/chart/errorBar.ts
@@ -1,0 +1,49 @@
+import { OPT_STRING, Validate } from '../util/validation';
+
+import { Path } from '../scene/shape/path';
+import type { Point } from '../scene/point';
+
+export interface ErrorBarDatum {
+    readonly yLowerValue: any; // TODO(olegat) this should be a "continuous" type
+    readonly yUpperValue: any;
+    readonly yLowerPoint: Point;
+    readonly yUpperPoint: Point;
+}
+
+export class ErrorBarConfig {
+    @Validate(OPT_STRING)
+    yLowerKey: string = '';
+
+    @Validate(OPT_STRING)
+    yLowerName?: string = undefined;
+
+    @Validate(OPT_STRING)
+    yUpperKey: string = '';
+
+    @Validate(OPT_STRING)
+    yUpperName?: string = undefined;
+
+    constructor(yLowerKey: string, yUpperKey: string) {
+        this.yLowerKey = yLowerKey;
+        this.yUpperKey = yUpperKey;
+    }
+}
+
+export class ErrorBar extends Path {
+    updatePath() {
+        // TODO(olegat) implement, this is just placeholder draw code
+        const { path } = this;
+        this.fill = 'black';
+        this.stroke = 'black';
+        this.strokeWidth = 1;
+        this.fillOpacity = 1;
+        this.strokeOpacity = 1;
+
+        const { yLowerPoint, yUpperPoint } = this.datum as ErrorBarDatum;
+
+        path.clear();
+        path.moveTo(yLowerPoint.x, yLowerPoint.y);
+        path.lineTo(yUpperPoint.x, yUpperPoint.y);
+        path.closePath();
+    }
+}
--- a/packages/ag-charts-community/src/chart/errorBar.ts
+++ b/packages/ag-charts-community/src/chart/errorBar.ts
@@ -1,7 +1,11 @@
-import { OPT_STRING, Validate } from '../util/validation';
-
-import { Path } from '../scene/shape/path';
+import type { Scale } from '../scale/scale';
+import { Group } from '../scene/group';
+import type { Node } from '../scene/node';
 import type { Point } from '../scene/point';
+import { Selection } from '../scene/selection';
+import { Path } from '../scene/shape/path';
+import { OPT_STRING, Validate } from '../util/validation';
+import type { ProcessedData } from './data/dataModel';
 
 export interface ErrorBarPoints {
     readonly yLowerPoint: Point;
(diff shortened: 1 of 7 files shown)


packages/ag-charts-community/src/chart/series/cartesian/lineSeries.ts
```diff
@@ -59,6 +59,26 @@ class LineSeriesLabel extends Label {
 type LineContext = SeriesNodeDataContext<LineNodeDatum>;
 type LineAnimationData = CartesianAnimationData<LineContext>;
 
+// TODO(olegat) move somewhere else?
```
olegat: Please advise whether these kind of `// TODO` comments are acceptable. If not, then I'll make a note of this somewhere else.

alantreadway: We don't typically use these - there is a JIRA plugin which adds checklist support, so for smaller things like this I would usually capture them there, assuming you are going to resolve this in the scope of your current ticket?

packages/ag-charts-community/src/chart/series/cartesian/lineSeries.ts
```diff
@@ -59,6 +59,26 @@ class LineSeriesLabel extends Label {
...
+export class ErrorBarDatum {
```
alantreadway: Nit: We usually reserve `Datum` as a term for actual data - I would probably just run with `ErrorBar` or `ErrorBarConfig` perhaps?

packages/ag-charts-community/src/chart/series/cartesian/lineSeries.ts
```diff
@@ -59,6 +59,26 @@ class LineSeriesLabel extends Label {
...
+    @Validate(OPT_STRING)
+    yLowerKey: string;
```
alantreadway: A quirk/smell of TS->JS is that if a field is not assigned a value, it doesn't exist at runtime - typically you'll see that fields that are `@Validate()`'ed also assign an initial value, even if that is `undefined` - this allows us to sanity check that options provided by users actually exist before assignment from the users options, rather than just blinding adding.

packages/ag-charts-community/src/chart/series/cartesian/lineSeries.ts
```diff
@@ -118,6 +138,8 @@ export class LineSeries extends CartesianSeries<LineContext> {
     @Validate(OPT_STRING)
     yName?: string = undefined;
 
+    errorBar?: ErrorBarDatum = undefined;
```
alantreadway: You may need to assign an instance here, otherwise the options assignment code won't know how to construct it - once we extract to an enterprise module, the enterprise module intiialisation will take care of this step intrinsically though.

packages/ag-charts-community/src/chart/series/cartesian/lineSeries.ts
olegat: Thanks for the explanation. That makes sense. What values should be used?

I'm thinking of using an empty string `''` for `yLowerKey` and `yUpperKey` because these properties are required, and `undefined` for `yLowerName` and `yUpperName` because these are optional.

packages/ag-charts-community/src/chart/series/cartesian/lineSeries.ts
olegat: ACK. I'll use `ErrorBarConfig`

packages/ag-charts-community/src/chart/series/cartesian/lineSeries.ts
olegat: If I assign this to a `new ErrorBarConfig()`, how will the `LineSeries` class know whether or not the `'errorBars'` property was given in the input JSON?

packages/ag-charts-community/src/chart/series/cartesian/lineSeries.ts
olegat: Understood. I'll remove this comment.

packages/ag-charts-community/src/chart/errorBar.ts
```diff
@@ -0,0 +1,125 @@
...
+        path.moveTo(yLowerPoint.x, yLowerPoint.y);
+        path.lineTo(yUpperPoint.x, yUpperPoint.y);
+        path.closePath();
+    }
+}
+
+export class ErrorBars {
+    public groupNode?: Group = undefined;
+    private selection?: Selection<ErrorBarNode> = undefined;
+    private nodeData: (ErrorBarPoints | undefined)[] = new Array(0);
```
alantreadway: Nit: `[]` is sufficient/convention for empty arrays.

packages/ag-charts-community/src/chart/errorBar.ts
```diff
@@ -0,0 +1,125 @@
...
+        path.clear();
+        path.moveTo(yLowerPoint.x, yLowerPoint.y);
+        path.lineTo(yUpperPoint.x, yUpperPoint.y);
+        path.closePath();
+    }
+}
+
+export class ErrorBars {
+    public groupNode?: Group = undefined;
+    private selection?: Selection<ErrorBarNode> = undefined;
```
alantreadway: Suggestion: If this field is initalised once in the constructor, you don't need to mark it as optional or assign an initial value inline here. You can also mark this field `readonly` if the reference never changes.

packages/ag-charts-community/src/chart/errorBar.ts
```diff
@@ -0,0 +1,125 @@
...
+        if (this.groupNode === undefined) {
+            this.groupNode = new Group({ name: `${parent.id}-series-errorBars` });
+            parent.appendChild(this.groupNode);
+        }
+
+        this.selection = Selection.select(this.groupNode, () => this.errorBarFactory());
+    }
+
+    update() {
+        if (this.selection != undefined) {
```
alantreadway: This guard should be redundant if you can apply my suggestion regarding the `selection` field above 🤞

alantreadway (commented):
Some suggestions, but this looks like a reasonable first pass :-)

packages/ag-charts-community/src/chart/errorBar.ts
```diff
@@ -0,0 +1,125 @@
...
+        xIndex?: number,
+        _yIndex?: number, // This is will be used when the scatterplot errorbars are implemented
+        xScale?: Scale<any, any, any>,
+        yScale?: Scale<any, any, any>,
+        config?: ErrorBarConfig
+    ) {
+        this.selection = undefined;
+
+        const { nodeData } = this;
+        const { yLowerKey = undefined, yUpperKey = undefined } = config ?? {};
```
iMoses:
```suggestion
        const { yLowerKey, yUpperKey } = config ?? {};
```
nit: it's redundant to assign `undefined` when spreading, that's the default anyhow

packages/ag-charts-community/src/chart/errorBar.ts
olegat: Done.

packages/ag-charts-community/src/chart/errorBar.ts
olegat: Done. I've removed the `parent: Node` parameter of `createNodeData` and made it a parameter of the constructor.

packages/ag-charts-community/src/chart/errorBar.ts
olegat: Done.

packages/ag-charts-community/src/chart/errorBar.ts
olegat: Done.

alantreadway (approved):
LGTM - thanks for making the updates!

PR #188, opened 2023-09-08
AG-9150: Add box-plot-column series

iMoses:
image


packages/ag-charts-enterprise/src/box-plot/boxPlotGroup.ts
```diff
@@ -77,45 +77,84 @@ export class BoxPlotGroup extends _Scene.Group {
...
-
-        whiskers[1].setProperties({
-            x1: q3Value,
-            x2: Math.round(maxValue - whiskerProperties.strokeWidth / 2),
-            y: axisValue + bandwidth / 2,
-        });
+        const capStart = axisValue + (bandwidth * (1 - cap.lengthRatio)) / 2;
+        const capEnd = axisValue + (bandwidth * (1 + cap.lengthRatio)) / 2;
+
+        if (invertAxes) {
```
alantreadway, line 83: Is there any way to avoid repeating both paths here?

E.g. could something like this reduce the repetition?

```
const categoryProp = invertAxes ? 'x' : 'y';
const valueProp = invertAxes ? 'y' : 'x';
const categoryWidthProp = invertAxes ? 'width' : 'height';
const valueDepthProp = invertAxes ? 'height' : 'width';

....

boxes[0].setProperties({
  [categoryProp]: axisValue,
  [valueProp]: q3Value,
  [categoryWidthProp]: bandwidth,
  [valueDepthProp]: Math.round(medianValue - q3Value + strokeWidth / 2),
});

```

iMoses: It's actually a problem because it isn't as simple as switching the axes, some of the math is also inverted because Y starts at the top left corner, unlike what the user expects which is the bottom left

PR #189, opened 2023-09-09
Enable Enterprise animation tests + add basic initial load case coverage.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-9281

Enables and adds basic test coverage of animations (initial load case) for enterprise series types.

Fixes:
- Ensure that both community and enterprise code is compiled by the same Jest compiler for `ag-charts-enterprise`.


manapeirov (approved):
LGTM, thank you

PR #190, opened 2023-09-11
Bug Fixes: AG-9024 / AG-9153 / AG-9026

iMoses:
- Skip item render if data is malformed/missing
- Fix line BBox in case of using a reverse direction


alantreadway (approved):
LGTM

packages/ag-charts-community/src/util/types.ts
```diff
@@ -1 +1,19 @@
 export type Has<P extends keyof T, T> = T & { [K in P]-?: T[P] };
+
+export type Mutable<T> = T extends object ? { -readonly [K in keyof T]: Mutable<T[K]> } : T;
+
+export type Defined<T> = T extends undefined ? never : T;
+
+export type DeepRequired<T> = T extends (...args: any[]) => any
```
alantreadway, line 7: Side-note: I'm a little skeptical about how well this will work, TS (3.x from memory) has historically had major issues with cyclic types like this - but maybe they have addressed this in more recent versions?

alantreadway, line 7: Side-note: I'm a little skeptical about how well this will work, TS (3.x from memory) has historically had major issues with cyclic types like this - but maybe they have addressed this in more recent versions?

alantreadway (approved):
LGTM

packages/ag-charts-community/src/util/types.ts
iMoses: I only started working with TS about a year ago, version 4, and I haven't encountered any issues with it.
But FYI, I have encountered some issues with built-in types and our interface extraction code for the docs. e.g. An interface extending `Array` crushed the script but a type declaration extending it didn't.
I'm more worried about how we handle complex type than how TS is handling it 🙃

PR #192, opened 2023-09-11
AG-9206 - Options reference round 4 - Adding theme link

taktran:
## Changes

* Add themes reference page
* Remove `theme` object from options reference replace with a top level link to the themes reference page
* Extract `framework`, `config` and `optionsData` as context that can get passed into react components


markdurrant (approved):
👍

PR #193, opened 2023-09-12
AG-8930 - Add automatic package version bump

alantreadway:
https://ag-grid.atlassian.net/browse/AG-8930

Adds support for automatic package version bumping.


tools/calculate-next-version.js
seanlandsman, line 1: So I think this is all fine but now I think about it we should probably add a time element too - I can see multiple versions being released on a single day if an issue is found

seanlandsman (approved):
Looks fine overall - I would however add a time element (or some other suffix) to allow for multiple versions being released on the same day

If you don't believe that this is necessary then it looks fine as is

tools/calculate-next-version.js
alantreadway: Yup, I can change to time, for now line 28 is adding a `.<index>` suffix if the current version is using an identical timestamp?

PR #194, opened 2023-09-12
AG-7600 Move Gradient Legend to options root

alexanderby:


packages/ag-charts-community/src/chart/agChartV2.ts
```diff
@@ -396,8 +397,12 @@ function applyChartOptions(chart: Chart, processedOptions: ProcessedOptions, use
 
     const seriesOpts = processedOptions.series as any[];
     const seriesDataUpdate = !!processedOptions.data || seriesOpts?.some((s) => s.data != null);
-    const otherRefreshUpdate = processedOptions.legend ?? processedOptions.title ?? processedOptions.subtitle;
-    forceNodeDataRefresh = forceNodeDataRefresh || seriesDataUpdate || !!otherRefreshUpdate;
+    const legendKeys = getLegendKeys();
+    const optionsHaveLegend = Object.values(legendKeys).some(
+        (legendKey) => (processedOptions as any)[legendKey] != undefined
+    );
+    const otherRefreshUpdate = Boolean(processedOptions.title ?? processedOptions.subtitle);
```
alantreadway: Nit: Can we just use a straightforward expression rather than relying on Truthy evaluation? e.g.:
```
const otherRefreshUpdate = processedOptions.title != null || processedOptions.subtitle != null;
```

packages/ag-charts-community/src/chart/agChartV2.ts
```diff
@@ -425,20 +430,36 @@ function applyModules(chart: Chart, options: AgChartOptions) {
...
-
-        if (shouldBeEnabled === isEnabled) continue;
-        modulesChanged = true;
-
-        if (shouldBeEnabled) {
-            chart.addModule(next);
-        } else {
-            chart.removeModule(next);
+    const processModules = <T extends Module<ModuleInstance>>(
+        moduleType: string,
```
alantreadway: Is there a stricter type we can use here rather than `string`? Would `Module['type']` work?

packages/ag-charts-community/src/chart/agChartV2.ts
```diff
@@ -507,10 +528,10 @@ function applyAxes(chart: Chart, options: { axes?: AgBaseAxisOptions[] }) {
...
-    chart.setLegendInit((legend) => {
-        applyOptionValues(legend, options.legend ?? {}, { skip });
-        if (options.legend?.listeners) {
-            Object.assign(chart.legend?.listeners!, options.legend.listeners ?? {});
+    chart.setLegendInit((legendKey, legend) => {
+        applyOptionValues(legend, (options as any)[legendKey] ?? {}, { skip });
+        if (chart.legend && options.legend?.listeners) {
+            Object.assign(chart.legend.listeners!, options.legend.listeners);
         }
     });
```
alantreadway: Do we need to use a callback here? Since the distinct legend types are now on distinct property paths, can't they just be assigned options directly from `AgChartV2` rather than defering to later?

packages/ag-charts-enterprise/src/gradient-legend/gradientLegendThemes.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const BOTTOM: AgChartLegendPosition = 'bottom';
+
+export const GRADIENT_LEGEND_THEME = {
+    position: BOTTOM,
+    spacing: 20,
+    stop: {
+        padding: 8,
+        label: {
+            color: 'rgb(70, 70, 70)',
```
alantreadway: Nit: Use `_Theme.DEFAULT_LABEL_COLOUR` here.

packages/ag-charts-community/src/chart/series/series.ts
```diff
@@ -740,7 +740,7 @@ export abstract class Series<C extends SeriesNodeDataContext = SeriesNodeDataCon
         return new SeriesNodeDoubleClickEvent(event, datum, this);
     }
 
-    abstract getLegendData(): ChartLegendDatum[];
+    abstract getLegendData(legendType: string): ChartLegendDatum[];
```
alantreadway: Can we properly restrict the type of `legendType` please?

alantreadway:
NOTE: Conflicts after the camelCase directory rename.

packages/ag-charts-community/src/chart/agChartV2.ts
```diff
@@ -397,16 +398,21 @@ function applyChartOptions(chart: Chart, processedOptions: ProcessedOptions, use
...
-    applyLegend(chart, processedOptions);
 
     const seriesOpts = processedOptions.series as any[];
     const seriesDataUpdate = !!processedOptions.data || seriesOpts?.some((s) => s.data != null);
-    const otherRefreshUpdate = processedOptions.legend ?? processedOptions.title ?? processedOptions.subtitle;
-    forceNodeDataRefresh = forceNodeDataRefresh || seriesDataUpdate || !!otherRefreshUpdate;
+    const legendKeys = getLegendKeys();
+    const optionsHaveLegend = Object.values(legendKeys).some(
+        (legendKey) => (processedOptions as any)[legendKey] != null
+    );
```
alantreadway, line 412: Just to confirm, this condition is only true when `legend` or it's child options get changed, not when `legend` is present but unchanged?

packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
```diff
@@ -645,17 +645,17 @@ export class AreaSeries extends CartesianSeries<AreaSeriesNodeDataContext> {
...
-        if (!data?.length || !xKey || !yKey) {
+        if (!data?.length || !xKey || !yKey || legendType !== 'category') {
             return [];
         }
 
         // Area stacks should be listed in the legend in reverse order, for symmetry with the
         // vertical stack display order.
         return [
-            {
+            <CategoryLegendDatum>{
```
alantreadway: Nit: Prefer `as CategoryLegendDatum` syntax (it's equivalent)?

alantreadway: Alternatively, I believe there is some type-safety lost with any cast like this.

It might be better to follow this pattern, which ensures that the inline object conforms to the interface:
```
const legendDatun: CategoryLegendDatum = {
  legendType: 'category',
  id,
  // ...etc....
};
```

alantreadway (commented):
Looking great, just a few niggles left :-)

packages/ag-charts-community/src/chart/agChartV2.ts
alexanderby: `legendKeys` means not the properties of the legend, but `legend`, `gradientLegend` option keys. The previous condition was:
```ts
const otherRefreshUpdate = processedOptions.gradientLegend ?? processedOptions.legend ?? processedOptions.title ?? processedOptions.subtitle;
forceNodeDataRefresh = forceNodeDataRefresh || seriesDataUpdate || !!otherRefreshUpdate;
```

packages/ag-charts-community/src/options/options/gradientLegendOptions.ts
alantreadway, line 1: NOTE: Can you check that `gradientLegend` appears in all the right places in the theme overrides?

alantreadway (approved):
LGTM - please notice my note above though.

PR #195, opened 2023-09-12
AG-8664 - Fix axis fallback to defaults not being themed.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-8864

Fixes application of axis themes to the default axis configuration when the user supplies an invalid `axes` config.


alexanderby (approved):
🎉

PR #196, opened 2023-09-12
Add lint rules to enforce kebab case for community + enterprise directories.

alantreadway:
Fixes some camel-case directory names too.


packages/ag-charts-community/.eslintrc.json
```diff
@@ -7,7 +7,21 @@
...
       "files": ["*.ts", "*.tsx", "*.js", "*.jsx"],
-      "rules": {}
+      "rules": {
+        // "check-file/filename-naming-convention": [
+        //   "error",
+        //   {
+        //     "**/*.{jsx,tsx}": "CAMEL_CASE",
+        //     "**/*.{js,ts}": "CAMEL_CASE"
+        //   }
+        // ],
```
lsjroberts, line 17: remove?

alexanderby (approved):

packages/ag-charts-community/.eslintrc.json
alantreadway: I want to enable these, there is just too many counter-case at the moment ) which we need to review).

olegat: nit: It might be possible to exclude these counter cases with a more restrictive rule than `**/*.{js.ts}`. Do they share a common pattern?

PR #198, opened 2023-09-13
AG-9024 - Add series clipping configuration and better default enablement

alantreadway:
https://ag-grid.atlassian.net/browse/AG-9024

Intent:
- Enable correct zoomed/constrained series rendering (series rendering should be clipped, rather than skipping or clamping datums).

Changes:
- There is now a `seriesArea` root object, with `clip` and `padding` sub-options to allow user override.
- I've removed the concept of `strict` from scales, and replaced with `mode: 'raw' | 'clamped'` - and we're using `raw` in all cases except a few edge cases for crosslines.
  - This should mean that when zooming all series render correctly as we're using `seriesRect` clipping rather than removal of the scene graph nodes.
  - There maybe a performance impact for larger data-sets, but it maybe a premature optimisation to deal with that now.
  - I've fixed several series implementation to NOT perform additional range checking, and removed redundant code around this aspect.
  - ***NOTE*** Due to complexities of getting `LogScale` to behave consistently, I've defaulted this to `'clamp'` behaviour for it.
- Improved series clipping auto-enablement to take axis `dataDomain` clipping into account - if axes have `min` and/or `max` set AND this reduces the visible data-set, we enable clipping.


packages/ag-charts-community/src/chart/series/cartesian/__image_snapshots__/area-series-test-ts-area-series-create-for-area-category-x-axis-fractional-log-y-axis-it-should-render-to-canvas-as-expected-1-snap.png
alantreadway, line 1: I'm aware of the issues here, I'm working on these.

packages/ag-charts-enterprise/src/polar-axes/radius-number/radiusNumberAxis.ts
```diff
@@ -202,7 +202,9 @@ export class RadiusNumberAxis extends _ModuleSupport.PolarAxis {
 
     normaliseDataDomain(d: number[]) {
         const { min, max } = this;
-        return normalisedExtent(d, min, max);
+        const domain = normalisedExtent(d, min, max);
+
+        return { domain, clipped: domain.clipped };
```
iMoses: Why in some of the places you create a copy of the domain array and in others (such as this) you don't?

packages/ag-charts-community/src/util/array.ts
```diff
@@ -29,18 +29,22 @@ export function extent(values: Array<number | Date>): [number, number] | undefin
...
     }
     if (!isNaN(max)) {
+        clipped ||= max < d[1];
         d = [d[0], max];
     }
     if (d[0] > d[1]) {
         d = [];
     }
-    return d;
+    return Object.assign(d, { clipped });
```
iMoses: should we really extend a native array to hold our property?
I think it's better if we return an object with `{ domain, clipped }`

packages/ag-charts-community/src/scale/scale.ts
```diff
@@ -1,9 +1,5 @@
-interface ScaleClampParams {
-    /**
-     * If `true` the values outside of the domain will become `NaN`.
-     * If `false` such values will be clamped to the domain edges.
-     */
-    strict: boolean;
+export interface ScaleClampParams {
+    mode: 'clamped' | 'raw';
```
iMoses: nit: for future proofing we may wanna make the interface more generic, e.g. `ScaleConvertParams` & `clampMode`

packages/ag-charts-community/src/scale/continuousScale.ts
```diff
@@ -80,6 +82,11 @@ export abstract class ContinuousScale<D extends number | Date, I = number> imple
         );
     }
 
+    /** Override point for special convert() handling by sub-classes. */
+    protected specialConvert(x: D, _domainStart: D, _domainEnd: D) {
```
iMoses: is this new? I'm not seeing it used anywhere (maybe I missed it)

alantreadway: Good spot - removed, this was a hang-over from my battle with `LogScale` :P

packages/ag-charts-community/src/scale/scale.ts
alantreadway: They are distinct concepts in my mind so wouldn't want to conflate them, but maybe I'm mis-understanding your suggestion?

packages/ag-charts-community/src/util/array.ts
alantreadway: That's fair, my original intent was to reduce impact of the addition by keeping a compatible return type - I've adjusted the approach here to be cleaner.

packages/ag-charts-enterprise/src/polar-axes/radius-number/radiusNumberAxis.ts
alantreadway: The need for the spread is now redundant, so I've cleaned this up.

packages/ag-charts-community/src/scale/scale.ts
iMoses: I'll try to clarify my suggestion.
Today the scale covert method signature is: `convert(value: D, params?: ScaleClampParams): R;`.
We explicitly declare the params of the covert method as being clamp params, but in a broader sense these are the convert method params which currently, for the time being, contain only the clamp mode option, but that may change in the future and hold other params which relate to the convert method, not necessarily clamp related. 
that's why I'm suggesting merely changing the names to make them more generic in advance. I also believe this reflect the intention better, so if tomorrow someone need another option they'll know in goes in those params and there's no need for yet another argument for it.
So instead of `ScaleClampParams` we could call it `ScaleConvertParams`, because these are params for the scale convert method, and instead of simply `mode` we could call it `clampMode` so the context will remain clear, that way we could still use the same params object to pass more options even if they're unrelated to clamping.

iMoses (approved):
My nit picking aside, LGTM 👍

packages/ag-charts-community/src/scale/scale.ts
alantreadway: Ok - I've renamed.

PR #200, opened 2023-09-13
Upgrade Nx to 16.8.1.

alantreadway:


package.json
```diff
@@ -127,6 +127,7 @@
       "packages/ag-charts-vue3",
       "packages/ag-charts-website",
       "packages/ag-charts-website-e2e",
+      "libraries/ag-charts-community-test",
```
olegat: It looks like this directory does not exist. Is that deliberate? Or are there missing files in this pull request?

alantreadway: My mistake, let me remove this - thanks for spotting this!

alantreadway: Fixed

olegat (approved):

PR #202, opened 2023-09-13
Ag 9215/waterfall series animation

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-9215


lsjroberts (approved):
LGTM, good spot on that updated indices bug

PR #204, opened 2023-09-13
AG-9159: Series highlighting for box-plot

iMoses:
Although we talked about a general solution I found a simpler local solution which I think will do just fine for now


packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
```diff
@@ -394,13 +394,18 @@ export class BoxPlotSeries extends CartesianSeries<
         return false;
     }
 
+    private highlightedIds: string[] = [];
```
alantreadway: Rather than keeping this state hanging around, could we route the selected item into `updateDatumSelection()` by modifying `CartesianSeries.updateSelections()` to pass the determined highlighted item from the call on line 275 into the call on line 293?
https://github.com/ag-grid/ag-grid/blob/663dc2e643b46026d5784615f351ff16fed0f2e6/charts-community-modules/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts#L274

iMoses: let's have a chat about it. because I think it's a bit more complicated than that.
Notice I've been changing the opacity of the node instead of removing it because I've come to find that all the events are attached to the original item and not the highlighted item that goes over it.

alantreadway (approved):
LGTM

PR #206, opened 2023-09-14
AG-9026: minor fix of default legend labels

iMoses:
+ small adjustment to try and avoid AA issues

### DIFF
--- a/packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
+++ b/packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
@@ -358,7 +358,7 @@ export class BoxPlotSeries extends CartesianSeries<
             seriesId: id,
             enabled: visible,
             label: {
-                text: legendItemName ?? yName ?? id,
+                text: legendItemName ?? yName ?? `Box Plot #${id.substring(id.lastIndexOf('-') + 1)}`,
             },
             legendItemName,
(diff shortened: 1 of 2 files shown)


packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts
```diff
@@ -358,7 +358,7 @@ export class BoxPlotSeries extends CartesianSeries<
             seriesId: id,
             enabled: visible,
             label: {
-                text: legendItemName ?? yName ?? id,
+                text: legendItemName ?? yName ?? `Box Plot #${id.substring(id.lastIndexOf('-') + 1)}`,
```
alantreadway, line 361: I would follow the convention of falliing back to `yKey` here rather than trying to parse an index out of a string (see bar, area, line, scatter and radar series?).

iMoses: There's no `yKey` in a box plot series 😅

iMoses: I think our options are either extract the index from the `id` and add it to the default name or simply have a static name as a default. Which one do you think is better?

iMoses:
After a discussion with Rob decided to keep the `id` as the default

PR #208, opened 2023-09-14
AG-8489 pie animations

lsjroberts:

### DIFF
--- a/packages/ag-charts-community/src/chart/data/processors.ts
+++ b/packages/ag-charts-community/src/chart/data/processors.ts
@@ -297,8 +297,7 @@ export function diff(
                 if (added.has(prevId)) {
                     if (updateMovedDatums || !arraysEqual(added.get(prevId).values, prev.values)) {
                         updated.set(prevId, prev);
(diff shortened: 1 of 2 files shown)


packages/ag-charts-community/src/chart/data/processors.ts
```diff
@@ -297,8 +297,7 @@ export function diff(
                 if (added.has(prevId)) {
                     if (updateMovedDatums || !arraysEqual(added.get(prevId).values, prev.values)) {
                         updated.set(prevId, prev);
-                        const updatedIndex = added.get(prevId);
-                        updatedIndices.set(prevId, updatedIndex);
+                        updatedIndices.set(prevId, i);
```
lsjroberts, line 300: While waterfall animations are pending, reverting this fixes an issue with the pie sectors which expect the other index here https://github.com/ag-grid/ag-charts/blob/latest/packages/ag-charts-community/src/chart/series/polar/pieSeries.ts#L1772-L1774

We should re-evaluate tracking both the before and after index later.

alantreadway (approved):
LGTM

PR #212, opened 2023-09-14
Isolate options interfaces from chart implementation modules

alantreadway:
As a first step to breaking up `ag-charts-community` dependencies, this PR:
- moves all option interface contracts into their own sub-folder.
- enforces isolation using `dependency-cruiser` rules.

See https://github.com/sverweij/dependency-cruiser


packages/ag-charts-community/.dependency-cruiser.js
```diff
@@ -0,0 +1,443 @@
...
+                dependencyTypes: ['npm-peer'],
+            },
+        },
+        {
+            name: 'ag-isolated-options',
+            comment: 'Options modules should be isolated from implementation modules.',
+            severity: 'error',
+            from: { path: 'src/options' },
+            to: { pathNot: '^src/options' },
+        },
```
alantreadway, line 183: FYI: The rest of the config here is boilerplate, this is the section I added for our particular use-case.

iMoses (approved):

PR #213, opened 2023-09-15
AG-8564 Remove delay from axis and bar animations

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-8564

This makes bar series consistent with the other series and prevents any complications around axis being different for each series.

https://github.com/ag-grid/ag-charts/assets/3817697/ad2980df-214c-472a-a04c-ace8334bbd0d


alantreadway (approved):
LGTM

PR #214, opened 2023-09-15
AG-9212 - Add missing theme override common options.

alantreadway:

Adds missing cartesian-specific chart-level options to the theme overrides `common` config structure.

These require special handling, as polar + hierarchy charts should never have these configured (they current console.warn() in the latest `8.1.0` release too).


olegat (approved):
LGTM

alexanderby (approved):

PR #219, opened 2023-09-18
AG-9212 - Improve SeriesModule and theme typings

alantreadway:

On the path to aligning enterprise and community series registration I found some cleanup that was required:
- Makes `SeriesModule` theme definitions type-safe.
- Adds missing `range-area` theme options.
- Removes the need to cast template symbols on every use (moves cast to `string` to declaration point).

### DIFF
--- a/packages/ag-charts-community/src/chart/mapping/prepareSeries.ts
+++ b/packages/ag-charts-community/src/chart/mapping/prepareSeries.ts
@@ -85,7 +85,7 @@ export function processSeriesOptions(_opts: AgChartOptions, seriesOptions: Serie
 
     const preprocessed = seriesOptions.map((series: SeriesOptions & { stacked?: boolean; grouped?: boolean }) => {
         // Change the default for bar/columns when yKey is used to be grouped rather than stacked.
-        const sType = series.type ?? 'line';
+        const sType = series.type ?? '';
         const groupable = isGroupableSeries(sType);
         const stackable = isStackableSeries(sType);
         const stackedByDefault = isSeriesStackedByDefault(sType);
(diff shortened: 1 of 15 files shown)


packages/ag-charts-community/src/util/module.ts
```diff
@@ -90,7 +94,7 @@ export type Module<M extends ModuleInstance = ModuleInstance> =
     | AxisModule
     | AxisOptionModule
     | LegendModule
-    | SeriesModule;
+    | SeriesModule<any>;
```
olegat: nit: All instantiations of `SeriesModule` use a string. Would it make more sense to use `SeriesModule<string>` instead?

packages/ag-charts-community/src/chart/mapping/prepareSeries.ts
```diff
@@ -85,7 +85,7 @@ export function processSeriesOptions(_opts: AgChartOptions, seriesOptions: Serie
 
     const preprocessed = seriesOptions.map((series: SeriesOptions & { stacked?: boolean; grouped?: boolean }) => {
         // Change the default for bar/columns when yKey is used to be grouped rather than stacked.
-        const sType = series.type ?? '';
+        const sType = series.type ?? 'line';
```
olegat: nit: This should cause a behavioural change (old behaviour: invalid type/error, new behaviour: default to 'line' type). That's probably okay, but it's unclear whether this is what's desired because the change doesn't seemed scoped in bug AG-9212.

packages/ag-charts-community/src/util/module.ts
alantreadway: Good call - unfortunately `string` can't be used since (ironically) it's more general than the specification for the parameter allows. I've changed to `RequiredSeriesType` since this matches the specification.

packages/ag-charts-community/src/chart/mapping/prepareSeries.ts
alantreadway: I've reverted this, I thought it was needed due to tightening of types for the functions called later here (but I've broken that to a distinct PR).

Series without `type` specified are implicitly assumed to be `line` so practically this wouldn't a breaking change.

olegat (approved):
LGTM

PR #222, opened 2023-09-18
AG-9015 Move error bars implementation into the enterprise package

olegat:
This PR adds an enterprise module for error bars. The modules follows the same pattern as the crosshair module for consistency (the crosshair module gets added in `Axis.addAxisModule`, and the error bar module gets added in `Series.addModule`).

Key changes:
- Moves _errorBars.ts_ from `ag-charts-community/` to `ag-chars-enterprise/`
- Adds & exports an error bar module in the enterprise package.
- Add _moduleMap.ts_ which is a generalised implementation of the module map in `packages/ag-charts-community/src/axis.ts`.
- Adds the ability to listen for data model events in _cartesianSeries.ts_ (currently, only _lineSeries.ts_ fires this event).
- _module.ts_ is split into three files (_baseModule.ts_, _coreModules.ts_, and _optionModules.ts_). This is to avoid a cyclic dependency: `Series` needs to about `SeriesOptionModule` without importing from _module.ts_ (because this file imports _series.ts_)


olegat:
@alantreadway @alexanderby @manapeirov This PR is ready to merge (checks passed). Please review when possible. Thanks.

packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts
```diff
@@ -236,6 +239,18 @@ export abstract class CartesianSeries<
         this.subGroups.splice(0, this.subGroups.length);
     }
 
+    public addListener(type: DataEventType, listener: (event: DataEvent) => void) {
+        return this.dataModelListeners.addListener(type, listener);
+    }
+
+    public removeListener(listenerSymbol: Symbol) {
+        this.dataModelListeners.removeListener(listenerSymbol);
+    }
```
alantreadway, line 248: Nit: We may want to tweak the method naming here, as there is already a `series: { listeners: {} }` set of public configuration points for callbacks about series events?

packages/ag-charts-community/src/util/coreModules.ts
alantreadway, line 1: Nit: Could we try to preserve some of the history by making this a git file rename rather than a delete and add?

packages/ag-charts-community/src/util/moduleMap.ts
alantreadway, line 1: Thanks for adding this, it was niggling at the back of my mind that we should pull out some common for this case whenever I saw it.

packages/ag-charts-enterprise/src/error-bar/errorBar.ts
```diff
@@ -0,0 +1,138 @@
...
+        this.destroyFns.push(() => this.cartesianSeries.removeListener(listener));
+    }
+
+    onDataProcessed(event: {
+        dataModel: _ModuleSupport.DataModel<any, any, any>;
+        processedData: _ModuleSupport.ProcessedData<any>;
+    }) {
+        if (event.dataModel !== undefined && event.processedData !== undefined) {
+            this.createNodeData(event.dataModel, event.processedData);
+            this.update();
```
alantreadway, line 83: I don't think it's a major issue, but this does mean that the `ErrorBars` `createNodeData()` and `update()` happen during the series `processData` phase, which seems a little inconsistent.

packages/ag-charts-enterprise/src/error-bar/errorBar.ts
```diff
@@ -0,0 +1,138 @@
...
+        nodeData.length = processedData.data.length;
+
+        for (let i = 0; i < processedData.data.length; i++) {
+            const { datum, values } = processedData.data[i];
+            const xDatum = values[xIndex];
+
+            if (yLowerKey in datum && yUpperKey in datum) {
+                nodeData[i] = {
+                    yLowerPoint: { x: convert(xScale, xDatum), y: convert(yScale, datum[yLowerKey]) },
+                    yUpperPoint: { x: convert(xScale, xDatum), y: convert(yScale, datum[yUpperKey]) },
```
alantreadway, line 113: Although accessing the `datum` properties directly here achieves the right outcome for line, I think it would be better to specify additional properties to the `DataModel` so that grouped/stacked cases can be handled correctly.

Maybe we can follow-up on this in the next PR?

packages/ag-charts-enterprise/src/error-bar/errorBar.ts
olegat: `createNodeData` is perhaps a misnomer. `CartesianSeries.createNodeData` is called by [`CartesianSeries.updateSeries`](https://github.com/ag-grid/ag-charts/blob/4286604b832df7197d966ec609c89a81fe3e0a42/packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts#L296) which is called by [`CartesianSeries.update`](https://github.com/ag-grid/ag-charts/blob/4286604b832df7197d966ec609c89a81fe3e0a42/packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts#L272). So this module consistent with the series that uses it.

packages/ag-charts-community/src/util/moduleMap.ts
olegat: Yes, I think that _axis.ts_ will need a rewrite to use this class too, but I'll do that in a separate PR.

packages/ag-charts-enterprise/src/error-bar/errorBar.ts
olegat: It's unclear to me what the group/stacked cases involve. Let's follow up on this during a 1-on-1.

packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts
olegat: Yes, when fixing the cyclical dependency I noticed that the `Series` class extends the `Observable` class and already has an `addEventListener` method. So these new methods are redundant. It shouldn't be too difficult to use the existing `addEventListener` method as part of this PR.

packages/ag-charts-community/src/util/coreModules.ts
olegat: Yes that should be doable. However I think that's only doable for one of the files (_module.ts_, _coreModules.ts_, _optionModules.ts_. If I try to tell git that _coreModules.ts_ is a rename of _module.ts_ then I think that the old _module.ts_ would loss its history. Please let me know your thoughts.

alantreadway: I would pick whichever one is retaining the most original lines from the source file (perhaps `coreModules.ts`)?

packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts
olegat: On second thoughts let's keep this as-is for now. `Observable` is a legacy class and `Listeners` is preferable. We can refactor / edit this when the time is right.

packages/ag-charts-community/src/util/coreModules.ts
olegat: Yes _coreModules.ts_ has the most lines. I've updated this PR so that it inherits the history
https://github.com/ag-grid/ag-charts/blame/075ee22022dace9eb3e00ba05fc9dcc4d1fd813f/packages/ag-charts-community/src/util/coreModules.ts

alantreadway (approved):
LGTM

PR #223, opened 2023-09-18
AG-9212 - Split scatter and bubble series apart.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-9212

Splits `scatter` into `scatter` and `bubble` series types.

PR #224, opened 2023-09-18
AG-8470 Animation add/update/remove tests for bar and scatter

lsjroberts:

This has revealed a bug in bubble update animation - the unchanged markers immediately change size when the scale changes. I'll fix this separately.


packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.test.ts
```diff
@@ -97,4 +97,83 @@ describe('ScatterSeries', () => {
             });
         }
     });
+
+    describe('remove animation', () => {
```
alantreadway: #223 just got merged, could you add/update this case and `bubbleSeries.test.ts` too so we don't end up with a gap between this PR and that one?

alantreadway (approved):
LGTM

lsjroberts:
Removed scatter and bubble while I figure out some changes to the animation spy for those. Will merge bars first then do a new PR for scatter and bubble.

PR #226, opened 2023-09-18
Move module files from util/ to module/

olegat:
https://ag-grid.atlassian.net/browse/AG-9015

The `packages/ag-charts-community/src/util/` mostly contains a bunch of files that aren't related to each other. Therefore it feels to make sense to move these files somewhere else.


alantreadway (approved):
LGTM

PR #227, opened 2023-09-18
AG-9332: Fix chart.data overrides series.data

iMoses:
WIP - next step, need to add tests


packages/ag-charts-community/src/chart/chart.test.ts
```diff
@@ -475,4 +464,50 @@ describe('Chart', () => {
...
+                        yKey: datasets.economy.valueKey2,
+                    },
+                ],
+            });
+            expect(chart.data).toEqual(moreData);
+            expect(chart.series[0].data).toEqual(moreData);
+            expect(chart.series[1].data).toEqual(lessData);
+
+            chart.series[0].data = lessData;
+            chart.series[1].data = null;
```
alantreadway: Rather than manually manipulating the implementation, could we instead use `AgChart.update()` to adjust the options appropriately?

iMoses: Done

alantreadway (approved):
LGTM

PR #229, opened 2023-09-18
Fix IDE automatic imports

iMoses:
This fixes the issue for me and doesn't seem to effect the build.

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -17,10 +17,10 @@
     "importsNotUsedAsValues": "error",
     "paths": {
       "ag-charts-build-tools": ["plugins/ag-charts-build-tools/src/index.ts"],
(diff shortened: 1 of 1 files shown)


alantreadway (approved):
LGTM - my only concern is whether this creates a mismatch between actual package exports and those that TS allows at compile time, but the `tsconfig` `paths` documentation is unclear on this point, and they seem to use wildcards/blobs in their examples for that use-case?

iMoses:
> LGTM - my only concern is whether this creates a mismatch between actual package exports and those that TS allows at compile time, but the `tsconfig` `paths` documentation is unclear on this point, and they seem to use wildcards/blobs in their examples for that use-case?

I used a wildcard at the end of the path at first but we seem to have a lint rule against it, and it seems to work just fine without it.
Considering that it doesn't effect build, we can try to push it and if we notice any issues we can always revisit or rewind these changes. Hopefully that won't be the case, I found this solution after looking at multiple nx issues online.

PR #234, opened 2023-09-19
AG-9212 - Convert cartesian series to use SeriesModule registration.

alantreadway:

Converts the remaining community cartesian series to use `SeriesModule` defintions.

Bonus:
- Cleanup some dangling optional options properties.


lsjroberts (approved):

packages/ag-charts-community/src/chart/series/cartesian/areaSeriesModule.ts
```diff
@@ -0,0 +1,59 @@
+import type { SeriesModule } from '../../../util/module';
```
olegat, line 1: Just an FYI: this type is moving into _coreModules.ts_ in PR #222; so the build errors / merge-conflicts will need to be resolved by someone.

PR #238, opened 2023-09-19
AG-9212 - Cleanup remaining series definitions.

alantreadway:

Converts `treemap` and `pie` series to use `SeriesModule` definitions.


olegat:
This will cause a compilation error if it's merged in with PR #222, so I'll hold those changes off until these additional SeriesModule change make it in.

packages/ag-charts-community/src/chart/themes/symbols.ts
```diff
@@ -13,3 +15,6 @@ export const DEFAULT_AXIS_GRID_COLOUR = Symbol('default-axis-grid-colour') as un
 export const DEFAULT_BACKGROUND_COLOUR = Symbol('default-background-colour') as unknown as string;
 export const DEFAULT_SHADOW_COLOUR = Symbol('default-shadow-colour') as unknown as string;
 export const DEFAULT_TREEMAP_TILE_BORDER_COLOUR = Symbol('default-treemap-tile-border-colour') as unknown as string;
+
+export const BOLD: FontWeight = 'bold';
+export const BOTTOM: AgChartLegendPosition = 'bottom';
```
iMoses: Why are there constants that aren't symbols in the symbols file?

alantreadway: These are additional constants needed by a couple of theme template cases, given these are used in conjunction with in similar places to the symbols above the grouping makes some sense?

alantreadway: I've split these into a `constants.ts` file now.

alantreadway:
@iMoses PTAL

iMoses (approved):
LGTM 👍

PR #239, opened 2023-09-19
Ag 9343 gallery design updates

markdurrant:
- Small refactor 
- Update layout to be more like other 'full screen' pages. (actually shows more charts per page because of aspect ratios)
- Update series icon colour to better match text
- Update background colour 
- Update chart thumbnails to remove box shadow but stay minimal 
- Re-order histogram in series list so it's with the other enterprise items


manapeirov (approved):
LGTM!

PR #241, opened 2023-09-20
AG-9022 Add initial error bars documentation

olegat:
https://ag-grid.atlassian.net/browse/AG-9022

### DIFF
--- a/packages/ag-charts-website/src/content/docs/error-series/index.mdoc
+++ b/packages/ag-charts-website/src/content/docs/error-series/index.mdoc
@@ -3,14 +3,14 @@ title: 'Error Series'
 enterprise: true
 ---
 
-Error bars show the uncertainty of continuous data values. They can be enabled
+Error Bars show the uncertainty of continuous data values. They can be enabled
 on Line, Bar and Scatter Series.
 
 ## Line Series Error Bars
 
 {% chartExampleRunner title="Line Series Error Bars" name="line-series-error-bars" type="generated" options="{ \"enterprise\": true }" /%}
 
-Error bars can be enabled on Line Series using the `errorBars` key.
+Error Bars can be enabled on Line Series using the `errorBars` key.
 
 The `yLowerKey` and `yUpperKey` defines the keys used to retrieve the error
 values from the data.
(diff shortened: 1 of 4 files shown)


packages/ag-charts-website/src/content/docs/error-series/index.mdoc
```diff
@@ -1,5 +1,37 @@
 ---
 title: 'Error Series'
+enterprise: true
 ---
 
-TODO
+Error bars show the uncertainty of continuous data values. They can be enabled
```
olegat: Should error bars be title cased ("Error Bars") throughout the documentation?

packages/ag-charts-website/src/content/docs/error-series/index.mdoc
```diff
@@ -1,5 +1,37 @@
...
+on Line, Bar and Scatter Series.
+
+## Line Series Error Bars
+
+{% chartExampleRunner title="Line Series Error Bars" name="line-series-error-bars" type="generated" options="{ \"enterprise\": true }" /%}
+
+Error bars can be enabled on Line Series using the `errorBars` key.
+
+The `yLowerKey` and `yUpperKey` defines the keys used to retrieve the error
+values from the data.
```
olegat, line 16: I'm inclined to use the word "dataset" instead of "data" to clarify that it's actual statistical data rather than config data like theme options. Thoughts?

packages/ag-charts-website/src/content/docs/error-series/index.mdoc
alantreadway: Correct

packages/ag-charts-website/src/content/docs/error-series/index.mdoc
alantreadway: I'm not sure drawing on that concept is entirely necessary - `yLowerKey` and `yUpperKey` are both part of the elements in `data`, so I would keep the language and concepts simple?

olegat: Ah okay, I understand. I thought that the input parameter for `AgEnterpriseCharts.create()` was also named `data`, but it's named `options`. So there indeed shouldn't be any confusion.

packages/ag-charts-website/src/content/docs/error-series/index.mdoc
olegat: Done.

alantreadway (approved):
LGTM

PR #242, opened 2023-09-20
AG-8470 Add scatter & bubble animation tests

lsjroberts:

This also changes the scatter and bubble update animations to do a clear and reload, since we can't track the datums by any unique id.

I made some changes to the selection to support this and so added some tests for that too.


packages/ag-charts-community/src/chart/__image_snapshots__/gallery-examples-test-ts-gallery-examples-ag-chart-v-2-create-for-simple-scatter-chart-example-it-should-render-to-canvas-as-expected-1-snap.png
lsjroberts, line 1: The scatter & bubble examples were missing nodes with identical x/y coords due to the issue with the non-unique datum id

alantreadway (approved):
LGTM

PR #244, opened 2023-09-20
AG-8470 Change histogram update animation to simultaneous and add tests

lsjroberts:


alantreadway (approved):
LGTM

PR #246, opened 2023-09-20
Cleanup listeners code

iMoses:


packages/ag-charts-community/src/util/listeners.ts
```diff
@@ -1,82 +1,51 @@
...
+                listener.handler(...params);
             } catch (e) {
                 Logger.errorOnce(e);
-                results.push(undefined as any);
             }
         }
-        return results;
-    }
-
-    public cancellableDispatch(
```
alantreadway, line 37: We should probably keep this behaviour here, I can see it being needed in other cases in the future potentially.

alantreadway (commented):
Nice cleanup of the listener removal case - I don't think we should push the cancellable dispatch behaviour down to `InteractionManager` though since there maybe other cases in the future where we want to have similar dispatch behaviour.

packages/ag-charts-community/src/util/listeners.ts
iMoses: let's talk about it offline

alantreadway: Discussed offline - the prior implementation was suspect on 2nd look, we'd want to implement it differently anyway.

alantreadway (approved):
LGTM - thanks, nice cleanup!

PR #248, opened 2023-09-21
Add ErrorBar styling options

olegat:

This adds the following options:
    
- visible (default: true)
- stroke (default: 'black')
- strokeWidth (default: 1)
- strokeOpacity (default: 1)
    
These options are duplicated for both the whiskers and caps (so that
these components can be customised individually). Cap inherit the
defaults from the whisky (e.g. setting whisker.stroke = 'red' will
also set cap.stroke = 'red')
    
The rendering of caps is not fully implemented yet (`lengthRatio` is ignored)

### DIFF
--- a/packages/ag-charts-enterprise/src/error-bar/errorBar.ts
+++ b/packages/ag-charts-enterprise/src/error-bar/errorBar.ts
@@ -1,5 +1,6 @@
 import type { _Scale } from 'ag-charts-community';
 import { _Scene, _Util, _ModuleSupport } from 'ag-charts-community';
+import { ErrorBarPoints, ErrorBarNode } from './errorBarNode';
 
 const { ChartAxisDirection, Validate, OPT_STRING } = _ModuleSupport;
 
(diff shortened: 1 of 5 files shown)


packages/ag-charts-enterprise/src/error-bar/errorBar.ts
```diff
@@ -123,11 +142,29 @@ export class ErrorBars extends _ModuleSupport.BaseModuleInstance implements _Mod
         this.selection.each((node, datum, i) => this.updateNode(node, datum, i));
     }
 
+    private inheritProperties(
```
alantreadway: Let's sync up, we normally handle these scenarios using theme configuration (error bars should work the same way).

alantreadway: https://github.com/ag-grid/ag-charts/blob/latest/packages/ag-charts-enterprise/src/box-plot/boxPlotSeries.ts#L479-L485

packages/ag-charts-enterprise/src/error-bar/errorBar.ts
```diff
@@ -126,8 +139,10 @@ export class ErrorBars extends _ModuleSupport.BaseModuleInstance implements _Mod
         const { nodeData } = this;
         const points = nodeData[index];
         if (points) {
-            node.points = points;
-            node.updatePath();
+            const defaults: ErrorBarNodeProperties = ERROR_BARS_THEME.errorBar;
```
olegat, line 142: Note: this a temporary workaround. Getting themes to work correctly for error bars will probably making changes to `packages/ag-charts-community/src/chart/factory/setupModules.ts` to support `'series-option'` type modules. I will do this in a follow up PR

alantreadway (approved):
LGTM

PR #249, opened 2023-09-21
AG-9212 - Move histogram and treemap to enterprise.

alantreadway:

Moves the `histogram` and `treemap` into the enterprise package. Moves as many tests from community as possible without too much complexity.


packages/ag-charts-enterprise/src/histogram/histogramSeries.test.ts
alantreadway, line 1: NOTE: This has been `git mv` in a discrete commit to preserve line history, despite appearing as a new addition here.

lsjroberts (approved):
Great stuff, lgtm

packages/ag-charts-community/src/chart/test/examples-gallery.ts
```diff
@@ -5,123 +5,159 @@ import { cartesianChartAssertions, hierarchyChartAssertions, polarChartAssertion
...
+        enterprise: false,
     },
 };
+
+export const EXAMPLES = Object.entries(COMMUNITY_AND_ENTERPRISE_EXAMPLES)
+    .filter(([, v]) => !v.enterprise)
+    .reduce((pv, [k, v]) => {
+        pv[k] = v;
+        return pv;
+    }, {} as typeof COMMUNITY_AND_ENTERPRISE_EXAMPLES);
```
iMoses:
```suggestion
export const EXAMPLES = Object.entries(COMMUNITY_AND_ENTERPRISE_EXAMPLES)
    .filter(([, v]) => !v.enterprise)
    .reduce<typeof COMMUNITY_AND_ENTERPRISE_EXAMPLES>((pv, [k, v]) => {
        pv[k] = v;
        return pv;
    }, {});
```
nit: I know it's just a test, so it's okay if we're less strict here, but generally I would avoid using TS `as` if possible to keep the type checks in place.

iMoses (approved):
Looks good 👍

packages/ag-charts-community/src/chart/test/examples-gallery.ts
alantreadway: Nice suggestion, thanks. 100% agree that `as` is a smell, hadn't realised that overriding the inferred type for a `reduce()` call was possible, there are probably a bunch of places we can apply this fix.

PR #250, opened 2023-09-21
AG-8488 Add line add/update/remove animation tests

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-8488

This doesn't truly fix the issue with animating in and out `undefined` values, but it'll make sure any fixes for that don't break the rest of it.


alantreadway (approved):
LGTM

PR #254, opened 2023-09-21
AG-9212 - Cleanup enterprise and options folder structures.

alantreadway:
Cleanup the folder structures under:
- `packages/ag-charts-enterprise/src/` - separate `series/`, `features/` and `axes/`.
- `packages/ag-charts-community/src/options/` - rename 1st-level nested `options/` folder, remove folders for two features that have their own feature folder with a single module.


lsjroberts (approved):
Nice, making features and series distinct is good. LGTM.

PR #256, opened 2023-09-22
AG-8260 Radial Bar series

alexanderby:


packages/ag-charts-enterprise/src/axes/angle-number/angleNumberAxis.ts
```diff
@@ -0,0 +1,468 @@
...
+    (v: any) => ANGLE_LABEL_ORIENTATIONS.includes(v),
+    `expecting a label orientation keyword such as 'fixed', 'parallel' or 'perpendicular'`
+);
+
+class AngleCategoryAxisLabel extends _ModuleSupport.AxisLabel {
+    @Validate(ANGLE_LABEL_ORIENTATION)
+    orientation: AgAngleAxisLabelOrientation = 'fixed';
+}
+
+export class AngleNumberAxis extends _ModuleSupport.PolarAxis<_Scale.LinearScale> {
```
alantreadway: This entire class looks very similar to `angleCategoryAxis.ts` - can we factor out the common behaviours?

packages/ag-charts-enterprise/src/axes/radius-category/radiusCategoryAxis.ts
```diff
@@ -0,0 +1,200 @@
...
+}
+
+type TickDatum = {
+    tickLabel: string;
+    tick: any;
+    tickId: string;
+    translationY: number;
+};
+
+export class RadiusCategoryAxis extends _ModuleSupport.PolarAxis {
```
alantreadway: There seems to be a bunch of commonality with `radiusNumberAxis.ts` - can we also factor out some common code?

packages/ag-charts-enterprise/src/axes/radius-category/radiusCrossLine.ts
```diff
@@ -0,0 +1,157 @@
...
+const { ChartAxisDirection, Validate, OPT_NUMBER } = _ModuleSupport;
+const { Path, Sector, Text } = _Scene;
+const { normalizeAngle360, toRadians, isNumberEqual } = _Util;
+
+class RadiusCrossLineLabel extends PolarCrossLineLabel {
+    @Validate(OPT_NUMBER(-360, 360))
+    positionAngle?: number = undefined;
+}
+
+export class RadiusCrossLine extends PolarCrossLine {
```
alantreadway: This seems to be an identical implementation to the `radius-number` axis implementation? Can we factor out the common code please?

packages/ag-charts-enterprise/src/axes/angle-number/angleCrossLine.ts
```diff
@@ -0,0 +1,173 @@
+import { _ModuleSupport, _Scale, _Scene, _Util } from 'ag-charts-community';
+import { PolarCrossLine } from '../polarCrossLine';
+
+const { ChartAxisDirection } = _ModuleSupport;
+const { Path, Sector, Text } = _Scene;
+const { normalizeAngle360, isNumberEqual } = _Util;
+export class AngleCrossLine extends PolarCrossLine {
```
alantreadway: This seems to be an identical implementation to the `angle-category` axis implementation? Can we factor out the common code please?

packages/ag-charts-enterprise/src/series/radial-bar/radialBarSeries.ts
```diff
@@ -0,0 +1,708 @@
...
+    maybeRefreshNodeData() {
+        const circleChanged = this.didCircleChange();
+        if (!circleChanged && !this.nodeDataRefresh) return;
+        const [{ nodeData = [] } = {}] = this.createNodeDataSync();
+        this.nodeData = nodeData;
+        this.nodeDataRefresh = false;
+    }
+
+    async createNodeData() {
+        return this.createNodeDataSync();
```
alantreadway, line 265: This seems completely unnecessary? Please maintain the async contract - all callers (specifically `maybeRefreshNodeData()`) are called by `async` functions so there should be no need for a synchronous version of this function.

alantreadway (changes requested):

packages/ag-charts-enterprise/src/series/radial-bar/radialBarSeries.ts
alexanderby: I saw it was implemented in Pie series that way. I will try making it async.

alexanderby: Unfortunately there is `computeLabelsBBox` method of `PolarSeries` that is `sync`. Maybe I could make a follow up PR to fix it for all the polar series?

alantreadway: Ok sounds good

alantreadway (approved):
LGTM - thanks for the updates, looks much better!

PR #258, opened 2023-09-25
AG0-9016 Add Error Bars to Scatter series

olegat:
https://ag-grid.atlassian.net/browse/AG-9016
Key changes

**Documentation:**
- Add volume/pressure scatter example.
- Tidy indentation of line series example (use 4-space indentation like waterfall doc).

**Community Code:**
- Add `errorBar` option for Scatter series.
- Fire data processing event in `ScatterSeries` class.

**Enterprise Code:**
- Generalise the Y-axis error bar logic.
- Implement (optional) X-axis error bar configuration & rendering.

### DIFF
--- a/packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
+++ b/packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
@@ -225,6 +225,7 @@ export class ScatterSeries extends CartesianSeries<SeriesNodeDataContext<Scatter
         }
 
         nodeData.length = actualLength;
+        this.fireDataProcessed(dataModel, processedData);
 
         return [{ itemId: this.yKey ?? this.id, nodeData, labelData: nodeData }];
     }
--- a/packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
+++ b/packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
@@ -225,7 +225,6 @@ export class ScatterSeries extends CartesianSeries<SeriesNodeDataContext<Scatter
         }
 
         nodeData.length = actualLength;
-        this.fireDataProcessed(dataModel, processedData);
 
         return [{ itemId: this.yKey ?? this.id, nodeData, labelData: nodeData }];
     }
--- a/packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
+++ b/packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
@@ -5,6 +5,7 @@ import { ErrorBarNode } from './errorBarNode';
 import { ERROR_BARS_THEME } from './errorBarTheme';
 
 const { mergeDefaults, ChartAxisDirection, Validate, BOOLEAN, OPT_STRING, NUMBER } = _ModuleSupport;
+const { Logger } = _Util;
 type AnyScale = _Scale.Scale<any, any, any>;
 
 type OptionalErrorBarNodeProperties = { [K in keyof ErrorBarNodeProperties]?: ErrorBarNodeProperties[K] };
@@ -137,7 +138,7 @@ export class ErrorBars
                             yDatum
                         );
                     } else {
-                        throw new Error(
+                        Logger.warnOnce(
                             `AG Charts - series type 'scatter' requires xLowerKey (= ${xLowerKey}) and xUpperKey (= ${xUpperKey}) to exist in all data points`
                         );
                     }
--- a/packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
+++ b/packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
@@ -5,11 +5,7 @@ import { ErrorBarNode } from './errorBarNode';
 import { ERROR_BARS_THEME } from './errorBarTheme';
 
 const { mergeDefaults, ChartAxisDirection, Validate, BOOLEAN, OPT_STRING, NUMBER } = _ModuleSupport;
-
-type CartesianSeries<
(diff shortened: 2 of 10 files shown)


packages/ag-charts-website/src/content/docs/error-bars/_examples/scatter-series-error-bars/main.ts
```diff
@@ -0,0 +1,25 @@
+import { AgChartOptions, AgEnterpriseCharts } from 'ag-charts-enterprise';
+import { getData } from './data';
+
+const options: AgChartOptions = {
+    container: document.getElementById('myChart'),
+    data: getData(),
+    title: {
+        text: 'Volume-Pressure Relationship with Confidence Intervals',
+    },
+    series: [
```
olegat, line 10: Note: This scatter plot doesn't show units for X (m³) and Y (Bar) axes.

Also this generated data is counter-intuitive from a physics point of view. It shows a positive correlation between Volume and Pressure, but this should be reversed ([Boyle's law](https://en.wikipedia.org/wiki/Boyle%27s_law#:~:text=Boyle%27s%20law%20is%20a%20gas,doubled%2C%20the%20pressure%20is%20halved.)).

We may want to consider tweaking this example chart & data for better presentation.

olegat: I've reverse the relationship between Volume & Pressure, but the units of the axes still isn't show. Please let me know whether that's acceptable for this example.

packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
```diff
@@ -225,6 +225,7 @@ export class ScatterSeries extends CartesianSeries<SeriesNodeDataContext<Scatter
         }
 
         nodeData.length = actualLength;
+        this.fireDataProcessed(dataModel, processedData);
```
alantreadway: Should/could we move this to the super-class so all `CartesianSeries` have this behaviour after running `processData()`?

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
```diff
@@ -100,36 +108,63 @@ export class ErrorBars
...
+                            datum[xLowerKey],
+                            datum[xUpperKey],
+                            yScale,
+                            yDatum,
+                            yDatum
+                        );
+                    } else {
+                        throw new Error(
+                            `AG Charts - series type 'scatter' requires xLowerKey (= ${xLowerKey}) and xUpperKey (= ${xUpperKey}) to exist in all data points`
+                        );
```
alantreadway: Nit: Since this is a data-validation style error, we should console.warn() about this and carry on if possible - `Logger.warnOnce()` should be the right choice here so we don't spam the user for every missing data-poiont, on every reprocessing of the data:
https://github.com/ag-grid/ag-charts/blob/c1b1c01dd577a516ee1eaeb816c2a5bc16b7747c/packages/ag-charts-community/src/util/logger.ts#L25

packages/ag-charts-website/src/content/docs/error-bars/_examples/scatter-series-error-bars/main.ts
alantreadway: So there are a few relevant configuration points depending on how you want to display the units:
- `series.label.formatter` - series label format
- `series.tooltip.format` - tooltip format
- `axes[].label.format` - axis label format
- `axes[].label.formatter` - axis label formatter
- `axes[].title.text` - axis title

For examples I would avoid adding too much clutter, so whichever gives a sensible result in the smallest number of additional lines is probably best.

olegat: Understood. I think that we can leave it like this for the documentation and leave the more stylised examples for the showcase.

packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
olegat: I initially tried doing this ([ba1739ec7](https://github.com/ag-grid/ag-charts/commit/ba1739ec7d47391ac01639538dad7fb4704cc2f0)).

I cannot remember exactly what went wrong with this approach, but AFAIK the `Scale.convert()` function was not returning the expected result when called by the event handler in `errorBar.ts`. This could perhaps be because the axes are not fully initialised at this point. On the other hand, firing the event during the `createNodeData()` function gives the expected `convert()` return values.

That said, I think he should try to make this work by firing the event in `processData()` because it would be more robust.

Should I tried to resolve this within the scope of this PR or leave it as an action item for a follow-up?

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
olegat: Good idea. Thanks for the feedback. I've changed this ([25c55792e](https://github.com/ag-grid/ag-charts/pull/258/commits/25c55792e1bf51c29bbc38fd15d637095ab2087e))

packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
olegat: Actually, I've found another way to fire the event in the `CartesianSeries` base class ([a0c519a1](https://github.com/ag-grid/ag-charts/pull/258/commits/a0c519a1a556aa803aebd09a1cf338b7f73f00c0)). Resolving.

alantreadway (approved):
LGTM - nice work!

PR #260, opened 2023-09-25
AG-9087 - Add doc internal links

taktran:
## Changes

* Remove all `charts-` prefix in links
* Replace absolute link urls (`/`) with relative path (`./`)
* Add link markdoc replacement to include base url and framework

## Notes

The following rules are used for markdoc links:

1. For internal docs links, use `./` prefix only eg, `[Axis Label](./axis-labels/)` -> `/ag-charts/vue/axes-labels/`
1. For website links use `/` eg, `[Gallery](/gallery)` -> `/ag-charts/gallery/`
1. For external links use `http/s` eg, `[Getting started](https://www.youtube.com/watch?v=KS-wg5zfCXc)` -> `https://www.youtube.com/watch?v=KS-wg5zfCXc`

### DIFF
--- a/packages/ag-charts-website/markdoc.config.ts
+++ b/packages/ag-charts-website/markdoc.config.ts
@@ -8,6 +8,10 @@ export default defineMarkdocConfig({
             ...nodes.heading, // Preserve default anchor link generation
             render: component('./src/components/Heading.astro'),
         },
+        link: {
+            ...nodes.link,
+            render: component('./src/components/Link.astro'),
+        },
     },
     functions: {
         isFramework: {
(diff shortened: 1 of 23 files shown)


packages/ag-charts-website/src/content/docs/api-explorer/index.mdoc
```diff
@@ -4,7 +4,7 @@ title: 'API Explorer'
 
 The API explorer can be used to gain familiarity with the AG Charts API and see in real-time how different `options` affect charts. A code snippet showing the use of the API is produced based on the point and click configurations specified in the explorer.
 
-<charts-api-explorer></charts-api-explorer>
+<charts-api-explorer><./api-explorer>
```
alantreadway: Looks like a mistake?

alantreadway (approved):
LGTM other than one issue

packages/ag-charts-website/src/content/docs/api-explorer/index.mdoc
taktran: Good catch!

PR #261, opened 2023-09-25
Remove expect.extend from test files (use jest.setup.ts)

iMoses:
Minor tests cleanup following the creation of a `jest.setup.ts` file


alantreadway (approved):
LGTM once we can iron out the CI gremlins.

PR #272, opened 2023-09-27
AG-8564 - Fix axis label zoom-out animation yielding missing labels.

alantreadway:
Attempts to fix axes labels disappearing on zoom out.

Before:
https://github.com/ag-grid/ag-charts/assets/17544187/86b60908-8165-4aee-bb3b-8af4ce621b57

After:
https://github.com/ag-grid/ag-charts/assets/17544187/627785c9-37f0-4c6f-8f96-03383b497bf5

### DIFF
--- a/packages/ag-charts-community/src/chart/axis/axis.ts
+++ b/packages/ag-charts-community/src/chart/axis/axis.ts
@@ -919,35 +919,32 @@ export abstract class Axis<S extends Scale<D, number, TickInterval<S>> = Scale<a
         return { minTickCount, maxTickCount, defaultTickCount };
     }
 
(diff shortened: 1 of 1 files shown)


lsjroberts (approved):
Nice spot. Though tbh I imagine we probably don't want to animate the axis here, since the zoom doesn't animate the series. But maybe we do want to animate them? Something for another time.

alantreadway:
@lsjroberts agreed there might be a wider question about axis animations; I think that since visibility is handled specially for the axis tick/label nodes we should still ensure the same logic is applied irrespective if there is animation code present.

PR #274, opened 2023-09-27
AG-8488 Add throttling to line animation updates

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-8488

Unthrottled:

https://github.com/ag-grid/ag-charts/assets/3817697/5ab595e2-f0cd-4b5d-b14f-952e4613207c

Throttled:

https://github.com/ag-grid/ag-charts/assets/3817697/49729ac4-a587-40f5-8a96-d52b2ccb53cd


packages/ag-charts-community/src/chart/interaction/animationManager.ts
```diff
@@ -53,6 +55,18 @@ export class AnimationManager extends BaseManager<AnimationEventType, AnimationE
...
+            const isThrottled = throttleTime && opts.duration && Date.now() - throttleTime < opts.duration;
+
+            if (isThrottled) {
+                opts.delay = 0;
+                opts.duration = 1;
+            }
+
+            this.throttles.set(opts.throttleId, Date.now());
+        }
+
```
lsjroberts, line 69: This is the approach from before the refactor, but if you think there is a better way to match the changes let me know @iMoses

packages/ag-charts-community/src/chart/series/cartesian/lineSeries.ts
```diff
@@ -156,7 +156,8 @@ export class LineSeries extends CartesianSeries<LineContext> {
         this.dataModel = dataModel;
         this.processedData = processedData;
 
-        if (processedData.reduced?.diff?.added.length > 0 && processedData.reduced?.diff?.removed.length > 0) {
+        // If the diff is too complex then just clear and redraw to prevent wonky line wobbling
+        if (processedData.reduced?.diff?.added.length > 1 && processedData.reduced?.diff?.removed.length > 1) {
```
lsjroberts, line 160: This is definitely a little hacky, but we need to cover two common use cases:

- Many points are changing -> do a clear and full redraw
- The first point is removed and a new point added onto the end -> animate normally (see the animation in the main PR description)

The other way to differentiate these would require a bunch of pre-processing of the data and diff before calling the transition. There may be a nicer way to do that, but I'll look at that when I do a greater refactor of the line animation out of the current big blob of logic.

lsjroberts:
And before, with throttling removed (while calling before the animation has finished):

https://github.com/ag-grid/ag-charts/assets/3817697/c464ee0f-4940-4f87-b810-bd3215d0b535

alantreadway (approved):
LGTM

PR #278, opened 2023-09-27
Enforce and cleanup mutually exclusive chart updates.

alantreadway:
Whilst investigating the source of competing chart updates for a specific large-scale heatmap case, I found that `AnimationManager` animation frame updates don't consider whether `Chart.performUpdate()` is mid-execution. These two processes should be mutually exclusive practically, as `Chart.performUpdate()` will likely affect the currently executed animation anyway.

Adds:
- `Mutex` type to isolate mutex-related behaviours.
- Use of a shared `Mutex` for factory options update, `Chart.performUpdate()` and `AnimationManager` animation-frame processing.
- More debug information for repeated/looping `Chart.performUpdate()` cases.

### DIFF
--- a/packages/ag-charts-community/src/chart/agChartV2.ts
+++ b/packages/ag-charts-community/src/chart/agChartV2.ts
@@ -346,8 +346,6 @@ abstract class AgChartInternal {
             };
         }
 
-        await chart.awaitUpdateCompletion();
-
         if (chart.destroyed) return;
 
         debug('AgChartV2.updateDelta() - applying delta', processedOptions);
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -26,7 +26,6 @@ import type {
 import { debouncedAnimationFrame, debouncedCallback } from '../util/render';
 import type { Point } from '../scene/point';
 import { BOOLEAN, OPT_BOOLEAN, STRING_UNION, Validate } from '../util/validation';
-import { sleep } from '../util/async';
 import type { TooltipMeta as PointerMeta } from './tooltip/tooltip';
 import { Tooltip } from './tooltip/tooltip';
 import { ChartOverlays } from './overlay/chartOverlays';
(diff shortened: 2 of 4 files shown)


packages/ag-charts-community/src/chart/agChartV2.ts
```diff
@@ -346,8 +346,6 @@ abstract class AgChartInternal {
             };
         }
 
-        await chart.awaitUpdateCompletion();
```
alantreadway, line 349: NOTE: No longer needed as the only call to `updateDelta()` is to a code-block which has already acquired the `Chart.updateMutex()` lock so there are no other pending updated.

packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1369,13 +1351,20 @@ export abstract class Chart extends Observable implements AgChartInstance {
     async waitForUpdate(timeoutMs = 5000): Promise<void> {
         const start = performance.now();
 
-        while (this._pendingFactoryUpdates.length > 0 || this.updatePending) {
+        if (this._pendingFactoryUpdatesCount > 0) {
+            // Await until any pending updates are flushed through.
+            await this.updateMutex.acquire(async () => undefined);
+        }
```
lsjroberts, line 1358: Is `_pendingFactoryUpdatesCount` needed if we are waiting for updates to be flushed anyway?

Also, is it worth making `.acquire(async () => undefined)` it's own function on `Mutex`, seems like it may be useful in situations opposite to `.acquireImmediately()`? Like perhaps `.acquireWhenReady()` or some better name.

lsjroberts (approved):
LGTM

packages/ag-charts-community/src/chart/chart.ts
alantreadway: So I tried to remove `_pendingFactoryUpdateCount`, but unfortunately since there is a gap between updates being applied and the `performUpdate()` cycle executing, this was the only reliable way to ensure that all applied updates have been applied AND then that the resulting `performUpdate()` has happened.

alantreadway: Added `Mutex.waitForClearAcquireQueue()` for this case as suggested.

PR #280, opened 2023-09-27
Types cleanup

iMoses:
Contains mainly 3 things:
- Use `StateMachine` directly, using inheritance
- Fix `getLegendData` types
- Separate `SeriesTooltip` to another file

imports changed only due to auto-sort/IDE cleanup


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -909,48 +908,41 @@ export abstract class Chart extends Observable implements AgChartInstance {
...
-            this.series
-                .filter((s) => s.showInLegend)
-                .forEach((series) => {
-                    const data = series.getLegendData(legendType);
-                    legendData.push(...data);
-                });
-
-            if (legendType === 'category') {
-                this.validateLegendData(legendData);
+            const isCategory = (_data: unknown): _data is CategoryLegendDatum[] => legendType === 'category';
```
alantreadway: Nit: I'd rather we cast using `as` if we're going this route, since it's clearer that we're making assumptions about the type?

Could we just add a real type-guard that actually checks the element types somewhere?

packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
```diff
@@ -151,8 +152,6 @@ export class AreaSeries extends CartesianSeries<AreaSeriesNodeDataContext> {
 
     shadow?: DropShadow = undefined;
 
-    protected highlightedDatum?: MarkerSelectionDatum;
```
alantreadway, line 154: Good spot :+1:

packages/ag-charts-community/src/chart/series/seriesTooltip.ts
```diff
@@ -0,0 +1,34 @@
...
+export class SeriesTooltip<P extends AgSeriesTooltipRendererParams> {
+    @Validate(BOOLEAN) enabled = true;
+    @Validate(BOOLEAN) showArrow = false;
+    @Validate(OPT_STRING) format?: string = undefined;
+    @Validate(OPT_FUNCTION) renderer?: TooltipRenderer<P> = undefined;
+
+    readonly interaction = new SeriesTooltipInteraction();
+    readonly position = new TooltipPosition();
+
+    toTooltipHtml(defaults: AgTooltipRendererResult, params: P, overrides?: TooltipOverrides<P>) {
```
alantreadway, line 23: 🎉

alantreadway (commented):
Generally looking great - just a couple of comments.

packages/ag-charts-community/src/chart/chart.ts
iMoses: we could if we look at the data `legendType`
is this the direction you prefer?
checkout the latest update

alantreadway (approved):
Perfect - LGTM

packages/ag-charts-community/src/chart/chart.ts
alantreadway: Great - thank you!

PR #281, opened 2023-09-27
AG-9015 - Error bars QA feedback

olegat:
https://ag-grid.atlassian.net/browse/AG-9015?focusedCommentId=53576

# Issues addressed:

1. Axis domain is extended to accommodate error bar values by providing additional properties to `DataController.request` using the new `''processData-prerequest'` series event.
2. Error bars are drawn in front of the markers/bars by increasing the `ZINDEX`.
3. Fixed console warnings because of misuses of `@Validate`
4. Error bars are hidden when the series is hidden by listening to the new `'visibility-changed'` events.

# Caveats

- Error bars are not drawn above the highlights.
- Axis domain is not extended for bar series (only works on line and scatter at the moment).


packages/ag-charts-community/src/chart/layers.ts
```diff
@@ -7,6 +7,7 @@ export enum Layers {
     AXIS_ZINDEX = 20,
     SERIES_CROSSLINE_RANGE_ZINDEX = 30,
     SERIES_LAYER_ZINDEX = 500,
+    SERIES_ERRORBAR_ZINDEX = 600,
```
olegat, line 10: It's unclear whether this is necessary or whether we could just use `SERIES_LAYER_ZINDEX` instead

alantreadway: 🤔 I'd need to look at the live scene-graph to work that out - let's come back to it.

packages/ag-charts-community/src/chart/series/seriesEvents.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export interface SeriesPrerequestDataEvent extends BaseSeriesEvent<'processData-prerequest'> {
+    props: PropertyDefinition<string>[];
+    isContinuousX: boolean;
+    isContinuousY: boolean;
+}
+
+export interface SeriesVisibilityEvent extends BaseSeriesEvent<'visibility-changed'> {
+    itemId: any;
```
alantreadway: Can we add a stricter typing here?

packages/ag-charts-community/src/chart/series/seriesEvents.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export interface BaseSeriesEvent<_T extends SeriesEventType> {}
+
+export interface SeriesDataUpdateEvent extends BaseSeriesEvent<'data-update'> {
+    dataModel: DataModel<any, any, any>;
+    processedData: ProcessedData<any>;
+}
+
+export interface SeriesPrerequestDataEvent extends BaseSeriesEvent<'processData-prerequest'> {
+    props: PropertyDefinition<string>[];
```
alantreadway: Just wondering if `string` is the right generic type here - maybe `unknown` would be better if feasible?

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
```diff
@@ -4,23 +4,34 @@ import type { ErrorBarPoints, ErrorBarNodeProperties } from './errorBarNode';
...
-type AnyScale = _Scale.Scale<any, any, any>;
 
+type AnyScale = _Scale.Scale<any, any, any>;
 type OptionalErrorBarNodeProperties = { [K in keyof ErrorBarNodeProperties]?: ErrorBarNodeProperties[K] };
 
 class ErrorBarCapConfig implements OptionalErrorBarNodeProperties {
-    @Validate(BOOLEAN)
-    visible?: boolean = undefined;
+    @Validate(OPT_BOOLEAN)
+    visible?: boolean;
```
alantreadway: Q: Does this work without the `= undefined` - i.e. can a user specify `visible` without an options binding error?

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
```diff
@@ -88,14 +99,52 @@ export class ErrorBars
...
+        this.destroyFns.push(
+            this.cartesianSeries.addListener('data-update', (event: _ModuleSupport.SeriesDataUpdateEvent) =>
+                this.onDataUpdate(event)
+            )
+        );
+        this.destroyFns.push(
+            this.cartesianSeries.addListener('visibility-changed', (event: _ModuleSupport.SeriesVisibilityEvent) =>
+                this.onToggleSeriesItem(event)
+            )
+        );
```
alantreadway: Nit: Maybe just condense this to a single `.push()` call with multiple arguments?

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
```diff
@@ -88,14 +99,52 @@ export class ErrorBars
...
+        event.props.push(
+            valueProperty(cartesianSeries, yLowerKey, isContinuousY, { id: 'yValue' }),
+            valueProperty(cartesianSeries, yUpperKey, isContinuousY, { id: 'yValue' })
+        );
+        if (xLowerKey !== undefined && xUpperKey !== undefined) {
+            event.props.push(
+                valueProperty(cartesianSeries, xLowerKey, isContinuousX, { id: 'xValue' }),
+                valueProperty(cartesianSeries, xUpperKey, isContinuousX, { id: 'xValue' })
+            );
+        }
```
alantreadway: I would prefer if we treated the event as immutable, and instead returned additional props to append to the data-model properties, as it's a bit easier to reason about?

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
```diff
@@ -88,14 +99,52 @@ export class ErrorBars
...
-    onDataProcessed(event: {
+    onPrerequestData(event: {
+        props: _ModuleSupport.PropertyDefinition<string>[];
+        isContinuousX: boolean;
+        isContinuousY: boolean;
+    }) {
+        const { cartesianSeries, xLowerKey, xUpperKey, yLowerKey, yUpperKey } = this;
+        const { isContinuousX, isContinuousY } = event;
+        event.props.push(
+            valueProperty(cartesianSeries, yLowerKey, isContinuousY, { id: 'yValue' }),
```
olegat, line 129: Note: I think that I better solution would be to use keys like `yValueLower` and add another series event for when `Series.getDomain()` is called so that `ErrorBars` can add the error values to the domain.

That would be a better solution because right now `ErrorBars` replies on private/internal implementation of the series. This solution would allow the `ErrorBars` class to function independently of the underlying series implementation (including bar series, which doesn't work at the moment).

I plan on doing this in a follow-up PR.

packages/ag-charts-community/src/chart/series/seriesEvents.ts
olegat: I'll try that.

packages/ag-charts-community/src/chart/series/seriesEvents.ts
olegat: I would like to, but this interface needs to be compatible with the parameters of [Series.toggleSeriesItem](https://github.com/ag-grid/ag-charts/blob/ce7c48ca175ae05309f50e8add6761b0b6f7f499/packages/ag-charts-community/src/chart/series/series.ts#L785).

I'm not sure what to use instead. Open to suggestions.

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
olegat: I agree; I thought about this too initially. It's will be a bit trickier to implement because currently `Listeners.dispatch` returns `void`, but I'll see whether it's feasible to change this.

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
olegat: Done.

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
olegat: Oddly enough yes (https://plnkr.co/edit/m2TZ2vDoTYoHO6ra?open=main.js), although I'm unsure why. IIRC from another conversation that we had, this property shouldn't exist unless its set to something 🤷🏻‍♂️

packages/ag-charts-community/src/chart/series/seriesEvents.ts
olegat: `unknown` works if I update errorBar.ts to reflect this. Done.

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
olegat: I've rebased & amended the commit that removes the explicit `= undefined` just to be safe.

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
olegat: Resolved in [5e1b3d780](https://github.com/ag-grid/ag-charts/commit/5e1b3d780a7dc1be647bc7937d2085bd99ca1424)

packages/ag-charts-community/src/chart/series/cartesian/lineSeries.ts
```diff
@@ -149,6 +149,11 @@ export class LineSeries extends CartesianSeries<LineContext> {
             props.push(diff(this.processedData));
         }
 
+        const listenerProps: (typeof props)[] =
```
olegat, line 153: Note: There's quite a lot of existing & new code duplication in _lineSeries.ts_ and _scatterSeries.ts_. I've filed [AG-9434](https://ag-grid.atlassian.net/browse/AG-9434) this morning for this.

alantreadway (approved):
LGTM

PR #282, opened 2023-09-28
Add automatic import sorting

iMoses:


olegat:
LGTM, but this will almost certainly cause merge conflict with all other open PRs.

Can we hold off on merging this and re-run the formatter at EoD when other PRs have been reviewed & merged? This would make more sense than having everyone resolve conflicts on an individual basis.

alantreadway (approved):
LGTM, thanks for this!

iMoses:
> LGTM, but this will almost certainly cause merge conflict with all other open PRs.
> 
> Can we hold off on merging this re-run the formatter at EoD when other PRs have been reviewed & merged? This would make more sense than having everyone resolve conflicts on an individual basis.

I don't think it matters much because you can decline the import changes and just run formatting again, but we can also delay the merge if the team prefers. I would advise to merge right after this is pushed anyhow.

Anyhow, there's still one bug which is making sure `mdoc` files are included, I'm looking into it now.

olegat:
> > LGTM, but this will almost certainly cause merge conflict with all other open PRs.
> > Can we hold off on merging this re-run the formatter at EoD when other PRs have been reviewed & merged? This would make more sense than having everyone resolve conflicts on an individual basis.
> 
> I don't think it matters much because you can decline the import changes and just run formatting again, but we can also delay the merge if the team prefers. I would advise to merge right after this is pushed anyhow.
> 
> Anyhow, there's still one bug which is making sure `mdoc` files are included, I'm looking into it now.

I'm not fussed either way (merge this PR now or merge it EoD), but as long as everyone is onboard with the decision then I'm okay.

iMoses:
Closing this PR, @alantreadway would push these changes at EoD.

PR #283, opened 2023-09-28
Ag 9343/gallery

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-9343


markdurrant (approved):
All LGTM 👍

packages/ag-charts-website/src/content/gallery/_examples/customised-waterfall/main.ts
```diff
@@ -0,0 +1,48 @@
...
+                },
+                negative: {
+                    fill: '#FF6B6B',
+                    stroke: '#FF6B6B',
+                },
+                total: {
+                    name: 'Total / Subtotal',
+                    fill: '#404066',
+                    stroke: '#404066',
+                },
```
alantreadway: Shouldn't we avoid having these inline styles?

alantreadway (approved):
LGTM other than one comment - I think it would be better to have a naked version of that chart which we intend to fix in the next phase or two?

packages/ag-charts-website/src/content/gallery/_examples/customised-waterfall/main.ts
manapeirov: Yes for sure, but this is a customisation example shifted from the docs page for waterfall, it will probably be changed again when we get to making the gallery examples good

manapeirov: I can remove the styles for now

PR #285, opened 2023-09-28
AG-9435 - Block redundant highlight maintenance.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-9435

Follow up to #278 to reduce redundant `Chart.performUpdate()` restarts/interruption from series highlighting.

The failure case in AG-9435 happens during drag-to-pan - it appears that removing the re-highlighting behaviour during this operation resolves the processing conflict that prevents `Chart.performUpdate()` to complete in a single pass.

Bonus:
- Cleanup of the `Chart` constructor & fix for listener destruction on `Chart.destroy()`.

### DIFF
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -327,7 +327,6 @@ export abstract class Chart extends Observable implements AgChartInstance {
             this.layoutService.addListener('start-layout', (e) => this.positionCaptions(e.shrinkRect)),
             this.layoutService.addListener('layout-complete', (e) => this.layoutComplete(e)),
 
-            // Add interaction listeners last so child components are registered first.
             this.interactionManager.addListener('click', (event) => this.onClick(event)),
             this.interactionManager.addListener('dblclick', (event) => this.onDoubleClick(event)),
             this.interactionManager.addListener('hover', (event) => this.onMouseMove(event)),
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -240,6 +240,7 @@ export abstract class Chart extends Observable implements AgChartInstance {
     mode: 'standalone' | 'integrated' = 'standalone';
 
     private _destroyed: boolean = false;
+    private readonly _destroyFns: (() => void)[] = [];
     get destroyed() {
         return this._destroyed;
     }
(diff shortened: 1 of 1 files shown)


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -318,46 +320,33 @@ export abstract class Chart extends Observable implements AgChartInstance {
...
-        this.zoomManager.addListener('zoom-change', (_) =>
-            this.update(ChartUpdateType.PROCESS_DATA, { forceNodeDataRefresh: true })
+        SizeMonitor.observe(this.element, (size) => this.rawResize(size));
+        this._destroyFns.push(
+            // eslint-disable-next-line sonarjs/no-duplicate-string
+            this.layoutService.addListener('start-layout', (e) => this.positionPadding(e.shrinkRect)),
+            this.layoutService.addListener('start-layout', (e) => this.positionCaptions(e.shrinkRect)),
+            this.layoutService.addListener('layout-complete', (e) => this.layoutComplete(e)),
+
+            // Add interaction listeners last so child components are registered first.
```
lsjroberts: This comment doesn't really make a lot of sense anymore

packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -318,46 +320,33 @@ export abstract class Chart extends Observable implements AgChartInstance {
...
+            this.interactionManager.addListener('click', (event) => this.onClick(event)),
+            this.interactionManager.addListener('dblclick', (event) => this.onDoubleClick(event)),
+            this.interactionManager.addListener('hover', (event) => this.onMouseMove(event)),
+            this.interactionManager.addListener('leave', (event) => this.onLeave(event)),
+            this.interactionManager.addListener('page-left', () => this.destroy()),
+            this.interactionManager.addListener('wheel', () => this.disablePointer()),
+            this.interactionManager.addListener('wheel', () => this.disablePointer()),
+
+            // Block redundant and interfering attempts to update the hovered element during dragging.
+            this.interactionManager.addListener('drag-start', () => (this.reapplyPointerOnPerformUpdate = false)),
```
lsjroberts, line 339: This is similar to the animation's `disableInteraction` feature. Definitely different in behaviour from that specific case atm, but maybe if we do more similar things in the future we may want to look into a way of generically disabling certain or all interactions?

lsjroberts (approved):
LGTM

packages/ag-charts-community/src/chart/chart.ts
alantreadway: Agreed, I spent a few minutes trying to imagine how we might tie these together, but I didn't see a clear approach that made sense from the two data-points so far without introducing more coupling.

packages/ag-charts-community/src/chart/chart.ts
alantreadway: Done

PR #287, opened 2023-09-28
AG-9429 Refactor Axis class to use ModuleMap

olegat:
https://ag-grid.atlassian.net/browse/AG-9429


alantreadway (approved):
Nice - thanks!

PR #289, opened 2023-09-28
AG-9130 Area & line animation refactor

lsjroberts:


lsjroberts:
Closing as we are going a different direction

PR #290, opened 2023-09-29
Improve types

iMoses:
WIP

### DIFF
--- a/packages/ag-charts-community/src/chart/agChartV2.ts
+++ b/packages/ag-charts-community/src/chart/agChartV2.ts
@@ -7,6 +7,7 @@ import type {
     AgBaseSeriesOptions,
     AgChartInstance,
     AgChartOptions,
+    AgChartProxy,
 } from '../options/agChartOptions';
 import { Debug } from '../util/debug';
 import { jsonApply, jsonDiff, jsonMerge } from '../util/json';
@@ -17,6 +18,7 @@ import { CartesianChart } from './cartesianChart';
 import type { Chart, SpecialOverrides } from './chart';
 import type { ChartAxis } from './chartAxis';
 import { getJsonApplyOptions } from './chartOptions';
+import { AgChartInstanceProxy } from './chartProxy';
 import { ChartUpdateType } from './chartUpdateType';
 import { getAxis } from './factory/axisTypes';
 import { getLegendKeys } from './factory/legendTypes';
@@ -79,7 +81,7 @@ export abstract class AgChart {
     /**
      * Create a new `AgChartInstance` based upon the given configuration options.
      */
-    public static create(options: AgChartExtendedOptions): AgChartInstance {
+    public static create(options: AgChartExtendedOptions): AgChartProxy {
         return AgChartInternal.createOrUpdate(options);
     }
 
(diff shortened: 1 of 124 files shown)


packages/ag-charts-community/src/chart/agChartV2.ts
```diff
@@ -79,7 +81,7 @@ export abstract class AgChart {
     /**
      * Create a new `AgChartInstance` based upon the given configuration options.
      */
-    public static create(options: AgChartExtendedOptions): AgChartInstance {
+    public static create(options: AgChartExtendedOptions): AgChartProxy {
```
alantreadway: `AgChart.create` and `AgChartInstance` are public, and I don't really want to advertise that it's a proxy as that is an implementation detail

packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
```diff
@@ -680,13 +680,17 @@ export class ScatterSeries extends CartesianSeries<SeriesNodeDataContext<Scatter
...
-    getDatumId(datum: ScatterNodeDatum) {
-        const keys = [`${datum.xValue}`, `${datum.yValue}`];
-        if (this.labelKey) keys.push(datum.label.text);
-        return createDatumId(keys);
-    }
+    // getDatumId(datum: ScatterNodeDatum) {
+    //     const keys = [`${datum.xValue}`, `${datum.yValue}`];
+    //     if (this.labelKey) keys.push(datum.label.text);
+    //     return createDatumId(keys);
+    // }
```
alantreadway, line 687: ?

packages/ag-charts-community/src/util/json.ts
```diff
@@ -246,6 +246,7 @@ export type JsonApplyParams = {
  *
  * @param target to apply source JSON properties into
  * @param source to be applied
+ * @param params
```
alantreadway, line 249: Needed?

packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -693,13 +687,13 @@ export abstract class Chart extends Observable implements AgChartInstance {
         return this._axes;
     }
 
-    protected _series: Series[] = [];
-    set series(values: Series[]) {
+    protected _series: Series<any>[] = [];
```
alantreadway, line 693: Guessing we can't use `Series<unknown>` here and elsewhere since we typically don't care about this parameter at this level of abstraction?

packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
iMoses: Unused code. I commented out because the code that used this method is also commented out in the same file.

packages/ag-charts-community/src/util/json.ts
iMoses: Not really. My IDE JSDoc analyzers just yells at me that it isn't defined as part of the JSDoc 😅

packages/ag-charts-community/src/chart/chart.ts
iMoses: I didn't wanna make this PR too big but that's definitely the next step. 
I want the user define `datum` to be `unknown` everywhere, but it'll require a bit more change in the codebase.

alantreadway: 👍

packages/ag-charts-community/src/options/chart/chartBuilderOptions.ts
```diff
@@ -0,0 +1,31 @@
...
+export interface AgPolarChartOptions extends AgBasePolarChartOptions, AgBaseChartOptions {
+    theme?: AgChartTheme | AgChartThemeName;
+}
+export interface AgHierarchyChartOptions extends AgBaseHierarchyChartOptions, AgBaseChartOptions<any> {
+    theme?: AgChartTheme | AgChartThemeName;
+}
+export type AgChartOptions = AgCartesianChartOptions | AgPolarChartOptions | AgHierarchyChartOptions;
+
+export interface AgChartInstance {
+    className?: string;
```
alantreadway: Is this new? It probably shouldn't be in the public contract as our classnames are part of our internal implementation.

packages/ag-charts-community/src/options/chart/chartBuilderOptions.ts
```diff
@@ -0,0 +1,31 @@
...
+    /** Get the `AgChartOptions` representing the current chart configuration. */
+    getOptions(): AgChartOptions;
+    /** Destroy the chart instance and any allocated resources to support its rendering. */
+    destroy(): void;
+}
+
+export interface AgChartProxy extends AgChartInstance {
```
alantreadway: This should also not be part of the public options. (Anything under src/options/ should be considered part of the public contract)

alantreadway (commented):
LGTM other than the specific cases pointed out - if you can remove the `AgChartInstance.className` and `AgChartProxy` elements from the public options contract I'm happy :+1:

alantreadway (approved):
LGTM - thanks!

PR #292, opened 2023-09-29
Enforce Typescript override modifier.

alantreadway:
Tighten up our our overrides so that renames on parents trigger compile failures on children.

Interestingly, also highlights a bunch of places we're overriding but probably shouldn't....


iMoses (approved):
LGTM 👍

olegat:
Good to see more compile-time checks being enforced. Thanks for making this change 👍

PR #293, opened 2023-09-29
Pull up animation state machine declaration to PolarSeries.

alantreadway:
Moving `PolarSeries` to follow a similar pattern as `CartesianSeries` for animation `StateMachine` handling.

### DIFF
--- a/packages/ag-charts-community/src/chart/series/polar/polarSeries.ts
+++ b/packages/ag-charts-community/src/chart/series/polar/polarSeries.ts
@@ -7,9 +7,9 @@ import type { DataModel, ProcessedData } from '../../data/dataModel';
 import type { SeriesNodeDataContext, SeriesNodeDatum } from '../series';
 import { Series, SeriesNodePickMode } from '../series';
 
-type PolarAnimationState = 'empty' | 'ready' | 'waiting' | 'clearing';
-type PolarAnimationEvent = 'update' | 'updateData' | 'clear';
-type PolarAnimationData = { duration: number };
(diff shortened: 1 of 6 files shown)


packages/ag-charts-community/src/chart/series/polar/polarSeries.ts
```diff
@@ -1,11 +1,16 @@
...
 import type { BBox } from '../../../scene/bbox';
 import type { PointLabelDatum } from '../../../util/labelPlacement';
 import { ChartAxisDirection } from '../../chartAxisDirection';
 import type { DataModel, ProcessedData } from '../../data/dataModel';
 import type { SeriesNodeDataContext, SeriesNodeDatum } from '../series';
 import { Series, SeriesNodePickMode } from '../series';
 
+type PolarAnimationState = 'empty' | 'ready' | 'waiting' | 'clearing';
+type PolarAnimationEvent = 'update' | 'updateData' | 'clear';
+type PolarAnimationData = { duration: number };
```
lsjroberts: Should be optional `duration?: number`, it's only used when `animateClearingUpdateEmpty()` calls `this.animationState.transition('update', { duration: updateDuration });`

lsjroberts (approved):
All good but one comment

packages/ag-charts-community/src/chart/series/polar/polarSeries.ts
alantreadway: Fixed

PR #294, opened 2023-10-02
AG-9014 Extend domain for Bar Series with Error Bars

olegat:

This PR fixes a bug where the domain was not being extended for Bar Series. The `errorBar.ts` module file now adds properties `'xValue-errors'` and `'yValue-errors'` to the data model, and uses these property IDs to determine the domain of the errors. The `Series.getDomain` method now calculates the union of the current series + any modules attached to this series.

I did some refactoring of the `Series` abstract class that affects the derived classes:

- Renamed `Series.getDomain` to `Series.getSeriesDomain` (see [commit 583cbe121](https://github.com/ag-grid/ag-charts/pull/294/commits/583cbe121dd625081a6931ab6c5d7c965bbf0159) for more info).
- Removed `fixNumericExtend` (see [commit 3fff2baf](https://github.com/ag-grid/ag-charts/pull/294/commits/3fff2baff9037bd20e82880b4bd1a43ef46a3652) for more info).


packages/ag-charts-community/src/chart/series/series.ts
```diff
@@ -569,7 +568,16 @@ export abstract class Series<
...
+    // The union of the series domain ('community') and series-option domains ('enterprise').
+    getDomain(direction: ChartAxisDirection): any[] {
+        const seriesDomain: any[] = this.getSeriesDomain(direction);
+        const moduleDomains: any[][] = this.dispatch('data-getDomain', { direction }) ?? [];
+        // Flatten the 2D moduleDomains into a 1D array and concatenate it with seriesDomain
+        return moduleDomains.reduce((total, current) => total.concat(current), seriesDomain);
+    }
+
+    // Get the 'community' domain (excluding any additional data from series-option modules).
+    abstract getSeriesDomain(direction: ChartAxisDirection): any[];
```
alantreadway, line 580: Maybe we should name this `getDataDomain()`, as the `Series` seems redundant otherwise in the context of a method on `Series`?

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
```diff
@@ -19,7 +20,20 @@ const {
 } = _ModuleSupport;
 const { Logger } = _Util;
 
+const XVALUE_ERRORS_ID = 'xValue-errors';
+const YVALUE_ERRORS_ID = 'yValue-errors';
+
+type AnyCartesianSeries = _ModuleSupport.CartesianSeries<any, any>;
```
alantreadway: Nit: Could you try `unknown` rather than `any` if these parameter types aren't important for us to know here?

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
```diff
@@ -108,51 +125,67 @@ export class ErrorBars
...
             );
         }
         return props;
     }
 
-    onDataUpdate(event: {
-        dataModel: _ModuleSupport.DataModel<any, any, any>;
-        processedData: _ModuleSupport.ProcessedData<any>;
-    }) {
+    onDataProcessed(event: SeriesDataProcessedEvent) {
```
alantreadway: Nit: Can this now be `private`?

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
```diff
@@ -108,51 +125,67 @@ export class ErrorBars
...
+    }
+
+    private hasAxis(direction: _ModuleSupport.ChartAxisDirection): boolean {
+        if (direction == ChartAxisDirection.X) {
+            return this.xLowerKey !== undefined && this.xUpperKey != undefined;
+        }
+        return true;
+    }
+
+    onGetDomain(event: SeriesDataGetDomainEvent) {
```
alantreadway: `private`?

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
olegat: `<unknown, unknown>` wouldn't compile, but I've used `<_Scene.Node, _ModuleSupport.CartesianSeriesNodeDatum>` which should work for this type. There doesn't to be much that I can do about the types though (`DataModel`, `ProcessedData`, `Scale`).

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
olegat: Done

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
olegat: Done

packages/ag-charts-community/src/chart/series/series.ts
olegat: I somewhat agree about the redundancy.

However the name `getDataDomain()` is a bit too vague; the lower/upper error bounds are also part of the input data, so this method name could be misunderstood to mean "domain of the data in the series and the series-options modules".

I've deliberately named this `getSeriesDomain()` to clarify that this is the domain of the series excluding any extra data points from series-options modules.

I initially used the name `getLocalDomain()` (series) and `getTotalDomain()` (series + series-options), but the wording `Local` felt too abstract.

alantreadway: Ok - let's roll with it then 👍

alantreadway (approved):
LGTM

PR #296, opened 2023-10-02
Extract fromToMotion utility function to reduce BarSeries animation boilerplate.

alantreadway:


packages/ag-charts-community/src/motion/fromToMotion.ts
```diff
@@ -0,0 +1,139 @@
...
+ * @param diff optional diff from a DataModel to use to detect added/moved/removed cases
+ */
+export function fromToMotion<N extends Shape, T extends AnimationValue & Partial<N>, D>(
+    id: string,
+    animationManager: AnimationManager,
+    selections: Selection<N, D>[],
+    fromFn: (node: N, datum: D, state: NodeUpdateState) => T,
+    toFn: (node: N, datum: D, state: NodeUpdateState) => T,
+    extraOpts: Partial<AnimationOptions<T> & AdditionalAnimationOptions> = {},
+    idFn?: (node: N, datum: D) => string,
```
lsjroberts: `idFn` is called `getDatumId` everywhere else, probably worth keeping that aligned

packages/ag-charts-community/src/motion/fromToMotion.ts
```diff
@@ -0,0 +1,139 @@
...
+        for (const node of selection.nodes()) {
+            let status: NodeUpdateState = 'added';
+            if (idFn && diff) {
+                status = calculateStatus(node, node.datum, idFn, ids);
+            }
+
+            cleanup ||= status === 'removed';
+
+            const from = typeof fromFn === 'function' ? fromFn(node, node.datum, status) : fromFn;
+            const to = typeof toFn === 'function' ? toFn(node, node.datum, status) : toFn;
```
lsjroberts: Are these `typeof` checks needed? Looks like they are always functions from the function param types.

lsjroberts (approved):
LGTM

packages/ag-charts-community/src/motion/fromToMotion.ts
alantreadway: Good call, done!

packages/ag-charts-community/src/motion/fromToMotion.ts
alantreadway: Fair - removed.

PR #298, opened 2023-10-02
Apply fromToMotion to HistogramSeries.

alantreadway:


packages/ag-charts-community/src/chart/series/cartesian/barUtil.ts
```diff
@@ -138,3 +141,34 @@ export function prepareAnimationFunctions<T extends AnimatableBarDatum>(
...
+
+    let startingX = Infinity;
+    let startingY = 0;
+
+    if (isColumnSeries) {
+        const isContinuousY = yAxis?.scale instanceof ContinuousScale;
+        if (isContinuousY) {
+            startingY = yAxis.scale.convert(0);
+        } else if (yAxis) {
+            startingY = yAxis.scale.convert(Math.max(...yAxis.range));
```
lsjroberts, line 162: Previously, this would only take into account values where `datum.yValue >= 0`, so it would get the correct line when values are negative as well (or both). Is that implied in the range/conversion? Not sure any of the tests cover this atm.

alantreadway: Discussed offline, I'm not 100% sure when this would come into play since this is only for non-continuous cases (we handle the `0` case for continuous axes in the previous if clause)

PR #300, opened 2023-10-02
AG-9434 Cleanup Duplicate Code in Series

olegat:
https://ag-grid.atlassian.net/browse/AG-9434


packages/ag-charts-community/src/chart/series/dataModelSeries.ts
olegat, line 1: Alternative names for this class could be `TabularSeries` or `RecordSeries` (something to emphasis that the data of this series is like a table, unlike a hierarchal series where each data point can have 0..n children).

olegat:
These changes use PR #294 as the base. So let's wait for that PR to merge first.

alantreadway (approved):
LGTM, nice cleanup!

PR #302, opened 2023-10-03
Apply fromToMotion to more series types.

alantreadway:
- Converts `RangeBarSeries` to use `fromToMotion`.
- Partially converts `WaterfallSeries` to use `fromToMotion`.
- Converts `BoxPlotSeries` to use `fromToMotion`.
- Factors out common selection reset handling to a new `resetMotion` function.
- Moves `Shape.setProperties()` up to `Node`.

### DIFF
--- a/packages/ag-charts-enterprise/src/series/box-plot/boxPlotSeries.ts
+++ b/packages/ag-charts-enterprise/src/series/box-plot/boxPlotSeries.ts
@@ -1,11 +1,12 @@
-import type {
-    AgBoxPlotSeriesFormatterParams,
-    AgBoxPlotSeriesStyles,
-    AgBoxPlotSeriesTooltipRendererParams,
+import {
+    type AgBoxPlotSeriesFormatterParams,
+    type AgBoxPlotSeriesStyles,
+    type AgBoxPlotSeriesTooltipRendererParams,
     _Scene,
 } from 'ag-charts-community';
 import { _ModuleSupport, _Scale, _Util } from 'ag-charts-community';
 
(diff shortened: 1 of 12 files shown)


packages/ag-charts-enterprise/src/series/box-plot/boxPlotSeries.ts
```diff
@@ -450,36 +452,17 @@ export class BoxPlotSeries extends CartesianSeries<
...
+        motion.resetMotion(datumSelections, resetBoxPlotSelectionsScalingCenterFn(isVertical));
+
+        const { from, to } = prepareBoxPlotFromTo(isVertical);
+        motion.staticFromToMotion(
+            `${this.id}_empty-update-ready`,
+            this.ctx.animationManager,
+            datumSelections,
+            from,
+            to
+        );
```
iMoses, line 465: We're missing the `onStop`

alantreadway: Added a default `onStop()` to the `fromToMotion` functions - @lsjroberts does that make sense?

lsjroberts: It may change some behaviours in the situation where new animations are started before the old ones have finished. But I can't tell from looking at the code. Best to check it against the examples on the animation page.

lsjroberts: But that may be based on outdated understanding of the animation manager too

alantreadway: 👍 All looks good on the animation page from what I can see.

lsjroberts (approved):

PR #304, opened 2023-10-03
Factor out common label animations.

alantreadway:


packages/ag-charts-community/src/chart/series/seriesLabelUtil.ts
```diff
@@ -0,0 +1,21 @@
...
+    labelSelections: Selection<NodeWithOpacity, T>[]
+) {
+    const duration = animationManager.defaultDuration;
+    staticFromToMotion(
+        `${id}_empty-update-ready_labels`,
+        animationManager,
+        labelSelections,
+        { opacity: 0 },
+        { opacity: 1 },
+        { delay: duration, duration: animationManager.defaultDuration * 0.2 }
```
lsjroberts: Current requirements put this as hardcoded to 200ms, regardless the duration of the rest of the animation. Pending further expansion of https://ag-grid.atlassian.net/browse/AG-9189

lsjroberts (approved):
lgtm

packages/ag-charts-community/src/chart/series/seriesLabelUtil.ts
alantreadway: Ok - I'll switch back 👍

alantreadway: Fixed - thanks for pointing that out

PR #307, opened 2023-10-04
Labels formatters refactor WIP

iMoses:


packages/ag-charts-community-examples/src/charts-overview/examples/bar-with-labels/main.ts
```diff
@@ -31,7 +31,7 @@ const options: AgChartOptions = {
         fontWeight: "bold",
         color: "white",
         formatter: params => {
-          return (params.value > 0 ? "+" : "") + params.value
+          return (params.defaultValue > 0 ? "+" : "") + params.defaultValue
```
alantreadway, line 34: Can you make sure to capture all breaking changes in the related JIRA `Breaking Changes` field?

packages/ag-charts-community/src/chart/chart.test.ts
```diff
@@ -2,21 +2,18 @@ import { afterEach, beforeEach, describe, expect, it, jest } from '@jest/globals
...
 import { Selection } from '../scene/selection';
 import { Rect } from '../scene/shape/rect';
 import { Sector } from '../scene/shape/sector';
 import { AgChart } from './agChartV2';
 import type { Chart } from './chart';
+import type { AgChartProxy } from './chartProxy';
 import { Circle } from './marker/circle';
 import {
     clickAction,
+    createChart as createChartUtil,
```
alantreadway: Nit: Import without the alias?

packages/ag-charts-community/src/chart/legend.ts
```diff
@@ -889,11 +888,10 @@ export class Legend {
         const calculateTranslationPerpendicularDimension = () => {
             switch (this.position) {
                 case 'top':
+                case 'left':
```
alantreadway, line 891: Nit: Can you either add a comment to indicate this is a deliberate fall-through, or undo this combination?

(SonarCloud will complain about this change in it's current state)

packages/ag-charts-community/src/util/validation.ts
```diff
@@ -215,21 +215,7 @@ export const OPT_BOOLEAN_ARRAY = predicateWithMessage(
...
-    'lighter',
-    '100',
-    '200',
-    '300',
-    '400',
-    '500',
-    '600',
-    '700',
-    '800',
-    '900',
```
alantreadway, line 231: This might be a breaking change.

packages/ag-charts-enterprise/src/series/heatmap/heatmapSeries.ts
```diff
@@ -219,16 +212,9 @@ export class HeatmapSeries extends _ModuleSupport.CartesianSeries<_Scene.Rect, H
...
-                    seriesId: this.id,
-                    value: labelValue,
-                };
-                text = callbackCache.call(label.formatter, labelFormatterParams) ?? '';
-            } else {
-                text = String(labelValue);
-            }
+            const text = label.formatter
+                ? callbackCache.call(label.formatter, { seriesId: this.id, datum, defaultValue: labelValue }) ?? ''
+                : String(labelValue);
```
alantreadway: Let's avoid these non-trivial ternary statements - SonarCloud is going to pull us up over it, and on review a simple `if...else...` block is typically more readable for these cases.

packages/ag-charts-community/src/util/validation.ts
iMoses: Added a note on Jira

packages/ag-charts-community/src/chart/legend.ts
iMoses: I'm not entirely sure what it'll complain about. 
These two values simply return the same result logically.
How does it differ from the rest of the code on the same file? 
e.g.
```
            switch (this.position) {
                case 'top':
                case 'bottom':
                    translationX = (shrinkRect.width - legendBBox.width) / 2;
                    translationY = calculateTranslationPerpendicularDimension();
                    newShrinkRect.shrink(legendBBox.height, this.position);
                    break;

                case 'left':
                case 'right':
                default:
                    translationX = calculateTranslationPerpendicularDimension();
                    translationY = (shrinkRect.height - legendBBox.height) / 2;
                    newShrinkRect.shrink(legendBBox.width, this.position);
            }
```

packages/ag-charts-community-examples/src/charts-overview/examples/bar-with-labels/main.ts
iMoses: Added a note in Jira. I'll go over all the changes with David as well to make sure everything is noted.

packages/ag-charts-enterprise/src/series/heatmap/heatmapSeries.ts
iMoses: reverted

alantreadway (approved):

packages/ag-charts-community/src/chart/legend.ts
alantreadway: Let's roll with it, I can't find the rule I'm thinking of in the current active SonarCloud ruleset, let's just deal with anything (or not) the report flags up - this is an example of what it has required previously:
https://github.com/ag-grid/ag-charts/blob/ce01f3cdeadc6bcd1d0a258c8b5f23df84413ef4/packages/ag-charts-community/src/scale/timeScale.ts#L154

...but maybe that is a different pattern since each clause is actually doing something.

PR #311, opened 2023-10-04
AG-8238 Add e2e tests for Error Bars

olegat:
https://ag-grid.atlassian.net/browse/AG-8238

Currently there's just 5 simple test cases:
1. One line series
2. One bar series
3. Two line series
4. Two bar series
5. One scatter series

I've noticed a bug in the "Two bar series" test case: the X coords of the error bars is incorrect. I'll address this in a follow-up


alantreadway (approved):
LGTM

PR #313, opened 2023-10-05
AG-9014 Error Bar improvements for Bar Charts

olegat:

Use the `Series.contextNodeData` to get the `midPoint` rendering value. This makes the `errorBar.ts` file render the error bars in the centre of the bars in Bar Charts.

Also flip the X/Y keys in horizontal Bar Charts so that horizontal whiskers are used. The vertical whiskers are now optional (which I suppose makes sense, it would also allow scatter series with confidence intervals in just 1 axis).


alantreadway (approved):
LGTM - thanks for reworking to use nodeData, looks much neater 👍

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
```diff
@@ -182,104 +181,79 @@ export class ErrorBars
...
-        const xIndex = 'xValue';
-        const yIndex = this.cartesianSeries.type == 'bar' ? 'yValue-end' : 'yValue';
-        return {
-            xIndex: dataModel.resolveProcessedDataIndexById(this.cartesianSeries, xIndex).index,
-            yIndex: dataModel.resolveProcessedDataIndexById(this.cartesianSeries, yIndex).index,
-        };
-    }
+    private shouldFlipXY(): boolean {
+        if (this.cartesianSeries.type === 'bar') {
+            return (this.cartesianSeries as unknown as _ModuleSupport.BarSeries).direction === 'horizontal';
```
olegat: I'm not sure what's the best way to check if the bar series is horizontal. In theory, this `as unknown as BarSeries` should in theory be safe because the `type` property is being checked. As said, they may be a cleaner way to happen this edge case.

Alternatives could be:

- Using `instanceof` instead of `as`
- Providing the `direction` value as part of the module context.

olegat: Actually I've move `shouldFlipXY()` has part of `CartesianSeries` (and `BarSeries` overrides it). I find that this a much cleaner approach than having _errorBar.ts_ dig into the implementation details of the underlying series.

alantreadway (approved):

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
alantreadway: SGTM - thanks!

PR #314, opened 2023-10-05
Refactor RadarColumnBaseSeries and sub-classes to use fromToMotion()

alantreadway:


packages/ag-charts-enterprise/src/series/nightingale/nightingaleSeries.ts
```diff
@@ -2,12 +2,18 @@ import { _ModuleSupport, _Scale, _Scene, _Util } from 'ag-charts-community';
...
 
-const { Sector } = _Scene;
+const { seriesLabelFadeInAnimation } = _ModuleSupport;
+const { Sector, motion } = _Scene;
 
 export class NightingaleSeries extends RadialColumnSeriesBase<_Scene.Sector> {
     static className = 'NightingaleSeries';
 
+    constructor(moduleCtx: _ModuleSupport.ModuleContext) {
+        super(moduleCtx, { animationResetFns: { item: resetRadialColumnSelectionFn } });
```
alexanderby, line 14: Why is `resetRadialColumnSelectionFn` not a class method to be overriden by sub-classes?

alexanderby (approved):

lsjroberts (approved):

packages/ag-charts-enterprise/src/series/nightingale/nightingaleSeries.ts
alantreadway: Ultimately we need to move towards a composition approach rather than class-hierarchy approach for series behaviours - for animation I'm attempting to use re-usable functions for this purpose.

PR #320, opened 2023-10-06
Improve BarSeries add/update/remove animation

alantreadway:
Adds the ability for `to`/`from` animation functions to influence animation delay/duration.


packages/ag-charts-enterprise/src/series/histogram/__image_snapshots__/histogram-series-test-ts-histogram-series-add-animation-for-simple-histogram-should-animate-at-25-1-snap.png
lsjroberts, line 1: Is this change what you were expecting to see? If it is only adding data, does it delay before doing the add animation? Or am I misinterpreting this?

alantreadway: So the add happens towards the end of the animation cycle, avoiding bar overlaps, so I think this look ok to me?

PR #321, opened 2023-10-06
AG-9017 Cap length configuration

olegat:

- Adds support for `series.errorBar.cap.length` and `series.errorBar.cap.lengthRatio`
- Adds the optional `capLength` property to `CartesianSeriesNodeDatum`.
  - This is the multiplier for `lengthRatio`
  - This allows the _errorBar.ts_ file to remain series-agnostic, because the multiplier is implemented by the underlying series.
  - Changes to some other datum types was needed to omit this property.
- Adds several test cases.
- Tested on line, scatter, bar (vertical) and bar (horizontal) series.


packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts
```diff
@@ -40,6 +40,10 @@ export interface CartesianSeriesNodeDatum extends SeriesNodeDatum {
     readonly yKey?: string;
     readonly xValue?: any;
     readonly yValue?: any;
+    // Caps can appear on bar, line and scatter series. The length is determined
+    // by the diameter of the marker (line, scatter), width of the bar (vertical
+    // bars), or height of the bar (horizontal bars)
+    readonly capLength?: number;
```
olegat: `capLength` is a bit of misnomer, because this is actually the multiplier for `cap.lengthRatio`. Some other names to consider:

1. `capLengthRatioMultiplier`
2. `capMaxLength`

The first option is preferred because it's more specific and less likely to be misinterpreted. It is however much more verbose.

The second option is acceptable, but can be misleading because there are no checks in place that prevent the user from inputting a `cap.length` > `capMaxLength`, and such a check is probably undesirable (e.g. someone might want error bar caps that are wider than the markers on a line/scatter series).

olegat: This has been changed to an object `capDefaults?: { lengthRatio: { defaultValue: number; multiplier: number } }`. Just using 1 number is insufficient because the defaults or line/scatter depends on the marker size, where bar series depend on the ratio size and rect width. So it's necessary to include also a "default ratio" to address this discrepancy.

plugins/ag-charts-build-tools/src/executors/generate-code-reference-files/generate-code-reference-files.mjs
alantreadway, line 1: Changes here generally make me a bit nervous - have you verified the impact of your changes on the `dist/packages/ag-charts-community/*.json` files by diff'ing them before and after this change?

packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
```diff
@@ -42,7 +42,7 @@ import type { CartesianAnimationData, CartesianSeriesNodeDatum } from './cartesi
 import { CartesianSeries, CartesianSeriesMarker, CartesianSeriesNodeClickEvent } from './cartesianSeries';
 import { getMarkerConfig, updateMarker } from './markerUtil';
 
-interface MarkerSelectionDatum extends Required<CartesianSeriesNodeDatum> {
+interface MarkerSelectionDatum extends Required<Omit<CartesianSeriesNodeDatum, 'capDefaults'>> {
```
alantreadway: Just wondering if it would be possible to go the other way around, and have an extension of `CartesianSeriesNodeDatum` which is the base for the affected cases, rather than modifying the unaffected cases?

plugins/ag-charts-build-tools/src/executors/generate-code-reference-files/generate-code-reference-files.mjs
olegat: I haven't checked the JSON files but I'll re-run a build before & after these changes to make sure.

packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
olegat: That sounds like a very good idea. We already do something similar with tooltip parameters (e.g. [AgLineSeriesTooltipRendererParams](https://github.com/ag-grid/ag-charts/blob/cbae49fb6a8b83624c10ba67112c86f392ec926e/packages/ag-charts-community/src/options/series/cartesian/lineOptions.ts#L12) extends [AgCartesianSeriesTooltipRendererParams](https://github.com/ag-grid/ag-charts/blob/cbae49fb6a8b83624c10ba67112c86f392ec926e/packages/ag-charts-community/src/options/series/cartesian/cartesianSeriesTooltipOptions.ts#L3) and [AgErrorBoundSeriesTooltipRendererParams](https://github.com/ag-grid/ag-charts/blob/cbae49fb6a8b83624c10ba67112c86f392ec926e/packages/ag-charts-community/src/options/series/cartesian/cartesianSeriesTooltipOptions.ts#L15))

Having a separate `ErrorBoundNodeDatum` which the line, bar and scatter series node types can extends would be a safer option

olegat:
I've squashed the commits to tidy up this PR. There was a lot of unnecessary changes in the previous chain of commits

Old commits: https://github.com/ag-grid/ag-charts/commits/951e54efcfcec5677ecedafba7b2a814446e69b6
Squashed commit: https://github.com/ag-grid/ag-charts/commit/1bf4b5e1215d1186fed905e66b474aac74f97591

I've also removed the changes to _generate-code-reference-files.mjs_

plugins/ag-charts-build-tools/src/executors/generate-code-reference-files/generate-code-reference-files.mjs
olegat: The output JSON files seems acceptable:
https://gist.github.com/olegat/bc4b3d955cc8a0d99c810d988a831df6

packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
olegat: Done. https://github.com/ag-grid/ag-charts/blob/1bf4b5e1215d1186fed905e66b474aac74f97591/packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts#L45

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
```diff
@@ -11,14 +11,56 @@ const {
...
+        _ModuleSupport.BarSeries,
+        _ModuleSupport.LineSeries,
+        _ModuleSupport.ScatterSeries,
+    ];
+    for (const ctor of supportedSeriesTypes) {
+        const series: ErrorBoundCartesianSeries | undefined = typeGuard(ctx, ctor);
+        if (series !== undefined) {
+            return series;
+        }
+    }
```
alantreadway, line 56: I think it would be sufficient here to use the string literals `'bar', 'line', 'scatter'`, since under the covers in `typeGuard()` you are just doing a string comparison anyway - and avoid directly referencing/depending on the implementation types?

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
```diff
@@ -91,26 +134,18 @@ export class ErrorBars
...
         super();
 
-        const supportedSeriesTypes = ['bar', 'line', 'scatter'];
-        if (!supportedSeriesTypes.includes(ctx.series.type)) {
-            throw new Error(
-                `AG Charts - unsupported series type '${
-                    ctx.series.type
-                }', error bars supported series types: ${supportedSeriesTypes.join(', ')}`
-            );
-        }
```
alantreadway, line 112: Regarding my previous comment, this looks good enough - perhaps we could just use the options typing for the `supportedSeriesTypes` array to get some compiler assistance that the elements are valid?

alantreadway (approved):
Generally LGTM - happy if you want to send a follow-up PR to address the outstanding comments?

olegat:
> Generally LGTM - happy if you want to send a follow-up PR to address the outstanding comments?

Acknowledged. I'll see if there's another way to enforce type safety without exporting the `LineSeries`, `BarSeries` and `ScatterSeries` types in the follow-up PR.

PR #323, opened 2023-10-09
AG-9272 Fix errorbar tooltip docs/example

olegat:
The xValue / yValue tooltip parameters are no longer provided to the formatter. Use datum[params.xKey] / datum[params.yKey] instead.

Also update errorBar.ts to reflect this (no longer provides the values for error data).

For more info see:
https://github.com/ag-grid/ag-charts/pull/307


alantreadway (approved):
LGTM

PR #333, opened 2023-10-11
Normalize label formatters, make sure params are received as expected

iMoses:


alantreadway (approved):
LGTM, thanks for all of the clean-up along the way!

Just to confirm, have you verified the generated API sections/API explorer/API docs are working, or is there follow-up work needed for these to be correct?

iMoses:
> LGTM, thanks for all of the clean-up along the way!
> 
> Just to confirm, have you verified the generated API sections/API explorer/API docs are working, or is there follow-up work needed for these to be correct?

I'm checking them now. The only follow up is reverting `defaultValue` on the label's formatter params to `value`

PR #336, opened 2023-10-11
AG-9022 ErrorBar Styling & Cap Length Documentation

olegat:
https://ag-grid.atlassian.net/browse/AG-9022

Add sections to error-bars/index.mdoc:
- Customisation.Styling
- Customisation.CapLength
- Customisation.CapLength.LengthRatio
- Customisation.CapLength.AbsoluteLength

Add examples:
-   packages/ag-charts-website/src/content/docs/error-bars/_examples/cap-length
-   packages/ag-charts-website/src/content/docs/error-bars/_examples/cap-lengthRatio
-   packages/ag-charts-website/src/content/docs/error-bars/_examples/styling/

Also make minor tweaks to the existing documentation.

### DIFF
--- a/packages/ag-charts-community/src/options/chart/errorBarOptions.ts
+++ b/packages/ag-charts-community/src/options/chart/errorBarOptions.ts
@@ -1,10 +1,46 @@
-export interface AgErrorBarOptions {
+import type { StrokeOptions } from '../series/cartesian/commonOptions';
+import type { PixelSize, Ratio } from './types';
+
+interface ErrorBarStylingOptions extends StrokeOptions {
+    /** Whether to display the error bars. */
+    visible?: boolean;
+}
+
+interface ErrorBarCapLengthOptions {
+    /** Absolute length of caps in pixels. */
+    length?: PixelSize;
+    /** Length of caps relative to the shape used by the series. */
+    lengthRatio?: Ratio;
+}
+
+interface ErrorBarDataOptions {
+    /** The key to use to retrieve lower bound error values from the x axis data. */
+    xLowerKey?: string;
+    /** Human-readable description of the lower bound error value for the x axis. This is the value to use in tooltips or labels. */
+    xLowerName?: string;
+    /** The key to use to retrieve upper bound error values from the x axis data. */
+    xUpperKey?: string;
+    /** Human-readable description of the upper bound error value for the x axis. This is the value to use in tooltips or labels. */
+    xUpperName?: string;
+
     /** The key to use to retrieve lower bound error values from the y axis data. */
-    yLowerKey: string;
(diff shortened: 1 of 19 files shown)


olegat:
This PR depends on https://github.com/ag-grid/ag-charts/pull/321

packages/ag-charts-community/src/options/chart/errorBarOptions.ts
```diff
@@ -1,10 +1,47 @@
...
+
+export interface AgErrorBarCapOptions extends ErrorBarCapLengthOptions, ErrorBarStylingOptions {}
+
+export interface AgErrorBarOptions extends AgErrorBarDataOptions, ErrorBarStylingOptions {
+    /** Options to style error bars' caps */
+    cap?: AgErrorBarCapOptions;
+}
+
+export interface AgErrorBarSeriesOptions {
+    // TODO: relative URL ./error-bars resolves to /error-bars instead of /[lang]/error-bars
```
olegat: We should consider addressing this bug. I feel that including links to the error bars page within the line, bar, and scatter series pages would enhance the documentation's overall navigation

alantreadway: So I wouldn't include links like this in the JSDocs - these JSDocs are presented in IDE auto-complete for users, and the Markdown style links don't make any sense there?

packages/ag-charts-website/src/content/docs/error-bars/index.mdoc
```diff
@@ -54,7 +57,7 @@ data: [
         pressure: 1.2,
         pressureLower: 1.1,
         pressureUpper: 1.4
-    },
+    }, ...
```
alantreadway, line 60: Intentional?

packages/ag-charts-community/src/options/chart/errorBarOptions.ts
olegat: That's unfortunate. In that case it would be better to remove these Markdown links for now.

But we may want to revisit this feature at some point, I've logged a backlog issue for this: https://ag-grid.atlassian.net/browse/AG-9562

packages/ag-charts-website/src/content/docs/error-bars/index.mdoc
olegat: Yes. The intent here is to say that the `data` array includes 1 or more entries that follow this pattern.

Open to feedback on how to better convey this.

packages/ag-charts-community/src/options/chart/errorBarOptions.ts
olegat: Link removed.

PR #340, opened 2023-10-12
PR #321 Follow Up

olegat:
Follow up for PR #321 

This prevents the need of exporting `BarSeries`, `LineSeries` and
`ScatterSeries`.

This losses the type safety check that the `TDatum` type of `CartesianSeries`
extends the `ErrorBoundSeriesNodeDatum` type. This introduces two risks:

1.  If the `BarSeries`, `LineSeries`, `ScatterSeries` are changed such that the
    `TDatum` type no longer extends the `ErrorBoundSeriesNodeDatum`, then
    this will cause a bug regression.

2.  If a new series types is added/modified with the intent of supporting
    error bars, then there will be runtime errors if the series does not
    respect the assumptions of _errorBar.ts_.

The first risk is acceptable because our tests should catch this.

The second risk is acceptable because the developer writing this should
notice the problem.

Note: the `AgErrorBarSupportedSeriesTypes` constant is required because the
_errorBar.ts_ file needs a `string[]` to perform its checks. Converting a
`string[]` to a union is trivial, but converting a union to a `string[]` is
more complicated. See:
https://github.com/microsoft/TypeScript/issues/13298#issuecomment-820634611


olegat:
This depends on PR https://github.com/ag-grid/ag-charts/pull/336

alantreadway:
I updated the base-branch to point to the branch for #336 to try and cleanup the diff a bit, but there are conflicts? Sorry if this was the wrong thing to do.

The approach looks good, ping me when the dependent branch is merged and/or conlifcts resolved?

olegat:
> I updated the base-branch to point to the branch for #336 to try and cleanup the diff a bit, but there are conflicts? Sorry if this was the wrong thing to do.
> 
> The approach looks good, ping me when the dependent branch is merged and/or conlifcts resolved?

Let's wait for #336 to get approved & merged. This PR isn't urgent so we can delay it.

olegat:
Rebase & resolved conflicts. Ready for review.

alantreadway (approved):
LGTM

PR #348, opened 2023-10-13
AG-9560 Do not apply errorBar module on unsupported series

olegat:
https://ag-grid.atlassian.net/browse/AG-9560

### DIFF
--- a/packages/ag-charts-community/src/chart/agChartV2.ts
+++ b/packages/ag-charts-community/src/chart/agChartV2.ts
@@ -510,7 +510,16 @@ function applySeriesOptionModules(series: Series<any>, options: AgBaseSeriesOpti
 
     for (const mod of seriesOptionModules) {
         if (mod.optionsKey in options) {
-            series.getModuleMap().addModule(mod);
+            const supportedSeriesTypes: readonly string[] = mod.seriesTypes;
+            if (!supportedSeriesTypes.includes(series.type)) {
+                console.warn(
+                    `AG Charts - series-option '${mod.optionsKey}' does not support series type '${
+                        series.type
+                    }' (allowed series include: '${supportedSeriesTypes.join("', '")}')`
+                );
+            } else {
+                series.getModuleMap().addModule(mod);
+            }
         }
     }
 }
--- a/packages/ag-charts-community/src/chart/agChartV2.ts
+++ b/packages/ag-charts-community/src/chart/agChartV2.ts
@@ -509,17 +509,9 @@ function applySeriesOptionModules(series: Series<any>, options: AgBaseSeriesOpti
     const seriesOptionModules = REGISTERED_MODULES.filter((m): m is SeriesOptionModule => m.type === 'series-option');
 
     for (const mod of seriesOptionModules) {
-        if (mod.optionsKey in options) {
-            const supportedSeriesTypes: readonly string[] = mod.seriesTypes;
-            if (!supportedSeriesTypes.includes(series.type)) {
-                console.warn(
-                    `AG Charts - series-option '${mod.optionsKey}' does not support series type '${
-                        series.type
-                    }' (allowed series include: '${supportedSeriesTypes.join("', '")}')`
-                );
-            } else {
-                series.getModuleMap().addModule(mod);
-            }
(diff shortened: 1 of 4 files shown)


packages/ag-charts-community/src/chart/agChartV2.ts
```diff
@@ -509,7 +509,8 @@ function applySeriesOptionModules(series: Series<any>, options: AgBaseSeriesOpti
     const seriesOptionModules = REGISTERED_MODULES.filter((m): m is SeriesOptionModule => m.type === 'series-option');
 
     for (const mod of seriesOptionModules) {
-        if (mod.optionsKey in options) {
+        const supportedSeriesTypes: readonly string[] = mod.seriesTypes;
```
alantreadway, line 512: Nit: Is the `: readonly string[]` needed? I would have thought the infered type from `mod.seriesTypes` should already be `string[]` or something stricter?

packages/ag-charts-community/src/module/optionModules.ts
```diff
@@ -10,7 +10,26 @@ export interface AxisOptionModule<M extends ModuleInstance = ModuleInstance> ext
...
+        | 'radar-area'
+        | 'radar-line'
+        | 'radial-bar'
+        | 'radial-column'
+        | 'range-area'
+        | 'range-bar'
+        | 'scatter'
+        | 'treemap'
+        | 'waterfall'
+    )[];
```
alantreadway: Nit: Let's consider adding a type alias for the series types rather than repeating them?

Perhaps something based on `AgCartesianSeriesOptions['type'] | AgPolarSeriesOptions['type'] | AgHierarchySeriesOptions['type']`?

packages/ag-charts-community/src/chart/agChartV2.ts
olegat: That's what I would have thought, but using `mod.seriesTypes` directly causes this compiler error:
```
packages/ag-charts-community/src/chart/agChartV2.ts:513:67 - error TS2345: Argument of type 'string' is not assignable to parameter of type '"line" | "scatter" | "bubble" | "area" | "bar" | "box-plot" | "histogram" | "heatmap" | "waterfall" | "range-bar" | "range-area" | "pie" | "radar-line" | "radar-area" | "radial-bar" | "radial-column" | "nightingale" | "treemap"'.

513         if (mod.optionsKey in options && mod.seriesTypes.includes(series.type)) {
                                                                      ~~~~~~~~~~~
```

packages/ag-charts-community/src/module/optionModules.ts
olegat: That's a good idea, although for some reason using this type alias triggers the same error as I've mentioned in the comment about `mod.seriesTypes` (but this time the error is with `supportedSeriesTypes`). I'll take a closer look at this on Monday.

olegat: Done. Added `AgSeriesType` in _packages/ag-charts-community/src/options/series/seriesOptions.ts_

packages/ag-charts-community/src/module/optionModules.ts
```diff
@@ -1,6 +1,14 @@
+import type { AgCartesianSeriesOptions } from '../options/series/cartesian/cartesianSeriesTypes';
+import type { AgHierarchySeriesOptions } from '../options/series/hierarchy/hierarchyOptions';
+import type { AgPolarSeriesOptions } from '../options/series/polar/polarOptions';
 import type { BaseModule, ModuleInstance } from './baseModule';
 import type { AxisContext, ModuleContextWithParent, SeriesContext } from './moduleContext';
 
+type RemoveUndefined<T> = T extends undefined ? never : T;
```
alantreadway: Nit: Try the inbuilt `NonNullable`?

alantreadway (approved):
LGTM other than one nit - thanks!

packages/ag-charts-community/src/module/optionModules.ts
olegat: Thanks, I didn't know about `NonNullable` but that's exactly what I was looking for.

PR #349, opened 2023-10-13
AG-8557 Change double click to not reset zoom if over a node

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-8301
https://ag-grid.atlassian.net/browse/AG-8557

Adds a `nodeDatumManager` to the module context. This allows zoom to check if double clicking over a node and then prevent a reset of the zoom.

It will also be useful for the context menu to give options based on the clicked datum later.

Questions:
- I'm very open to name suggestions, like something with the words of `node`, `datum`, `event` manager?
- There's a `chartEventsManager` that I was originally going to use for this, but I'm not sure is a great fit given the tracked state (which will likely be expanded upon by other features). Which also begs the question if the chart events manager can be improved upon and not just a random collection of things.


alantreadway:
This seems to overlap a bit conceptually with `HighlightManager` - should we merge the two?

lsjroberts:
I've updated it to use the `highlightManager`, I haven't renamed that yet. I'll wait to see what happens with context menu to see if that inspires a better name for any extra functionality.

Also added resetting on a specific axis when clicked.

alantreadway (approved):
LGTM

PR #351, opened 2023-10-13
AG-8238 Bugfix ErrorBar Visibility Toggle

olegat:
https://ag-grid.atlassian.net/browse/AG-8238

Keeping the ErrorBarNode instance in the Series.contentGroup breaks the
rendering when an item is highlighted.

Keeping the ErrorBarNode instance in the Series.highlightGroup fixes this
rendering bug when an item is highlighted but breaks visibility toggling.

Therefore, we need to move the ErrorBarNode back and forth between the
contentGroup and highlightGroup when the highlighting changes.


alantreadway (commented):
Discussed offline.

packages/ag-charts-community/src/chart/layers.ts
```diff
@@ -7,7 +7,6 @@ export enum Layers {
     AXIS_ZINDEX = 20,
     SERIES_CROSSLINE_RANGE_ZINDEX = 30,
     SERIES_LAYER_ZINDEX = 500,
-    SERIES_ERRORBAR_ZINDEX = 600,
```
olegat, line 10: Note: this is no longer necessary now that the error bar are in their own group.

olegat:
> Discussed offline.

I've added an `annotationGroup` to _series.ts_ as we've discussed.

alantreadway (approved):
LGTM

PR #359, opened 2023-10-16
AG-8681 - Use user provided palette colors for waterfall series items

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-8681


alexanderby (approved):
The code looks alright, but why do we have both negative and positive totals green?

PR #364, opened 2023-10-17
AG-8556 Update bar-like series to zoom on the y-axis when horizontal

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-8556

This adds a way for series to change their defaults when their axes are swapped.


lsjroberts:
Closing as this approach won't work.

PR #366, opened 2023-10-17
AG-9272 Error Bar Highlight Style

olegat:


olegat:
This has a merge conflict with https://github.com/ag-grid/ag-charts/pull/351 and may need a rebase

packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts
```diff
@@ -47,6 +47,9 @@ export interface ErrorBoundSeriesNodeDatum {
     // by the size of the marker (line, scatter), width of the bar (vertical
     // bars), or height of the bar (horizontal bars).
     readonly capDefaults: { lengthRatio: number; lengthRatioMultiplier: number; lengthMax: number };
+    // The error bar module needs to know which node data to retrieve when
+    // events on are triggered on a datum.
+    readonly index: number;
```
alantreadway: Can we discuss the use of `index` for this use case - we haven't needed it for any other cases, so I'm a little suspicious that this is needed?

olegat: The `index` is required for the `ErrorBar` class to know which `ErrorBarNode` scene object to update.

However, I've removed this property. Instead, the `ErrorBar.restyleHightlightChange` method will determine the index by searching sequentially for a matching `datum` object.

packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
```diff
@@ -170,7 +170,8 @@ export class ScatterSeries extends CartesianSeries<Group, ScatterNodeDatum> {
         const nodeData: ScatterNodeDatum[] = [];
 
         const font = label.getFont();
-        for (const { values, datum } of processedData.data ?? []) {
+        for (let i = 0; i < processedData.data.length; i++) {
+            const { values, datum } = processedData.data[i];
```
alantreadway: Nit: Can we just revert this line to remove the use of `i`?

alantreadway (approved):
LGTM other than one nit - thanks for the updates :-)

packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts
olegat: Well spotted. Yes of course. I'll revert this.

PR #369, opened 2023-10-17
AG-8564 Calculate Axis layout before rendering

alexanderby:


alantreadway (commented):
I would have expected some code to have been removed? Is the layout logic replicated in the application phase?

alexanderby:
Some calculation results can be reused. I will try finding the cause for a little axis shift and refactor a bit.

alantreadway (approved):
LGTM

alexanderby:
I can't believe it!

PR #370, opened 2023-10-17
Line series rework

alantreadway:
First rough cut of the `LineSeries` rework.

Reworks:
- Initial load animation (now clip-rect based for line).
- Add/update/remove animations.
  - Continuous X-axis uses coordinate shift + match.
  - Category X-axis uses X-value matching.
- Moves to the new pattern for animation setup, introduces `lineUtil.ts`.

Fixes:
- Legend toggle.


packages/ag-charts-community/src/chart/series/cartesian/lineSeries.test.ts
```diff
@@ -249,32 +249,32 @@ describe('LineSeries', () => {
...
+                    AgChart.updateDelta(chart, { data: changedData });
                     await waitForChartStability(chart);
                     await compare();
                 });
             }
         }
 
         for (const ratio of [0, 0.25, 0.5, 0.75, 1]) {
-            it(`should animate with short-circuiting when run again before it finishes at ${
+            it.skip(`should animate with short-circuiting when run again before it finishes at ${
```
alantreadway: @lsjroberts any pointers on how to reproduce this case, I put together a Plnkr which did reproduce the failure successfully - then made a fix, but the test is still failing?

packages/ag-charts-community/src/chart/interaction/animationManager.ts
```diff
@@ -148,6 +149,25 @@ export class AnimationManager extends BaseManager<AnimationEventType, AnimationE
...
+            if (id != null && this.controllers.has(id)) {
+                this.controllers.get(id)?.stop();
+            }
+        } catch (error: unknown) {
+            this.failsafeOnError(error);
+            return;
+        }
+    }
+
+    public stopByAnimationIdPrefix(prefix: string) {
```
alexanderby: image

packages/ag-charts-community/src/chart/series/cartesian/markerUtil.ts
```diff
@@ -134,11 +136,37 @@ export function markerScaleInAnimation<T>(
...
+
+export function markerSwipeScaleInAnimation<T extends CartesianSeriesNodeDatum>(
+    { id }: { id: string },
+    animationManager: AnimationManager,
+    markerSelections: Selection<Node, T>[],
+    seriesWidth: number
+) {
+    const fromFn = (_: Node, datum: T) => {
+        const x = datum.midPoint?.x ?? seriesWidth;
+        const delayRatio = easing.easeInOut(x / seriesWidth) - 0.1;
```
alexanderby: Maybe some comment is needed

packages/ag-charts-community/src/chart/series/cartesian/markerUtil.ts
```diff
@@ -134,11 +136,37 @@ export function markerScaleInAnimation<T>(
...
+        const delay = Math.max(Math.min(delayRatio, 1), 0);
+        return { scalingX: 0, scalingY: 0, animationDelay: delay, animationDuration: 0.2 };
```
alexanderby: Should `0.2` be some constant?

alexanderby (approved):

packages/ag-charts-community/src/chart/interaction/animationManager.ts
lsjroberts: It does feel a little iffy to rely on a string like this.

If we need grouping to stop a bunch of animations I think handling that in the specific case with an array of ids iterated into `stopByAnimationId()` or some grouping in the animation manager itself would be better.

packages/ag-charts-community/src/chart/series/cartesian/__image_snapshots__/line-series-test-ts-line-series-animation-should-animate-adding-removing-and-updating-simultaneously-by-reloading-at-100-1-snap.png
lsjroberts, line 1: This test should be renamed since it no longer reloads

packages/ag-charts-community/src/chart/series/cartesian/__image_snapshots__/line-series-test-ts-line-series-multiple-overlapping-lines-should-render-line-series-with-the-correct-relative-z-index-1-snap.png
lsjroberts, line 1: Lost the right end markers?

packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts
```diff
@@ -799,8 +831,37 @@ export abstract class CartesianSeries<
...
+    }): Promise<void> {
+        // Override point for sub-classes.
+    }
+
     protected resetAllAnimation(data: CartesianAnimationData<TNode, TDatum, TLabel, TContext>) {
-        const { datum, label, marker } = this.opts?.animationResetFns ?? {};
+        const { path, datum, label, marker } = this.opts?.animationResetFns ?? {};
+
+        // Stop any running animations by prefix convention.
+        this.ctx.animationManager.stopByAnimationIdPrefix(`${this.id}_`);
```
lsjroberts: Yeah so perhaps here `stopAnimationsInGroup(this.id)` or something?

packages/ag-charts-community/src/chart/series/cartesian/lineSeries.test.ts
lsjroberts: Does this need to be removed or made to work?

packages/ag-charts-community/src/chart/series/cartesian/lineSeries.ts
```diff
@@ -444,607 +502,77 @@ export class LineSeries extends CartesianSeries<Group, LineNodeDatum> {
...
-                from: 0,
-                to: 1,
-                delay: duration,
-                duration: 200,
-                onUpdate: (opacity) => {
-                    labelSelections[contextDataIndex].each((label) => {
-                        label.opacity = opacity;
-                    });
-                },
-            });
```
lsjroberts, line 918: 🪦 RIP convoluted code

packages/ag-charts-community/src/chart/series/cartesian/lineUtil.ts
```diff
@@ -0,0 +1,402 @@
...
+            }
+
+            return { opacity: status === 'removed' ? 0 : path.opacity, ...FROM_TO_MIXINS[status] };
+        },
+    };
+
+    return { status, path: { addPhaseFn, updatePhaseFn, removePhaseFn }, pathProperties, marker: { fromFn, toFn } };
+}
+
+function findPointOnLine(a: { x: number; y: number }, b: { x: number; y: number }, targetX: number) {
```
lsjroberts: nit: Technically this function is no longer finding the point on the line, it is finding the y-point on the line given a fixed x. Probably worth changing the comment or renaming it so it doesn't get used in the wrong circumstance in the future.

packages/ag-charts-community/src/chart/series/cartesian/markerUtil.ts
lsjroberts: There are plans to make this a user option on animations, but it remains to be seen how that is done. Previously this was just a number of `200` as milliseconds in various places. But yeah, probably.

lsjroberts (changes requested):
Excellent stuff. Few minor points.

I think it would be super helpful for future devs if we had a comment block that gives an overview of the approach for the line animation, and perhaps some reasoning as to why this method was chosen. Given there's a lot of unstated thought behind this code.

packages/ag-charts-community/src/chart/interaction/animationManager.ts
alantreadway: Replaced with `Animation.groupId` and a `AnimationManager.stopAnimationByGroupId()`.

packages/ag-charts-community/src/chart/series/cartesian/markerUtil.ts
alantreadway: Added

packages/ag-charts-community/src/chart/series/cartesian/markerUtil.ts
alantreadway: Added a more general constant for these 'fast' ad-hoc transitions.

packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts
alantreadway: Fixed

packages/ag-charts-community/src/chart/series/cartesian/lineUtil.ts
alantreadway: Updated

alantreadway (commented):
@lsjroberts PTAL - I'm working on the tests next.

packages/ag-charts-community/src/chart/series/cartesian/lineSeries.test.ts
alantreadway: Discussed offline - we can't reproduce the issue, and adjusting the test doesn't seem to allow us to reproduce the real behaviour more accurate - I've removed the test case(s) for now.

PR #374, opened 2023-10-18
AG-9343 - register error bar theme template

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-6686


packages/ag-charts-enterprise/src/features/error-bar/errorBarTheme.ts
```diff
@@ -3,12 +3,12 @@ import { _Theme } from 'ag-charts-community';
...
     errorBar: {
         visible: true,
-        stroke: 'black',
+        stroke: _Theme.DEFAULT_LABEL_COLOUR,
         strokeWidth: 1,
         strokeOpacity: 1,
         cap: {
             length: undefined,
-            lengthRatio: undefined,
+            lengthRatio: 0.3,
```
olegat: Keep this as `undefined`.

The default `cap.lengthRatio` is intentionally different for line/scatter (`1.0`) and bar series (`0.5`).
- https://github.com/ag-grid/ag-charts/blob/ec1dd33a1c245d7065fb91e3724e380af3d60f92/packages/ag-charts-community/src/chart/series/cartesian/lineSeries.ts#L230
- https://github.com/ag-grid/ag-charts/blob/ec1dd33a1c245d7065fb91e3724e380af3d60f92/packages/ag-charts-community/src/chart/series/cartesian/scatterSeries.ts#L201
- https://github.com/ag-grid/ag-charts/blob/ec1dd33a1c245d7065fb91e3724e380af3d60f92/packages/ag-charts-community/src/chart/series/cartesian/barSeries.ts#L416

Judging from the updated snapshots, it looks like this `0.3` is overriding the defaults defined by the series, which is not desired.

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
```diff
@@ -320,4 +317,9 @@ export class ErrorBars
     private errorBarFactory(): ErrorBarNode {
         return new ErrorBarNode();
     }
+
+    private getWhiskerProperties() {
+        const { stroke, strokeWidth, visible, strokeOpacity, lineDash } = this;
```
olegat: The `lineDashOffset` value also needs to be returned.

nit: It might be a good idea to explicitly specify the `ErrorBarWhiskerTheme` return type for `getWhiskerProperties()` to hopefully catch some of these errors at compiler-time.

alantreadway (approved):
LGTM once you can resolve feedback from Oli

packages/ag-charts-community/src/chart/series/cartesian/barSeriesModule.ts
```diff
@@ -57,5 +57,12 @@ export const BarSeriesModule: SeriesModule<'bar'> = {
             blur: 5,
         },
     },
+    enterpriseThemeTemplate: {
+        errorBar: {
+            cap: {
+                lengthRatio: 0.2,
```
olegat, line 63: The default should be `0.5` https://ag-grid.atlassian.net/browse/AG-9014

olegat:
Ideally, this PR should pass the tests without updating the snapshots. The theme default values are not changing, they are just being initialised differently. Therefore the generated images should stay the same.

alantreadway:
Re-opened as https://github.com/ag-grid/ag-charts/pull/376 due to some CI issues (GH Actions seem to be using stale state - trying to track down the source now).

PR #375, opened 2023-10-18
AG-8556 Update zoom to flip xy in accordance with the series

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-8556

A better approach than https://github.com/ag-grid/ag-charts/pull/364.

This adds a `shouldFlipXY?: boolean` prop to the `layout-complete` event, which is then populated by `CartesianChart` and set to `true` if all the series in the chart agree.


alantreadway (approved):
LGTM

PR #379, opened 2023-10-18
Normalize markers formatter

iMoses:


packages/ag-charts-community/src/chart/series/series.ts
```diff
@@ -775,4 +782,53 @@ export abstract class Series<
...
+        params: TParams & Omit<AgSeriesMarkerFormatterParams<TDatum>, 'seriesId'>,
+        defaultStyle: AgSeriesMarkerStyle = marker.getStyle()
+    ) {
+        const { point } = params.datum;
+        const activeStyle = this.getMarkerStyle(marker, params, defaultStyle);
+
+        markerNode.setProperties({
+            visible: this.visible && activeStyle.size > 0 && point && !isNaN(point.x) && !isNaN(point.y),
+            translationX: point?.x,
+            translationY: point?.y,
```
alantreadway: Should these be present in this call, given the guard in the following statement?

packages/ag-charts-community/src/chart/series/seriesMarker.ts
```diff
@@ -43,4 +60,15 @@ export class SeriesMarker extends ChangeDetectable {
     @Validate(OPT_NUMBER(0, 1))
     @SceneChangeDetection({ redraw: RedrawType.MAJOR })
     strokeOpacity: number = 1;
+
+    @Validate(OPT_FUNCTION)
+    // @SceneChangeDetection({ redraw: RedrawType.MAJOR })
```
alantreadway: Intended?

alantreadway (approved):
Couple of comments, but otherwise LGTM when addressed

packages/ag-charts-community/src/chart/series/seriesMarker.ts
iMoses: No :X good catch

alantreadway (approved):
LGTM

PR #381, opened 2023-10-18
AG-8238 Increase the sensitivity of the tests

olegat:
Some of these tests (e.g. those that test the cap length) can fail with only very few pixels differing.


packages/ag-charts-enterprise/src/features/error-bar/errorBar.test.ts
```diff
@@ -127,7 +127,7 @@ describe('ErrorBars', () => {
         await waitForChartStability(chart);
 
         const imageData = extractImageData(ctx);
-        expect(imageData).toMatchImageSnapshot(IMAGE_SNAPSHOT_DEFAULTS);
+        expect(imageData).toMatchImageSnapshot({ ...IMAGE_SNAPSHOT_DEFAULTS, failureThreshold: 0 });
```
alantreadway, line 130: So the reason we don't set this as the default threshold is that we're storing compressed images for the snapshot, and there is a (perhaps only perceived) risk that compression artefacts may start to come into play and introduce false negative test results (aka flakiness). Happy to try this, but we should be prepared to roll this back if we observe any flakiness.

alantreadway: As a side-note - if using `nx run-many -t test -c update` to update snapshots, we should be defaulting to `failureThreshold: 0` to make sure all changes are picked up - are you using this command?

alantreadway (approved):

packages/ag-charts-enterprise/src/features/error-bar/errorBar.test.ts
olegat: I understand. However I think that this risk is small because we're storing images in PNG format which is lossless. I also feel that false negatives are less of a problem than false positives. We can raise the tolerance on a case-by-case basis if we observe false negatives, whereas false positives will quietly slip under the radar.

PR #382, opened 2023-10-19
Consolidated line series page plus other small edits

AG-DavidG:
Consolidated the line series docs page and examples.
@rmc-software - let's go over this before I do the same for area


olegat (approved):
LGTM, subject to approval from Rob.

PR #383, opened 2023-10-19
AG-9272 Error Bar mouse hover implementation

olegat:

This PR:
- Highlights the datum when the mouse hovers over an error bars scene node.
- Shows the tooltip when the mouse hovers over an error bars scene node.
- Renders the tooltip higher up to avoid covering the error bar.

Notes:
- Interaction range `'exact'` and `'nearest'` are treated the same as `range = 0`
- `ErrorBarNode` uses a hierarchical bounding box structure for precise & high-performance hit-testing.
- Tested on https://plnkr.co/edit/p9mib2yv6YYY97uL?open=main.js

**Before:**
Screenshot 2023-10-20 at 14 44 29

**After:**
Screenshot 2023-10-20 at 14 44 49

### DIFF
--- a/packages/ag-charts-community/src/chart/interaction/tooltipManager.ts
+++ b/packages/ag-charts-community/src/chart/interaction/tooltipManager.ts
@@ -29,6 +29,10 @@ export class TooltipManager {
         this.destroyFns.push(interactionManager.addListener('hover', (e) => this.checkExclusiveRects(e)));
     }
 
+    public getRange() {
+        return this.tooltip.range;
+    }
+
     public updateTooltip(callerId: string, meta?: TooltipMeta, content?: string) {
         if (content == null) {
             content = this.states[callerId]?.content;
--- a/packages/ag-charts-community/src/chart/interaction/tooltipManager.ts
+++ b/packages/ag-charts-community/src/chart/interaction/tooltipManager.ts
@@ -1,4 +1,8 @@
 import type { BBox } from '../../scene/bbox';
+import type { HdpiCanvas } from '../../scene/canvas/hdpiCanvas';
+import type { Point } from '../../scene/point';
+import type { ErrorBoundSeriesNodeDatum } from '../series/cartesian/cartesianSeries';
+import type { SeriesNodeDatum } from '../series/series';
 import type { Tooltip, TooltipMeta } from '../tooltip/tooltip';
 import type { InteractionEvent, InteractionManager } from './interactionManager';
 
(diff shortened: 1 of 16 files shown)


packages/ag-charts-community/src/chart/interaction/tooltipManager.ts
```diff
@@ -106,4 +110,46 @@ export class TooltipManager {
 
         this.appliedState = { content: contentToApply, meta: metaToApply };
     }
+
+    public static makeTooltipMeta(
```
olegat, line 117: Should this static method be moved into its own file (e.g. _tooltipUtils.ts_)?

The new imports are creating a lot of warnings in `nx run ag-charts-community:lint-depcruise` (although this should be okay because it's only `import type` statements).

packages/ag-charts-community/src/chart/interaction/tooltipManager.ts
```diff
@@ -106,4 +110,46 @@ export class TooltipManager {
...
+            offsetX,
+            offsetY,
+            event: event,
+            showArrow: datum.series.tooltip.showArrow,
+            position,
+        };
+
+        // On line and scatter series, the tooltip covers the top of errorbars when using
+        // datum.midPoint. Using datum.yBar.upperPoint renders the tooltip higher up.
+        const refPoint: Point | undefined = datum.yBar?.upperPoint ?? datum.midPoint;
```
olegat, line 140: Note: see PR description for a preview of what the output looks like with `midPoint` and `upperPoint`

packages/ag-charts-enterprise/src/features/error-bar/errorBarNode.ts
```diff
@@ -85,5 +95,49 @@ export class ErrorBarNode extends _Scene.Group {
...
+
+    override containsPoint(x: number, y: number): boolean {
+        if (this.bboxes.union.containsPoint(x, y)) {
+            for (const bbox of this.bboxes.components) {
+                if (bbox.containsPoint(x, y)) {
+                    return true;
+                }
+            }
+        }
+        return false;
```
alantreadway: Nit: Could we reduce the nesting here by short-circuiting the outer condition?

e.g.:
```
if (!this.bboxes.union.containsPoint(x, y)) return false;
```

packages/ag-charts-community/src/chart/interaction/tooltipManager.ts
alantreadway: So I don't think moving this method to a new file will help with the circular deps; the problem is the dependency back to `SeriesNodeDatum` and `ErrorBoundSeriesNodeDatum ` - we need to either not use these, or break them into their own minimally dependent modules?

I'd prefer if we resolve this now if straightforward.

packages/ag-charts-enterprise/src/features/error-bar/errorBarNode.ts
olegat: Sure, happy to make this change. I generally believe that having few `return` statements makes the code more readable, but I don't think that it matters much in this case.

packages/ag-charts-community/src/chart/interaction/tooltipManager.ts
olegat: Okay. I think that moving `SeriesNodeDatum` and `ErrorBoundSeriesNodeDatum` is the preferred option. So I'll see whether that's viable.

Not using them (i.e. redefining the type contract in _tooltipManager.ts_) will be our Plan B.

packages/ag-charts-enterprise/src/features/error-bar/errorBarNode.ts
olegat: Done

packages/ag-charts-community/src/chart/interaction/tooltipManager.ts
olegat: I've moved these types into _seriesTypes.ts_ just below the `ISeries` interface. This seems like a decent solution. Please let me know your thoughts.

alantreadway (approved):
Nice, thanks for the follow-up!

PR #384, opened 2023-10-20
Line animation cleanup

alantreadway:
Grab-bag of changes + cleanup whilst trying to improve animation test accuracy:
- Partial rework of `spyOnAnimationManager()` to allow `requestAnimationFrame` mocking - this is not practically used right now, see the (partially reverted) commit for the intended pattern I think we should move towards: 35d5ed0fddb83b6ea679e937111ff33d4fa7c926
- Support in `fromToMotion` and `staticFromToMotion` for `start` and `finish` properties, allowing animation-related boolean flags to be set at start/end of animation as needed.
- Attempt to fix markers jumping to final state without being animated in `LineSeries`.
- Cleanup of `Path` clipping code; skips transform mutation if `clipScalingX/Y` properties are set to `1`.
- Partial rework of `HdpiCanvas.context` to allow use of `Proxy` in the future, rather than mutating the context object directly.


packages/ag-charts-community/src/chart/test/utils.ts
```diff
@@ -331,9 +342,22 @@ export async function createChart(options: AgChartOptions) {
     return chart;
 }
 
+let activeAnimateCb: ((totalDuration: number, ratio: number) => Promise<void>) | undefined;
 export function spyOnAnimationManager() {
```
alexanderby, line 346: 🕵️‍♂️

alexanderby (approved):

lsjroberts (approved):
Think I understand what's going on 😅, looks good

PR #392, opened 2023-10-23
AG-9022 a11y: Change error bar docs customisation style

olegat:
Generally speaking, it's more readable for the docs to have CSS colour names in plain English. The colours should also be accessible on to wide range of colour blindnesses and vision loses. The colours should ideally also be visible on both the light and dark themes.

'pink' and 'red' seem to be the best colours for these requirements.

Tested using the 'A11Y - Color blindness empathy test' Chrome Extension: https://chrome.google.com/webstore/detail/a11y-color-blindness-empa/idphhflanmeibmjgaciaadkmjebljhcc

Other good colours could include:

-   Colour 1: 'blue', 'lightblue', 'cyan'
-   Colour 2: 'red', 'orange', 'gold'

However, many of these colours are not so distinguishable to people with Achromatopsia. Some of these colour like blue and orange don't look great on a dark background either.

### DIFF
--- a/packages/ag-charts-website/src/content/docs/error-bars/_examples/customisation/main.ts
+++ b/packages/ag-charts-website/src/content/docs/error-bars/_examples/customisation/main.ts
@@ -17,11 +17,11 @@ const options: AgChartOptions = {
                 xUpperKey: 'volUpper',
                 yLowerKey: 'presLower',
                 yUpperKey: 'presUpper',
-                stroke: '#4A4A93',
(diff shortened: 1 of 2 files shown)


alantreadway (approved):
LGTM - I would suggest running this past Rob once it's on the testing server?

olegat:
> LGTM - I would suggest running this past Rob once it's on the testing server?

Will do 👍

PR #393, opened 2023-10-23
AG-9017 Error Bar Formatters

olegat:

Implement formatters on error bar whiskers and caps.

Depends on PR #383


olegat:
unit/e2e tests pending... but some rudimentary tests were manually done in plunker: https://plnkr.co/edit/S29uKixcbgb0yW2h?open=main.js

packages/ag-charts-community/src/util/object.ts
```diff
@@ -59,3 +59,16 @@ export function mergeDefaults<TSource extends Record<string, any>, TArgs extends
 
     return target as Intersection<Exclude<TArgs[number], FalsyType>>;
 }
+
+// Similar to Object.assign, but only copy an explicit set of keys.
+export function partialAssign<T>(keysToCopy: (keyof T)[], target: T, source?: Partial<T>): T {
+    if (source !== undefined) {
```
alantreadway: Nit: Maybe short-circuit the `source === undefined` case up-front to remove a level of nesting? Also I would prefer using a standard `for ... of` loop over `.forEach()` if the difference is trivial, since it's both a bit more readable and less overhead?

packages/ag-charts-enterprise/src/features/error-bar/errorBarNode.ts
```diff
@@ -44,14 +65,64 @@ export class ErrorBarNode extends _Scene.Group {
...
+            if (formatters.formatter !== undefined) {
+                const result = formatters.formatter(params);
+                this.applyStyling(whiskerPath, result);
+                this.applyStyling(capsPath, result);
+                this.applyStyling(capsPath, result?.cap);
+            }
+
+            if (formatters.cap.formatter !== undefined) {
+                const result = formatters.cap.formatter(params);
+                this.applyStyling(capsPath, result);
```
alantreadway: `capsPath` is having settings applied multiple times - it would be better to calculate the styles and apply once, as the current approach has the possibility of triggering false-positive change-detection due to flip-flopping on style changes between the different inputs.

packages/ag-charts-community/src/util/object.ts
olegat: Done.

packages/ag-charts-enterprise/src/features/error-bar/errorBarNode.ts
olegat: That's a good idea. I've repurposed `mergeDefaults` from the box-plot series to avoid calling `applyStyling` more than once for each node.

alantreadway (approved):
LGTM - thanks for following up, note there are some linter failures.

olegat:
> LGTM - thanks for following up, note there are some linter failures.

Linter failures resolved.

PR #396, opened 2023-10-24
AG-8636 Add min visible items options to zoom

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-8636

This replaces `minRatioX` with `minVIsibleItemsX`.


packages/ag-charts-community/src/chart/cartesianChart.ts
```diff
@@ -56,6 +56,7 @@ export class CartesianChart extends Chart {
                 paddedRect: seriesPaddedRect,
                 visible: visibility.series,
                 shouldFlipXY: this.shouldFlipXY(),
+                minRect: (this.series[0] as CartesianSeries<any, any>).getMinRect?.(),
```
lsjroberts, line 59: This could be improved if we passed the series type into `Chart<CartesianSeries>` perhaps?

packages/ag-charts-community/src/chart/cartesianChart.ts
```diff
@@ -47,6 +47,14 @@ export class CartesianChart extends Chart {
 
         this.hoverRect = seriesPaddedRect;
 
+        const minRects = this.series.map((series) => (series as CartesianSeries<any, any>).getMinRect?.());
```
lsjroberts: This could be improved if we passed the series type into `Chart<CartesianSeries>` perhaps?

packages/ag-charts-enterprise/src/features/zoom/__image_snapshots__/zoom-test-ts-zoom-scrolling-should-zoom-in-then-out-1-snap.png
lsjroberts, line 1: This is expected as a result of changing the zoom scrolling speed in c35a55e9ce61bd4026e503aef98e2cbe351332a6. It now scales the speed based on the zoom level, so zooming back out one step will be a smaller amount than the original zoom in by one step.

This feels more natural overall as it slows down the rate of zooming as you go further in.

packages/ag-charts-community/src/chart/cartesianChart.ts
alantreadway: Maybe add the `getMinRect()` definition to `ISeries` to avoid the need to cast?

packages/ag-charts-enterprise/src/features/zoom/zoom.ts
```diff
@@ -288,11 +291,30 @@ export class Zoom extends _ModuleSupport.BaseModuleInstance implements _ModuleSu
...
+            this.minRatioX ||= Math.min(1, round(normalisedWidthRatio, DECIMALS));
+        }
+
+        if (this.isScalingY()) {
+            const heightRatio = (minRect.height * this.minVisibleItemsY) / paddedRect.height;
+            const normalisedHeightRatio = heightRatio * (zoom.y.max - zoom.y.min);
+            this.minRatioY ||= Math.min(1, round(normalisedHeightRatio, DECIMALS));
+        }
+
+        this.minRatioX ||= this.minRatioY || 0;
```
alantreadway, line 316: Nit: I'm a little confused by the use of `||` and `||=` on `number` properties here - could we use `??` and `??=` instead?

lsjroberts: When the chart first loads, the `getMinRect()` looks at the `contextNodeData` but it is an empty array initially. So the width and height of the rect are 0.

When the next layout completes, i.e. next interaction, it looks again and calculates it correctly.

Using `||=` instead of `??=` ensures it gets set to something after the first 0 but also prevents it from fluctuating due to fractions as you zoom in, causing inconsistencies in the zoom limits.

Perhaps instead it'd be better if the layout event returns `undefined` if there's no `contextNodeData`.

packages/ag-charts-community/src/chart/cartesianChart.ts
lsjroberts: This is only needed for cartesian series for now, but I presume we'll be making it work for other series at some point

packages/ag-charts-enterprise/src/features/zoom/zoom.ts
lsjroberts: I've updated this to use `undefined` for empty `contextNodeData`, but it remains simpler to use `||=` for these ratios and default them to 0. The effect of it being 0 or undefined would be the same, i.e. infinite zoom depth allowed, but defaulting to 0 skips a bunch of checking elsewhere.

alantreadway (approved):
LGTM - thanks for the updates + explaination.

PR #399, opened 2023-10-24
AG-9272 Add unit test for error bar cursor

olegat:

Tested manually in [Plunker](https://plnkr.co/edit/Ufro6XZAiV0HO6B8?open=main.js)

### DIFF
--- a/packages/ag-charts-enterprise/src/features/error-bar/errorBar.test.ts
+++ b/packages/ag-charts-enterprise/src/features/error-bar/errorBar.test.ts
@@ -656,7 +656,7 @@ describe('ErrorBars', () => {
 
     it('should use correct cursor', async () => {
         const getCursor = () => {
-            return chart['cursorManager']['element'].style.cursor;
+            return chart.getModuleContext().cursorManager.getCursor();
         };
         chart = deproxy(
             AgEnterpriseCharts.create({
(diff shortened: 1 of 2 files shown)


packages/ag-charts-enterprise/src/features/error-bar/errorBar.test.ts
```diff
@@ -602,4 +602,29 @@ describe('ErrorBars', () => {
         await clickAction(x + width - 1, y)(chart);
         await compare();
     });
+
+    it('should use correct cursor', async () => {
+        const getCursor = () => {
+            return chart['cursorManager']['element'].style.cursor;
+        };
```
alantreadway: Seems a little fragile, but I'm ok with this as long as we don't repeat it elsewhere (or have one place we define this) - could you move this to a test utility function, perhaps as a sibling to where `hoverAction()` is declared?

I think it might be better to also use a cast as early as possible so that trivial refactors highlight a break here?

e.g. `(chart as Chart).getModuleContext().cursorManager.element?.style.cursor` might be a better way to go (make `element` a `public readonly` field on `CursorManager` for test purposes perhaps?

olegat: I agree, it's unsafe code. In this case I think it's okay because it's just a test. It will fail if we refactor the cursor manager but won't affect the integrity of the charts library.

That said, the `getModuleContext()` approach that you're suggesting is better, because the failure will be raised sooner by the compiler / IDE.

I'll make these changes.

olegat: I've replaced this with `chart.getModuleContext().cursorManager.getCursor()`

olegat:
Rebased & Resolved merge conflict in _errorBar.test.ts_

PR #410, opened 2023-10-26
AG-9254 Update animation docs and remove old examples

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-9254

I imagine we will go through a few interations of the text, but this is a first pass with the two main sections of the initial load and data updates animations, plus the only option currently of `duration`.

If we need test cases from the deleted examples, we can create more "hidden" pages to exercise those.


alantreadway (approved):
LGTM - I'll review the examples once they are on testing.

PR #411, opened 2023-10-26
AG-9272 QA Feedback: Click Events

olegat:

### DIFF
--- a/packages/ag-charts-community/src/module/moduleMap.ts
+++ b/packages/ag-charts-community/src/module/moduleMap.ts
@@ -55,4 +55,10 @@ export class ModuleMap<M extends Module<C, I>, C, I extends ModuleInstance = Mod
     isModuleEnabled(module: M) {
         return this.modules[module.optionsKey] != null;
     }
+
+    *[Symbol.iterator](): IterableIterator<I> {
+        for (const { instance } of Object.values(this.modules)) {
+            yield instance;
+        }
+    }
 }
--- a/packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
+++ b/packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
@@ -302,8 +302,15 @@ export class ErrorBars
     }
 
     private pickDatum(event: _ModuleSupport.InteractionEvent): ErrorBarNodeDatum | undefined {
-        const node = this.groupNode.pickNode(event.offsetX, event.offsetY);
-        return node?.datum;
+        // The error bars cover part of series datum node (marker/bar). If this part is clicked on, then
+        // that triggers the event twice (once in the Series class, and once in this ErrorBars class).
+        // We don't want that, so check whether the event hits a series node datum:
+        const seriesNode = this.cartesianSeries.contentGroup.pickNode(event.offsetX, event.offsetY);
+        if (seriesNode !== undefined) {
+            return undefined; // Ignore this event, it's already being handled by the series.
+        }
+        const errorBarNode = this.groupNode.pickNode(event.offsetX, event.offsetY);
+        return errorBarNode?.datum;
     }
 
     private onHoverEvent(event: _ModuleSupport.InteractionEvent<'hover'>) {
--- a/packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
+++ b/packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
@@ -1,4 +1,4 @@
-import type { AgErrorBarOptions, AgErrorBarThemeableOptions, _Scale } from 'ag-charts-community';
+import type { AgErrorBarOptions, AgErrorBarThemeableOptions, InteractionRange, _Scale } from 'ag-charts-community';
 import { AgErrorBarSupportedSeriesTypes, _ModuleSupport, _Scene } from 'ag-charts-community';
 
 import type {
@@ -14,6 +14,7 @@ const {
     mergeDefaults,
     valueProperty,
     ChartAxisDirection,
+    SeriesNodePickMode,
     Validate,
     OPT_BOOLEAN,
     OPT_COLOR_STRING,
@@ -45,6 +46,8 @@ type AnyDataModel = _ModuleSupport.DataModel<any, any, any>;
 type AnyProcessedData = _ModuleSupport.ProcessedData<any>;
 type AnyScale = _Scale.Scale<any, any, any>;
 type HighlightNodeDatum = NonNullable<_ModuleSupport.HighlightChangeEvent['currentHighlight']>;
+type InteractionEvent<T extends _ModuleSupport.InteractionEvent['type'] = _ModuleSupport.InteractionEvent['type']> =
+    _ModuleSupport.InteractionEvent<T>;
 
 type SeriesDataPrerequestEvent = _ModuleSupport.SeriesDataPrerequestEvent;
 type SeriesDataProcessedEvent = _ModuleSupport.SeriesDataProcessedEvent;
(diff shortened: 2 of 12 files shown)


packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
```diff
@@ -297,25 +303,89 @@ export class ErrorBars
...
+    private checkInteractionEventOverlap(event: InteractionEvent, range: InteractionRange): boolean {
+        if (range === 'nearest') {
+            return false;
+        }
+        // The error bars cover part of series datum node (marker/bar). If this part is clicked on, then
+        // that triggers the event twice (once in the Series class, and once in this ErrorBars class).
+        // We don't want that because there is an API for adding listeners to 'nodeClick' and
+        // 'nodeDoubleClick' events. Therefore, check whether the event hits a series node datum:
+        if (range === 'exact' || range === 0) {
+            const seriesNode = this.cartesianSeries.contentGroup.pickNode(event.offsetX, event.offsetY);
```
alantreadway: `pickNode()` is a potentially expensive operation, we should avoid executing it multiple times (it's already in-play from a series perspective?) here and below.

Maybe we could consider 'disarming' the error-bar clickable state based upon the currently highlighted datum state?

packages/ag-charts-enterprise/src/features/error-bar/errorBarNode.ts
```diff
@@ -170,52 +200,57 @@ export class ErrorBarNode extends _Scene.Group {
         }
         caps.path.closePath();
         caps.markDirtyTransform();
+    }
+
+    updateBBoxes(...ranges: InteractionRange[]): void {
```
alantreadway: I'm just wondering if we should improve the scene-graph support for this type of operation, rather than have all of this behaviour as `ErrorBar`-specific functionality?

`Group.computeBBox()` already exists, and I think we could improve the `Group.containsPoint()` handling for this case (specifically, allow `Group` to configurably have accurate bounding-box handling)?

packages/ag-charts-enterprise/src/features/error-bar/errorBarNode.ts
```diff
@@ -29,17 +30,46 @@ export type ErrorBarCapFormatter = (params: AgErrorBarFormatterParams) => AgErro
 
 type CapDefaults = NonNullable<ErrorBarNodeDatum['capDefaults']>;
 
+function toPixelRange(range: InteractionRange): PixelSize {
+    return typeof range === 'string' ? 0 : range;
+}
+
+class HierarchialBBox {
```
alantreadway, line 34: See my other comment, but I'm thinking these behaviours are not error-bar specific?

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
olegat: Yes, `pickNode()` is not ideal. However, we cannot use the highlight state to determine whether a node is clickable because we provide the option to customise the highlight and click range separately:
```
{
  tooltip: { range: "exact" },
  series: [ { nodeClickRange: 10 } ],
}
```
This example makes it possible to click an unhighlighted node if the user clicks within a range of 10 pixels.

Essentially this function just needs to check whether the series node datum that the error bars map to is handling the event. So it doesn't need to do hit-testing for all of the nodes in `contentGroup`. We might be able to add a optimised method to just do hit-testing on a single a datum. Example:
```typescript
private checkInteractionEventOverlap(datum: ErrorBarNodeDatum, event: InteractionEvent, range: InteractionRange): boolean {
    if (this.cartesianSeries.checkDatumHit(datum, event, range)) {
        return false;
    }
    return true;
}
```

Ideally, we want either the series datum node or the error bar node to consume the click event to avoid firing the `nodeClick` event twice. There is an `InteractionEvent.consume()` function. I tried calling that function in the _errorBar.ts_ file but it didn't prevent both nodes from firing the `nodeClick` event.

packages/ag-charts-enterprise/src/features/error-bar/errorBarNode.ts
olegat: I agree that this should be a built-in scene graph feature. However, adding such a feature to the scene has the potential of causing many bug and performance regressions across the entire ag-charts framework.

Are we confident enough with our test coverage to mitigate this risk so close the initial release? Or should we leave this in the backlog?

packages/ag-charts-enterprise/src/features/error-bar/errorBarNode.ts
olegat: We can safely make `HierarchialBBox` part of the ag-chart-community package as a utility class. Although I don't see much benefit to this. The `HierarchialBBox` class only supports a tree of depth 1, so it's not such a useful utility class. A more useful version of `HierarchialBBox` would support trees with multiple levels:
```diff
 class HierarchialBBox {
     private readonly union: BBox;
-    private readonly components: BBox[];
+    private readonly components: HierarchialBBox[];
     ...
 }
```

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
olegat: Okay, so I've discovered that the `InteractionEvent.consume()` method is intended to suppress the event to other "internal" listeners like the legend and the error bars modules. The Series class could in theory consume this to hide it from all the other listener, but I don't think that this is desirable (it might unintentionally break things with the legend, for example).

Instead, I've added the `InteractionEvent.callbackComplete` property to keep track of whether or not the nodeClick / nodeDoubleClick have been broadcasted to the library user. https://github.com/ag-grid/ag-charts/commit/8193d3cdc36f2308da8121e1937069fc9b6aa8a6

I think that this is the best solution, because it avoids the need to recalculate the hit-tests and keeps a clear distinction between internal & external listeners.

packages/ag-charts-community/src/module/moduleMap.ts
```diff
@@ -55,4 +55,10 @@ export class ModuleMap<M extends Module<C, I>, C, I extends ModuleInstance = Mod
     isModuleEnabled(module: M) {
         return this.modules[module.optionsKey] != null;
     }
+
+    *[Symbol.iterator](): IterableIterator<I> {
+        for (const { instance } of Object.values(this.modules)) {
+            yield instance;
```
alantreadway, line 61: Interesting, this is the first practical use of `yield` I've seen in a codebase :D

olegat: I actually think that it might make more sense for `Series` to extend `ModuleMap` and have `this.modules` be protected, but that's a discussion for another time 😄

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
olegat: This has been changed. `ErrorBars` now implements `pickNode` for exact, nearest and main-axis-first pick modes. PTAL.

packages/ag-charts-enterprise/src/features/error-bar/errorBarNode.ts
olegat: Some this has been moved into the community package (new file _nearest.ts_), but the `ErrorBarNode` still cares take of calculating the size & position of bounding boxes. PTAL.

packages/ag-charts-enterprise/src/features/error-bar/errorBarNode.ts
olegat: What shall we do with this `HierarchialBBox` class? I can move it into the community package as a scene utility and have `ErrorBarNode` extend it implement hit testing. Please let me know what you think.

alantreadway (approved):
LGTM

PR #415, opened 2023-10-26
Animation fixes

alantreadway:
Fixes:
- PieSeries radius animation - radius changes are now animated.

Bonus:
- Removes a lot of redundant radius calculations :-P

Notes:
- Pie outer-radius calculation is slightly changed - it's not clear why but the changes seem acceptable.
- Removal case for a pie segment doesn't know the radius to use, so the prior radius is used (this seems like an edge case, legend toggle works as expected).


packages/ag-charts-community/src/chart/series/polar/pieSeries.ts
```diff
@@ -1522,9 +1523,11 @@ export class PieSeries extends PolarSeries<PieNodeDatum, Sector> {
     }
 
     getDatumIdFromData(datum: any) {
-        const { calloutLabelKey, sectorLabelKey } = this;
+        const { calloutLabelKey, sectorLabelKey, legendItemKey } = this;
```
lsjroberts, line 1526: Does this conflict/supersede with the fallback to `legendItemValue` in `getDatumId()`?

lsjroberts (approved):
LGTM, with one question

packages/ag-charts-community/src/chart/series/polar/pieSeries.ts
alantreadway: Fair point - `legendItemValue`/`legendItemKey` needs to take precedence over the other options, and then `legendItemValue` is a fallback is redundant, so I've removed it.

PR #417, opened 2023-10-27
AG-9017 QA Feedback: Apply formatter cap lengths

olegat:

This joins the ErrorBarNode updateStyle and updateTranslation. The cap
length is part of the style, which unfortunately affects the translation.
In fact, the stroke width can also affect the translation by growing the
bounding box. So having separate update methods for style and translation
is no longer feasible.

Depends on PR #411


alantreadway (commented):
Feedback discussed offline.

alantreadway (approved):
LGTM

PR #420, opened 2023-10-30
Dg/update bubble docs

AG-DavidG:
Updated docs and examples for Bubble Series.

Removed defaults for cartesian axes in the comments of the options, as the defaults are theme and series specific.


alantreadway (approved):
LGTM

PR #421, opened 2023-10-30
Tooltip titles

jacobp100:
Improve type consistency for sanitizeHtml calls, ensure tooltip title is not displayed if no title is returned from the renderer fn

The issue looked to be `sanitizeHtml` always returns a string - even if the value was undefined. Later on, there was some nil-defaulting, which would not be run because the value was now an empty string in every case except the pie chart. This follows the pie chart case, where if you don't return a title, it won't render a title (so users have a mechanism to remove the title)

### DIFF
--- a/packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
+++ b/packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
@@ -596,7 +596,7 @@ export class AreaSeries extends CartesianSeries<
 
         const xString = xAxis.formatDatum(xValue);
         const yString = yAxis.formatDatum(yValue);
-        const title = yName != null ? sanitizeHtml(yName) : undefined;
+        const title = sanitizeHtml(yName);
         const content = sanitizeHtml(xString + ': ' + yString);
 
         const baseStyle = mergeDefaults({ fill: this.fill }, marker.getStyle(), {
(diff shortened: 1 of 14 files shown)


packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
```diff
@@ -596,7 +596,7 @@ export class AreaSeries extends CartesianSeries<
 
         const xString = xAxis.formatDatum(xValue);
         const yString = yAxis.formatDatum(yValue);
-        const title = sanitizeHtml(yName);
+        const title = yName != null ? sanitizeHtml(yName) : undefined;
```
alantreadway: Nit: Just wondering if we should change `sanitizeHtml()` to return `undefined` if it's input is `undefined`, since it looks like the majority of uses are guarding against this - so it might be cleaner overall?

alantreadway: You could probably change the signature of `sanitizeHtml()` using conditional return types to help flush out any issues with that change? E.g.:

```
export function sanitizeHtml<T extends string | undefined>(text: T): T extends undefined ? undefined : string {
```

alantreadway (approved):
LGTM - feel free to merge once `latest` is green.

PR #427, opened 2023-10-31
AG-8648 Add module conflicts handling for zoom and navigator

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-8648


alantreadway (approved):
LGTM - the approach makes sense, just to confirm that this is in alignment with David + Robs thinking?

PR #428, opened 2023-10-31
More AreaSeries fixes.

alantreadway:
Fixes:
- AreaSeries now shrinks to/from `0` height on legend toggle.
- LineSeries now fades in/out on legend toggle in all cases.

Non-visible fixes:
- Animations now have the `visible` state as part of the context data to simplify add/remove detection.
- `NaN` is now handled more gracefully for path animation - practically this is the 'all series hidden' case when the to/from points are all `NaN` since the axes + scales don't exist.

### DIFF
--- a/packages/ag-charts-community/src/chart/series/cartesian/lineUtil.ts
+++ b/packages/ag-charts-community/src/chart/series/cartesian/lineUtil.ts
@@ -50,6 +50,7 @@ function calculateMoveTo(from = false, to = false): PathPoint['moveTo'] {
 type LineContextLike = {
     scales: CartesianSeriesNodeDataContext['scales'];
     nodeData: PathNodeDatumLike[];
+    visible: boolean;
 };
 
 export function pairContinuousData(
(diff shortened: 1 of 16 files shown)


packages/ag-charts-community/src/chart/series/cartesian/lineUtil.ts
```diff
@@ -223,9 +228,14 @@ export function pairCategoryData(
...
-    if (oldData.nodeData.length === 0 && newData.nodeData.length > 0) {
+
+    const visible = (data: LineContextLike) => {
+        return data.visible;
+    };
+
+    if (!visible(oldData) && visible(newData)) {
         status = 'added';
-    } else if (oldData.nodeData.length > 0 && newData.nodeData.length === 0) {
+    } else if (visible(oldData) && !visible(newData)) {
```
lsjroberts, line 238: Presumably `visible` is set to `false` on a series when its `nodeData` is empty?

lsjroberts (approved):
LGTM

packages/ag-charts-community/src/chart/series/cartesian/lineUtil.ts
alantreadway: So previously we were using `nodeData.length` as a way to determine visibility status, we're now being a bit more direct by driving directly from the `Series.visible` boolean flag 🤞

PR #431, opened 2023-10-31
AG-8648 Update zoom and navigator to reset on enable/disable

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-8648

Also makes the module conflict stuff actually override the options... I was just seeing the effect of navigator itself blocking zoom before. The conflict stuff didn't actually work.


alantreadway (approved):
LGTM

PR #432, opened 2023-10-31
Ensure consistency with bar chart axes

jacobp100:


packages/ag-charts-community/src/module/coreModules.ts
```diff
@@ -74,7 +74,8 @@ export type ExtensibleDefaults<
     ChartOptions = AgChartOptions & { series?: { type: SeriesType } },
 > = Extensible<ChartOptions>;
 
-export interface SeriesModule<SeriesType extends RequiredSeriesType = RequiredSeriesType> extends BaseModule {
+export interface SeriesModule<SeriesType extends RequiredSeriesType = RequiredSeriesType, SeriesOptions = any>
```
alantreadway: Nit: Could we derive `SeriesOptions` from `SeriesType` for better type-safety/consistnecy in the module declaration?

alantreadway (commented):
Approach looks good for the core problem, just one nit to avoid mismatching series declarations.

packages/ag-charts-community/src/module/coreModules.ts
jacobp100: Updated

jacobp100: Think I just about got this to work, but had to stick an `any` over [here](https://github.com/ag-grid/ag-charts/pull/432/files#diff-0065008268af620210c1f943b2a92e6ed142dd2529e2083978ba120e6e476353R10)

alantreadway: Ok - yeah the need to use `any` for these cases is annoying, but I think that is fine on balance since the likelihood is that the module is declared incorrectly rather than the infrastructure code doing the wrong thing 🤞

alantreadway (approved):
LGTM

alantreadway (approved):
LGTM

PR #434, opened 2023-10-31
AG-8524 Change zoom to skip animation on that update

lsjroberts:
This fixes the conflicts between zoom and animation. When a zoom change triggers an update, that particular update skips animations. Once an update completes it will reset animation skipping back to its initial state if it has been changed.


lsjroberts:
As discussed, gunna look at moving batch state to animation manager.

PR #436, opened 2023-11-01
AG-8524 Fix zoom with animation

lsjroberts:
Adds animation batches that contain the controllers. These can be skipped as a batch together.

So the chart update can just call `this.animationManager.skipCurrentBatch()` and the state of that will be handled internally.


alantreadway (approved):
Excellent, LGTM!

PR #437, opened 2023-11-01
Community options cleanup

iMoses:
- Cleanup options jsDocs
- Change axis type `groupedCategory` into kebab-case, to keep consistency with other types
- Change treemap data into an array to keep consistency with other series types


alantreadway (approved):
LGTM

PR #438, opened 2023-11-01
AG-8238 Mute circular dependency warning

olegat:
The import dependency was:
bbox.ts -> nearest.ts -> node.ts -> bbox.ts

nearest.ts doesn't really need to import the `Node` type, it just needs the `transformPoint` method. So redeclare that as part of the `DistantContainer` interface.


alantreadway (approved):
Excellent, thanks for this!

PR #439, opened 2023-11-01
AG-8238 Improve ErrorBars.pickNodeExact

olegat:
We don't need to calculate the distance in this pick-mode because it's always 0. Therefore just use the cheaper containsPoint/pickNode.


alantreadway (approved):
LGTM

PR #440, opened 2023-11-01
AG-9195 Fix bar stacking to obey stacked option when false

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-9195

It appeared to be an intentional feature, but I'm not sure if it was just for backwards compatibility? Probably best to remove though still?


alantreadway (approved):
LGTM - I agree this is probably legacy behaviour.

PR #446, opened 2023-11-02
Pie and Doughnut Series docs and examples cleanup

AG-DavidG:


packages/ag-charts-website/src/content/docs/doughnut-series/index.mdoc
```diff
@@ -2,104 +2,107 @@
...
+                text: '$100,000',
+                margin: 4,
+                fontSize: 48,
+                color: 'green',
+            },
+        ],
+        innerCircle: {
+            fill: '#c9fdc9',
+        },
+    }]
```
lsjroberts: The indentation here is a bit off. The `...` trips up the formatter. Same issue in a few other places too.

lsjroberts (approved):
LGTM. Though we should definitely get the formatter running on the `ag-charts-website` package too.

packages/ag-charts-website/src/content/docs/doughnut-series/index.mdoc
AG-DavidG: I've commented the `...`, formatted and then uncommented it again.
Seems to have worked.

PR #449, opened 2023-11-03
Cleanup and improve docs navigation

iMoses:


packages/ag-charts-community/src/options/chart/chartBuilderOptions.ts
```diff
@@ -16,7 +16,7 @@ export interface AgCartesianChartOptions extends AgBaseCartesianChartOptions, Ag
     /**
      * A predefined theme name or an object containing theme overrides.
      *
-     * See: [Themes Reference](./themes-api/)
+     * See: [Themes Reference](../themes-api/)
```
taktran, line 19: I wonder if it's more future proof if we use abs urls for this, so there aren't assumptions about which page these docs are processed.

```suggestion
     * See: [Themes Reference](/options/themes-api/)
```

taktran (approved):
LGTM 👍🏼

packages/ag-charts-community/src/options/chart/chartBuilderOptions.ts
iMoses: Hmm. we might but then we'll have to make sure the `SITE_BASE_URL` is merged to it otherwise the links will be wrong. I'll look into it in a separate PR, I'll see if we can configure the markdown component to add it.

taktran: Yea, I was thinking of doing it at the markdown component layer rather than in the js docs

PR #450, opened 2023-11-03
AG-9254 Add extra interactions to animation data updates example

lsjroberts:
https://github.com/ag-grid/ag-charts/assets/3817697/b7e35ff8-e979-4382-b159-4aeff20e622d


alantreadway (approved):
LGTM

I really like the example - I wonder if it's too complex for the main docs, but would be better suited as a gallery example?

I defer to Rob/David on that one.

PR #452, opened 2023-11-03
AG-9702 Fix area series zoom reset on double click

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-9702

For the purposes of checking if clicking a node should block the double click zoom reset, this changes it to only check if the node is exactly matched. Since area series by default use `'nearest'` matching and would thus always block double click otherwise.


alantreadway (approved):
LGTM

PR #454, opened 2023-11-06
Add highlighting animation to selected api properties

iMoses:


packages/ag-charts-website/src/features/api-documentation/components/ApiReference.module.scss
```diff
@@ -174,3 +174,12 @@
...
     }
 }
+
+.highlightAnimate {
+    animation: highlight-animation 3000ms ease-out;
+}
+
+@keyframes highlight-animation {
+    0%,100% { background-color: inherit }
+    10%, 50% { background-color: var(--light-water-blue) }
```
markdurrant: It's best to avoid using the specific colours variables E.g. `--color-name` and prefer the colour "scale" variables. E.g. `--scale-XXX`. As those will also work for darkmode. 

E.g. Here we might use `--background-400`.

markdurrant (commented):
One comment on colours. Otherwise all LGTM 👍

PR #458, opened 2023-11-06
AG-9640 Add error for invalid theme name

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-9640

### DIFF
--- a/packages/ag-charts-community/src/chart/mapping/themes.ts
+++ b/packages/ag-charts-community/src/chart/mapping/themes.ts
@@ -1,5 +1,6 @@
 import type { AgChartTheme, AgChartThemeName, AgChartThemeOverrides } from '../../options/agChartOptions';
 import { jsonMerge } from '../../util/json';
+import { Logger } from '../../util/logger';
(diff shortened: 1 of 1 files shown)


packages/ag-charts-community/src/chart/mapping/themes.ts
```diff
@@ -46,9 +47,13 @@ export function getChartTheme(value?: string | ChartTheme | AgChartTheme): Chart
...
 
-    const stockTheme = themes[value as AgChartThemeName];
-    if (stockTheme) {
-        return stockTheme();
+    if (typeof value === 'string') {
+        const stockTheme = themes[value as AgChartThemeName];
+        if (stockTheme) {
+            return stockTheme();
+        }
+        Logger.errorOnce(`the theme [${value}] is invalid, using [ag-default] instead.`);
```
alantreadway: Nit: We usually use warnings for non-fatal errors - I think this should probably qualify?

alantreadway (approved):
LGTM

PR #459, opened 2023-11-06
Api options refs design

markdurrant:


packages/ag-charts-website/src/design-system/_typography.scss
```diff
@@ -102,19 +102,19 @@ strong,
...
 }
 
-p:not(:last-child),
-ul:not(:last-child),
-ol:not(:last-child),
-blockquote:not(:last-child) {
+p:where(:not(:last-child)),
+ul:where(:not(:last-child)),
+ol:where(:not(:last-child)),
+blockquote:where(:not(:last-child)) {
```
iMoses, line 117: just wondering, why do you need to use `where` here?

packages/ag-charts-website/src/features/api-documentation/components/SearchBox.tsx
```diff
@@ -80,6 +80,18 @@ export function SearchBox({
...
+                <div className={styles.searchDropdown} onMouseDown={(e) => e.preventDefault()}>
+                    <div className={styles.searchOptions}>
+                        <div className={styles.searchOption}>
+                            <span className="font-size-small">
+                                We couldn't find any matches for "<b>{searchQuery}</b>"
+                            </span>
+                        </div>
+                    </div>
+                </div>
+            )}
```
iMoses, line 90: I can't so a multiline suggestion for some reason...
I think you can simplify it by doing this:
```
            {searchQuery.length > 0 && inFocus && (
                <div className={styles.searchDropdown} onMouseDown={(e) => e.preventDefault()}>
                    <div className={styles.searchOptions}>
                        {data.length ? data.map((data, index) => (
                            <div
                                key={data.label}
                                ref={
                                    index === selectedIndex
                                        ? (ref) => ref?.scrollIntoView({ block: 'nearest', inline: 'start' })
                                        : null
                                }
                                className={classnames(styles.searchOption, {
                                    [styles.selected]: index === selectedIndex,
                                })}
                                onClick={() => handleClick(data)}
                                onMouseEnter={() => setSelectedIndex(index)}
                            >
                                {markResults && searchQuery ? (
                                    <Markdown>{data.label.replace(searchRegexp, `**$1**`)}</Markdown>
                                ) : (
                                    data.label
                                )}
                            </div>
                        )) : (
                            <span className="font-size-small">
                                We couldn't find any matches for "<b>{searchQuery}</b>"
                            </span>
                        )}
                    </div>
                </div>
            )}
```

packages/ag-charts-website/src/components/link-icon/LinkIcon.module.scss
```diff
@@ -9,7 +9,7 @@
     transition: opacity 0.33s ease-in-out;
 
     &:hover {
-        opacity: 1;
+        opacity: 1 !important;
```
iMoses: is this a must? can we find a way to not use `!important`?

packages/ag-charts-website/src/design-system/_typography.scss
markdurrant: It's to reduce the specificity of the last child stuff. 

`p:not(:last-child)` has higher specificity than a class, but `p:where(:not(:last-child))` doesn't.

packages/ag-charts-website/src/features/api-documentation/components/SearchBox.tsx
markdurrant: This looks good. I'll update.

taktran (approved):
LGTM 👍🏼

packages/ag-charts-website/src/components/link-icon/LinkIcon.module.scss
markdurrant: Depends on how much you hate `!important` 😄 

We could add an extra rule to `.apiReference tr` ... 

```
.apiReference {
    position: relative;
    table-layout: fixed;

    tr {
        scroll-margin-top: var(--anchor-offset, 0);

        &:hover a:global(.docs-header-icon) {
            opacity: 0.5;
        }

        a:global(.docs-header-icon):hover {
            opacity: 1;
        }
    }
...
```

But to me that's a cure that's worse than the disease though.

iMoses: I personally really hate `!important` 😅 
I agree that it's ugly to artificially increase specificity but when you eventually need to override an important value down the line it's a nightmare to work with IMO and you never know what else you're breaking...

PR #462, opened 2023-11-07
AG-9711 - PieSeries animation fixes.

alantreadway:
Fixes:
- PieSeries callout labels now always match the sector (AG-9711).
- Segment ordering is no longer fixed as per feedback here: https://ag-grid.atlassian.net/browse/AG-8489?focusedCommentId=54662
- Consequently, segments now transition colour (fill/stroke) to better support datum re-ordering/update cases.


lsjroberts (approved):
lgtm

PR #466, opened 2023-11-07
AG-8564 - Fix LineSeries and AreaSeries short-circuit

alantreadway:
Fixes short-circuiting of axes animation when line/area series rendering are not animatable.

### DIFF
--- a/packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
+++ b/packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
@@ -690,13 +690,7 @@ export class AreaSeries extends CartesianSeries<
 
         super.resetAllAnimation(animationData);
 
-        if (
-            contextData.length === 0 ||
-            !previousContextData ||
-            previousContextData.length === 0 ||
-            !contextData.every((d) => d.animationValid) ||
-            !previousContextData.every((d) => d.animationValid)
-        ) {
+        if (contextData.length === 0 || !previousContextData || previousContextData.length === 0) {
             this.updateAreaPaths(paths, contextData);
             return;
         }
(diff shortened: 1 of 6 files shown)


packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
```diff
@@ -707,6 +701,7 @@ export class AreaSeries extends CartesianSeries<
 
         const fns = prepareAreaPathAnimation(newData, oldData, this.processedData?.reduced?.diff);
         if (fns === undefined) {
+            animationManager.skipCurrentBatch();
```
lsjroberts, line 704: Would this present issues when using on combined series types charts?

packages/ag-charts-website/src/content/docs/line-series-test/_examples/category-animation/main.ts
```diff
@@ -1,5 +1,7 @@
 import { AgChartOptions, AgEnterpriseCharts, time } from 'ag-charts-enterprise';
 
+window.agChartsDebug = 'animation';
```
lsjroberts: intentional?

lsjroberts (approved):

packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
alantreadway: So it was always the intent that if one series didn't animate, that the others shouldn't either - so I believe this is desirable and consistent with that idea?

packages/ag-charts-website/src/content/docs/line-series-test/_examples/category-animation/main.ts
alantreadway: Good spot, fixed

lsjroberts (approved):

PR #468, opened 2023-11-07
AG-8492 Add prettier formatting to the generated examples on build

lsjroberts:
Kinda related to https://ag-grid.atlassian.net/browse/AG-8492

This has picked up some jsx/tsx errors, namely adjacent unwrapped elements and comments in html files create invalid react.

This only runs on build, and typically takes 38s on my machine. Not massively ideal.

Requires us to run `find packages/ag-charts-website/src/content/docs -type d -empty -delete` once merged to delete empty folders and prevent old deleted examples from being generated and then failing the prettier formatting.

Also formats all the examples with prettier. Check this commit https://github.com/ag-grid/ag-charts/pull/468/commits/e451b7dd8470b92c143806818db50dacfb1b8bed for the actual changes.


packages/ag-charts-website/src/astro/integrations/agCharts.ts
```diff
@@ -0,0 +1,61 @@
...
+                        'prettier',
+                        './dist/packages/ag-charts-website/**/examples/**/*.{html,js,ts,jsx,tsx}',
+                        '--write',
+                        '--ignore-path',
+                        './dist/.prettierignore',
+                        '--log-level',
+                        'error',
+                    ],
+                    { stdio: 'inherit' }
+                );
```
alantreadway, line 54: Rather than running prettier as a post-processing step on `dist/` outputs, could we instead use the JS API during generation, which may remove some of the overhead and other drawbacks to this approach?

lsjroberts: As discussed.

It is possible to do it like that for the js parts, but not for the html. Astro puts the example's html into an `.astro` template which then gets generated. There doesn't appear to be an integrated way in astro to run after that step other than in the `astro:build:done` hook.

It does mean the serve and build are not in parity, but that can be a future investigation.

lsjroberts:
I found a method!

PR #473, opened 2023-11-07
AG-9310 - Fix AreaSeries category axis handling.

alantreadway:
Adds working support for `category` axes and data for `AreaSeries`.

### DIFF
--- a/packages/ag-charts-community/src/chart/series/cartesian/areaUtil.ts
+++ b/packages/ag-charts-community/src/chart/series/cartesian/areaUtil.ts
@@ -82,12 +82,20 @@ function pairFillCategoryData(
     const oldPoints = splitFillPoints(oldData);
     const newPoints = splitFillPoints(newData);
 
+    const pairOpts = { multiDatum: true, datumIndexMultiplier: 2 };
+
     return {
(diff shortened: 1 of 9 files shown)


packages/ag-charts-community/src/chart/series/cartesian/areaUtil.ts
```diff
@@ -82,12 +82,20 @@ function pairFillCategoryData(
     const oldPoints = splitFillPoints(oldData);
     const newPoints = splitFillPoints(newData);
 
+    const pairOpts = { multiDatum: true, datumIndexMultiplier: 2 };
```
lsjroberts, line 85: Do we expect `datumIndexMultiplier` to be anything other than `1` or `2`? It's not a particularly clear param. Would a boolean flag be possible?

lsjroberts (approved):
LGTM. One small question, but good to merge.

packages/ag-charts-community/src/chart/series/cartesian/areaUtil.ts
alantreadway: So the idea is that for particular series, there are N points rendered at a given X value - this multiplier helps to align the index of points back to datum indices for the purpose of being able to use the `diff` indices. A `boolean` I think is less clear in that respect?

I've added a comment to the property to explain it's use - is this ok?

PR #474, opened 2023-11-07
Ag 9571/gallery examples

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-9570


packages/ag-charts-enterprise/src/test/load-example.ts
```diff
@@ -0,0 +1,46 @@
...
+        // Remove grossly unsupported lines.
+        if (filters.some((f) => f.test(line))) continue;
+        // Remove declares.
+        line = line.replace(/declare var.*;/g, '');
+
+        lines.push(line);
+    }
+
+    return lines;
+};
```
alantreadway: Could we reference the versions of these functions in `ag-charts-community` rather than repeating them here?

packages/ag-charts-enterprise/src/test/load-example.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export function parseExampleOptions(evalFn: string, exampleJs: string, dataJs?: string) {
+    const evalExpr = [dataJs ? dataJs : '', ...cleanJs(exampleJs), `return ${evalFn};`].join('\n');
+    // @ts-ignore - used in the eval() call.
```
alantreadway: Nit: `// @ts-ignore` should be redundant.

PR #483, opened 2023-11-08
AG-9749 - Fix BarSeries(-like) rapid legend toggles.

alantreadway:
Since the animation for `BarSeries` and similar series somewhat rely on the scene-graph state being in a consistent state when calculating `from` and `to` properties, we need to actively finish any running animations before calculating new animations.

I've added explicit `AnimationManager.stopByAnimationGroupId(this.id)` calls in strategic places to ensure the affected series types all achieve scene-graph node consistency prior to new animation calculations.


lsjroberts (approved):
lgtm

PR #486, opened 2023-11-09
AG-8492 Add prettier formatting middleware and on generated contents.json

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-8492

This method works on both dev server and build.

It does not format the generated ts/js files, instead it formats the text in the `contents.json` files. Since that is what is actually displayed in the code blocks.

Then there is a middleware to catch the `index.html` files which work slightly differently due to astro templates.

Trying to push it all through the middleware did not work as you can not call `fetch()` to the localhost in the astro build process. It gives an `ECONNREFUSED` error.

### DIFF
--- a/packages/ag-charts-website/src/utils/format.ts
+++ b/packages/ag-charts-website/src/utils/format.ts
@@ -1,0 +1,55 @@
+import * as prettier from 'prettier';
+
+/**
+ * Format a block of text by the inferred file type.
+ * @param fileName Name of the file, including the extension.
+ * @param text Text to format.
+ * @param extensionsToFormat optional Only format the text if its extension is in this array.
+ * @returns Promise<string>
+ */
+export function format(fileName: string, text: string, extensionsToFormat?: Array<string>) {
+    const parts = fileName.split('.');
+    const maybeExtension = parts[parts.length - 1];
+
+    let extension: PrettierExtension = 'html';
+    if (Object.keys(prettierParsers).includes(maybeExtension)) {
+        extension = maybeExtension as PrettierExtension;
+    }
+
+    if (extensionsToFormat && !extensionsToFormat.includes(extension)) {
+        return text;
+    }
+
+    return prettier.format(text, {
+        parser: prettierParsers[extension],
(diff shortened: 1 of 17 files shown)


packages/ag-charts-website/src/utils/format.ts
```diff
@@ -0,0 +1,55 @@
...
+const prettierConfig = {
+    tabWidth: 4,
+    printWidth: 120,
+    singleQuote: true,
+    importOrder: ['^ag-charts-(.*)$', '^[./]'],
+    importOrderParserPlugins: ['typescript', 'decorators-legacy'],
+    importOrderSeparation: true,
+    importOrderSortSpecifiers: true,
+    plugins: ['@trivago/prettier-plugin-sort-imports'],
+};
```
alantreadway: Nit: Could we load `.prettierrc` and trim it rather than saying we do (I suspect there will be drift over time, so the comment will go crusty)?

alantreadway (approved):
LGTM, but @taktran should review too.

packages/ag-charts-website/src/utils/format.ts
```diff
@@ -0,0 +1,59 @@
...
+
+type PrettierExtension = 'html' | 'json' | 'js' | 'jsx' | 'ts' | 'tsx';
+
+const prettierParsers: { [T in PrettierExtension]: string } = {
+    html: 'html',
+    js: 'babel',
+    ts: 'typescript',
+    jsx: 'babel',
+    tsx: 'babel-ts',
+    json: 'json',
```
taktran: Do we need to add `css` here as well?

taktran (approved):
Great work 👏🏼 👏🏼 👏🏼 

As mentioned before, I would love to remove the need for generating the `html` using Astro, and instead have some other templating solution like the other code files. That would remove the need for the custom middleware. Something for another day though.

packages/ag-charts-website/src/utils/format.ts
lsjroberts: Good spot

PR #487, opened 2023-11-09
AG-5670 Bullet themes and defaults

olegat:
https://ag-grid.atlassian.net/browse/AG-9301
https://ag-grid.atlassian.net/browse/AG-9321
https://ag-grid.atlassian.net/browse/AG-9323

- Fixes incorrect param&return types for `customDefaultAxesSwapper` https://github.com/ag-grid/ag-charts/blob/83422ce66b91c212454f6d89f25410fc8a25ff07/packages/ag-charts-community/src/module/coreModules.ts#L101
- Adds files _bulletDefaults.ts_ and _bulletThemes.ts_.
- Adds styling options to `BulletSeries`
- Adds `scale.max` property to `BulletSeries`. Clip the colour ranges to not exceed `scale.max`

Test file:
https://plnkr.co/edit/BqtSjxH0R7fgHi6g?open=main.js

### DIFF
--- a/packages/ag-charts-community/src/options/series/cartesian/bulletOptions.ts
+++ b/packages/ag-charts-community/src/options/series/cartesian/bulletOptions.ts
@@ -1,5 +1,6 @@
 import type { AgSeriesTooltip, AgSeriesTooltipRendererParams } from '../../chart/tooltipOptions';
 import type { AgBaseSeriesOptions, AgBaseSeriesThemeableOptions } from '../seriesOptions';
+import type { AgBarSeriesStyle } from './barOptions';
 
 interface BulletSeriesKeysAndNames {
     /** The key to use to retrieve the gauge value from the data. */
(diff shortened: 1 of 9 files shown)


packages/ag-charts-community/src/options/series/cartesian/bulletOptions.ts
```diff
@@ -16,7 +17,17 @@ export interface AgBulletSeriesTooltipRendererParams<TDatum = any>
     extends AgSeriesTooltipRendererParams<TDatum>,
         BulletSeriesKeysAndNames {}
 
-export interface AgBulletSeriesThemeableOptions extends AgBaseSeriesThemeableOptions {}
+export interface AgBulletSeriesStyle extends AgBarSeriesStyle {}
+
+export interface AgBulletScaleOptions {
+    /** Maximum value of the scale. Any values exceeding this number will be clipped to this maximum */
+    max?: number;
```
alantreadway, line 24: The way to achieve this with other chart-types is via the axis options - we should follow the same approach for this.

See https://ag-grid.com/javascript-charts/axes-domain/#domain-min--max

olegat: I thought of that, but @AG-DavidG and I discussed this. We came to the conclusion that bullet graphs are gauges that don't conceptually have any axes (even though the underlying implementation uses axes). So having a dedicated `scale.max` makes sense in this case.

Although we could perhaps just have `scale.max` be an alias for `axes[0].domain.max` to make it easier to maintain.

alantreadway (approved):
LGTM

PR #490, opened 2023-11-09
Fix docs snippets syntax to enable prettier formatting

lsjroberts:
Using `...` in code snippets breaks the formatting, so I replaced that with `// ...`.

Also, need to use `ts` not `js` for snippets with types in them.

I haven't adjusted any of the formatted results so authors can do that to their tastes.


alantreadway (approved):
Nice, thanks for the cleanup!

JamesSwinton:
Thanks!

PR #491, opened 2023-11-09
Enable omitting labelKey or setting it to a numeric field

jacobp100:
AG-9424

### DIFF
--- a/packages/ag-charts-community/src/chart/series/hierarchy/hierarchySeries.ts
+++ b/packages/ag-charts-community/src/chart/series/hierarchy/hierarchySeries.ts
@@ -150,11 +150,7 @@ export abstract class HierarchySeries<S extends SeriesNodeDatum> extends Series<
     }
 
     getDatumIdFromData(node: HierarchyNode) {
-        const { labelKey } = this;
-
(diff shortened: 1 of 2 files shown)


packages/ag-charts-community/src/chart/series/hierarchy/hierarchySeries.ts
```diff
@@ -150,11 +150,7 @@ export abstract class HierarchySeries<S extends SeriesNodeDatum> extends Series<
     }
 
     getDatumIdFromData(node: HierarchyNode) {
-        const { labelKey } = this;
-
-        if (labelKey != null) {
-            return node.datum?.[labelKey];
-        }
+        return `${node.index}`;
```
alantreadway, line 153: We should keep the `labelKey` case (with fallback to `node.index`) since it will be less fragile than the `node.index` when there are data-changes?

alantreadway: Discussed offline, we'll revisit once we get to Sunburst animation cases.

alantreadway (approved):
LGTM

PR #495, opened 2023-11-09
AG-9649 - Disable animations for unsupported use-cases

alantreadway:
Disables animations in the following use-cases:
- Layout change which adjusts the position/dimensions of the `Chart.animationRect`.
- Series option changes that affect `type`, `direction`, `xKey`, `yKey`, `sizeKey`, `angleKey`, `stacked` or `stackGroup` properties.
  - Additionally forces axis recreation if series recreation happens to force the initial-load animation.

New concepts:
- `Chart.animationRect` - the `BBox` containing all animated elements (series + axes) - allows detection of un-animatable layout changes.

Bonus:
- Moves `Chart` layout processing into a distinct class and sub-folder `src/chart/update/` - the intent would be to split more `Chart.performUpdate()` processing out here incrementally.


lsjroberts (approved):
Nice one.

PR #496, opened 2023-11-09
Initial sunburst

jacobp100:


packages/ag-charts-website/src/content/docs/sunburst-series/_examples/org-chart/data.ts
```diff
@@ -0,0 +1,101 @@
+let rowData: any[] = [
```
jacobp100, line 1: This is duplicated from the treemap - maybe there's a better way

packages/ag-charts-community/src/options/series/hierarchy/sunburstOptions.ts
```diff
@@ -0,0 +1,110 @@
...
+    minimumFontSize?: FontSize;
+    /**
+     * Text wrapping strategy for treemap labels.
+     * `'always'` will always wrap text to fit within the tile.
+     * `'hyphenate'` is similar to `'always'`, but inserts a hyphen (`-`) if forced to wrap in the middle of a word.
+     * `'on-space'` will only wrap on white space. If there is no possibility to wrap a line on space and satisfy the tile dimensions, the text will be truncated.
+     * `'never'` disables text wrapping.
+     * Default: `'on-space'`
+     */
+    wrapping?: TextWrap;
```
alantreadway: Nit: Maybe this JSDoc belongs on the declaration of `TextWrap`?

packages/ag-charts-community/src/scene/shape/sector.ts
```diff
@@ -54,11 +58,23 @@ export class Sector extends Path {
                 path.arc(centerX, centerY, innerRadius, endAngle, startAngle, true);
             }
         } else {
-            path.moveTo(centerX + innerRadius * Math.cos(startAngle), centerY + innerRadius * Math.sin(startAngle));
-            path.arc(centerX, centerY, outerRadius, startAngle, endAngle);
+            const innerAngleOffset = inset !== 0 && innerRadius !== 0 ? inset / innerRadius : 0;
```
alantreadway: Could you verify that the PieSeries animations are happy after these changes? (Sector rendering changes are tingling my spidey-sense)

packages/ag-charts-website/src/content/docs/sunburst-series/_examples/org-chart/data.ts
alantreadway: This is ok, we don't share code between examples so they are as standalone as possible.

packages/ag-charts-enterprise/src/series/sunburst/sunburstSeries.ts
```diff
@@ -0,0 +1,502 @@
...
+        });
+    }
+
+    override async createNodeData() {
+        return [];
+    }
+
+    override getSeriesDomain() {
+        // FIXME: Is this right?
+        return [0, 1];
```
alantreadway: This is used for the axis scales - if there are no axes maybe `[NaN, NaN]` might be more appropriate (or omit it completely - `HierarchySeries` could probably have this as the default implementation?

packages/ag-charts-enterprise/src/series/sunburst/sunburstSeries.ts
```diff
@@ -0,0 +1,502 @@
...
+            sizeKey,
+            title,
+            color,
+            seriesId,
+        });
+    }
+
+    override async createNodeData() {
+        return [];
+    }
```
alantreadway, line 500: Seeing what is going on in the update processing above, I still think it would be better to break some of that out into this method, and keep the `updateNodes()` and `updateSelections()` as a purer 'apply calculated state to scene-graph'?

When it comes to animation, the separation of calculations vs. application will help I think.

packages/ag-charts-community/src/scene/shape/sector.ts
jacobp100: The radial bar works - should be fine?

alantreadway (approved):
LGTM

PR #497, opened 2023-11-09
AG-5670 Join duplicate bar-like series code into AbstractBarSeries

olegat:
https://ag-grid.atlassian.net/browse/AG-5670

Cut down duplicate code in the follow series classes:
- `BarSeries`
- `BoxPlotSeries`
- `BullerSeries`
- `RangeBar`
- `WaterfallSeries`

This merges common features into a new `AbstractBarSeries`. Right now, this class just contains a `direction` property and methods to help with category/number axes management.

### DIFF
--- a/packages/ag-charts-community/src/chart/series/cartesian/abstractBarSeries.ts
+++ b/packages/ag-charts-community/src/chart/series/cartesian/abstractBarSeries.ts
@@ -1,0 +1,44 @@
+import type { Direction } from '../../../options/chart/types';
+import type { Node } from '../../../scene/node';
+import { DIRECTION, Validate } from '../../../util/validation';
+import type { ChartAxis } from '../../chartAxis';
+import { ChartAxisDirection } from '../../chartAxisDirection';
+import type { SeriesNodeDatum } from '../seriesTypes';
+import { CartesianSeries } from './cartesianSeries';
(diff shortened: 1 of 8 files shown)


packages/ag-charts-enterprise/src/series/box-plot/blotPlotUtil.ts
alantreadway, line 1: Please undo this change, this was following he same pattern as other series for animation handling, and once add/update/remove/legend-toggle cases are added the module is likely to be much more complex.

alantreadway (commented):
If you could address the one comment, this looks good - thanks for the clean-up.

As a heads-up, the intent is to move to composed series implementation longer-term, so I would avoid pulling more behaviour into these classes where possible (our general approach has been to break-out reusable functions for a bunch of cases).

packages/ag-charts-enterprise/src/series/box-plot/blotPlotUtil.ts
olegat: Ah, OK. Thanks for providing context. I'll roll this change back.

olegat:
> If you could address the one comment, this looks good - thanks for the clean-up.
> 
> As a heads-up, the intent is to move to composed series implementation longer-term, so I would avoid pulling more behaviour into these classes where possible (our general approach has been to break-out reusable functions for a bunch of cases).

Noted. I'll do that where applicable.

PR #500, opened 2023-11-09
Ag 9571/gallery examples

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-9570


packages/ag-charts-website/src/content/gallery/_examples/grouped-horizontal-bar/data.ts
```diff
@@ -1,9 +1,10 @@
-export function getData() {
+export function getData(): any[] {
```
alantreadway: Is the `any[]` needed, it looks like it should be inferred ok?

alantreadway (approved):
One nit, but otherwise LGTM

PR #501, opened 2023-11-10
AG-9361: Fixed when pie data includes negative values, colours do not…

iMoses:
… display correctly


alantreadway (approved):
LGTM - thanks!

Is it worth adding a single visual test-case for this?

iMoses:
> Is it worth adding a single visual test-case for this?

I think you're right. I'll do it before I merge.

PR #504, opened 2023-11-10
AG-9715 - Rewrite tabs as astro component

taktran:
The react component was getting server side and client side data mixed up and causing the page not to load. It might be that it's a complicated use case that involves parsing markdoc into html, extracting tab data from the markdoc data and then rendering the tabs as react. Removing the react part of it simplifies the tech stack and allows Astro to optimize the server rendering part, while the client side can be handled by a standard Web Component. In theory, because most of the logic is a Web Component, it can be ported to other future tech stacks.


packages/ag-charts-website/src/components/tabs/TabsWithHtmlChildren.astro
```diff
@@ -9,6 +10,110 @@ const tabsData = extractTabsData(html);
...
+            this.updateTabs();
+            this.listenToHashChanges();
+        }
+
+        get selected(): TabData {
+            return this.tabsData[this.selectedIndex];
+        }
+
+        listenToHashChanges() {
+            browserHistory?.listen(({ location }) => {
```
iMoses: should we "unlisten" when the component is removed, or is that not a real use case?

iMoses (approved):
One question, other than that LGTM 👍

packages/ag-charts-website/src/components/tabs/TabsWithHtmlChildren.astro
taktran: I figured the site would be loading new pages since it's more of a Multipage app with Astro, so would clean up naturally. However, it's a good shout 👍🏼 , and I should really be using `connectedCallback` and `disconnectedCallback` for this.

PR #507, opened 2023-11-13
AG-9649 - Stop animations in unsupported cases

alantreadway:
Blocks animations when:
- Chart layout changes.
- Data-validations don't pass:
  - Categories for category data must be unique.
  - The main continuous data axis must be ordered.


lsjroberts (approved):
lgtm

PR #509, opened 2023-11-13
Use HSL for colour interpolation for color scales

jacobp100:


alantreadway (approved):
LGTM once test breaks are fixed.

packages/ag-charts-community/src/scene/shape/linearGradientFill.ts
```diff
@@ -0,0 +1,72 @@
+import { ColorScale } from '../../scale/colorScale';
+import { BBox } from '../bbox';
+import type { RenderContext } from '../node';
+import { RedrawType, SceneChangeDetection } from '../node';
+import type { Path } from './path';
+import { Shape } from './shape';
+
+export class LinearGradientFill extends Shape {
+    @SceneChangeDetection({ redraw: RedrawType.MAJOR })
+    direction: 'to bottom' | 'to right' = 'to right';
```
alantreadway, line 10: Nit: could you use an identifier format that doesn't include whitespace?

packages/ag-charts-community/src/scene/scene.ts
```diff
@@ -308,6 +308,7 @@ export class Scene {
 
         const renderCtx: RenderContext = {
             ctx,
+            devicePixelRatio: this.overrideDevicePixelRatio ?? 1,
```
alantreadway: Use `this.canvas.pixelRatio` as otherwise the fallback of `1` is incorrect (and semantically makes more sense since it's the pixel ration reported for the canvas)?

packages/ag-charts-community/src/scale/colorScale.test.ts
```diff
@@ -64,3 +64,81 @@ test('multi-value domain', () => {
     expect(scale.convert(500)).toBe('#ffffff');
     expect(scale.convert(501)).toBe('#ffffff');
 });
+
+test('heatmap multistop', () => {
```
alantreadway: Nit: Could you add an enclosing `describe()` block for this entire module whilst you're here?

alantreadway (approved):
LGTM

PR #510, opened 2023-11-13
AG-9649 - Fix initial load for several series.

alantreadway:
#507 blocked initial load animation for a few series (scatter, histogram, radar area/line and range-area are affected) - this change rectifies that state by making the validation checks introduced skippable for series-types where the validations don't make sense.

Additionally adds animation validation to a few cases that were missed in the previous PR.


lsjroberts (approved):
LGTM.

But do bubble & scatter need the `diff()` processor if they are no longer checking the processed data for the animation validation? Or do they use it elsewhere?

alantreadway:
> LGTM.
> 
> But do bubble & scatter need the `diff()` processor if they are no longer checking the processed data for the animation validation? Or do they use it elsewhere?

Fair point - I've removed it.

PR #511, opened 2023-11-13
AG-9301 Set overrideable lengths of bullet graphs node shapes

olegat:
https://ag-grid.atlassian.net/browse/AG-9301
The bar width defaults to 50% of the width/height of the series area. The target line defaults to 75% of the width/height of the series area.

### DIFF
--- a/packages/ag-charts-community/src/options/series/cartesian/bulletOptions.ts
+++ b/packages/ag-charts-community/src/options/series/cartesian/bulletOptions.ts
@@ -25,8 +25,13 @@ export interface AgBulletScaleOptions {
 }
 
 export interface AgBulletSeriesThemeableOptions extends AgBulletSeriesStyle, AgBaseSeriesThemeableOptions {
+    /* Width of the bar relative to the width/height of the series area. */
+    widthRatio?: number;
     /** Styling options for the target node. */
-    target?: AgBulletSeriesStyle;
+    target?: AgBulletSeriesStyle & {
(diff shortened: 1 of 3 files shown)


alantreadway (approved):
LGTM

packages/ag-charts-community/src/options/series/cartesian/bulletOptions.ts
```diff
@@ -25,8 +25,13 @@ export interface AgBulletScaleOptions {
 }
 
 export interface AgBulletSeriesThemeableOptions extends AgBulletSeriesStyle, AgBaseSeriesThemeableOptions {
+    /* Width of the bar relative to the width/height of the series area. */
+    widthRatio?: number;
     /** Styling options for the target node. */
-    target?: AgBulletSeriesStyle;
+    target?: AgBulletSeriesStyle & {
```
iMoses, line 31: @olegat FYI, we should avoid using type literals inside the options files because it breaks the documentation. We cannot display the type of this property because a type literal has no name.

olegat: Okay noted. The names of these options are still TBD which is why I haven't bothered creating separate typename yet.

iMoses: I get it... but in future builds this will throw an error and fail the build.
Better to have a temp name than no name at all :)

PR #512, opened 2023-11-13
AG-8492 Add examples typescript check to build process

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-8492

The main changes are https://github.com/ag-grid/ag-charts/commit/9f820f2574f1ae16b4158bbe631ca09648015a5b & https://github.com/ag-grid/ag-charts/commit/b42868e7fb1f84a88ea9046d004bda470f8b0a05. The rest are fixes these caught.

### DIFF
--- a/packages/ag-charts-website/project.json
+++ b/packages/ag-charts-website/project.json
@@ -17,14 +17,9 @@
       }
     },
     "build-tsc": {
-      "executor": "@nx/js:tsc",
-      "inputs": ["tsDefaults"],
-      "outputs": ["{options.outputPath}"],
+      "executor": "nx:run-commands",
       "options": {
-        "main": "packages/ag-charts-website/src",
-        "outputPath": "dist/packages/ag-charts-website/out-tsc",
-        "tsConfig": "packages/ag-charts-website/tsconfig.json",
-        "compiler": "tsc"
+        "command": "tsc --project packages/ag-charts-website/tsconfig.json"
       }
     },
     "dev": {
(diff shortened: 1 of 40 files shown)


packages/ag-charts-website/tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "extends": "../../tsconfig.base.json",
+  "include": ["./src/content/docs/**/_examples/**/*.ts", "./src/content/gallery/_examples/**/*.ts"],
```
lsjroberts, line 3: This excludes the rest of the website package, however there are currently 800 ts errors in that. But none of those are used in the library so not a big deal.

This does though hinder some behaviour in vscode around our usage of `import foo from @design-system/foo` etc. Not a major issue and we can fix up all the ts errors in the website at some point later.

packages/ag-charts-website/project.json
```diff
@@ -17,14 +17,9 @@
...
-      "inputs": ["tsDefaults"],
-      "outputs": ["{options.outputPath}"],
+      "executor": "nx:run-commands",
       "options": {
-        "main": "packages/ag-charts-website/src",
-        "outputPath": "dist/packages/ag-charts-website/out-tsc",
-        "tsConfig": "packages/ag-charts-website/tsconfig.json",
-        "compiler": "tsc"
+        "command": "tsc --project packages/ag-charts-website/tsconfig.json"
       }
```
alantreadway, line 23: What was the motivation to switch to the `tsc` command-line tool? I'd like to try and minimise this scenario, as it adds more edge cases on upgrade vs. following the 'standard' Nx approach?

lsjroberts: I'm not sure what the nx/js:tsc is running, but it doesn't output any of the errors and passes while `tsc` fails correctly.

alantreadway: We bounced around a bit on this offline, and are unable to find a way to use the Nx approach to resolve this issue - lets move ahead with the proposed change :+1:

alantreadway (approved):
LGTM

PR #515, opened 2023-11-13
AG-9717 - Rework Axis label selection datums ready for animation.

alantreadway:
Attempts to rework the axis label processing flow so that we end up with most label properties as a `AxisLabelNodeDatum` managed by the label-related `Selection` - this should make animating the individual labels easier in a follow-up PR.

### DIFF
--- a/packages/ag-charts-enterprise/src/axes/angle/angleAxis.ts
+++ b/packages/ag-charts-enterprise/src/axes/angle/angleAxis.ts
@@ -194,7 +194,7 @@ export abstract class AngleAxis<
         const { label, tickLabelGroupSelection } = this;
 
         const ticks = this.tickData;
-        tickLabelGroupSelection.update(label.enabled ? ticks : []).each((node, _, index) => {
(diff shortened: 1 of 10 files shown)


packages/ag-charts-enterprise/src/axes/angle/angleAxis.ts
```diff
@@ -194,7 +194,7 @@ export abstract class AngleAxis<
         const { label, tickLabelGroupSelection } = this;
 
         const ticks = this.tickData;
-        tickLabelGroupSelection.update(label.enabled ? ticks : []).each((node, _, index) => {
+        tickLabelGroupSelection.update(label.enabled ? (ticks as any[]) : []).each((node, _, index) => {
```
lsjroberts, line 197: 👀 any?

lsjroberts (approved):

packages/ag-charts-enterprise/src/axes/angle/angleAxis.ts
alantreadway: Will see if I can address in the follow-up PR I'll send shortly.

PR #517, opened 2023-11-14
AG-8555 Add visible range scale to axis tick calculation

lsjroberts:

### DIFF
--- a/packages/ag-charts-community/src/chart/axis/axis.ts
+++ b/packages/ag-charts-community/src/chart/axis/axis.ts
@@ -711,8 +711,6 @@ export abstract class Axis<S extends Scale<D, number, TickInterval<S>> = Scale<a
             return;
         }
 
-        this.setTickCount();
-
         scale.nice = nice;
         scale.update();
     }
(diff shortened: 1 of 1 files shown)


packages/ag-charts-community/src/chart/axis/axis.ts
```diff
@@ -711,8 +711,6 @@ export abstract class Axis<S extends Scale<D, number, TickInterval<S>> = Scale<a
             return;
         }
 
-        this.setTickCount();
```
lsjroberts, line 714: This call doesn't do anything as it short circuit returns if no count param or not continuous (skipped on line 711).

packages/ag-charts-community/src/chart/axis/axis.ts
```diff
@@ -1145,14 +1143,21 @@ export abstract class Axis<S extends Scale<D, number, TickInterval<S>> = Scale<a
...
 
         return max - min;
     }
 
+    protected calculateVisibleRange() {
+        const { visibleRange } = this;
+        const visibleScale = 1 / (visibleRange[1] - visibleRange[0]);
+
+        return this.calculateAvailableRange() * visibleScale;
+    }
```
lsjroberts, line 1159: The `calculateAvailableRange()` function is still used for the label collision checking

packages/ag-charts-community/src/chart/axis/axis.ts
alantreadway: Nice spot!

alantreadway (approved):
LGTM

PR #520, opened 2023-11-14
Adjust ag-charts-enterprise entry point to inc. community

iMoses:
Mimic interfaces like `ag-grid`


packages/ag-charts-community/src/chart/test/load-example.ts
```diff
@@ -4,7 +4,7 @@ import { Logger } from '../../util/logger';
 
 export const filters = [
     /.*AgChart\.(update|create)/,
-    /.*AgEnterpriseCharts\.(update|create)/,
+    /.*AgChart\.(update|create)/,
```
iMoses: i'll remove this duplication before merged

iMoses:
BTW, do we wanna change the main object from `AgChart` to `AgCharts`?

rmc-software:
@iMoses "BTW, do we wanna change the main object from AgChart to AgCharts?" - yes we should - breaking change obv. Perhaps we can also keep AgChart and deprecate it?

packages/ag-charts-community/project.json
```diff
@@ -15,7 +15,7 @@
         "rollupConfig": "packages/ag-charts-community/rollup.config.cjs",
         "project": "packages/ag-charts-community/package.json",
         "compiler": "tsc",
-        "format": ["cjs"]
+        "format": ["cjs", "esm"]
```
alantreadway: I'm a bit concenred about the performance impact here - is it doubling our reload times?

packages/ag-charts-enterprise/src/main.ts
```diff
@@ -60,19 +60,11 @@ _ModuleSupport.registerModule(TreemapSeriesModule);
 _ModuleSupport.registerModule(WaterfallModule);
 _ModuleSupport.registerModule(ZoomModule);
 
-export class AgEnterpriseCharts {
-    public static create(options: AgChartOptions): AgChartInstance {
+export class AgChart extends AgChartCommunity {
```
alantreadway: Does this approach work for the FW cases, i.e. do we currently get the license message on opening an enterprise React example?

iMoses: Haven't checked specific libraries, just plain Javascript. I'll verify tomorrow.

packages/ag-charts-community/project.json
iMoses: I don't think it's that bad. I'm also not sure if we have any other options.

PR #524, opened 2023-11-15
AG-8489 - Pie Series animation fixes.

alantreadway:
Fixes:
- Sector radius animation on add/remove (now in step with updated sectors).
- An error on skipped updates, where Selection ids were lost due to animation-validations only being applied selectively. (`PieSeries.getDatumId()` was short-circuiting without the validation result being present).


lsjroberts (approved):
LGTM.

PR #528, opened 2023-11-15
Fix options typing + throw error on docs unknown member types

iMoses:


taktran (approved):
Noice 👍🏼 Good that you're adding an exception for future encounters too 🚀

PR #533, opened 2023-11-15
Fix drawing of linear gradients

jacobp100:
The total transform of the mask would be set to a non-integer value, so would break drawing at the pixel level. Now we compute the bounding box from the perspective of the root node, reset the transform, then draw in the transformed bounding box

### DIFF
--- a/packages/ag-charts-community/src/scene/shape/linearGradientFill.ts
+++ b/packages/ag-charts-community/src/scene/shape/linearGradientFill.ts
@@ -1,5 +1,4 @@
 import { ColorScale } from '../../scale/colorScale';
-import { BBox } from '../bbox';
 import type { RenderContext } from '../node';
 import { RedrawType, SceneChangeDetection } from '../node';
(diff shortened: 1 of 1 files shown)


alantreadway (approved):
LGTM

packages/ag-charts-community/src/scene/shape/linearGradientFill.ts
```diff
@@ -12,40 +11,58 @@ export class LinearGradientFill extends Shape {
...
+    get mask(): Path | undefined {
+        return this._mask;
+    }
+    set mask(newMask: Path | undefined) {
+        if (this._mask != null) {
+            this.removeChild(this._mask);
+        }
+
+        if (newMask != null) {
+            this.appendChild(newMask);
```
jacobp100, line 23: Am I right in thinking that because the `group`'s render method is responsible for rendering children - and we're not inheriting from group - we won't draw this mask?

PR #536, opened 2023-11-15
Cleanup & simplify load-example logic

iMoses:
Note: this is already included in #537


alantreadway (approved):
LGTM, thanks for the cleanup!

PR #538, opened 2023-11-15
Revert "AG-9223 - Fix trailing slash infinite loop"

iMoses:
Reverts ag-grid/ag-charts#534

locally the problem seems worse after the fix so I'm attempting to debug the previous version


alantreadway (commented):
I'm not close enough to this to understand what is going on - happy to proceed if Tak is happy.

taktran (approved):
Go for it. Will need to investigate further

PR #541, opened 2023-11-16
AG-9223: Fix docs infinite loop bug

iMoses:


packages/ag-charts-website/src/stores/frameworkStore.ts
```diff
@@ -3,22 +3,23 @@ import { persistentAtom, persistentMap } from '@nanostores/persistent';
...
 
 const LOCALSTORAGE_PREFIX = 'documentation';
 const DEFAULT_INTERNAL_FRAMEWORK: InternalFramework = 'vanilla';
 
-export const $internalFramework = persistentAtom(
+export const $internalFramework = persistentAtom<InternalFramework>(
     `${LOCALSTORAGE_PREFIX}:internalFramework`,
-    DEFAULT_INTERNAL_FRAMEWORK
+    DEFAULT_INTERNAL_FRAMEWORK,
+    { listen: false }
```
taktran, line 17: Nice catch 🌟 

I wonder if we need this for the other stores too ie, `themeStore` and `darkmodeStore`. Although, the issue isn't as bad since it doesn't cause a url changes, they might also be in infinite loops but we just don't know about it

taktran (approved):
Great work 👍🏼 

Just a comment about other stores.

packages/ag-charts-website/src/stores/frameworkStore.ts
iMoses: It doesn't seem to cause the same bug, I'm assuming because the logic around components using it is much simpler, but it's probably worth fixing anyhow. Basically what we need to do is also listen to store changes and update local state accordingly, so we won't have a mismatch between tabs. So if a user changes dark/light mode or the theme it'll automatically apply to all other tabs as well (unless we don't want this feature, in which case `listen: false` would do just fine)
WDYT?

PR #542, opened 2023-11-16
Update defaults for diverging color scale

jacobp100:
Updates the heatmap colour scale to use more conventional colours, and copies the colour scale over to hierarchy series

###### Before
Screenshot 2023-11-16 at 09 54 14

###### After
Screenshot 2023-11-16 at 09 53 54


alantreadway (approved):
LGTM

PR #548, opened 2023-11-16
AG-9824 Updated charts website search bar and framework selector

salmenus:


packages/ag-charts-website/src/components/site-header/SiteHeader.astro
```diff
@@ -1,12 +1,32 @@
...
+import Search from '../search/Search';
+
+export interface Props {
+    showSearchBar?: boolean;
+}
+
+const {showSearchBar} = Astro.props;
+
+const path = Astro.url.pathname;
+const frameworks = FRAMEWORKS;
```
taktran: Don't think you need this line here as it isn't used?

packages/ag-charts-website/src/pages/[framework]/[pageName].astro
```diff
@@ -28,13 +32,24 @@ const extendedHeadings = addTabsToHeadings({
 });
 
 const menuEntry = await getEntry('menu', 'data');
+const frameworks = FRAMEWORKS;
+
+const frameworksData = frameworks.map((framework) => ({
```
Comment (taktran) [inline packages/ag-charts-website/src/pages/[framework]/[pageName].astro]: This isn't used anywhere?

packages/ag-charts-website/src/pages/[framework]/[pageName].astro
```diff
@@ -47,11 +62,31 @@ const menuEntry = await getEntry('menu', 'data');
                 },
             ]}
         >
-            <Header title={title} framework={framework} isEnterprise={enterprise} />
+            <div
+                id="doc-content"
```
Comment (taktran) [inline packages/ag-charts-website/src/pages/[framework]/[pageName].astro]: This looks like it was duplicated from above? I think it can be removed

taktran (changes requested):

packages/ag-charts-website/src/components/framework-selector-inside-doc/FrameworkSelectorInsideDocs.tsx
```diff
@@ -0,0 +1,43 @@
...
+                <span classnames={styles.divider}></span>
+                <select
+                    value={currentFramework}
+                    onChange={(event) => handleFrameworkChange(event.target.value)}
+                    onClick={(event) => event.stopPropagation()} // Prevent event propagation
+                    className={styles.select}
+                >
+                    {FRAMEWORKS.map((framework) => (
+                        <option value={framework} key={framework}>
+                            {framework.charAt(0).toUpperCase() + framework.slice(1)}
```
taktran: You can use `getFrameworkDisplayText` from `@utils/framework` here, which does a similar thing, but we can change it all in one place

taktran (approved):
LGTM, just a minor comment 👍🏼

PR #554, opened 2023-11-16
AG-9665 Fix console errors caused by error bars

olegat:
https://ag-grid.atlassian.net/browse/AG-9665

This adds a guard for contextNodeData[0], which could be an empty array.


alantreadway (approved):
LGTM

PR #556, opened 2023-11-16
AG-9323 Add optionConstructor for 'series[].colorRanges'

olegat:
https://ag-grid.atlassian.net/browse/AG-9323

This enables console warning when invalid color options are used. Without this, the @Validate annotations of the BulletColorRange class have no effect.


alantreadway (approved):
LGTM

PR #557, opened 2023-11-17
Treemap fixes

jacobp100:


iMoses:
Looking good, thanks! One note...
We also want to make sure it works in `OptionsNavigation.tsx` so that these options won't appear on the options ref page.
Add we'll need to add these exclusions to the `ApiReferencePage.tsx` using the `ApiReferenceConfigContext.Provider` (I think it's better if we pass a config property to the component and configure this from the `*.astro` files instead of hard-coding it.

packages/ag-charts-website/src/features/api-documentation/apiReferenceHelpers.ts
```diff
@@ -87,14 +87,20 @@ export function normalizeType(refType: TypeNode, includeGenerics?: boolean): str
...
+    const prefix = prefixPath.join('.');
     const { prioritise, include, exclude } = config;
     const isInterface = interfaceRef.kind === 'interface';
     if (include?.length || exclude?.length) {
-        members = members.filter(
-            (member) => !exclude?.includes(member.name) && (include?.includes(member.name) ?? true)
-        );
+        members = members.filter((member) => {
+            const fullMemberName = prefix !== '' ? `${prefix}.${member.name}` : member.name;
+            return exclude?.includes(fullMemberName) !== true && include?.includes(fullMemberName) !== false;
```
taktran: nitpick: Couldn't you remove the boolean comparison?

```suggestion
            return !exclude?.includes(fullMemberName) && include?.includes(fullMemberName);
```

packages/ag-charts-website/src/features/api-documentation/components/OptionsNavigation.tsx
```diff
@@ -80,7 +80,7 @@ export function OptionsNavigation({
                     <NavBreadcrumb breadcrumbs={breadcrumbs} rootInterface={rootInterface} basePath={basePath}>
                         {interfaceRef.kind === 'interface' && (
                             <NavGroup depth={breadcrumbs?.length ?? 0}>
-                                {processMembers(interfaceRef, config).map((member) => (
+                                {processMembers(interfaceRef, config, breadcrumbs).map((member) => (
```
iMoses: `breadcrumbs` aren't actually a part of the path, it's just for visuals

packages/ag-charts-website/src/features/api-documentation/components/OptionsNavigation.tsx
```diff
@@ -186,7 +186,7 @@ function NavProperty({
                 <>
                     <NavGroup depth={depth + 1}>
                         {isInterface
-                            ? processMembers(interfaceRef, config).map((childMember) => (
+                            ? processMembers(interfaceRef, config, []).map((childMember) => (
```
iMoses: you'll need to map the path and pull out the current config from the default otherwise this won't work
image

taktran (approved):
@iMoses will be better at looking through the options ref/api docs code, but LGTM.

Just a minor comment.

packages/ag-charts-community/src/chart/series/hierarchy/hierarchySeries.ts
```diff
@@ -79,10 +79,10 @@ export abstract class HierarchySeries<S extends SeriesNodeDatum> extends Series<
     colorKey?: string = undefined;
 
     @Validate(OPT_COLOR_STRING_ARRAY)
-    fills?: string[] = undefined;
+    fills: string[] = ['#c16068', '#a2bf8a', '#ebcc87', '#80a0c3', '#b58dae', '#85c0d1'];
```
alantreadway: Is this pushing the default theme fills as a fallback in? If so, could we use a constant defined by the default theme to capture this relationship?

packages/ag-charts-community/src/chart/series/hierarchy/hierarchySeries.ts
```diff
@@ -79,10 +79,10 @@ export abstract class HierarchySeries<S extends SeriesNodeDatum> extends Series<
...
 
     @Validate(OPT_COLOR_STRING_ARRAY)
-    strokes?: string[] = undefined;
+    strokes: string[] = ['#874349', '#718661', '#a48f5f', '#5a7088', '#7f637a', '#5d8692'];
```
alantreadway: Ditto

alantreadway (commented):
A few comments on the inline styles.

packages/ag-charts-website/src/features/api-documentation/components/OptionsNavigation.tsx
```diff
@@ -113,18 +117,29 @@ function NavGroup({ depth = 0, className, ...props }: { depth?: number } & AllHT
...
+    depth,
+    prefixPath: basePrefixPath,
     path,
+    defaultOptions,
     onClick,
 }: {
     member: MemberNode;
-    depth?: number;
+    depth: number;
+    prefixPath: string[];
```
iMoses: is `prefixPath` really needed here?
can't it be extracted from `path`?

packages/ag-charts-website/src/features/api-documentation/components/OptionsNavigation.tsx
```diff
@@ -113,18 +117,29 @@ function NavGroup({ depth = 0, className, ...props }: { depth?: number } & AllHT
...
+    defaultOptions: Record<string, ApiReferenceConfig>;
     onClick?: (navData: NavigationData) => void;
 }) {
     const selection = useContext(SelectionContext);
     const reference = useContext(ApiReferenceContext);
-    const config = useContext(ApiReferenceConfigContext);
+
+    const lastType = path[path.length - 1].type;
+    const defaultConfig = defaultOptions[lastType];
+    const prefixPath = defaultConfig != null ? [] : basePrefixPath;
```
iMoses: I'm a bit worried about this check, it may work under the current config and data conditions but seems error prune and can break easily

packages/ag-charts-community/src/chart/series/hierarchy/hierarchySeries.ts
jacobp100: > If an invalid colour is provided, please default to the theme, or hard-coded default theme colours instead of not rendering anything.

I didn't find a way to achieve this - but this matches the behaviour of the pie series

alantreadway (approved):
LGTM

packages/ag-charts-community/src/chart/series/hierarchy/hierarchySeries.ts
alantreadway: Discussed offline.

PR #568, opened 2023-11-20
AG-9649 - Stop animations in unsupported series cases.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-9649

Makes animation skipping the default for unimplemented animation cases.


lsjroberts (approved):
lgtm

PR #575, opened 2023-11-20
AG-9321 Bullet multiple series validation

olegat:
https://ag-grid.atlassian.net/browse/AG-9321

Only use the first bullet series when multiple bullet series are specified.

Filter out bullet series if the first series is anything other than a bullet series.

In both cases, `console.warn` the user about this validation error.


packages/ag-charts-community/src/chart/mapping/prepare.ts
```diff
@@ -176,6 +176,47 @@ function sanityCheckOptions<T extends AgChartOptions>(options: T) {
...
+function validateSoloSeries<T extends AgChartOptions>(options: T): T {
+    function isSoloSeries(series: SeriesOptionsTypes) {
+        return series.type === 'bullet';
+    }
+    function hasSoloSeries(options: SeriesOptionsTypes[]) {
+        for (const series of options) {
+            if (isSoloSeries(series)) return true;
+        }
+        return false;
+    }
```
alantreadway: We should probably encode this as part of `SeriesModule` to avoid having a series-specific behaviour.

olegat: Good suggestion. I'll change this.

packages/ag-charts-community/src/chart/mapping/prepare.ts
```diff
@@ -174,6 +175,44 @@ function sanityCheckOptions<T extends AgChartOptions>(options: T) {
...
+                Logger.warn(
+                    `series[${userIndex}] of type '${series[i].type}' is incompatible with other series types. Ignoring series[${userIndex}]`
+                );
+                series.splice(i, 1);
+            } else {
+                i++;
+            }
+            userIndex += 1;
+        }
+    }
```
alantreadway: I would keep the code here simpler by providing a less specific warning:
```
const rejects = new Set(series.filter(s => isSoloSeries(s.type)).map(s => s.type));
Logger.warnOnce(`Unable to mix these series types with the lead series type: ${rejects}`);

series = series.filter(s => !isSoloSeries(s.type));
```

olegat: I've got no objections with that. I'll change it.

alantreadway (approved):
LGTM - thanks!

PR #586, opened 2023-11-21
AG-8598 Fix double validation warning with tooltip

olegat:
https://ag-grid.atlassian.net/browse/AG-8598

Although `options.tooltip?.position` is of type AgTooltipPositionOptions, this isn't guaranteed because this value is user data. mergeSeriesOptions expands the globalTooltipPositionOptions using ..., which doesn't do what we want when the position isn't a valid type (such as a string in the example AG-8598).

Therefore, force the globalTooltipPositionOptions to be the correct type by converting it to `unknown` and explicitly validating the assumptions.


alantreadway (approved):
LGTM

PR #595, opened 2023-11-22
AG-8555 Fix zoom tick performance

lsjroberts:

One part of this is clipping / only creating ticks for the sliding window when panning, as they are the only ones visible.

The other part limits the number of ticks on a continuous axis such that the ticks are never closer than the closest two datums.

Unzoomed ~115fps
Panning/zooming before ~30fps and stuttering
Panning/zooming after ~115fps no stuttering

### DIFF
--- a/packages/ag-charts-community/src/chart/axis/axis.ts
+++ b/packages/ag-charts-community/src/chart/axis/axis.ts
@@ -246,6 +246,8 @@ export abstract class Axis<S extends Scale<D, number, TickInterval<S>> = Scale<a
 
     private destroyFns: Function[] = [];
 
+    private minRect?: BBox;
+
     constructor(
         protected readonly moduleCtx: ModuleContext,
         readonly scale: S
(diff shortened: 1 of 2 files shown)


packages/ag-charts-community/src/chart/axis/axis.ts
```diff
@@ -275,15 +277,21 @@ export abstract class Axis<S extends Scale<D, number, TickInterval<S>> = Scale<a
         this.assignCrossLineArrayConstructor(this._crossLines);
 
         let previousSize: { width: number; height: number } | undefined = undefined;
-        this.destroyFns = [
+        this.destroyFns.push(
```
lsjroberts, line 280: There's a hover listener on line 260 that was not being destroyed as it was being dumped by this. Probably wasn't causing any perf issues though.

alantreadway (approved):
LGTM

PR #604, opened 2023-11-23
AG-8555 Hide ticks in the range bleed outside the series rect

lsjroberts:

### DIFF
--- a/packages/ag-charts-community/src/chart/axis/axis.ts
+++ b/packages/ag-charts-community/src/chart/axis/axis.ts
@@ -1129,7 +1129,7 @@ export abstract class Axis<S extends Scale<D, number, TickInterval<S>> = Scale<a
             this.defaultTickMinSpacing,
             rangeWithBleed / ContinuousScale.defaultMaxTickCount
         );
-        const clampMaxTickCount = !isNaN(maxSpacing);
+        let clampMaxTickCount = !isNaN(maxSpacing);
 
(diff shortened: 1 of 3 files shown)


packages/ag-charts-community/src/chart/axis/axis.ts
```diff
@@ -1069,6 +1071,18 @@ export abstract class Axis<S extends Scale<D, number, TickInterval<S>> = Scale<a
...
+            if (seriesRect) {
+                if (this.direction === ChartAxisDirection.X && (translationY < 0 || translationY > seriesRect.width)) {
+                    continue;
+                }
+
+                if (this.direction === ChartAxisDirection.Y && (translationY < 0 || translationY > seriesRect.height)) {
+                    continue;
+                }
+            }
+
```
alantreadway, line 1085: Is this going to play well with animations?

We might want to defer this culling until the animation processing, somewhere around here?
https://github.com/ag-grid/ag-charts/blob/1066f9ada5083f31f3d82c5c448cdb21d154a5e6/packages/ag-charts-community/src/chart/axis/axisUtil.ts#L60-L63

lsjroberts: This all appears to work with animation, including adding nodes before and after and with ticking updates.

alantreadway (approved):
LGTM

PR #614, opened 2023-11-23
Use webp for examples

jacobp100:
2.6mb -> 770kb


alantreadway (approved):
Nice - caniuse.com seems to indicate support for `webp` is basically universal, so LGTM
https://caniuse.com/webp

taktran (approved):
Nice work ✨

PR #617, opened 2023-11-24
Gallery updates

jacobp100:
* Changes the gallery examples on the homepage to use the gallery example images rather than having Mark generate this in Photoshop
* Allows grouping of gallery sections without starting new lines (so Radar Line and Radar Area are in the same group)
* Adds an additional example for both treemap and sunburst


alantreadway (approved):
LGTM - nice improvements!

PR #619, opened 2023-11-24
AG-9588 Add validation for user theme

olegat:
This isn't an ideal solution, but it's an improvement to just naively converting user data to a theme with `as AgChartTheme` with no validation whatsoever.

In long term, we should find some way to integrate this with the _validatation.ts_ utility.

This will solve the bug where nothing is rendered, but it’s not an ideal solution. Unfortunately it seems like user themes are not integrated with our _validation.ts_ utility (which would be an ideal solution).

This PR implements some rudimentary validation to try to prevent unexpected behaviours like no rendering. However, as a consequence, the entire `options.theme` value is thrown away if an error is found. Migrating to the _validation.ts_ will enhance our ability to control the defaults when validation fails.

I think that the current state of the solution is acceptable; nevertheless, it is tech debt. I don’t believe that making further enhancements to this solution is a good idea.

### DIFF
--- a/packages/ag-charts-community/src/chart/mapping/themes.ts
+++ b/packages/ag-charts-community/src/chart/mapping/themes.ts
@@ -101,6 +101,11 @@ function validateChartTheme(value: unknown): string | ChartTheme | AgChartTheme 
 }
 
 export function getChartTheme(unvalidatedValue: unknown): ChartTheme {
+    // unvalidatedValue is either a built-in theme (`string | ChartTheme`) or a user defined
+    // theme (`AgChartTheme`). In the latter case, we can't make any assumption about the
+    // property types, hence why the input parameter is `unknown`. This abnormal validation
+    // is tech debt; the ideal solution would be to integrate user themes with the @Validate
+    // decorator like other chart options.
     let value = validateChartTheme(unvalidatedValue);
 
     if (value instanceof ChartTheme) {
--- a/packages/ag-charts-community/src/chart/mapping/themes.ts
+++ b/packages/ag-charts-community/src/chart/mapping/themes.ts
@@ -1,4 +1,9 @@
-import type { AgChartTheme, AgChartThemeName, AgChartThemeOverrides } from '../../options/agChartOptions';
+import type {
+    AgChartTheme,
+    AgChartThemeName,
+    AgChartThemeOverrides,
+    AgChartThemePalette,
+} from '../../options/agChartOptions';
 import { jsonMerge } from '../../util/json';
 import { Logger } from '../../util/logger';
 import { ChartTheme } from '../themes/chartTheme';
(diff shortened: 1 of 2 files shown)


packages/ag-charts-community/src/chart/mapping/themes.ts
```diff
@@ -56,8 +116,6 @@ export function getChartTheme(value?: string | ChartTheme | AgChartTheme): Chart
         return lightTheme();
     }
 
-    value = value as AgChartTheme;
```
olegat, line 59: Note: this expression is no longer necessary because the two if-statements above remove `ChartTheme | undefined | string` from the union, leaving `AgChartTheme` the only remaining type.

packages/ag-charts-community/src/chart/mapping/themes.ts
```diff
@@ -42,12 +47,72 @@ export const themes: ThemeMap = {
...
+    if (baseTheme !== undefined && typeof baseTheme !== 'string' && typeof baseTheme !== 'object') {
+        Logger.warn(`invalid theme.baseTheme type ${typeof baseTheme}, expected (string | object).`);
+        valid = false;
+    }
+    if (overrides !== undefined && typeof overrides !== 'object') {
+        Logger.warn(`invalid theme.overrides type ${typeof overrides}, expected object.`);
+        valid = false;
+    }
+    if (typeof palette === 'object') {
+        if (palette !== null) {
```
alantreadway, line 68: Nit: The `typeof palette === 'object'` check makes this guard redundant?

packages/ag-charts-community/src/chart/mapping/themes.ts
```diff
@@ -42,12 +47,72 @@ export const themes: ThemeMap = {
...
+type Unvalidate<T> = { [K in keyof T]?: unknown };
+function validateChartThemeObject(unknownObject: object | null): AgChartTheme | undefined {
+    if (unknownObject === null) {
+        return undefined;
+    }
+
+    let valid = true;
+    const { baseTheme, palette, overrides } = unknownObject as Unvalidate<AgChartTheme>;
+
+    if (baseTheme !== undefined && typeof baseTheme !== 'string' && typeof baseTheme !== 'object') {
```
alantreadway, line 59: We aren't handling the `typeof baseTheme === 'object'` case 100% correctly here, but I think on balance that is fine since strictly the recursive handling is an undocumented feature.

alantreadway (approved):
LGTM, please see if you can address the single nit before merge?

packages/ag-charts-community/src/chart/mapping/themes.ts
olegat: You'd think so, but my `tsserver` tells me that the type of `palette` is `object | null` after the first guard.

This might be okay, but I don't know how the `const { fills, strokes } = palette` deconstruction will behave if the palette is `null`.

alantreadway (approved):
LGTM

PR #623, opened 2023-11-24
Fix hydration issues

jacobp100:


packages/ag-charts-website/src/utils/hooks/useIsSsr.ts
```diff
@@ -0,0 +1,12 @@
+import { useEffect, useState } from 'react';
+
+// Manages SSRs and the hydration of a page
+export const useIsSsr = () => {
+    const [isSsr, setIsSsr] = useState(false);
```
taktran: Shouldn't this be `true`, since it should be SSR until client side is loaded? But I'm not sure if this will work in all cases, so the `window` check might be more reliable

```suggestion
    const [isSsr, setIsSsr] = useState(true);
```

jacobp100: 🤦‍♂️ Good catch

PR #626, opened 2023-11-24
update links

AG-DavidG:


packages/ag-charts-website/src/content/docs/licensing/index.mdoc
```diff
@@ -23,7 +23,7 @@ Please email [info@ag-grid.com](mailto:info@ag-grid.com) to receive a trial lice
 
 ### Installing AG Charts Enterprise
 
-The [Installation](installation) page contains instructions on how to import AG Charts Enterprise and set your license key.
+The [Installation](./installation/) page contains instructions on how to import AG Charts Enterprise and set your license key.
```
jacobp100, line 26:
```suggestion
The [Installation](/installation/) page contains instructions on how to import AG Charts Enterprise and set your license key.
```

alantreadway: I'm not sure which way is more correct, but searching the code base we seem to have many more uses of `./` than `/`, so I'm going to merge David's changes and verify in staging if it's working or not.

AG-DavidG: It's working on staging.
There seems to be differences between behaviour on local and staging.

PR #631, opened 2023-11-24
Fix example flicker on load

jacobp100:
URL would be `undefined` on load, and JS would helpfully turn this into a string to give a url that does not exist


alantreadway (approved):
LGTM

PR #636, opened 2023-11-27
Fix broken links

jacobp100:
Fixes broken links in multiple tickets

Adds a terminal-only warning for when internal links don't have a trailing slash so the development experience is closer to staging. I've fixed a few of these, but it's too risky to attempt to fix everything now. Running the website build will log all links that need updating. This link checking only works for astro and mdoc pages. React links will need to call out to the `validateUrl` function to get this validation

### DIFF
--- a/packages/ag-charts-website/src/features/docs/utils/urlPaths.ts
+++ b/packages/ag-charts-website/src/features/docs/utils/urlPaths.ts
@@ -10,7 +10,7 @@ export function getFrameworkFromPath(path: string) {
 }
 
 export const getExamplePageUrl = ({ framework, path }: { framework: Framework; path: string }) => {
-    return pathJoin(SITE_BASE_URL, framework, path);
+    return pathJoin(SITE_BASE_URL, framework, path) + '/';
 };
 
 /**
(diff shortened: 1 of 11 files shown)


alantreadway (approved):
LGTM

packages/ag-charts-website/src/features/docs/utils/urlPaths.ts
```diff
@@ -10,7 +10,7 @@ export function getFrameworkFromPath(path: string) {
 }
 
 export const getExamplePageUrl = ({ framework, path }: { framework: Framework; path: string }) => {
-    return pathJoin(SITE_BASE_URL, framework, path);
+    return pathJoin(SITE_BASE_URL, framework, path) + '/';
```
taktran, line 13: In future, we'll want to fix `pathJoin` to support this scenario like in https://github.com/ag-grid/ag-charts/pull/534/files#diff-3aaaae2a7d6e9da3ba436acbc17177367fbe44281536edff80c69d399ee2d99c

taktran (approved):
Good stuff 👍🏼 

As we mentioned, we'll want to add the trailing slash config to astro in future: https://github.com/ag-grid/ag-charts/pull/534

packages/ag-charts-website/src/features/docs/utils/urlPaths.ts
jacobp100: Most likely. I was worried about the blast radius

PR #639, opened 2023-11-27
CRT-36: Simplify tooltip position type in external options

iMoses:


alantreadway (approved):
LGTM if you can fix the linting break?

alantreadway (approved):

PR #641, opened 2023-11-28
Getting started guides

jacobp100:


packages/ag-charts-website/src/features/example-runner/components/RemoveLogoOnInitScript.astro
```diff
@@ -23,14 +23,17 @@ const loadingIFrameId = getLoadingIFrameId({ pageName, exampleName });
...
         if (data?.type === 'init') {
-            document.getElementById(loadingLogoId)?.remove();
-            const iframe = document.getElementById(loadingIFrameId);
-            if (!iframe) return;
-
-            iframe.style.visibility = 'visible';
-            if (document.documentElement.dataset['darkMode'] === 'true') {
-                iframe.contentDocument.documentElement.dataset.darkMode = true;
-            }
+            // FIXME - some pages have the same example on the page twice
```
taktran, line 26: 👍🏼 Might be able to get away with a class, but can check later

taktran (approved):
I have a feeling that `const chart` will break something, but I don't know what 🤷🏼‍♂️ I can't think of any instances of multiple examples causing issues, but it has caused issues in other places. Can try it and see, and revert if necessary.

PR #646, opened 2023-11-28
CRT-87 Fix "Water Usage" doughnut gallery example

olegat:
https://ag-grid.atlassian.net/browse/CRT-87

This is the test result when the fix in _sector.ts_ is removed:
pie-series-test-ts-doughnut-multiple-1-snap-diff


alantreadway (approved):
LGTM

PR #648, opened 2023-11-28
Remove use strict from modules

jacobp100:
Removes `'use strict'` from examples. Any files with imports and exports implicitly has this anyway


alantreadway (approved):
LGTM

PR #650, opened 2023-11-28
CRT-84 Fix Vue Crosshairs example

olegat:
The '?' operator causes problem with the exampleRunner generator for the Vue framework. So use the '!' operator instead to tell TypeScript to assume that the types on non-nullable.


alantreadway (approved):
LGTM

PR #662, opened 2023-11-29
Fix plunkr links in Safari for local dev

jacobp100:
The first time you run `nx dev`, you'll need to enter your sudo password so the certs get installed and set up

You'll have to use http**s**://localhost:4600 too

There's no out-of-the-box way to run on mobile yet - but if you edit your `.env.dev` to replace localhost with your network IP (hold option and press the WiFi icon in the toolbar), it will load. You'll have to manually trust the certificate with [these docs](https://www.npmjs.com/package/vite-plugin-mkcert#mobile-devices)


taktran (approved):
Noice 👍🏼 Got it working on my machine too

alantreadway (approved):
LGTM

PR #665, opened 2023-11-29
AG-9954 Animate markers using inverseEaseOut on line/area series

olegat:
https://ag-grid.atlassian.net/browse/AG-9954


alantreadway (approved):
LGTM - you probably need to update snapshots by the looks of it?

olegat:
> LGTM - you probably need to update snapshots by the looks of it?

Yes, I've just updated them. I've also checked the range-area animation in this Plunker example: https://plnkr.co/edit/W8mNeZuat5ll4JF7?open=main.ts. It looks correct.

PR #666, opened 2023-11-29
AG-9956: Improved scrollspy

iMoses:


packages/ag-charts-website/src/components/pages-navigation/useScrollSpy.ts
```diff
@@ -0,0 +1,20 @@
...
+    const location = useLocation();
+
+    function handleScrollSpy(slug: string) {
+        if (menuRef.current == null) return;
+        for (const navItem of menuRef.current.querySelectorAll('a')) {
+            navItem.classList.toggle('active', navItem.getAttribute('href') === `#${slug}`);
+        }
+    }
+
+    useEffect(() => scrollspy(headings, handleScrollSpy, { offset: 120 }), [location?.hash, headings]);
```
taktran: Is the `120` the header or something? Should make it a shared variable somewhere.

packages/ag-charts-website/src/utils/scrollspy.ts
```diff
@@ -0,0 +1,42 @@
...
+        const { scrollTop, scrollHeight, clientHeight } = container ?? document.documentElement;
+
+        let selectedSlug: string | undefined;
+        let lastOffsetTop: number | undefined;
+
+        for (const [slug, node] of entries) {
+            if (node.offsetTop > scrollTop + (offset ?? 0) && scrollTop + clientHeight < scrollHeight) {
+                break;
+            }
+            // duplicate node's offsetTop indicates a multiple tab headings
```
taktran:
```suggestion
            // duplicate node's offsetTop indicates multiple tab headings
```

taktran (approved):
Works very nicely ✨ Great job 👍🏼 

Just some minor comments.

packages/ag-charts-website/src/components/pages-navigation/useScrollSpy.ts
iMoses: it's the header + buffer, because we want the items to activate when nearing the top, not necessarily when exactly at the top. I'm not sure if it should be shared though as different pages may wanna set different offsets, though I can make it configurable using the hook.

PR #667, opened 2023-11-30
Use OKLCH for gradients

jacobp100:
# Before

Screenshot 2023-11-30 at 09 54 14

# After

Screenshot 2023-11-30 at 09 54 12


alantreadway (approved):
LGTM

PR #669, opened 2023-11-30
AG-9976: [Docs] Use specific type alias for degrees instead of `number`

iMoses:
Seems that we're using degrees everywhere in our API (no radians)


packages/ag-charts-community/src/options/chart/types.ts
```diff
@@ -31,7 +31,10 @@ export type PixelSize = number;
 export type Ratio = number;
 
 /** Alias to denote that a value is a duration in milliseconds */
-export type MilliSeconds = number;
+export type Duration = number;
```
alantreadway: Nit: Could we name this `DurationMs` so it's clearer what the unit of time is?

alantreadway (approved):
LGTM other than one nit.

PR #668, opened 2023-11-30
Ensure sector highlights show inside the unhighlighted bounds

jacobp100:
Screenshot 2023-11-30 at 11 02 38
Screenshot 2023-11-30 at 11 02 22
Screenshot 2023-11-30 at 11 02 08
Screenshot 2023-11-30 at 11 01 55

PR #670, opened 2023-11-30
Enable a default stroke width iff strokes are present

jacobp100:
Off the back of https://github.com/ag-grid/ag-charts/pull/668

Allows `undefined` stroke, which will implicitly set `strokeWidth` to zero for the purposes of sector spacing

Currently, because the default style doesn't have a stroke, the stroke width is set to zero. This means when a user wants to add a stroke, they also need to update the stroke width

Alternatively, we could set the default stroke to 'transparent' - although there will be more overhead doing that, since there will almost definitely be some computation involved there

Needed animation updates to support this - might be a bridge too far - but this functionality might also be useful for other things?


alantreadway (approved):
LGTM

alantreadway:
@jacobp100 could you resolve the conflicts?

alantreadway (approved):
LGTM

PR #671, opened 2023-11-30
AG-9428 Remove listener handler return types

olegat:
https://ag-grid.atlassian.net/browse/AG-9428

This was only required for the series classes to get values from the 'error-bars' series-option module. However, the series classes also get values through the `SeriesOptionInstance` interface (pickNode).

This replaces the following events with SeriesOptionInstance methods:
- `SeriesDataPrerequestEvent`:   getPropertyDefinitions
- `SeriesDataGetDomainEvent`:    getDomain
- `SeriesTooltipGetParamsEvent`: getTooltipParams

This makes _listeners.ts_ and _observable.ts_ more similar to one another, which will therefore make the migration easier.


alantreadway (approved):
LGTM

PR #677, opened 2023-12-01
Allow innerRadius and outerRadius to be flipped for a sector shape

jacobp100:
https://ag-grid.atlassian.net/browse/AG-9883


alantreadway (approved):
LGTM

PR #678, opened 2023-12-01
AG-10007 - Upgrade website to Astro 3

taktran:
## Changes

* Upgrade to Astro 3.x
* Uppercase http endpoints - [notes](https://docs.astro.build/en/guides/upgrade-to/v3/#changed-http-request-methods-case)
* Update `svgr` and how SVGs are imported
* Use plain astro commands instead of `@nxextensions/astro` - as it was stalling on some commands
  * Add `dotenv-expand` to use `PORT` env var files
* Patches
  * Remove Astro 2.8.0 patch for angular `.mjs` files being deleted, since it wasn't being deleted anymore
  * Remove patch to deal with Buffer chunk
  * Add patch to use working directory when building static files, see https://github.com/withastro/astro/pull/9378

## Blockers

As at 2023-12-07:

* [x] Dev server (`nx dev`) works as expected, however on build (`nx preview ag-charts-website`), the markdoc styles of `Heading.astro` and custom tags don't come through. Potentially related to https://github.com/withastro/astro/issues/9347, however not able to replicate the markdoc styles issue on a clean Astro 3.x install. Needs more investigation
  * To see the issue, run preview and look at http://localhost:4601/javascript/licensing/, where the heading link icon and enterprise icon don't have styles
  * This is resolved with a patch from @jacobp100 : https://github.com/withastro/astro/pull/9378


packages/ag-charts-website/project.json
```diff
@@ -7,7 +7,7 @@
     "build": {
       "outputs": ["{workspaceRoot}/dist/{projectRoot}"],
       "dependsOn": ["^build", "^docs-resolved-interfaces", "generate-examples"],
-      "executor": "@nxtensions/astro:build",
+      "command": "cd packages/ag-charts-website && astro build",
```
jacobp100: Do we need to change this?

patches/astro+3.6.4.patch
```diff
@@ -0,0 +1,26 @@
...
+diff --git a/node_modules/astro/dist/runtime/server/render/common.js b/node_modules/astro/dist/runtime/server/render/common.js
+index 90534bd..e4da3b5 100644
+--- a/node_modules/astro/dist/runtime/server/render/common.js
++++ b/node_modules/astro/dist/runtime/server/render/common.js
+@@ -39,6 +39,8 @@ function stringifyChunk(result, chunk) {
+         }
+         return renderAllHeadContent(result);
+       }
++      case "Buffer":
++        return chunk.toString();
```
jacobp100: Should this be here?

jacobp100 (approved):

packages/ag-charts-website/project.json
taktran: So when I tried it with the normal nx commands, it would stall on `nx dev`. I haven't looked too much into it yet, but thought I'd do that after this gets merged.

patches/astro+3.6.4.patch
taktran: This needs to be reviewed again too with @alantreadway , don't know all the reasons for this patch on astro 2.8.0, but keeping it for now.

packages/ag-charts-website/project.json
alantreadway: FYI: There is an option to specify the CWD to use, that might be a slightly better way to go?

E.g: https://github.com/ag-grid/ag-charts/blob/latest/packages/ag-charts-community/project.json#L148

patches/astro+3.6.4.patch
alantreadway: Maybe we should test if examples/gallery cases work without this? (Also thumbnails)?

packages/ag-charts-website/project.json
taktran: Ah I did try this before, but for some reason `--host` doesn't get passed through, so you don't get `https` in the server. Not sure how this would relate to `cwd` 🤷🏼‍♂️ 

I tried:

```
"cwd": "packages/ag-charts-website",
"command": "astro dev --port=${PORT} --host",
```

and

```
"executor": "nx:run-commands",
"options": {
  "command": "astro dev"
},
"args": "--port=${PORT} --host",
```

and also putting `--host` before `--port` (for both above), but none of it works. Any ideas @alantreadway ?

taktran: I guess the issue with `--host` doesn't apply to `build` and `preview`, so can add it for those

taktran: 🤦🏼‍♂️ nvm, it works. It needs to be `options.cwd` as per the example @alantreadway posed

patches/astro+3.6.4.patch
taktran: Gallery seems to be working fine on dev/build, so have removed this.

PR #679, opened 2023-12-01
Scale perf

jacobp100:


packages/ag-charts-community/src/scale/invalidating.ts
alantreadway, line 1: There is some overlap with `ChangeDetectable` and `@SceneChangeDetection()` - perhaps we could merge these concepts?

jacobp100: I did look - but scene change detection is pretty wired into the scene graph, and even has some codegen just for that use-case too

alantreadway: Alright - I think it might be worth generalising this case further down the line (the scene-graph handling is a more specific use-case of invalidation) - but let's roll with this for now.

alantreadway (approved):
LGTM

PR #680, opened 2023-12-01
AG-10004 - Tactical improvement of website build thumbnail generation speed.

alantreadway:
The changes in this PR are based upon observing that the Chart rendering time is the largest time component of thumbnail rendering (the split was about 20% example generation, 70% rendering, 10% compression to PNG/WEBP).

Consequently attempts to improve performance of thumbnail generation (primarily during a CI build) by:
- Cache Chart instances between thumbnail generations, then...
- Adjusting generation order so that `format` changes happen sequentially, meaning that there are no option contract changes and the cached chart instance performs a `no-op`.
- Keeping the existing behaviour of rendering theme-permutations sequentially, meaning only the theme deltas are applied to the cached chart instance.

Locally this took `nx build ag-charts-website --prod` from over 7m to 4m06s.


packages/ag-charts-website/src/utils/generateExample.ts
```diff
@@ -39,9 +40,57 @@ export async function generateExample({ exampleName, theme, format }: Params) {
...
+        console.log('Waiting for pool to become available...');
+        await new Promise((resolve) => setTimeout(resolve, (10 * count) ^ 2));
+        count++;
+    }
+
+    if (pool.length === 0) {
+        throw new Error('No JSDOM instance available to borrow.');
+    }
+
+    return pool.splice(0, 1)[0];
```
jacobp100: Is this not the same as:-

```suggestion
    return pool.shift();
```

packages/ag-charts-website/src/utils/generateExample.ts
```diff
@@ -39,9 +40,57 @@ export async function generateExample({ exampleName, theme, format }: Params) {
...
+    }
+}
+initPool();
+
+async function borrowFromPool() {
+    let count = 0;
+    while (pool.length === 0 && count < 5) {
+        // eslint-disable-next-line no-console
+        console.log('Waiting for pool to become available...');
+        await new Promise((resolve) => setTimeout(resolve, (10 * count) ^ 2));
```
jacobp100:
```suggestion
        await new Promise((resolve) => setTimeout(resolve, (10 * count) ** 2));
```

taktran (approved):
Nice one, LGTM 👍🏼

PR #681, opened 2023-12-01
AG-8641 Calculate how many decimal points to use when `binCount` is specified

olegat:
https://ag-grid.atlassian.net/browse/AG-8641


alantreadway (approved):
LGTM - please ping back to me once ready to merge?

PR #682, opened 2023-12-01
Generate each example image once in production build

jacobp100:


packages/ag-charts-website/src/pages/gallery/examples/[exampleName]/[theme]-plain.[format].ts
```diff
@@ -13,18 +15,61 @@ interface Params {
...
+                const uncompressed = await generateExample({ exampleName, theme });
+                return uncompressed;
+            })
+        );
+        const subPages = chunk.flatMap(({ params }, index): Page[] => {
+            const { exampleName, theme } = params;
+            const uncompressed = examples[index];
+            return formats.map((format) => ({
+                params: { exampleName, theme, format },
+                props: { uncompressed },
```
Comment (alantreadway) [inline packages/ag-charts-website/src/pages/gallery/examples/[exampleName]/[theme]-plain.[format].ts:47]: I'm a bit concerned with the memory overhead this might imply - can we come back to this once I've had time to dig into a Nx-based task decomposition for example generation, which may include thumbnail generation coming from a completely different approach?

PR #683, opened 2023-12-01
AG-9914 Make userPalette fills and strokes optional

olegat:
https://ag-grid.atlassian.net/browse/AG-9914

If fills/strokes are undefined, then default to fills/strokes of the base theme.


alantreadway (approved):
LGTM so far, ping back to me once ready to merge?

alantreadway (approved):

PR #686, opened 2023-12-04
Fix angular build

jacobp100:
Screenshot 2023-12-04 at 16 16 26


StephenCooper:
This is not the correct fix. The wrong level of the package has been published in v9.0.0.

alantreadway (approved):
LGTM

alantreadway:
> This is not the correct fix. The wrong level of the package has been published in v9.0.0.

PR updated in this respect.

PR #687, opened 2023-12-04
AG-8777 Use theme default axis grid color

olegat:
https://ag-grid.atlassian.net/browse/AG-8777


packages/ag-charts-community/src/chart/mapping/prepare.ts
```diff
@@ -350,6 +354,19 @@ function prepareAxis<T extends AxesOptionsTypes>(
...
+        const { style } = axis.gridLine;
+        if (Array.isArray(style)) {
+            axis.gridLine.style = style.map((value) => {
+                const { stroke, lineDash } = value;
+                if (stroke === undefined && lineDash !== undefined) {
+                    return { stroke: defaultGridStroke, lineDash };
+                }
+                return value;
+            });
+        }
```
alantreadway: Would it be possible to take a similar approach to `crossLines` above (i.e. JSON merge of array elements with theme options), rather than treating `gridLine.style` as an exception that is passed in as a parameter?

olegat: Done. There are some nuanced differences between `crossLines` and `gridLine`, which I've explained in code comments.

packages/ag-charts-community/src/chart/mapping/prepare.ts
```diff
@@ -345,9 +347,27 @@ function prepareAxis<T extends AxesOptionsTypes>(
...
+        axis.gridLine.style = axis.gridLine.style.map((userStyle, index) => {
+            // An empty gridLine (e.g. `gridLine: { style: [ {} ] }`) means "draw nothing". So ignore theme
+            // defaults if this is the case:
+            if (userStyle.stroke === undefined && userStyle.lineDash === undefined) {
+                return userStyle;
+            }
+            // Themes will normally only have one element in gridLineStyle[], but cycle through the array
+            // with `mod` anyway to make sure that we honour the theme's grid line style sequence.
+            const themeStyle: typeof userStyle = gridLineStyle[mod(index, gridLineStyle.length)];
```
alantreadway: Nit: What is `mod()`? It's more conventional to use the `%` operator in JS?

alantreadway: (Just looked, `mod()` is overkill in this scenario where we are guaranteed that the values are positive integers)

olegat: Done.

alantreadway (approved):
LGTM

PR #696, opened 2023-12-05
AG-9963 Mute duplicate warnings when axis position is valid.

olegat:
https://ag-grid.atlassian.net/browse/AG-9963

The bug here that the position always defaults to 'left'. This that if the axes configuration something like 0: 'left' and 1: 'bottomasdf' then this will default to 0: 'left' and 1: 'left'.

This is what the causes the duplicate warning, because the library cannot determine which is the X axis.

To get around this, we'll change the default based on any valid positions that have been identified.

### DIFF
--- a/packages/ag-charts-community/src/chart/axis/axisUtil.ts
+++ b/packages/ag-charts-community/src/chart/axis/axisUtil.ts
@@ -1,8 +1,15 @@
 import { FROM_TO_MIXINS } from '../../motion/fromToMotion';
 import type { FromToMotionPropFn, NodeUpdateState } from '../../motion/fromToMotion';
+import type {
+    AgBaseAxisOptions,
+    AgBaseCartesianAxisOptions,
+    AgCartesianAxisPosition,
+    AgCartesianAxisType,
+} from '../../options/agChartOptions';
 import type { Group } from '../../scene/group';
 import type { Line } from '../../scene/shape/line';
 import type { Text } from '../../scene/shape/text';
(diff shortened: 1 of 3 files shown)


packages/ag-charts-community/src/chart/axis/axisUtil.ts
```diff
@@ -212,3 +219,61 @@ export function resetAxisLineSelectionFn() {
...
+}
+
+function isAxisPosition(position: unknown): position is AgCartesianAxisPosition {
+    const allowedPositions: string[] = CARTESIAN_AXIS_POSITIONS;
+    return typeof position === 'string' && allowedPositions.includes(position);
+}
+
+// If axis[].position, then we cannot always default to the same value. We need
+// to default to an 'untaken' position (see AG-9963 for more info).
+export class AxisPositionGuesser {
```
alantreadway: Hmmm... `axisUtil.ts` is really related to the axis implementation - I would keep this 'validation'-esque behaviour somewhere closer to the mapping/validation phase of processing. perhaps in `charts/mapping/prepareAxis.ts` similar to `prepareSeries.ts`?

olegat: No objections. I'll move these functions.

olegat: Done.

alantreadway (approved):
LGTM

PR #698, opened 2023-12-05
AG-9879 Add warning for missing pie chart data

olegat:
https://ag-grid.atlassian.net/browse/AG-9879


packages/ag-charts-community/src/chart/series/polar/pieSeries.test.ts
```diff
@@ -117,3 +107,42 @@ describe('Doughnut', () => {
         await compare();
     });
 });
+
+/* eslint-disable no-console */
+describe('Validation', () => {
```
alantreadway: Nit: We should aim for one root `describe()` block for a module, with a name describing the module under test (either `pieSeries.ts` (module filename) or `PieSeries` (main module export) - as test reports for the repo will merge this with other `Validation` describe blocks which are also declared at the top level.

packages/ag-charts-community/src/chart/series/polar/pieSeries.ts
```diff
@@ -413,6 +414,16 @@ export class PieSeries extends PolarSeries<PieNodeDatum, Sector> {
             ],
         });
 
+        // AG-9879 Warning about missing data.
+        for (const valueDef of this.processedData?.defs?.values ?? []) {
+            // The 'angleRaw' is an undocumented property for the internal implementation, so ignore this.
+            // If any 'angleRaw' values are missing, then we'll also be missing 'angleValue' values and
+            // will log a warning anyway.
+            if (valueDef.id !== 'angleRaw' && valueDef.missing !== undefined && valueDef.missing > 0) {
+                Logger.warn(`Missing ${valueDef.missing} ${String(valueDef.property)} value(s)`);
```
alantreadway: Nit: Use `Logger.warnOnce()` to avoid spamming the console.

packages/ag-charts-community/src/chart/series/polar/pieSeries.test.ts
olegat: Okay, noted. The blocks have different `beginEach` / `afterEach` handlers, but there's no reason that the two cannot be placed under a common `describe()` parent. I'll do that.

packages/ag-charts-community/src/chart/series/polar/pieSeries.ts
olegat: Done.

alantreadway (approved):
LGTM

PR #700, opened 2023-12-06
AG-9492 - Add caption tooltip when captions are truncated

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-9492

### DIFF
--- a/packages/ag-charts-community/src/chart/agChartV2.ts
+++ b/packages/ag-charts-community/src/chart/agChartV2.ts
@@ -659,15 +659,16 @@ function applySeriesValues(
     const skip: string[] = ['series[].listeners', 'series[].seriesGrouping'];
     const jsonApplyOptions = getJsonApplyOptions(moduleContext);
     const ctrs = jsonApplyOptions.constructors ?? {};
+    // Allow context to be injected and meet the type requirements
+    class PieTitleWithContext extends PieTitle {
+        constructor() {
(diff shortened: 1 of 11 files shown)


packages/ag-charts-community/src/chart/agChartV2.ts
```diff
@@ -1,3 +1,4 @@
+import type { ModuleContext } from '../module-support';
```
alantreadway: Nit: Could you use a more-specific import, rather than this large bundle? (e.g. from `./module/moduleContext` directly)?

(inside the same conceptual build unit, it's better to avoid bundles from a potential circular dependency perspective).

packages/ag-charts-community/src/chart/agChartV2.ts
```diff
@@ -390,7 +391,8 @@ function applyChartOptions(chart: Chart, processedOptions: ProcessedOptions, use
         registerListeners(chart, processedOptions.listeners);
     }
 
-    applyOptionValues(chart, processedOptions, { skip });
+    const moduleContext = chart.getModuleContext();
```
alantreadway: Nit: Maybe just inline this and avoid a variable declaration?

alantreadway (commented):
Some minor nits, otherwise LGTM

alantreadway (approved):
LGTM

PR #703, opened 2023-12-06
AG-9184 Update data model & controller to basic support data per series

lsjroberts:


packages/ag-charts-community/src/chart/series/cartesian/__image_snapshots__/bar-series-test-ts-bar-series-create-for-stacked-column-category-data-per-series-clashing-it-should-render-to-canvas-as-expected-1-snap.png
lsjroberts, line 1: This is the current failure case that will need to be fixed later

packages/ag-charts-community/src/chart/data/dataController.ts
```diff
@@ -97,6 +107,31 @@ export class DataController {
         }
     }
 
+    private extractScopedData(id: string, processedData: UngroupedData<any>) {
```
alantreadway, line 113: If there are not multiple data-sources in use, it would be good to short-circuit this processing, to avoid adding more overhead to the 'normal' case.

(Additionally if clients are concerned about performance, we can offer them the option of using a single data-source in order to optimise our processing)

packages/ag-charts-community/src/chart/data/dataModel.ts
```diff
@@ -612,27 +619,52 @@ export class DataModel<
...
+            }
+
+            for (const [valueDefIdx, def] of valueDefs.entries()) {
+                for (const scope of def.scopes ?? scopes) {
+                    const source = sourcesById[scope];
+                    const valueDatum = source?.data[datumIdx] ?? datum;
+
+                    value = processValue(def, valueDatum, value);
+
+                    if (value !== INVALID_VALUE && values) {
```
alantreadway: Nit: Outdent this block by changing this to `if (value === INVALID_VALUE || !values) continue;` ?

packages/ag-charts-community/src/chart/data/dataModel.ts
```diff
@@ -612,27 +619,52 @@ export class DataModel<
...
 
             const values = dataVisible && valueDefs.length > 0 ? new Array(valueDefs.length) : undefined;
-            let valueIdx = 0;
             let value;
-            for (const def of valueDefs) {
-                value = processValue(def, datum, value);
+
+            const sourcesById: { [key: string]: { id: string; data: D[] } } = {};
+            if (sources) {
+                for (const source of sources) {
```
alantreadway: Nit: Maybe get rid of the outer `if` here and use `const source of sources ?? []`?

alantreadway (commented):
Looks great so far - would be good to verify that for single `data` cases the performance of processing is unaffected?

lsjroberts:
> would be good to verify that for single `data` cases the performance of processing is unaffected?

Appears to be the same. With the short-circuit added there's really no extra processing going on, since the new loops will be just the one iteration.

alantreadway (approved):
Nice work, LGTM!

PR #706, opened 2023-12-07
Use Angular packages directly

StephenCooper:
On the current version and with the switch to standalone we can use the Angular compiler directly. This is a very nice thing to be able to get rid off as it is a totally external lib that we are depending on currently. 

I have made the same change to Grid Examples which previously required this for cell renderers. As charts is not even using custom Angular components it probably never needed these anyway.


alantreadway (approved):
LGTM

PR #707, opened 2023-12-07
AG-10003 - Split example gen out from astro build.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-10003

First step on trying to reduce the cost of `nx build ag-charts-website` - move example generation into Nx tasks.

Summary of changes:
- Move example generator into a Nx executor.
- Register `ag-charts-build-tools` as an Nx extension.
  - The extension will auto-register `_example/<name>/main.ts` files as dynamic Nx projects with a `generate-example` task, and add a dependency from `ag-charts-website:generate-examples`.
  - The `generate-example` task uses the new Nx executor.
- Updates Astro example setup to read generated outputs.
- Adjusts `nx dev` to watch example changes and trigger example-specific rebuild.

Bonus:
- Attempts to improve the level of parallel task execution in Nx builds.


packages/ag-charts-website/project.json
```diff
@@ -80,6 +80,17 @@
           "fix": true
         }
       }
+    },
+    "test-example-gen": {
```
taktran: Did you intend to keep this test in here?

packages/ag-charts-website/src/features/docs/utils/filesData.ts
```diff
@@ -1,30 +1,35 @@
...
     const pageFolders = page.slug.split('/');
     const pageName = pageFolders[pageFolders.length - 1];
     return pageName && !pageName.startsWith('_');
 }
 
-export const getExamplesPathUrl = ({ pageName }: { pageName: string }) => {
+export const getExamplesPath = ({ pageName }: { pageName: string }) => {
     const contentRoot = getContentRootFileUrl();
-    const sourceExamplesPath = pathJoin(contentRoot.pathname, 'docs', pageName, '_examples');
+    const sourceExamplesPath = path.join(contentRoot.pathname, 'docs', pageName, '_examples');
```
taktran, line 23: Was there a reason you needed `path.join` here? It's fine for this file since it'll be run in node, but I created `pathJoin` so that it can be used on client side code as well node. I'm also in the process of updating how we use `pathJoin`, so wondering if there is something missing in it.

packages/ag-charts-website/src/utils/pages.ts
```diff
@@ -90,6 +90,11 @@ export const getContentRootFileUrl = ({ isDev }: { isDev?: boolean } = {}): URL
     return new URL(contentRoot, import.meta.url);
 };
 
+export const getExampleRootFileUrl = ({ isDev = getIsDev() }: { isDev?: boolean } = {}): URL => {
+    const root = isDev ? '../../../..' : '../../../../..';
```
taktran: There's a method above for this

```suggestion
    const root = getRootUrl().pathname;
```

patches/astro+2.8.0.patch
```diff
@@ -11,3 +11,17 @@ index c80c10d..abf3411 100644
...
+diff --git a/node_modules/astro/dist/runtime/server/render/common.js b/node_modules/astro/dist/runtime/server/render/common.js
+index ee22ac3..87087e8 100644
+--- a/node_modules/astro/dist/runtime/server/render/common.js
++++ b/node_modules/astro/dist/runtime/server/render/common.js
+@@ -38,7 +38,9 @@ function stringifyChunk(result, chunk) {
+         }
+         return renderAllHeadContent(result);
+       }
++      case "Buffer":
++        return chunk.toString();
```
taktran, line 23: 😅 Was going to ask how you fixed this one

plugins/ag-charts-build-tools/src/executors/generate-example-files/generator/examplesGenerator.ts
```diff
@@ -1,11 +1,11 @@
...
-import fs from 'node:fs/promises';
+import fs from 'fs';
+import path from 'path';
 
-import { getIsDev } from '../../utils/env';
+import { readFile } from '../../../executors-utils';
 import { ANGULAR_GENERATED_MAIN_FILE_NAME, SOURCE_ENTRY_FILE_NAME } from './constants';
 import chartVanillaSrcParser from './transformation-scripts/chart-vanilla-src-parser';
-import type { GeneratedContents } from './types.d';
+import type { InternalFramework } from './types';
```
taktran: Can combine this

```suggestion
import type { InternalFramework, GeneratedContents } from './types';
```

taktran (approved):
Epic work 🚀 Great stuff 👍🏼 Just minor things and a q about the use of `path.join`

packages/ag-charts-website/project.json
alantreadway: Removed - thanks!

packages/ag-charts-website/src/features/docs/utils/filesData.ts
alantreadway: I was err'ing on the side of simplifying things as I went, it didn't seem necessary for the extra overhead of potentially handling URLs when these paths are always on the local FS - I don't mind reverting if you have a preference that way though?

patches/astro+2.8.0.patch
alantreadway: Not 100% sure about this, but it seemed to fix the issue - I manually inspected the Buffer contents and they were a HTML snippet (from an example) when it failed; I haven't completed a test-pass with `nx preview` to verify the final result of this yet.

packages/ag-charts-website/src/features/docs/utils/filesData.ts
taktran: Don't mind it using the node version for this specific file. I do want to figure out a better way to distinguish between which utils can only be used on node/build envs, but that's a task for another time.

patches/astro+2.8.0.patch
alantreadway: Ok, everything looks good with `nx preview ag-charts-website --prod`, so once CI is green I'm going for merge.

PR #714, opened 2023-12-08
AG-10004 - Split thumbnail generation into Nx tasks.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-10004

Moves thumbnail generation out to Nx tasks, similarly to example generation.

Summary of changes:
- Adds `plugins/ag-charts-task-autogen` Nx plugin to be responsible for adding dynamic tasks to Nx projects.
  - This simplifies initial bootstrap as it has no dependencies on any build artefacts (unlike the thumbnail generation executor which depends on `ag-charts-community`).
- Breaks some basic example loading + canvas setup into a new `ag-charts-test` shared library.
- Adds thumbnail generator Nx executor.
- Automatically adds Nx thumbnail generation tasks for all gallery example, except the large-dataset example which is failing (but equally is not visible in the gallery).
- Removes all thumbnail generation code from Astro.

TODO:
- [x] Improve batched e2e time (it is much more efficient, but due to being  single-threaded seems to take longer e2e than the default 3x parallelisation Nx uses for non-batched mode).
- [x] Add `nx dev` watch support for thumbnails.
- [x] Add a file to trigger thumbnail cache invalidation.


plugins/ag-charts-task-autogen/src/index.ts
```diff
@@ -0,0 +1,74 @@
...
+
+import * as generateChartThumbnails from './generate-chart-thumbnails';
+import * as generateExampleFiles from './generate-example-files';
+
+const PROJECTS = ['ag-charts-website'];
+const NON_UNIQUE_PATH_ELEMENTS = ['src', 'content', 'docs', '_examples'];
+const NON_UNIQUE_PATH_ELEMENTS_MAP = NON_UNIQUE_PATH_ELEMENTS.reduce((r, n) => {
+    r[n] = true;
+    return r;
+}, {});
```
jacobp100: Minor

```suggestion
const NON_UNIQUE_PATH_ELEMENTS = new Set(['src', 'content', 'docs', '_examples']);
```

plugins/ag-charts-build-tools/src/executors/generate-chart-thumbnail/executor.ts
```diff
@@ -0,0 +1,56 @@
...
+    for (const theme of Object.keys(THEMES) as AgChartThemeName[]) {
+        for (const format of ['png' as const, 'webp' as const]) {
+            const result = await generateExample({
+                name,
+                example: generatedExample,
+                format,
+                theme,
+            });
+
+            writeFile(path.join(options.outputPath, `${theme}.${format}`), result);
```
jacobp100: If we moved the `writeFile` part to `generateExample`, then re-use the canvas output for the different formats

plugins/ag-charts-build-tools/src/executors/generate-chart-thumbnail/generator/environment.ts
```diff
@@ -0,0 +1,61 @@
...
+    console.log(`Creating thumbnail pool of size ${THUMBNAIL_POOL_SIZE}`);
+    for (let i = 0; i < THUMBNAIL_POOL_SIZE; i++) {
+        pool.push(buildPoolEntry());
+    }
+}
+initPool();
+
+export async function borrowFromPool() {
+    let count = 0;
+    while (pool.length === 0 && count < 5) {
```
taktran: Should the `5` here be `THUMBNAIL_POOL_SIZE`?

```suggestion
    while (pool.length === 0 && count < THUMBNAIL_POOL_SIZE) {
```

taktran (approved):
LGTM, makes website a lot cleaner 👍🏼 Just a minor comment.

plugins/ag-charts-build-tools/src/executors/generate-chart-thumbnail/executor.ts
alantreadway: Sounds reasonable - let's do this as a follow-up as the change is quite unwieldy and is going to conflict with the Astro 3 change.

jacobp100 (approved):

PR #715, opened 2023-12-08
Fixes for gradient legend axis

jacobp100:
https://ag-grid.atlassian.net/browse/AG-9450


packages/ag-charts-community/src/scale/logScale.ts
```diff
@@ -103,11 +103,12 @@ export class LogScale extends ContinuousScale<number> {
     }
 
     ticks() {
+        this.refresh();
```
alantreadway: Is this ordering change consistent with the other scales? (Looks like the others were consistently doing this after the `this.domain` check before)?

jacobp100: Updated to remove this change

alantreadway (approved):
LGTM

PR #719, opened 2023-12-11
Firefox bugs

olegat:
The recommended approach is to include styles in <head> and then scripts after that. See https://bugzilla.mozilla.org/show_bug.cgi?id=1404468#c16


taktran (approved):
I'm good with this, if the dark mode doesn't flicker, but @markdurrant would be best to know how to test all that stuff.

markdurrant:
@olegat No issues with darkmode as far as I can see. 

However when I run the branch locally it looks like it fixes the Firefox console error (AG-9991), but doesn't display the background image (AG-9990). 

Unless there's something I'm missing.

olegat:
> However when I run the branch locally it looks like it fixes the Firefox console error (AG-9991), but doesn't display the background image (AG-9990).
> 
> Unless there's something I'm missing.

Ah right... I misread my own bug description 🙈 I thought that the background color was broken rather than the background image.

I'll update the bug & PR descriptions.

alantreadway (approved):
LGTM

PR #720, opened 2023-12-11
Refactor validation

iMoses:
A bit of cleanup but the main chain is to support the marking of optional properties


alantreadway (approved):
LGTM

PR #722, opened 2023-12-11
Fix HMR within plunker running locally

jacobp100:
HMR was broken in Plunker because Astro adds a HMR script like `<script src="/their-hmr-script.js">` - but since we're on plunker.com, that didn't work. This adds our local host only in development

There was some other code where Astro was generating bad tags in prod, so I've moved that into this logic too


alantreadway (approved):
LGTM

taktran (approved):
Nice clean up ✨ 👍🏼

PR #724, opened 2023-12-12
AG-9950 Fix polar crosslines

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-9950

Adds the padding to the polar crosslines to match cartesian. Also fixes issues with polar category axes & crosslines and cleans it up a little.


packages/ag-charts-enterprise/src/axes/polar-crosslines/__image_snapshots__/polar-cross-line-test-ts-polar-cross-line-create-for-radial-bar-polygon-crosslines-it-should-render-to-canvas-as-expected-1-snap.png
lsjroberts, line 1: This example exists to test the crosslines, and since we don't need a polygonal radial bar (at least at the moment) I left that part as-is.

packages/ag-charts-enterprise/src/axes/polar-crosslines/__image_snapshots__/polar-cross-line-test-ts-polar-cross-line-create-for-radial-bar-circle-crosslines-it-should-render-to-canvas-as-expected-1-snap.png
lsjroberts, line 1: There's a slight issue with the positioning on the radial bars and the radii when `axes['radius-category'].paddingInner` is provided. Note the mismatch on the inner circles. The crosslines are clamped to the inner and outer radius, but the axis lines are not accurate to those, and there isn't enough padding on the bars.

This is probably a larger fix and something for a follow up PR.

packages/ag-charts-community/src/chart/axis/axis.ts
```diff
@@ -301,10 +301,12 @@ export abstract class Axis<S extends Scale<D, number, TickInterval<S>> = Scale<a
 
     private attachCrossLine(crossLine: CrossLine) {
         this.crossLineGroup.appendChild(crossLine.group);
+        if (crossLine.labelGroup) this.crossLineGroup.appendChild(crossLine.labelGroup);
```
alantreadway: Is there any risk that `labelGroup` will be added/removed between the attach and detach? (It might be better to just have a fixed/static group on the `CrossLine` type, and always add/remove it - which is the approach we've taken elsewhere to give better consistency guarantees).

lsjroberts: I wouldn't have thought so, but I'll change it to match the rest

alantreadway (approved):
LGTM

PR #730, opened 2023-12-12
AG-9966 Validate Axis.title type

olegat:
https://ag-grid.atlassian.net/browse/AG-9966

This logs a warning if the property is set to to a non-object.


alantreadway (approved):
LGTM

PR #734, opened 2023-12-13
Fix link validation

jacobp100:


taktran (approved):
Nice one 👍🏼 

No need for trailing slash config anymore.

PR #736, opened 2023-12-13
Refactor nx tasks to remove some dependencies on the ag-charts builds

jacobp100:


.github/workflows/ci.yml
```diff
@@ -70,14 +70,14 @@ jobs:
                       packages/*/node_modules/
                       plugins/*/node_modules/
                       libraries/*/node_modules/
-                  key: node_modules-${{hashFiles('yarn.lock','patches/*.patch')}}
+                  key: node_modules-${{hashFiles('yarn.lock','patches/*.patch', 'packages/*/package.json', 'plugins/*/package.json')}}
```
jacobp100, line 73: This is required because dependencies on internal projects don't change the yarn lockfile, but they do need yarn to run to set up symlinks

PR #737, opened 2023-12-13
AG-9184 Fix area with data per series

lsjroberts:


alantreadway (approved):
LGTM once build fixed

PR #739, opened 2023-12-13
Ag 9970

jacobp100:
* Removes empty methods from Angular and Vue components
* Returns an object with duplicated properties with the dark mode comments - this is fine under whichever ES spec included `{ [objectKey]: something }` syntax, the second declaration takes precedence


alantreadway (approved):
LGTM

PR #742, opened 2023-12-13
Prepare decorators and validation to support properties instances

iMoses:


tsconfig.base.json
```diff
@@ -11,7 +11,7 @@
     "esModuleInterop": true,
     "target": "es2015",
     "module": "esnext",
-    "lib": ["es2020", "dom"],
+    "lib": ["es2022", "dom"],
```
alantreadway, line 14: Is this for `Object.hasOwn()`? caniuse.com looks good from a support perspective, but can you confirm with Sean that this aligned with the Grid's expectations for minimum browser API level we support?

alantreadway (approved):
LGTM - I leave to your discretion to merge when the accompanying mechanical changes are ready to merge.

tsconfig.base.json
iMoses: Sure, I'll ask him about it... (yes, that's mainly for `Object.hasOwn`)

PR #743, opened 2023-12-13
AG-10005 - nx dev speed-up

alantreadway:
https://ag-grid.atlassian.net/browse/AG-10005

Main changes:
- Splits `.d.ts` file generation to a distinct `build:types` target for `ag-charts-community` and `ag-charts-enterprise`, and updates `package.json` references.
- Switches to `esbuild` from `rollup` (80-90% speedup!)
- Fixes `nx dev` watch execution.
- Removes indirect dependency of thumbnail generation on library implementation (package version bump will trigger update).

Bonus:
- Cleanup of thumbnail generator console output:
  - hide license manager message.
  - addition of task name to improve traceability of example warnings.

Preview:

https://github.com/ag-grid/ag-charts/assets/17544187/e71687f7-5d4e-4eba-93c5-4cdad152a99c


jacobp100 (approved):

patches/@nx+esbuild+17.0.3.patch
```diff
@@ -0,0 +1,41 @@
...
+-            ? '.mjs'
+-            : format === 'esm'
+-                ? ESM_FILE_EXTENSION
+-                : CJS_FILE_EXTENSION;
++    const defaultExt = format === 'cjs' ? CJS_FILE_EXTENSION : format === 'esm' ? ESM_FILE_EXTENSION : '.js';
++    if (userDefExts[defaultExt]) {
++        return userDefExts[defaultExt];
++    }
++
++    return defaultExt;
```
jacobp100, line 27: Is nx esbuild overwriting the path extensions?

jacobp100 (approved):

patches/@nx+esbuild+17.0.3.patch
alantreadway: So it had some dubious logic (being patched here) about how to derive an extension, even if the configured `outputFileName` had a valid extension - so the patch:
1) Doesn't calculate a new extension if `outputFileName` already has an extension of `js`, `cjs` or `mjs`.
2) Uses the user-configured extension mapping in the esbuild options if they are configured.
3) Only then falls back to the defaults the Nx plugin decides.

alantreadway:
Merging, might need to clear CI cache after merge for `latest`, but I'll keep an eye on things.

PR #745, opened 2023-12-14
AG-10089 Unique generated series

olegat:
https://ag-grid.atlassian.net/browse/AG-10089

Make sure that generated series IDs do not clash with user specified series IDs.

Also add tests to verify this behaviour.


packages/ag-charts-community/src/chart/series/series.ts
```diff
@@ -301,7 +300,7 @@ export abstract class Series<
     protected readonly NodeClickEvent: INodeClickEventConstructor<TDatum, any> = SeriesNodeClickEvent;
 
     @Validate(STRING)
-    readonly id = createId(this);
```
alantreadway, line 304: Feedback left in JIRA.

olegat:
Closing PR. The long term plan is to separate user-specified IDs and internal auto-generated IDs.

PR #746, opened 2023-12-14
AG-6193 Draw rings in pie charts when the sum of datums in 0

olegat:
https://ag-grid.atlassian.net/browse/AG-6193


packages/ag-charts-community/src/chart/series/polar/pieSeries.ts
```diff
@@ -334,7 +340,14 @@ export class PieSeries extends PolarSeries<PieNodeDatum, Sector> {
...
         this.sectorLabelSelection = Selection.select(pieSectorLabels, Text);
         this.innerLabelsSelection = Selection.select(innerLabels, Text);
-        this.innerCircleSelection = Selection.select(this.innertCircleGroup, Circle);
+        this.innerCircleSelection = Selection.select(this.innerCircleGroup, Circle);
+
+        for (const circle of [this.zerosumInnerRing, this.zerosumOuterRing]) {
+            circle.fillOpacity = 0;
+            circle.stroke = 'black';
+            circle.strokeWidth = 1;
+            circle.strokeOpacity = 1;
```
alantreadway: 🤔 Should these relate to the pie series options? I.e. if I set a `series[].strokeWidth` should it be applied here?

olegat: Done. The `circle.stroke` property now depends on the theme-able option `calloutLabel.color`. The strike width & opacity can stay as is.

alantreadway (approved):
LGTM

PR #748, opened 2023-12-14
Ag 10044 design refactor phase 2

markdurrant:
Smaller design refactor


packages/ag-charts-website/src/design-system/_root.scss
```diff
@@ -1,10 +1,2 @@
 @use 'sass:map';
```
taktran: Do we still need this file?

taktran (approved):
LGTM 👍🏼 Just a minor comment

packages/ag-charts-website/src/design-system/_root.scss
markdurrant: Good spot. I'll remove.

PR #749, opened 2023-12-14
AG-10005 - Add multi threaded thumbnail generation

alantreadway:
https://ag-grid.atlassian.net/browse/AG-10005

Uses `process.fork()` to spawn multiple workers to process thumbnails in parallel, when `--batch` is used.

Bonus:
- Small cleanup of `yarn postinstall`.


taktran (approved):
LGTM 👍🏼

plugins/ag-charts-generate-chart-thumbnail/src/executors-utils.ts
```diff
@@ -97,6 +99,83 @@ export function batchExecutor<ExecutorOptions>(
...
+};
+
+process.on('exit', () => {
+    terminateWorkers();
+});
+
+export function batchWorkerExecutor<ExecutorOptions>(worker: () => ChildProcess) {
+    const maxChildCount = Math.max(1, os.cpus().length - 1);
+    const results: Record<string, Promise<BatchExecutorTaskResult>> = {};
+    const resolvers: Record<string, (res: BatchExecutorTaskResult) => void> = {};
```
jacobp100: Minor - but a `Map` would be better here

plugins/ag-charts-generate-chart-thumbnail/src/executors/generate/batch-instance.ts
```diff
@@ -0,0 +1,42 @@
...
+            await generateFiles(options, context);
+            result = { task: taskName, result: { success: true, terminalOutput: '' } };
+        } catch (e) {
+            result = { task: taskName, result: { success: false, terminalOutput: `${e}` } };
+        }
+
+        process.send(result);
+    }
+
+    processorRef = undefined;
```
jacobp100, line 36: Did this need a `cancelImmediate`?

jacobp100 (approved):

plugins/ag-charts-generate-chart-thumbnail/src/executors/generate/batch-instance.ts
alantreadway: I think it's redundant, if `processor()` is invoked again and the queue is empty it will just exit quickly.

plugins/ag-charts-generate-chart-thumbnail/src/executors-utils.ts
alantreadway: Fixed

PR #754, opened 2023-12-15
AG-9774 Return empty domain when there is no data

olegat:
https://ag-grid.atlassian.net/browse/AG-9774

This stops the chart from rendering the y-Axis.


alantreadway (approved):
LGTM

PR #766, opened 2023-12-19
AG-9184 Update data controller to filter and error invalid requests

lsjroberts:

For bar and area series, this filters out datasets with varying lengths and logs an error. It rejects the invalid requests to ensure the promises are completed and the chart can update.


alantreadway (approved):
LGTM

PR #769, opened 2023-12-19
AG-8696 Fix bar series to render every context with duplicate datum ids

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-8696

For bar series, iterates each of the `values` in the `processedData.data` array and creates a full array of contexts, rather than just the first.

Screenshot 2023-12-19 at 13 06 46

PR #770, opened 2023-12-19
AG-9519 Fix range bar series to render every context with dup datum ids

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-9519

Same change as #769.


alantreadway (approved):
LGTM

PR #773, opened 2023-12-19
AG-8880 Fix legend and node click to only trigger when mouse down is over same element

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-8880

For legend item clicks and series node clicks, tracks the element that the `mousedown` event targeted then compares it to the element the `click` event targets. Only fires the listener events if these match.

### DIFF
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -297,6 +297,8 @@ export abstract class Chart extends Observable implements AgChartInstance {
 
     private readonly processors: UpdateProcessor[] = [];
 
+    private mouseDownIds?: { seriesId: string; itemId: string };
+
     protected constructor(specialOverrides: ChartExtendedOptions, resources?: TransferableResources) {
         super();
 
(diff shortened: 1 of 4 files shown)


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1200,6 +1203,13 @@ export abstract class Chart extends Observable implements AgChartInstance {
         }
     }
 
+    protected onMouseDown(event: InteractionEvent<'mouse-down'>) {
+        this.mouseDownIds = undefined;
+        this.checkSeriesNodeRange(event, (series, datum) => {
```
alantreadway, line 1207: How expensive is this operation? If expensive, maybe we could just use some measurement of distance between the mouse down and mouse up events to determine if a 'click' should be generated?

lsjroberts: I thought about using `mousemove` or distance here, but I don't think it is universally applicable as some nodes are a few pixels across and some are half the chart.

In terms of performance, I don't think adding one more call to this will make a noticeable difference. But I believe it is `O(n)` to the series + nodes + pick modes?

lsjroberts:
Closing in favour of improvements to interactions

PR #776, opened 2023-12-20
Refactor series properties management

iMoses:


packages/ag-charts-community/src/chart/axis/axisTitle.ts
```diff
@@ -20,19 +20,19 @@ import { Caption } from '../caption';
 
 export class AxisTitle implements AgAxisCaptionOptions {
     @Validate(BOOLEAN)
-    enabled = false;
+    enabled: boolean = false;
 
     @Validate(STRING, { optional: true })
-    text?: string = undefined;
+    text?: string;
```
jacobp100, line 26: The JS engine will cope better if you pre-initialize these to undefined so the shape of the object doesn’t change

alantreadway: Interesting - we should probably benchmark this to work out the practical impact.

packages/ag-charts-community/src/chart/caption.ts
```diff
@@ -58,24 +62,18 @@ export class Caption {
...
 
-    private destroyFns: Function[] = [];
-
-    constructor(protected readonly moduleCtx: ModuleContext) {
-        const node = this.node;
-        node.textAlign = 'center';
-        node.pointerEvents = PointerEvents.None;
-
-        this.destroyFns.push(moduleCtx.interactionManager.addListener('hover', (e) => this.handleMouseMove(e)));
+    registerInteraction(moduleCtx: ModuleContext) {
```
alantreadway, line 75: Nit: I don't like this as a general DI-like approach, constructor injection is still preferable IMO - caveat we need a post-constructor initialiser call for actual use of injected dependencies, which is I guess the real reason you needed to do this?

packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
```diff
@@ -347,7 +294,7 @@ export class AreaSeries extends CartesianSeries<
                         point,
                         fill: marker.fill ?? seriesFill,
                         stroke: marker.stroke ?? seriesStroke,
-                        strokeWidth: marker.strokeWidth ?? this.getStrokeWidth(this.strokeWidth),
+                        strokeWidth: marker.strokeWidth ?? this.getStrokeWidth(this.properties.strokeWidth),
```
alantreadway, line 297: Thought: Do we think there is any significant performance penalty to the extra level of nesting here? Should we destructure these outside these loops as a general pattern?

packages/ag-charts-community/src/util/string.ts
alantreadway, line 1: Good spot, this was being exported as part of `sparkline-utils.ts` but it's not actually used in the Grid codebase it appears.

alantreadway (commented):
Generally LGTM - nice work!

packages/ag-charts-community/src/chart/axis/axisTitle.ts
iMoses: I'm all in favor of testing it.
Personally I don't think it matters because the decorator makes sure the property is declared onto the class prototype, so even if  not initialised the structure never changes.

packages/ag-charts-community/src/chart/caption.ts
iMoses: Hostenly, I did it as a hack ;)
I don't think that in this case there's a need for any of these. I see it as an example of DI abuse.
a caption shouldn't register itself to a service, that's bad coupling. 
We need to invert the control, move it to a higher level.
My patch moves it only one step up the chain :)

packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
iMoses: Do you know for a fact that it has influence over performance?
I don't think it does...

PR #788, opened 2023-12-22
AG-10110 Center multiple chart example in gallery

jacobp100:
In v10 we might want to revisit the css & html structure. They have the style

```css
position: relative;
display: block;
width: 100%;
height: 100%;
cursor: default;
```

But that 100% width and height aren't necessarily of the parent


alantreadway (approved):
LGTM

PR #792, opened 2024-01-02
AG-5716

jacobp100:
https://ag-grid.atlassian.net/browse/AG-5716

* Revert changes to connector lines in waterfall series when items have corner radii


olegat (approved):
LGTM

PR #797, opened 2024-01-02
AG-8196 QA Feedback for showSeriesStroke

olegat:
https://ag-grid.atlassian.net/browse/AG-8196

* Hide legend markers in line series when `marker.enabled=false` and `legend.item.showSeriesStroke=true`
* Add e2e tests for `legend.item.showSeriesStroke`


lsjroberts (approved):
LGTM

PR #798, opened 2024-01-02
AG-8097 Fix alignment of axis lines to better match the position of markers

jacobp100:
https://ag-grid.atlassian.net/browse/AG-8097

The markers aren't aligned with the pixel grid

There was an issue where we'd align try to align the axis lines to the pixel grid, but did so ignoring DPI scaling - so things would look off centre

Before -

* Line nodes would align to a pixel grid (although DPI scaling was ignored)
* The axis lines were set to `y = 0` and `translateY` was used to position them
* This would mean the line node alignment wasn't used
* However, `translateY` was `Math.round`ed to make things crisp, but this means the DPI scaling was ignored

Now -

* Line nodes correctly align to the DPI-adjusted pixel grid
* Axis lines are set with an explicit `y` value instead of `translateY`
* However, for animations, we still use `translateY` - otherwise the line node will re-compute the path data on every frame

Alternative is something like https://github.com/ag-grid/ag-charts/pull/799


packages/ag-charts-community/src/chart/crossline/__image_snapshots__/cross-line-test-ts-cross-lines-create-for-line-crosslines-it-should-render-to-canvas-as-expected-1-snap.png
lsjroberts, line 1: Lost a crossline border on the left edge here

jacobp100: For posterity - this was the case on `latest` too, but because we need to many pixels to change for it to be considered 'different', this was missed until now

PR #800, opened 2024-01-03
AG-10148 Hidden line series showing on load if animations enabled

olegat:
https://ag-grid.atlassian.net/browse/AG-10148


lsjroberts (approved):
LGTM

PR #803, opened 2024-01-03
AG-10212 Rename esm artefacts to .mjs to ensure correct resolution, esp with vitest

seanlandsman:


jacobp100 (approved):
Did you validate the generated `"main"` etc. in `dist/package.json` still matches up correctly after this change?

alantreadway:
> nx run all:"pack:verify"
> —————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————
>
 >  NX   Successfully ran target pack:verify for project all and 33 tasks it depends on (45s)

^ `nx pack:verify` passes, so this should cover off any mismatching of build outputs vs. generated `package.json` (I added checks before the 9.0.0 launch specifically to catch this class of mismatch) 🤞

alantreadway (approved):
LGTM

PR #805, opened 2024-01-03
AG-10152 Fix disappearing markers combination in charts

olegat:
https://ag-grid.atlassian.net/browse/AG-10152

When animations are disabled and bar series are drawn with line/area, the processedData.reduced.diff of the line/area series is undefined. This will cause the `prepareLinePathAnimation` function to use the `pairContinuousData` function, even when the data is categoric. This has the affect of adding all the markers into the both `pairMap.added` and `pairMap.removed`.

The `prepareMarkerAnimation` functions considers the status of the marker status to be 'removed'; it seems to assume that a marker cannot be both 'removed' and 'added' (because logically this would mean that the marker status is 'updated'). This causes it to generate a `toFn` to { opacity: 0 } instead of the expected { opacity: 1 }.

Allowing the `pairCategoryData` function to take an optional diff value ensures that the marker status is set to 'updated' and the correct `toFn` is used.


lsjroberts (approved):
LGTM

PR #806, opened 2024-01-03
Refactor chart options mechanism (part 2)

iMoses:


packages/ag-charts-community/src/chart/chartAxesTypes.ts
```diff
@@ -12,7 +12,7 @@ const AXES_THEME_TEMPLATES: Record<string, {}> = {};
 
 export const CHART_AXES_TYPES = {
     has(axisType: string) {
-        return Object.prototype.hasOwnProperty.call(TYPES, axisType);
+        return Object.hasOwn(TYPES, axisType);
```
jacobp100, line 15: What's our browser support?

packages/ag-charts-community/src/chart/themes/constants.ts
```diff
@@ -7,14 +7,16 @@ import type {
     FontWeight,
 } from '../../options/agChartOptions';
 
-export const FONT_SIZE = {
+export const FONT_SIZE: Record<string, number> = {
```
jacobp100: This removes the type safety so if you do `FONT_SIZE.NotASize`, it won't complain

jacobp100 (approved):

packages/ag-charts-community/src/chart/chartAxesTypes.ts
iMoses: we compile to `es2015` but locally work on `es2022`, so it doesn't effect the output - only our code
Sean previously approved the (/a similar) change :)

PR #808, opened 2024-01-03
AG-10138 Allow format to remove blank lines in generated examples

jacobp100:
https://ag-grid.atlassian.net/browse/AG-10138

Previously, the darkmode logic was integrated in the idiomatic way into the frameworks, and we used comment markers to work out what code to strip when it's presented to the user. The issue with this is that after these are stripped, it can leave excess whitespace in the examples

This PR un-tangles the darkmode logic so it's no longer dependent on each framework - and is now just a single snippet at the bottom of each `main.js` file. Technically, it could conflict with updates in the frameworks - so this is something we'll have to be careful with. But as long as we're careful to never include 'theme' in the options, we should be ok

Note - we had `bullet-chart-test` that generated invalid JS for React before, and now generates something that won't render anything, but is valid JS. This will be fixed in https://ag-grid.atlassian.net/browse/AG-9843


plugins/ag-charts-generate-example-files/src/executors/generate/generator/transformation-scripts/react-utils.ts
```diff
@@ -45,6 +45,7 @@ export function convertTemplate(template: string) {
 }
 
 export function convertFunctionalTemplate(template: string) {
+    if (template.includes('grid-area')) return 'null';
```
taktran, line 48: What does this do here?

taktran (approved):
Nice clean up work with the dark mode snippets ✨ 👏🏼 👏🏼 👏🏼 

> Technically, it could conflict with updates in the frameworks - so this is something we'll have to be careful with. But as long as we're careful to never include 'theme' in the options, we should be ok

As in when updating the `transformation-scripts` we don't update `theme` in the options there?

plugins/ag-charts-generate-example-files/src/executors/generate/generator/transformation-scripts/react-utils.ts
jacobp100: If there's multiple charts, give up. It wasn't supported earlier, and made syntax errors - and now those syntax errors were now causing the build to fail

PR #809, opened 2024-01-04
AG-9066 Fix test names

jacobp100:
https://ag-grid.atlassian.net/browse/AG-9066

* All test files now have one describe block
* Changes `for (...) { test('...') }` to `it.each` etc. so we get errors if there are no test cases (we had one example of this)

Note this regenerates image snapshots. If it looks like an image changed, it's because it never met the threshold before to be considered 'changed'


packages/ag-charts-community/src/chart/galleryExamplesUSTimezone.test.ts
```diff
@@ -50,15 +50,21 @@ describe('Gallery Examples (US TZ)', () => {
     });
 
     const ctx = setupMockCanvas();
-    for (const [exampleName, example] of Object.entries(TIME_AXIS_EXAMPLES)) {
-        it(`for ${exampleName} it should create chart instance as expected`, async () => {
+    // Missing time axis examples
+    it.skip.each(Object.entries(TIME_AXIS_EXAMPLES))(
```
lsjroberts, line 54: why skipped?

packages/ag-charts-community/src/chart/mapping/themes.test.ts
```diff
@@ -17,7 +16,7 @@ import { themes } from './themes';
 expect.extend({ toMatchImageSnapshot });
 
 /* eslint-disable no-console */
-describe('ThemesValidation', () => {
+describe('validateChartThemeObject', () => {
```
lsjroberts: Should this be called `'Themes'` to match the other changes matching their file names?

lsjroberts (approved):
LGTM apart from minor questions

packages/ag-charts-community/src/chart/galleryExamplesUSTimezone.test.ts
jacobp100: There weren't any examples. It was missed before in the for loop, but now jest throws an error

packages/ag-charts-community/src/chart/mapping/themes.test.ts
jacobp100: Updated

PR #823, opened 2024-01-08
AG-10162 QA Feedback: Range Area Legend Shape

olegat:
https://ag-grid.atlassian.net/browse/AG-10162


alantreadway (commented):
Approach LGTM

PR #833, opened 2024-01-09
AG-8112 Add connectMissingData option for radar series

lsjroberts:

Renames `connectNulls` to `connectMissingData` and adds it for radar series.

Also fixes area series to replace invalid values with `0` when stacked and `connectMissingData: true`.

I removed the snapshot image diff threshold since some new tests were not updating when the radar line was changing. The other tests are the same with just a slight diff on the text, probably from a while back.

edit: looks like those snapshot updates have been done elsewhere after I rebased from latest


alantreadway (approved):
LGTM so far

PR #834, opened 2024-01-09
AG-10268 - Add ag-charts-react UMD bundle.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-10268

### DIFF
--- a/packages/ag-charts-react/project.json
+++ b/packages/ag-charts-react/project.json
@@ -9,6 +9,20 @@
       "executor": "nx:noop"
     },
     "build": {
+      "executor": "nx:noop",
+      "dependsOn": ["build:package", "build:umd"],
+      "inputs": [
+        {
+          "externalDependencies": ["npm:typescript"]
+        }
+      ],
+      "cache": false,
+      "configurations": {
+        "watch": {},
+        "production": {}
+      }
+    },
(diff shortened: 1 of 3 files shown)


packages/ag-charts-react/project.json
```diff
@@ -36,6 +50,34 @@
...
+        "tsConfig": "packages/ag-charts-react/tsconfig.lib.json",
+        "esbuildConfig": "esbuild.config.cjs",
+        "platform": "browser",
+        "target": "es6",
+        "bundle": true,
+        "thirdParty": true,
+        "skipTypeCheck": true,
+        "external": ["react", "react-dom", "react/jsx-runtime", "ag-charts-community"],
+        "format": ["cjs"],
+        "sourcemap": true
```
jacobp100, line 70: How many of these do we need if we set the esbuildConfig?

jacobp100 (approved):

packages/ag-charts-react/project.json
alantreadway: The esbuild config is setup for `ag-charts-community` and `ag-charts-enterprise` non-UMD modules, so these overrides are needed - for now I only want to encode the things we can't configure through Nx in those specialised Javascript config files.

PR #839, opened 2024-01-09
AG-10251 - Fix consistent handling of DataModel scoped extracted values.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-10251

Fixes inconsistencies in data-model processing so that internal value representation doesn't leak back to series implementations.


packages/ag-charts-community/src/chart/data/dataController.ts
```diff
@@ -84,6 +84,10 @@ export class DataController {
         }
 
         for (const { opts, data, resultCbs, rejects, ids } of merged) {
+            const needsValueExtraction = opts.props.some((p) =>
+                p.type === 'value' || p.type === 'key' ? p.useScopedValues ?? false : false
+            );
```
lsjroberts: perhaps nicer as:

```suggestion
            const needsValueExtraction = opts.props.some((p) => {
                if (p.type !== 'value' && p.type !== 'key') return false;
                return p.useScopedValues ?? false;
            });
```

alantreadway: Fix applied, thanks!

lsjroberts (approved):
lgtm

PR #847, opened 2024-01-10
AG-10274 Fix tick size to ignore disabled theme override

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10274


alantreadway (approved):
LGTM

PR #850, opened 2024-01-10
AG-8112 Add connectMissingData option for range area series

lsjroberts:

Also fixes radar to connect when the first datum is invalid.


alantreadway (approved):
LGTM

PR #856, opened 2024-01-11
AG-8112 Update range area missing data docs to be simpler

lsjroberts:

Also fixes missing data examples for frameworks (variables outside the button functions aren't lifted into the framework classes).


alantreadway (approved):
LGTM

PR #858, opened 2024-01-11
AG-9965 Add warning for invalid values that are not replaced

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-9965

This logs a warning if there is an invalid value that does not have a fallback value provided in the data model.

Currently this only occurs on continuous axes where the value is not a date or number.


alantreadway (approved):
LGTM

PR #859, opened 2024-01-11
AG-10143

jacobp100:
The issue was two parts. Firstly, the axis computes fractional digits, but these weren't used on the first layout. Secondly, the fraction digit logic was prone to floating point weirdness

Fix axis fraction digits not being used on first render, improve countFractionDigits logic

https://ag-grid.atlassian.net/browse/AG-10143


alantreadway (approved):
LGTM

PR #860, opened 2024-01-11
AG-9915 Add legend.preventHidingAll

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-9915

Also a bit of refactoring.

### DIFF
--- a/packages/ag-charts-community/src/chart/legend.ts
+++ b/packages/ag-charts-community/src/chart/legend.ts
@@ -212,6 +212,9 @@ export class Legend {
     @Validate(UNION(['horizontal', 'vertical'], 'an orientation'), { optional: true })
     orientation?: AgChartLegendOrientation;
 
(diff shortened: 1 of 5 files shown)


packages/ag-charts-community/src/chart/legend.ts
```diff
@@ -765,6 +769,17 @@ export class Legend {
...
+            if (preventHidingAll && !newEnabled) {
+                const numVisibleItems = chartService.series
+                    .flatMap((series) => series.getLegendData('category'))
+                    .filter((datum) => datum.enabled).length;
+
+                if (numVisibleItems < 2) {
+                    newEnabled = true;
+                }
+            }
+
```
lsjroberts, line 782: This will fire the event even if the item was already enabled, but if for whatever reason all the legend items are disabled, this will allow them to be toggled back on again.

alantreadway (approved):
LGTM

PR #862, opened 2024-01-11
AG-10242 Fix crossline ranges coinciding with chart edge

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10242


alantreadway (approved):
LGTM

PR #871, opened 2024-01-12
Ag 10245/gallery examples

manapeirov:
PR for gallery examples, added another bar example to make it a multiple of 3 so that the line series doesn't display on the same row as the bars in the 3 column layout. Plus minor improvements to other examples.

Screenshot 2024-01-12 at 18 01 08
Screenshot 2024-01-12 at 18 01 37


alantreadway:
Closing for now, please re-open if/when relevant.

PR #877, opened 2024-01-15
AG-10326 Set default bullet colorRange based on user-defined series fill

olegat:
https://ag-grid.atlassian.net/browse/AG-10326


packages/ag-charts-enterprise/src/series/bullet/bulletSeriesProperties.ts
```diff
@@ -103,4 +102,7 @@ export class BulletSeriesProperties extends AbstractBarSeriesProperties<AgBullet
 
     @Validate(OBJECT)
     readonly tooltip = new SeriesTooltip<AgBulletSeriesTooltipRendererParams>();
+
+    @Validate(COLOR_STRING) // Internal: Set by paletteFactory.
+    backgroundFill: CssColor = 'white';
```
olegat, line 108: This isn't a good solution IMHO. Is there an easy to retrieve the `themeTemplatePalette` from the module's palette factory?

alantreadway: The WaterfallSeries has some special handling in this respect, does this look relevant?
https://github.com/ag-grid/ag-charts/blob/65c5647df895858398b131ee07ac30e28c7eb12b/packages/ag-charts-enterprise/src/series/waterfall/waterfallModule.ts#L19-L46

olegat: Doesn't look relevant. The `positive`, `negative` and `total` options are documented in the API: https://charts.ag-grid.com/options/series/waterfall/#reference-AgWaterfallSeriesOptions-item

The issue that I'm facing here is that `backgroundFill` is not documented and is not meant to be. It's only used for internal implementation in _bulletSeries.ts_. The user is not meant to be able to reconfigured `backgroundFill`, but they still can simply by set `options.series[].backgroundFill`

alantreadway: Discussed offline, I've got the wrong end of the stick

packages/ag-charts-community/src/util/properties.ts
```diff
@@ -72,6 +73,58 @@ export class PropertiesArray<T extends BaseProperties> extends Array {
     }
 }
 
+export class OptionalPropertiesArray<T extends BaseProperties> {
```
alantreadway: Maybe sync with @iMoses about the use-case for this?

olegat: Resolved. We've decided to use an empty `PropertiesArray` instance instead.

alantreadway (approved):
LGTM

PR #883, opened 2024-01-16
AG-10290 Run Charts Outside Angular CD

StephenCooper:
The idea of this PR is to stop the following. When moving the mouse over the chart you can see the Angular is constantly running CD.

https://github.com/ag-grid/ag-charts/assets/20125490/b24e8406-7d21-44f4-8faf-4a9cb19c12c3

After this change this no longer happens.

https://github.com/ag-grid/ag-charts/assets/20125490/e05e6175-d672-46ed-9c00-1411e4eba460

The final part is to make sure that any listeners provided by the user are patched so that their listener runs inside Angular meaning that they can update their templates from within the listeners.


StephenCooper:
@alantreadway this is a plunker that can be used to validate that the patching of event listeners is correct.

https://plnkr.co/edit/FJLZAr7A8a931E7m?open=app.component.ts

alantreadway (approved):
LGTM once code-freeze is off

PR #884, opened 2024-01-17
AG-10368 Convert to standalone component

StephenCooper:
Angular now supports standalone components which means you do not need to use modules. 

By declaring the `AgChartAngular` as `standalone` it can directly be imported instead of via the `AgChartAngularModule`.

This change is backwards compatible.

I have updated the examples to use the component directly as there is no advantage to using the module.


alantreadway (approved):
LGTM

PR #885, opened 2024-01-17
AG-10339 - Animate overlay appearance.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-10339

Animates overlay show/hide using animation phases, so that overlays don't overlap the final/first series on legend toggle.


lsjroberts (approved):
lgtm

PR #886, opened 2024-01-17
AG-10341 Fix line labels not animating when visibility changes

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10341 

Label animation was skipped when the visibility changed from the legend toggle as neither the diff nor the scales had changed.

This is the simplest fix, but may be missing something?


alantreadway (approved):
LGTM

PR #888, opened 2024-01-18
AG-8219 Fix zoom selection to consider series area padding

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-8219

Fixes zoom to correctly consider series area padding, in particular zoom selection.

Also fixes `zoomManager.getZoom()` to return the first axis zoom (per direction), considering that to the be the primary axis. Previously it returned the last axis zoom.

Note in this example, we select the area from 80 to 40 on the y-axis. It zooms such that those are the points at the top and bottom of the series area, but the line can overflow into the padded area.

https://github.com/ag-grid/ag-charts/assets/3817697/dcc538c2-772a-487b-b7ee-6614f928c33f


alantreadway (approved):
LGTM

PR #889, opened 2024-01-18
AG-10337 Count missing datums per series

olegat:
The fundamental difference is that `DatumPropertyDefinition.missing` is now a map type instead of number type.

When `missing` was a number type it would globally keep track of all missing values (i.e. so the missing values in one series would affect the miss-count in another series and vice-versa).

This commit uses a miss-map to count to the number of missing values that each series has.


lsjroberts (approved):
Nice, LGTM. I've got some tests I can refactor a bit with your mock console.

PR #890, opened 2024-01-18
AG-10225 Change context menu to use highlight node for charts without markers

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10225

When you right click, the `mousedown` event is fired first, then the `contextmenu` event. When a `mousedown` event is triggered we start dragging, which clears the highlighted node. So we have to reject any `mousedown` events not from the primary button.


alantreadway (approved):
LGTM

PR #896, opened 2024-01-23
AG-10442 Expose BoxPlotSeries type via static property

timkendrick:


alantreadway (approved):
LGTM - thanks @timkendrick

PR #897, opened 2024-01-23
AG-10395 - Add initial and basic window lazy loading

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10395

Example of window loading. It was easiest in the end to just use the zoom change event and not worry about making a specific singular window. Though note it does have issues with subsequent windows when dragged, outside the scope of this spike.

https://github.com/ag-grid/ag-charts/assets/3817697/850569eb-55e7-4854-b2c7-ba87f605cbba

### DIFF
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -37,6 +37,7 @@ import { ChartHighlight } from './chartHighlight';
 import type { ChartMode } from './chartMode';
 import { ChartUpdateType } from './chartUpdateType';
 import { DataController } from './data/dataController';
+import { DataService } from './data/dataService';
 import { AnimationManager } from './interaction/animationManager';
 import { ChartEventManager } from './interaction/chartEventManager';
 import { CursorManager } from './interaction/cursorManager';
@@ -60,6 +61,7 @@ import { SeriesStateManager } from './series/seriesStateManager';
 import type { ISeries, SeriesNodeDatum } from './series/seriesTypes';
 import { Tooltip } from './tooltip/tooltip';
 import { BaseLayoutProcessor } from './update/baseLayoutProcessor';
+import { DataWindowProcessor } from './update/dataWindowProcessor';
 import type { UpdateProcessor } from './update/processor';
 import { UpdateService } from './updateService';
 
@@ -162,13 +164,6 @@ export abstract class Chart extends Observable implements AgChartInstance {
     })
     container: OptionalHTMLElement = undefined;
 
-    @ActionOnSet<Chart>({
-        newValue(value) {
-            this.series?.forEach((series) => {
-                series.setChartData(value);
-            });
-        },
-    })
     public data: any = [];
 
     @ActionOnSet<Chart>({
@@ -277,6 +272,7 @@ export abstract class Chart extends Observable implements AgChartInstance {
     protected readonly interactionManager: InteractionManager;
     protected readonly tooltipManager: TooltipManager;
     protected readonly zoomManager: ZoomManager;
+    protected readonly dataService: DataService<any>;
     protected readonly layoutService: LayoutService;
     protected readonly updateService: UpdateService;
     protected readonly axisGridGroup: Group;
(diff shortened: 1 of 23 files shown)


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -346,7 +350,14 @@ export abstract class Chart extends Observable implements AgChartInstance {
...
 
-        this.processors = [new BaseLayoutProcessor(this, this.layoutService)];
+        this.dataLazyLoader = new DataLazyLoader<any>();
+
+        this.processors = [
+            new BaseLayoutProcessor(this, this.layoutService),
+            new DataWindowProcessor(this, this.dataLazyLoader, this.updateService, this.zoomManager, (data) => {
+                this.data = data;
+            }),
+        ];
```
lsjroberts, line 360: I like extracting this out of `Chart`, but this data setting callback is a little meh. On the other hand, a data service to change it seems too much?

packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -930,14 +941,18 @@ export abstract class Chart extends Observable implements AgChartInstance {
...
+
         if (this.series.some((s) => s.canHaveAxes)) {
             this.assignAxesToSeries();
             this.assignSeriesToAxes();
         }
 
         const dataController = new DataController(this.mode);
         const seriesPromises = this.series.map((s) => s.processData(dataController));
-        await dataController.execute();
+        dataController.execute();
```
lsjroberts, line 946: This function is not actually asynchronous

packages/ag-charts-community/src/chart/chart.ts
lsjroberts: I like extracting this out of `Chart`, but this data setting callback is a little meh. On the other hand, a data service to change it seems too much?

packages/ag-charts-website/src/content/docs/lazy-test/_examples/lazy/main.ts
```diff
@@ -0,0 +1,78 @@
...
+            const diff = timeAxis.max.getTime() - timeAxis.min.getTime();
+            if (diff <= DAY * 7) {
+                return new Promise((resolve) =>
+                    setTimeout(() => resolve(getWindowData(timeAxis.min, timeAxis.max)), 500)
+                );
+            }
+        }
+
+        return new Promise((resolve) => setTimeout(() => resolve(getCoarseData()), 500));
+    },
```
lsjroberts: The relevant bit of the example

packages/ag-charts-community/src/chart/chart.ts
alantreadway: I think it would be fine to have a `DataService` - the scope of 'tracking the current dataset(s)' seems fine even if it's the only responsibility?

packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -930,14 +941,18 @@ export abstract class Chart extends Observable implements AgChartInstance {
     }
 
     async processData() {
+        if (typeof this.data === 'function') {
+            this.data = await this.dataLazyLoader.full(this.data);
+        }
+
```
alantreadway: I would perhaps abstract this away or always invoke it (and have a no-op behaviour if `data` is a static array)?

packages/ag-charts-community/src/chart/chart.ts
alantreadway: Also, if we could avoid having mutating state on `Chart`, that might be preferable anyway?

packages/ag-charts-community/src/options/chart/chartOptions.ts
alantreadway, line 1: Nit: Could we hold off on making options contract changes for the moment?

alantreadway, line 1: Could we hold-back the options contract changes for the moment?

packages/ag-charts-website/src/content/docs/lazy-test/_examples/lazy/data.ts
```diff
@@ -0,0 +1,48 @@
...
+const week = day * 7;
+
+type Datum = { time: number; value: number };
+
+// Simple seeded randomisation for consistent data - https://stackoverflow.com/a/19303725
+let seed = 1234;
+function random() {
+    const x = Math.sin(seed++) * 10000;
+    return x - Math.floor(x);
+}
```
alantreadway: Let's come up with our own algorithm (we have other examples which already do this too?)

alantreadway (approved):
LGTM

PR #904, opened 2024-01-24
AG-10323 Fix tooltips on mobile devices

olegat:
https://ag-grid.atlassian.net/browse/AG-10323

The bug is caused because because we disabled the pointer whenever a drag event starts (i.e. when 'mousedown' or 'touchstart' is received). Now we only disable the pointer when the drag actually starts moving.

This also fixes a bug regression where clicking on a node with the mouse (on desktop) hides the tooltip and highlighting.


alantreadway (approved):
LGTM

PR #906, opened 2024-01-25
AG-10409 Pointer interactions refactor

olegat:

### DIFF
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -1084,10 +1084,6 @@ export abstract class Chart extends Observable implements AgChartInstance {
     };
 
     protected onMouseMove(event: InteractionEvent<'hover'>): void {
-        if (this.interactionManager.state !== InteractionState.Default) {
-            return;
-        }
-
         this.lastInteractionEvent = event;
         this.pointerScheduler.schedule();
 
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -1107,7 +1107,7 @@ export abstract class Chart extends Observable implements AgChartInstance {
         }
     });
     protected handlePointer(event: InteractionEvent<'hover'>) {
-        if (this.interactionManager.state !== InteractionState.Default) {
+        if (this.interactionManager.getState() !== InteractionState.Default) {
             return;
         }
 
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -46,7 +46,7 @@ import { CursorManager } from './interaction/cursorManager';
 import { GestureDetector } from './interaction/gestureDetector';
 import type { HighlightChangeEvent } from './interaction/highlightManager';
 import { HighlightManager } from './interaction/highlightManager';
-import type { InteractionEvent } from './interaction/interactionManager';
+import { InteractionEvent, InteractionState } from './interaction/interactionManager';
 import { InteractionManager } from './interaction/interactionManager';
 import { SyncManager } from './interaction/syncManager';
 import { TooltipManager } from './interaction/tooltipManager';
@@ -328,7 +328,7 @@ export abstract class Chart extends Observable implements AgChartInstance {
         this.seriesLayerManager = new SeriesLayerManager(this.seriesRoot);
         this.callbackCache = new CallbackCache();
 
-        this.animationManager = new AnimationManager(this.updateMutex);
+        this.animationManager = new AnimationManager(this.interactionManager, this.updateMutex);
         this.animationManager.skip();
         this.animationManager.play();
 
@@ -1084,6 +1084,10 @@ export abstract class Chart extends Observable implements AgChartInstance {
     };
 
     protected onMouseMove(event: InteractionEvent<'hover'>): void {
+        if (this.interactionManager.state !== InteractionState.Default) {
+            return;
+        }
+
         this.lastInteractionEvent = event;
         this.pointerScheduler.schedule();
 
@@ -1107,6 +1111,10 @@ export abstract class Chart extends Observable implements AgChartInstance {
         }
     });
     protected handlePointer(event: InteractionEvent<'hover'>) {
+        if (this.interactionManager.state !== InteractionState.Default) {
+            return;
+        }
+
         const { lastPick, hoverRect } = this;
         const { offsetX, offsetY } = event;
 
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -328,7 +328,7 @@ export abstract class Chart extends Observable implements AgChartInstance {
         this.seriesLayerManager = new SeriesLayerManager(this.seriesRoot);
         this.callbackCache = new CallbackCache();
 
-        this.animationManager = new AnimationManager(this.interactionManager, this.updateMutex);
+        this.animationManager = new AnimationManager(this.updateMutex);
         this.animationManager.skip();
         this.animationManager.play();
 
@@ -355,10 +355,9 @@ export abstract class Chart extends Observable implements AgChartInstance {
             this.interactionManager.addListener('hover', (event) => this.onMouseMove(event)),
             this.interactionManager.addListener('leave', (event) => this.onLeave(event)),
             this.interactionManager.addListener('page-left', () => this.destroy()),
-            this.interactionManager.addListener('wheel', () => this.disablePointer()),
 
-            // Block redundant and interfering attempts to update the hovered element during dragging.
-            this.interactionManager.addListener('drag', () => this.disablePointer()),
+            this.interactionManager.addListener('wheel', () => this.resetPointer()),
+            this.interactionManager.addListener('drag', () => this.resetPointer()),
 
             this.animationManager.addListener('animation-frame', () => {
                 this.update(ChartUpdateType.SCENE_RENDER);
@@ -517,7 +516,7 @@ export abstract class Chart extends Observable implements AgChartInstance {
         return result;
     }
 
-    disablePointer(highlightOnly = false) {
+    resetPointer(highlightOnly = false) {
         if (!highlightOnly) {
             this.tooltipManager.removeTooltip(this.id);
         }
@@ -628,7 +627,7 @@ export abstract class Chart extends Observable implements AgChartInstance {
 
             case ChartUpdateType.PROCESS_DATA:
                 await this.processData();
-                this.disablePointer(true);
+                this.resetPointer(true);
                 splits['🏭'] = performance.now();
             // fallthrough
 
@@ -903,7 +902,7 @@ export abstract class Chart extends Observable implements AgChartInstance {
         if (!width || !height || !isFiniteNumber(width) || !isFiniteNumber(height)) return;
 
         if (this.scene.resize(width, height)) {
-            this.disablePointer();
+            this.resetPointer();
             this.animationManager.reset();
 
             let skipAnimations = true;
@@ -1095,7 +1094,7 @@ export abstract class Chart extends Observable implements AgChartInstance {
 
     protected onLeave(event: InteractionEvent<'leave'>): void {
         if (!this.tooltip.pointerLeftOntoTooltip(event)) {
-            this.disablePointer();
+            this.resetPointer();
             this.update(ChartUpdateType.SCENE_RENDER);
         }
     }
@@ -1113,8 +1112,7 @@ export abstract class Chart extends Observable implements AgChartInstance {
 
         const disablePointer = (highlightOnly = false) => {
             if (lastPick) {
-                // Cursor moved from a non-marker node to empty space.
-                this.disablePointer(highlightOnly);
+                this.resetPointer(highlightOnly);
             }
         };
 
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -358,6 +358,7 @@ export abstract class Chart extends Observable implements AgChartInstance {
 
             this.interactionManager.addListener('wheel', () => this.resetPointer()),
             this.interactionManager.addListener('drag', () => this.resetPointer()),
+            this.interactionManager.addListener('contextmenu', () => this.resetPointer()),
 
             this.animationManager.addListener('animation-frame', () => {
                 this.update(ChartUpdateType.SCENE_RENDER);
--- a/packages/ag-charts-community/src/chart/interaction/interactionManager.ts
+++ b/packages/ag-charts-community/src/chart/interaction/interactionManager.ts
@@ -77,6 +77,13 @@ const CSS = `
 
 type SupportedEvent = MouseEvent | TouchEvent | Event;
 
+export enum InteractionState {
+    Default = -Infinity,
+    ZoomPan = 1,
+    ContextMenu = 2,
+    Animation = 3,
+}
+
 const DEBUG_SELECTORS = [true, 'interaction'];
 
 /**
@@ -97,6 +104,12 @@ export class InteractionManager extends BaseManager<InteractionTypes, Interactio
     private mouseDown = false;
     private touchDown = false;
     private dragStartElement?: HTMLElement;
+
+    private stateQueue: Set<InteractionState> = new Set();
+    private _state: InteractionState = InteractionState.Default;
+    public get state() {
+        return this._state;
+    }
 
     public constructor(element: HTMLElement, document: Document, window: Window) {
         super();
@@ -135,6 +148,20 @@ export class InteractionManager extends BaseManager<InteractionTypes, Interactio
         for (const type of EVENT_HANDLERS) {
             this.element.removeEventListener(type, this.eventHandler);
         }
+    }
+
+    public pushState(state: InteractionState) {
+        this.stateQueue.add(state);
+        this.updateCurrentState();
+    }
+
+    public popState(state: InteractionState) {
+        this.stateQueue.delete(state);
+        this.updateCurrentState();
+    }
+
+    private updateCurrentState() {
+        this._state = Array.from(this.stateQueue).reduce((max, s) => Math.max(max, s), InteractionState.Default);
     }
 
     private processEvent(event: SupportedEvent) {
--- a/packages/ag-charts-community/src/chart/interaction/interactionManager.ts
+++ b/packages/ag-charts-community/src/chart/interaction/interactionManager.ts
@@ -77,11 +77,14 @@ const CSS = `
 
 type SupportedEvent = MouseEvent | TouchEvent | Event;
 
+// These interaction state are both bitflags and priorities.
+// Smaller numbers have higher priority, because it is possible to find the least
+// significant bit in O(1) complexity using a bitwise operation.
 export enum InteractionState {
-    Default = 1,
-    ZoomDrag = 2,
-    ContextMenu = 4,
-    Animation = 8,
+    Default = 8,
+    ZoomDrag = 4,
+    ContextMenu = 2,
+    Animation = 1,
     All = Default | ZoomDrag | ContextMenu | Animation,
 }
 
@@ -106,8 +109,7 @@ export class InteractionManager extends BaseManager<InteractionTypes, Interactio
     private touchDown = false;
     private dragStartElement?: HTMLElement;
 
-    private stateQueue: Set<InteractionState> = new Set();
-    public state: InteractionState = InteractionState.Default;
+    private stateQueue: InteractionState = InteractionState.Default;
 
     public constructor(element: HTMLElement, document: Document, window: Window) {
         super();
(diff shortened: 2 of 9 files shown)


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -624,7 +615,6 @@ export abstract class Chart extends Observable implements AgChartInstance {
 
             case ChartUpdateType.PROCESS_DATA:
                 await this.processData();
-                this.disablePointer(true);
```
alantreadway, line 631: NOTE: The use-case here is that the highlighted node + tooltip should update when the data changes underneath it - one of the real-time examples is a good test-case for this.

packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -914,7 +904,6 @@ export abstract class Chart extends Observable implements AgChartInstance {
         if (!width || !height || !isFiniteNumber(width) || !isFiniteNumber(height)) return;
 
         if (this.scene.resize(width, height)) {
-            this.disablePointer();
```
alantreadway, line 906: NOTE: On resize, highlighted node + tooltip should be removed.

packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1100,7 +1093,6 @@ export abstract class Chart extends Observable implements AgChartInstance {
 
     protected onLeave(event: InteractionEvent<'leave'>): void {
         if (!this.tooltip.pointerLeftOntoTooltip(event)) {
-            this.disablePointer();
```
alantreadway, line 1098: NOTE: On mouse leaving the canvas bounding-box, the highlighted selection + tooltip should be cleared.

packages/ag-charts-enterprise/src/features/zoom/zoom.ts
```diff
@@ -186,6 +186,7 @@ export class Zoom extends _ModuleSupport.BaseModuleInstance implements _ModuleSu
 
     private onDoubleClick(event: _ModuleSupport.InteractionEvent<'dblclick'>) {
         if (!this.enabled || !this.enableDoubleClickToReset) return;
+        if (this.ctx.interactionManager.state !== InteractionState.Default) return;
```
alantreadway: Nit: Could we move these guards (which seem to be a common pattern) back into the listener config? This would also remove the direct coupling to `InteractionManager.state`.

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
```diff
@@ -104,7 +98,11 @@ export class InteractionManager extends BaseManager<
     private touchDown = false;
     private dragStartElement?: HTMLElement;
 
-    private pausers: Record<PauseType, number> = { animation: 0, 'context-menu': 0 };
+    private stateQueue: Set<InteractionState> = new Set();
+    private _state: InteractionState = InteractionState.Default;
+    public get state() {
```
alantreadway: This is probably not needed if we make filtering by state for listeners a responsibility of `InteractionManager`?

alantreadway (commented):
FYI: I've added some notes on use-cases to test, as these are things we've explicitly bug-fixed previously.

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
```diff
@@ -79,19 +73,19 @@ const CSS = `
 `;
 
 type SupportedEvent = MouseEvent | TouchEvent | Event;
-type ListenerMeta = { bypassPause?: PauseType[] };
+
+export enum InteractionState {
+    Default = -Infinity,
```
iMoses: nit: why do we need to set specific values on the enum?
Something to consider, we could use 2^X values to support "bitwise" conditions.

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
iMoses: nit: I personally try to avoid unnecessary setters and getters.
I think we can simply expose this property, it's internal anyhow, or maybe just replace it with a method `getState` and keep the private property named `state` and avoid the underscore prefix and duplicated definition.

packages/ag-charts-enterprise/src/features/zoom/zoom.ts
iMoses: Another option is to expose a `checkState(state: InteractionState)` and/or `isDefaultState()`

packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts
```diff
@@ -209,13 +204,13 @@ export class ContextMenu extends _ModuleSupport.BaseModuleInstance implements _M
         event.consume();
         event.sourceEvent.preventDefault();
 
-        this.show();
+        this.show(pickedNode);
     }
 
-    public show() {
-        if (!this.coverElement) return;
+    public show(pickedNode: ContextPickedNode) {
```
lsjroberts: I believe `pickedNode` could be `undefined`?

packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts
```diff
@@ -285,23 +267,27 @@ export class ContextMenu extends _ModuleSupport.BaseModuleInstance implements _M
...
         if (id && ContextMenu.disabledActions.has(id)) {
             return this.createDisabledElement(label);
         }
-        return this.createButtonElement(label, action);
+        return this.createButtonElement(pickedNode, label, action);
     }
 
-    private createButtonElement(label: string, callback: (params: ContextMenuActionParams) => void): HTMLElement {
+    private createButtonElement(
+        pickedNode: ContextPickedNode,
```
lsjroberts: So down to here should be `pickedNode?: ContextPickedNode`

packages/ag-charts-enterprise/src/features/zoom/zoom.ts
```diff
@@ -199,7 +200,11 @@ export class Zoom extends _ModuleSupport.BaseModuleInstance implements _ModuleSu
     }
 
     private onDrag(event: _ModuleSupport.InteractionEvent<'drag'>) {
+        const state = this.ctx.interactionManager.state;
         if (!this.enabled || !this.paddedRect || !this.seriesRect) return;
+        if (state !== InteractionState.Default && state !== InteractionState.ZoomPan) return;
+
+        this.ctx.interactionManager.pushState(InteractionState.ZoomPan);
```
lsjroberts: nit: it's not just pan here, can also be selection or axis zoom, so perhaps a more generic `InteractionState.ZoomDrag`?

lsjroberts (commented):
From a context menu and zoom point of view this all looks great, cheers!

packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts
olegat: I'm not entirely sure why these changes are even necessary TBH. I think that keeping `pickedNode` as property of `ContextMenu` should be fine, so I'll revert these changes

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
olegat: I have no strong opinions about this. I would however prefer making `_state` public than having a trivial `getState()` method.

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
olegat: We don't really need to. The initial idea that I had in mind was that the number would represent priorities, hence why `Default = -Infinity` has the lowest priority and `Animation = 3` has the highest priority. But this was unnecessary in the end. I think that using a string union might actually make the code more readable.

packages/ag-charts-community/src/chart/chart.ts
olegat: Okay. Thanks for the info. I'll restore these. Although I think that renaming the function to something like `resetPointer()` might make more sense.

packages/ag-charts-enterprise/src/features/zoom/zoom.ts
olegat: We could move these guards into the interaction manager. That is one of the approaches that was talking about on Tuesday. What I had in mind was some kind of state metadata that was be used to determine which interactions are allowed.

E.g.: (pseudo-code)

**animationManager.ts**
```typescript
interactionManager.pushState({
  stateName: 'animation',
  contextMenuEnabled: false,
  highlightAndTooltipsEnabled: false,
  dragEnabled: false,
  legendInteractionEnabled: false,
});
/**  run the animation */
interactionManager.popState('animation')
```

**contextMenu.ts**
```typescript
interactionManager.pushState({
  stateName: 'context-menu',
  contextMenuEnabled: true,
  highlightAndTooltipsEnable: false,
  dragEnabled: false,
  legendInteractionEnable: false,
});
/** wait for user to dismiss context-menu */
interactionManager.popState('state-menu')
```

But I think that this adds too much complexity and isn't very different to the mechanism that we've currently got. 

Event listeners really shouldn't know about other events listeners, but it makes it very difficult to coordinate things. I still believe that the best approach is to have individual listeners determine whether or not an event should be processed.

Perhaps a better option to reduce code duplication is to add a utility function for the `if (state === thing)` logic. e.g.:

```typescript
export function addStateConditionedEventListener(
  interactionManager: InteractionManager,
  eventType: string,
  eventHandler: (e) => any,
  requiredState: InteractionState[]
) {
  const wrapperFn = (e) => {
    if (requiredState.contains(interactionManage.state)) {
      return eventHandler(e);
    }
  }
  interactionManager.addEventListener( eventType, wrapperFn );
```

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
olegat: Actually powers of 2 is a good idea, because it would be much faster than iterating through an array 🤔

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
olegat: This is related to https://github.com/ag-grid/ag-charts/pull/906/files#r1472880057, so I'll resolve this convo.

packages/ag-charts-enterprise/src/features/zoom/zoom.ts
iMoses: I like your last proposal, and continuing our conversation about the state enum it could be a great place to use these flags (instead of an array) to support multiple states at once.

```
export enum InteractionState {
    Default = 0,
    ZoomPan = 1,
    ContextMenu = 2,
    Animation = 4,
    All = 7
}
// ...
addEventListener(eventName: string, eventHandler: (...args) => void, eventState = InteractionState.Default) // or InteractionState.All
// ...
interactionManager.addEventListener(eventType, (e) => {
  if (this.state === eventState || (this.state & eventState)) {
    eventHandler(e);
  }
});
// ...
pushState(state: InteractionState) { this.state |= state; } 
// ...
popState(state: InteractionState) { this.state &= ~state; }
```

olegat: Okay. SGTM 👍

olegat: Done.

packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts
olegat: Done. This change was reverted.

packages/ag-charts-enterprise/src/features/zoom/zoom.ts
olegat: Done. Renamed `ZoomPan` to `ZoomDrag`

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
olegat: Done. Changed number values to 1, 2, 4, 8

packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts
olegat: Done. This change was reverted.

packages/ag-charts-community/src/chart/navigator/navigator.ts
```diff
@@ -86,13 +86,11 @@ export class Navigator extends BaseModuleInstance implements ModuleInstance {
...
-        const interactionOpts = { bypassPause: [PauseType.ANIMATION] };
         this.destroyFns.push(
-            ctx.interactionManager.addListener('drag-start', (event) => this.onDragStart(event), interactionOpts),
-            ctx.interactionManager.addListener('drag', (event) => this.onDrag(event), interactionOpts),
-            ctx.interactionManager.addListener('hover', (event) => this.onDrag(event), interactionOpts),
-            ctx.interactionManager.addListener('drag-end', () => this.onDragStop(), interactionOpts),
+            ctx.interactionManager.addListener('drag-start', (event) => this.onDragStart(event)),
+            ctx.interactionManager.addListener('drag', (event) => this.onDrag(event)),
+            ctx.interactionManager.addListener('hover', (event) => this.onDrag(event)),
+            ctx.interactionManager.addListener('drag-end', () => this.onDragStop()),
```
olegat: Can someone clarify what's the intended behaviour of the navigator? Do we want this listener to receive drag & hover events when an animation is in progress?

packages/ag-charts-enterprise/src/features/zoom/zoom.ts
```diff
@@ -131,14 +133,13 @@ export class Zoom extends _ModuleSupport.BaseModuleInstance implements _ModuleSu
...
-            ctx.interactionManager.addListener('dblclick', (event) => this.onDoubleClick(event), interactionOpts),
-            ctx.interactionManager.addListener('drag', (event) => this.onDrag(event), interactionOpts),
-            ctx.interactionManager.addListener('drag-end', () => this.onDragEnd(), interactionOpts),
-            ctx.interactionManager.addListener('wheel', (event) => this.onWheel(event), interactionOpts),
-            ctx.interactionManager.addListener('hover', () => this.onHover(), interactionOpts),
+            ctx.interactionManager.addListener('dblclick', (event) => this.onDoubleClick(event)),
+            ctx.interactionManager.addListener('drag', (event) => this.onDrag(event), dragState),
+            ctx.interactionManager.addListener('drag-end', () => this.onDragEnd(), dragState),
+            ctx.interactionManager.addListener('wheel', (event) => this.onWheel(event)),
+            ctx.interactionManager.addListener('hover', () => this.onHover()),
```
olegat: Can someone clarify what's the intended behaviour of the zoom module? Do we want this listener to receive drag, wheel & hover events when an animation is in progress?

alantreadway: I believe so, and any state changes should cause the animation to skip. @lsjroberts thoughts?

alantreadway (approved):
Generally LGTM - let's get input from @lsjroberts and @iMoses before merging.

packages/ag-charts-enterprise/src/features/zoom/zoom.ts
lsjroberts: Yes I'd agree on this for zoom and navigator. If the user zooms or navigates the animation should skip to the end. That is the current behaviour.

packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts
```diff
@@ -80,10 +79,13 @@ export class ContextMenu extends _ModuleSupport.BaseModuleInstance implements _M
         // Module context
         this.highlightManager = ctx.highlightManager;
         this.interactionManager = ctx.interactionManager;
-        this.tooltipManager = ctx.tooltipManager;
         this.scene = ctx.scene;
 
-        this.destroyFns.push(ctx.interactionManager.addListener('contextmenu', (event) => this.onContextMenu(event)));
+        const contextState = InteractionState.Default & InteractionState.ContextMenu;
```
lsjroberts: I may be wrong, but I believe this should be `Default | ContextMenu`?

`Default & ContextMenu === 0` so will never trigger, right?

packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts
```diff
@@ -26,8 +26,9 @@ export type ContextMenuActionParams = {
     event: MouseEvent;
 };
 
+const InteractionState = _ModuleSupport.InteractionState;
+
 const { BOOLEAN, Validate } = _ModuleSupport;
```
iMoses:
```suggestion
const { BOOLEAN, InteractionState, Validate } = _ModuleSupport;
```
nit

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
```diff
@@ -147,12 +143,31 @@ export class InteractionManager extends BaseManager<
...
+        this.updateCurrentState();
+    }
+
+    public popState(state: InteractionState) {
+        this.stateQueue.delete(state);
+        this.updateCurrentState();
+    }
+
+    private updateCurrentState() {
+        this.state = Array.from(this.stateQueue).reduce((max, s) => Math.max(max, s), InteractionState.Default);
```
iMoses:
```suggestion
        this.state = Math.max(...this.stateQueue, InteractionState.Default);
```
nit: I personally think it's easier to read without using `reduce`.

do take into account that this forces the interaction manager to have only one state, while in reality using bitwise flags could mean that if we wanted to we could be in multiple states at once. just something to think about, not sure if it even makes sense in this context.

packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts
```diff
@@ -186,6 +172,16 @@ export class ContextMenu extends _ModuleSupport.BaseModuleInstance implements _M
         this.disabledActions.add(actionId);
     }
 
+    private isShown(): boolean {
+        return this.menuElement !== undefined;
```
iMoses, line 175: nit: 
```suggestion
        return this.menuElement != null;
```
nit: I think we should avoid using the `undefined` keyword as a value.

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
```diff
@@ -147,12 +143,31 @@ export class InteractionManager extends BaseManager<
...
     }
```
iMoses:
```suggestion
    public pushState(state: InteractionState) {
        this.state |= state;
    }

    public popState(state: InteractionState) {
        this.state &= ~state;
    }
```

Going over the entire PR again, I wonder if we really need the `stateQueue` property. I think that since we're using bitwise flags it's unnecessary.
It does however means that it's better if we set the `Default` state to `0` because technically `Default` should mean "no specific state was chosen".

packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts
alantreadway: Fixed on branch,

alantreadway (changes requested):
After some preliminary testing, there are a few gaps which I've documented in JIRA which we should address prior to merge.

packages/ag-charts-enterprise/src/features/zoom/zoom.ts
olegat: Done. Zoom & navigation listen for some events during animation.

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
olegat: The `stateQueue` is necessary to facilitate coordination. Intuitively, we might want to use simple setters & getters instead, but imagine this possible use-case:

1. State is initialised to `Default`.
2. AnimationManager starts and sets the state to `Animation`.
3. During the animation, a `'content-menu'` event is fired.
4. The context menu module hears this event, sets the state to `ContextMenu` and the `AnimationManager` to stop.
5. AnimationManager stops and removes the `Animation` state.

The key challenge is that after step 5, we want the state to be `ContextMenu` and not `Default`. The AnimationManager would need extra logic to figure out which state to transition to when the animation is completed or canceled.

Using the `InteractionState` values as priority number simplifies this problem significantly, because we just need to calculate the maximum to figure out the current state.

That said however, I agree that using a `Set` is unnecessary. We could just store the same information of `stateQueue` in a `number`.

olegat: `Default` cannot be `0` because if it is then listeners will not be able to specify whether or not they want to watch this state. Example:

```
const { Default, Animation } = InteractionState;
interactionManager.addListener('hover', onHover, Default);
interactionManager.addListener('drag', onHover, Default & Animation);
```

For `'hover'` events, the bitwise & on `Default` will always be 0 if `Default=0`, so the `onHover` listener will never receive events.

packages/ag-charts-community/src/chart/navigator/navigator.ts
olegat: Done. These navigator listeners receive events during animation.

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
olegat: Done. The `stateQueue` is now a bitfield rather than a `Set`.

alantreadway (approved):
Thanks, LGTM!

PR #908, opened 2024-01-25
Numbers util methods extension

iMoses:


packages/ag-charts-community/src/chart/axis/axis.ts
```diff
@@ -26,7 +26,7 @@ import { Selection } from '../../scene/selection';
 import { Line } from '../../scene/shape/line';
 import type { TextSizeProperties } from '../../scene/shape/text';
 import { Text, measureText, splitText } from '../../scene/shape/text';
-import { jsonDiff } from '../../sparklines-util';
+import { findMinMax, findRangeExtent, jsonDiff } from '../../sparklines-util';
```
alantreadway: Nit: `sparklines-util` is an export bundle, we shouldn't import from it (or the other bundles at that level) internally in `ag-charts-community`.

alantreadway (approved):
LGTM, could you fix the one nit and merge?

packages/ag-charts-community/src/chart/axis/axis.ts
iMoses: Good catch! I may be trusting the auto importer too much 😅

PR #910, opened 2024-01-26
AG-10483 - Fix consideration of `seriesGrouping` for series recreation cases.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-10483

Fixes for Integrated Charts:
- switching between bar-series types causing an unexpected 'series removal' animation, rather than immediately initiating an 'initial load' animation.
- seriesRect translation animation state is reset (so there is no animation of the seriesRect location during the series initial load animation).
- skipping of animation on layout changes (Chart.animationRect is now reset on series changes, which caused skipping here: https://github.com/ag-grid/ag-charts/blob/db3dced5aeca0f9f2abcb634e33eee8a069d499b/packages/ag-charts-community/src/chart/chart.ts#L1004)

### DIFF
--- a/packages/ag-charts-community/src/chart/agChartV2.ts
+++ b/packages/ag-charts-community/src/chart/agChartV2.ts
@@ -464,7 +464,7 @@ function applySeries(chart: Chart, options: AgChartOptions) {
         return false;
     }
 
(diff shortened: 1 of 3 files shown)


lsjroberts (approved):
lgtm

PR #918, opened 2024-01-29
AG-10524 Avoid firing two events when legend is clicked

olegat:
https://ag-grid.atlassian.net/browse/AG-10524

Before this change, the 'click' events were dispatched first to the chart and then to the legend.

The legend.ts file consumes the click events, but this consume() call can only affects the global chart if the legend listeners receive the 'click' message.

After this change, the 'click' events get dispatched first to the legend and then to the chart.


alantreadway (approved):
LGTM

PR #922, opened 2024-01-30
 Allow example generator to render multiple charts in one example

jacobp100:
https://ag-grid.atlassian.net/browse/AG-10020

Some changes were needed for this to happen:-

* Any div with an ID is considered a chart - some examples had to change to work around this
* The container must be written as `container: document.getElementById('a string literal here')`
* Multi-examples don't support methods
* We've now diverged quite a bit from the grid example generator - previously the options had to be called `options`, and that would match the property passed in as a prop (i.e. `<AgCharts options={options} />`) - but now you can call it whatever you like. This likely can't be back-ported to the grid


alantreadway (approved):
LGTM

PR #925, opened 2024-01-31
AG-10546 + AG-10547 - Dark mode/theme context menu and overlay support.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-10546
https://ag-grid.atlassian.net/browse/AG-10547

Adds dark-theme detection for overlays and context menu; this activates an extra CSS class to enable overrides of specific styles (background + text color typically).

I've laid the groundwork for tooltips too, but the logic is much more complicated due to user overrides and series overrides, so we may need to evaluate the priority of that.

### DIFF
--- a/packages/ag-charts-community/src/chart/overlay/overlay.ts
+++ b/packages/ag-charts-community/src/chart/overlay/overlay.ts
@@ -19,7 +19,7 @@ export class Overlay {
     text?: string;
 
     @Validate(BOOLEAN)
-    darkMode = false;
+    darkTheme = false;
 
     show(rect: BBox) {
         if (!this.element) {
@@ -27,11 +27,7 @@ export class Overlay {
             this.element.className = this.className;
         }
 
-        if (this.darkMode) {
-            this.element.classList.add(`${this.className}-dark`);
-        } else {
-            this.element.classList.remove(`${this.className}-dark`);
-        }
+        this.element.classList.toggle(`${this.className}-dark`, this.darkTheme);
 
         const { element } = this;
         element.style.position = 'absolute';
--- a/packages/ag-charts-community/src/chart/overlay/overlay.ts
+++ b/packages/ag-charts-community/src/chart/overlay/overlay.ts
@@ -1,6 +1,6 @@
 import { ADD_PHASE, REMOVE_PHASE } from '../../motion/animation';
 import type { BBox } from '../../scene/bbox';
-import { FUNCTION, STRING, Validate } from '../../util/validation';
+import { BOOLEAN, FUNCTION, STRING, Validate } from '../../util/validation';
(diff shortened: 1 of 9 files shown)


packages/ag-charts-community/src/chart/tooltip/tooltip.ts
```diff
@@ -136,12 +136,12 @@ const defaultTooltipCss = `
 }
 
 .${DEFAULT_TOOLTIP_CLASS}.${DEFAULT_TOOLTIP_DARK_CLASS} {
-    color: white;
-    background: #15181c;
+    // color: white;
```
lsjroberts: nit: technically this is invalid css, should be `/* color: white; */`

manapeirov: do we want to delete these?

packages/ag-charts-community/src/chart/overlay/overlay.ts
```diff
@@ -18,12 +18,21 @@ export class Overlay {
...
+    @Validate(BOOLEAN)
+    darkMode = false;
+
     show(rect: BBox) {
         if (!this.element) {
             this.element = this.createElement('div');
             this.element.className = this.className;
         }
 
+        if (this.darkMode) {
```
manapeirov: for these if else statements maybe we could use this one line instead? `this.element.classList.toggle(`${this.className}-dark`, this.darkMode);`

packages/ag-charts-community/src/chart/overlay/overlay.ts
```diff
@@ -18,12 +18,21 @@ export class Overlay {
     @Validate(STRING, { optional: true })
     text?: string;
 
+    @Validate(BOOLEAN)
+    darkMode = false;
```
manapeirov: `darkTheme`

manapeirov (approved):
LGTM thanks for this

packages/ag-charts-community/src/chart/tooltip/tooltip.ts
alantreadway: Removed

alantreadway: Removed

packages/ag-charts-community/src/chart/overlay/overlay.ts
alantreadway: Good call, done

packages/ag-charts-community/src/chart/overlay/overlay.ts
alantreadway: Renamed here and elsewhere

manapeirov (approved):

PR #926, opened 2024-01-31
AG-10428 Add throttling to lazy load callback

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10428

Throttles the lazy load callback to once every 100ms at the trailing edge.

I tested out a few options to see how they feel, and this seemed the best. But I've kept the other util functions as they will likely prove useful in the future.

I looked at using the scheduler from the render sequence, but it has different usage.

Note: When the throttle time is shorter than the request time, requests can come back out of order or on top of other throttled requests. So this can look a bit janky. This will be fixed by https://ag-grid.atlassian.net/browse/AG-10429


alantreadway (approved):
LGTM

PR #927, opened 2024-01-31
AG-10525 Fix tooltip positioning

olegat:
https://ag-grid.atlassian.net/browse/AG-10525

The bug was happening only when the enterprise package was used. It happened on line/scatter series when error bars were disabled.

This happened because ErrorBars.getDatum was adding an `undefined` and number value together, which apparently results in NaN. So the module was creating error bars with NaN translations.

To safe-guard against this, ErrorBars.getDatum now explicitly returns `number | undefined` types instead of `any` types.

Additionally, ErrorBars.createNodeData() has been modified to avoid creating error bar node-data when error bars are disabled.

### DIFF
--- a/packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
+++ b/packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
@@ -1,5 +1,5 @@
 import type { AgErrorBarThemeableOptions, _Scale } from 'ag-charts-community';
-import { AgErrorBarSupportedSeriesTypes, _ModuleSupport, _Scene } from 'ag-charts-community';
+import { AgErrorBarSupportedSeriesTypes, _ModuleSupport, _Scene, _Util } from 'ag-charts-community';
 
 import type { ErrorBarNodeDatum, ErrorBarStylingOptions } from './errorBarNode';
 import { ErrorBarGroup, ErrorBarNode } from './errorBarNode';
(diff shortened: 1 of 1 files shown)


packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
```diff
@@ -242,6 +247,27 @@ export class ErrorBars extends _ModuleSupport.BaseModuleInstance implements _Mod
...
+        }
+        const value: unknown = datum.datum[key];
+        if (value === undefined) {
+            return undefined;
+        }
+
+        // The datum has an error value for `key`. Validate this user input value:
+        if (typeof value !== 'number') {
+            _Util.Logger.warnOnce(
+                `Found [${key}] error value (${value}) of type ${typeof value}. Expected number type`
```
alantreadway: Nit: I would not include the raw `value` in the message, as the deduplication effect of `warnOnce` is likely to be negated by different non-numeric numbers? (`typeof value` seems sufficient)

alantreadway (commented):
LGTM aside from one nit

packages/ag-charts-enterprise/src/features/error-bar/errorBar.ts
olegat: Done. I've removed the `(${value})` part of the warning message.

alantreadway (approved):
LGTM

PR #929, opened 2024-01-31
Misc cleanup

iMoses:


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -213,52 +215,32 @@ export abstract class Chart extends Observable implements AgChartInstance {
...
-        return this._seriesArea;
-    }
-    set seriesArea(newArea: SeriesArea) {
-        if (!newArea) {
-            this._seriesArea = new SeriesArea();
-        } else {
-            this._seriesArea = newArea;
-        }
-    }
+    @Default(() => new SeriesArea())
```
alantreadway: Nit: Do we need this and the inline assignment on the next line?

packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -730,79 +715,68 @@ export abstract class Chart extends Observable implements AgChartInstance {
...
             series.removeEventListener('nodeDoubleClick', this.onSeriesNodeDoubleClick);
             series.destroy();
 
             series.chart = undefined;
         });
-        this._series = []; // using `_series` instead of `series` to prevent infinite recursion
-
-        // Reset animation state.
-        this.animationRect = undefined;
-        this.animationManager.reset();
```
alantreadway, line 805: 🚨 This was added last week to correctly handle Integrated Charts animation scenarios (specifically, reset initial load animation on series changes) - we can't delay this until `Chart.destroy()` because that doesn't always get invoked (we attempt to re-use the `Chart` instance whenever possible)

packages/ag-charts-community/src/util/sizeMonitor.ts
```diff
@@ -13,76 +13,55 @@ export class SizeMonitor {
...
-        } else {
-            // polyfill (more reliable even in browsers that support ResizeObserver)
-            const step = () => {
-                this.elements.forEach((entry, element) => {
-                    this.checkClientSize(element, entry);
-                });
-            };
-            this.pollerHandler = document.defaultView?.setInterval(step, 100);
-        }
+        if (typeof ResizeObserver === 'undefined') return;
```
alantreadway: I have some concerns that we still need the fallback behaviour.

iMoses: How come? It's been widely support for about 4 years now. The only use case we have in which it doesn't exist is running server-side IMO.

packages/ag-charts-community/src/chart/chart.ts
iMoses: You may be right, I'll test it.

packages/ag-charts-community/src/chart/chart.ts
iMoses: So we want to reset the animation manager on both the destruction of a chart OR when replacing the series array?

packages/ag-charts-community/src/chart/chart.ts
iMoses: it seems to be need because `Default` only triggers on `set`.

packages/ag-charts-community/src/util/sizeMonitor.ts
alantreadway: Discussed offline

packages/ag-charts-community/src/chart/chart.ts
alantreadway: Resolved offline

alantreadway (approved):
LGTM

PR #930, opened 2024-01-31
AG-10544 Fix continuous scale bandwidth calculation for range offset from origin

jacobp100:
https://ag-grid.atlassian.net/browse/AG-10544


alantreadway (approved):
LGTM

PR #933, opened 2024-02-01
AG-10537 - Fix range bar update always triggering initial load.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-10537

Fixes range-bar updates via `AgChart.update()` always triggering an animation reset; this was due to `direction` being unset in the default config, so a difference in this property was always detected.

Bonus: Cleaned up `AgChartV2.applySeries()` a bit, to make logging of mismatching properties quicker to identify.


manapeirov (approved):
LGTM!

PR #936, opened 2024-02-01
AG-10477 Do not clip lines on the left during swipe-in animations

olegat:
https://ag-grid.atlassian.net/browse/AG-10477


alantreadway (approved):
LGTM

PR #937, opened 2024-02-01
AG-10426 Change lazy load to delay requesting until axes are ready

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10426


alantreadway (approved):
LGTM

PR #938, opened 2024-02-01
AG-10528 - Rework factory series matching

alantreadway:
https://ag-grid.atlassian.net/browse/AG-10528

This is a prerequisite change to being able to support series add/remove using the add/update/remove animations (aka legend-toggle) - we need to be able to match and apply series options not based on array index, this change attempts to compare our pre-existing matching properties to match up options with existing series.

Bonus:
- Fixes a recently introduced bug with any `Chart.series[]` change leading to incorrect call of all `Series.destroy()` methods, including those we want to keep alive.


packages/ag-charts-community/src/chart/agChartV2.ts
```diff
@@ -458,67 +459,55 @@ function applyModules(chart: Chart, options: AgChartOptions) {
 function applySeries(chart: Chart, options: AgChartOptions) {
     const optSeries = options.series;
     if (!optSeries) {
-        return false;
+        return 'no-change';
```
lsjroberts, line 464: I'd be inclined to make these an enum

packages/ag-charts-community/src/chart/series/series.ts
```diff
@@ -480,6 +480,7 @@ export abstract class Series<
 
     destroy(): void {
         this.destroyFns.forEach((f) => f());
+        this.destroyFns.length = 0;
```
iMoses:
```suggestion
        this.destroyFns = [];
```
nit: this is a hacky way to reset an array, I think we should avoid doing it. 
It's an open door to memory leaks because the values aren't really removed they're simply not being iterated.

lsjroberts (commented):
Looks good but yeah, there's enough string consts going on that I think an enum would be nicer

packages/ag-charts-community/src/chart/agChartV2.ts
alantreadway: I don't see the benefit of an enum for these cases?

alantreadway: Discussed offline, I've added `SeriesChangeType` as an explicit type to cleanup use of `ReturnType`.

lsjroberts (approved):

PR #943, opened 2024-02-05
AG-10565/10567: Add an enterprise SyncModule

iMoses:
Missing validations but otherwise working


packages/ag-charts-community/src/chart/interaction/syncManager.ts
```diff
@@ -0,0 +1,44 @@
+import type { Chart } from '../chart';
+import { BaseManager } from './baseManager';
+
+type GroupId = string | symbol;
+
+export class SyncManager extends BaseManager {
+    static chartsGroups = new Map<GroupId, Set<Chart>>();
+    static DEFAULT_GROUP = Symbol('sync-group-default');
```
alantreadway, line 8: Minor nit: Any reason to not just make this a `const` outside the class definition?

packages/ag-charts-community/src/module-support.ts
```diff
@@ -38,6 +38,7 @@ export * from './chart/interaction/interactionManager';
 export * from './chart/interaction/gestureDetector';
 export * from './chart/interaction/tooltipManager';
 export * from './chart/interaction/zoomManager';
+export { type Chart } from './chart/chart';
```
alantreadway: I don't like exporting `Chart`, as it's interface contract is a bit too wild and unrestricted - could we introduced a narrower interface contract for sync-related methods which `Chart` implements?

packages/ag-charts-community/src/options/chart/chartOptions.ts
```diff
@@ -147,6 +154,8 @@ export interface AgBaseThemeableChartOptions<TDatum = any> {
     zoom?: AgZoomOptions;
     /** Configuration for the Navigator. */
     navigator?: AgNavigatorOptions;
+    /** Configuration for synchronizing multiple charts. */
+    sync?: AgChartSyncOptions;
```
alantreadway: Could we hold off adding this to the options contract until we have done a few iterations with David and agreed the options contract?

packages/ag-charts-enterprise/src/features/sync/chartSync.ts
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+
+    private enabledZoomSync() {
+        const { syncManager, zoomManager } = this.moduleContext;
+        this.disableZoomSync = zoomManager.addListener('zoom-change', (e) => {
+            if ('stopPropagation' in e && e.stopPropagation) return;
+            for (const chart of syncManager.getSiblings(this.groupId)) {
+                chart.zoomManager.updateZoom('zoom', zoomManager.getZoom());
+                (chart.zoomManager as any).listeners.dispatch('zoom-change', { ...e, stopPropagation: true });
```
alantreadway: Shouldn't `updateZoom()` percolate the change itself?

packages/ag-charts-enterprise/src/features/sync/chartSync.ts
```diff
@@ -0,0 +1,106 @@
...
+        const { syncManager } = this.moduleContext;
+        const chart = syncManager.getChart();
+        const syncSeries = syncManager.getGroup(this.groupId).flatMap((chart) => chart.series);
+
+        for (const axis of chart.axes) {
+            if (this.axes !== 'xy' && this.axes !== axis.direction) continue;
+            axis.boundSeries = syncSeries.filter((series) => {
+                const seriesKeys = series.getKeys(axis.direction);
+                return axis.keys.length ? axis.keys.some((key) => seriesKeys?.includes(key)) : true;
+            });
```
alantreadway, line 73: I have some concerns about the scalability of this, every chart axis instance has references to every other charts series?

packages/ag-charts-community/src/chart/interaction/syncManager.ts
iMoses: Matter of opinion, nothing more. I like that consts of specific classes are attached to them, creates sort of a namespace for the relevant consts. Downside is that you cannot import them directly.
You want me to change it?

packages/ag-charts-community/src/options/chart/chartOptions.ts
iMoses: I'll comment it out

packages/ag-charts-community/src/module-support.ts
iMoses: I think I ended up not using it anyhow. I'll remove it.

packages/ag-charts-enterprise/src/features/sync/chartSync.ts
iMoses: It is currently only used for calculating the domain, and there's nothing more to it.
I agree it's not ideal but I believe the solution is a refactor of the axis and chart implementation.

packages/ag-charts-enterprise/src/features/sync/chartSync.ts
iMoses: you're right. good catch 👍

packages/ag-charts-community/src/chart/interaction/syncManager.ts
alantreadway: :+1:

packages/ag-charts-enterprise/src/features/sync/chartSync.ts
alantreadway: Discussed offline, we should follow up with a refactor to try and decouple axis & scales, so we could avoid O(n^3) complexity in the domain calculation.

alantreadway (approved):
LGTM

PR #947, opened 2024-02-05
AG-10475 Change wheel deltaMode factor

olegat:
https://ag-grid.atlassian.net/browse/AG-10475


packages/ag-charts-community/src/chart/interaction/interactionManager.ts
```diff
@@ -361,7 +361,9 @@ export class InteractionManager extends BaseManager<InteractionTypes, Interactio
 
         let [deltaX, deltaY] = [NaN, NaN];
         if (this.isWheelEvent(event)) {
-            const factor = event.deltaMode === 0 ? 0.1 : 1;
+            // AG-10475 On Chrome (Windows), wheel clicks send deltaMode: 0 events with deltaY: -100 or +100.
+            // So we divide this by 100 to give us the desired step.
+            const factor = event.deltaMode === 0 ? 0.01 : 1;
```
alantreadway: Should we vary this behaviour, depending on whether we detect Windows as the host OS?

olegat: I don't know whether that's desirable. I believe that Windows does support smooth scrolling like on macOS if the correct hardware is installed, so checking the host OS seems like a bad idea.

What we really want to check is whether the hardware is a trackpad or a wheel, but I don't know how we can check this. I was under the impression that this is what `deltaMode` was meant for, but apparently `deltaMode` is always 0 but the `deltaY` is either constant (wheel) or variable (trackpad). Furthermore, the constant is browser dependant (±100 on Chrome, ±102 on Firefox).

This solution is far from ideal, but I don't see a better to solve this without a reliable way to distinguish between smooth and wheel scrolling.

PR #948, opened 2024-02-06
AG-10518 Update logic for hiding a donut series with no inner radius …

jacobp100:
…props

https://ag-grid.atlassian.net/browse/AG-10518

### DIFF
--- a/packages/ag-charts-community/src/chart/series/polar/donutSeries.ts
+++ b/packages/ag-charts-community/src/chart/series/polar/donutSeries.ts
@@ -310,7 +310,18 @@ export class DonutSeries extends PolarSeries<DonutNodeDatum, Sector> {
 
     async createNodeData() {
         const { id: seriesId, processedData, dataModel, angleScale } = this;
-        const { rotation } = this.properties;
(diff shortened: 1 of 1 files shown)


packages/ag-charts-community/src/chart/series/polar/donutSeries.ts
```diff
@@ -310,7 +310,18 @@ export class DonutSeries extends PolarSeries<DonutNodeDatum, Sector> {
...
-        const { rotation } = this.properties;
+        const { rotation, innerRadiusRatio, innerRadiusOffset } = this.properties;
+
+        if (innerRadiusRatio == null && innerRadiusOffset == null) {
+            Logger.warnOnce(
+                'Either an [innerRadiusRatio] or an [innerRadiusOffset] must be set to render a donut series.'
+            );
+
+            this.zerosumOuterRing.visible = true;
+            this.zerosumInnerRing.visible = true;
```
alantreadway, line 321: It might be better to add some flags to the result of this method, and apply this visibility change in a later update phase (ideally we wouldn't have side-effects in `processData()` and `createNodeData()` calls, defering updating the scene graph to later in the processing flow in the `update()` call).

PR #955, opened 2024-02-06
Fix theme defaults resolution

iMoses:


packages/ag-charts-enterprise/src/series/bullet/bulletDefaults.ts
```diff
@@ -9,7 +9,6 @@ export const BULLET_DEFAULTS = {
             position: POSITION.LEFT,
             nice: false,
             max: undefined as number | undefined,
```
alantreadway: Should these move too?

alantreadway (approved):
LGTM

packages/ag-charts-enterprise/src/series/bullet/bulletDefaults.ts
iMoses: It's the only place I wasn't sure about. WDYT?

PR #957, opened 2024-02-07
AG-10514 - Fix BarSeries legend toggle using stale from position.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-10514

Fixes BarSeries legend toggle animation for Integrated Charts cases, and any case where a resize happens between legend toggles.

The root cause was that because `CartesianSeries.updateSelections()` was not running for hidden series, the `Node.datum` was left in a stale state, meaning when animating a visibility toggle that bars scaled in from stale X/Y coordinates.


lsjroberts (approved):
LGTM

PR #956, opened 2024-02-06
AG-10412 AG-10426 Lazy loading fixes

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10412
https://ag-grid.atlassian.net/browse/AG-10426

- Fix data window processor to ignore unchanged axis domains
- Fix data window processor to map invalid dates to undefined
- Update lazy loading to skip animations

PR #959, opened 2024-02-07
AG-10430 Add overlay when lazy loading data

lsjroberts:

This is targeting https://github.com/ag-grid/ag-charts/pull/956.

### DIFF
--- a/packages/ag-charts-community/src/chart/overlay/chartOverlays.ts
+++ b/packages/ag-charts-community/src/chart/overlay/chartOverlays.ts
@@ -1,3 +1,4 @@
+import { ADD_PHASE } from '../../motion/animation';
 import { injectStyle } from '../../util/dom';
 import type { AnimationManager } from '../interaction/animationManager';
 import { DEFAULT_OVERLAY_CLASS, DEFAULT_OVERLAY_DARK_CLASS, Overlay } from './overlay';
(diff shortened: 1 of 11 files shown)


packages/ag-charts-community/src/chart/overlay/chartOverlays.ts
```diff
@@ -26,11 +32,55 @@ export class ChartOverlays {
...
+
+        const label = this.createElement(container, 'div');
+        label.style.marginTop = '1em';
+        label.innerText = this.loading.text ?? 'Loading data...';
+
+        const animationStyles = this.createElement(container, 'style');
+        animationStyles.innerText = [
+            '@keyframes ag-charts-loading { from { opacity: 0 } to { opacity: 1 } }',
+            '@keyframes ag-charts-loading-spinner { to { transform: rotate(0.5turn); } }',
+        ].join('');
```
lsjroberts, line 76: This uses css animations rather than the animation manager as the overlay api requires a `string` not `HTMLElement`.

alantreadway: SGTM, let's make use of CSS to do the heavy-lifting where possible.

packages/ag-charts-community/src/chart/update/overlaysProcessor.ts
```diff
@@ -0,0 +1,41 @@
+import type { BBox } from '../../integrated-charts-scene';
```
alantreadway: Nit: don't import top-level bundles.

alantreadway (commented):
LGTM, one nit

packages/ag-charts-community/src/chart/update/overlaysProcessor.ts
lsjroberts: ah doh, auto import got me there

alantreadway (approved):
LGTM

PR #960, opened 2024-02-07
AG-10518 Hide legend on donut with no inner radius properties

jacobp100:
https://ag-grid.atlassian.net/browse/AG-10518

### DIFF
--- a/packages/ag-charts-community/src/chart/series/polar/donutSeriesProperties.ts
+++ b/packages/ag-charts-community/src/chart/series/polar/donutSeriesProperties.ts
@@ -92,6 +92,8 @@ class DonutSeriesCalloutLine extends BaseProperties {
 
 export class DonutSeriesProperties extends SeriesProperties<AgDonutSeriesOptions> {
     override isValid(): boolean {
+        const superIsValid = super.isValid();
+
         if (this.innerRadiusRatio == null && this.innerRadiusOffset == null) {
             Logger.warnOnce(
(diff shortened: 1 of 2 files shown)


packages/ag-charts-community/src/chart/series/polar/donutSeriesProperties.ts
```diff
@@ -90,6 +91,20 @@ class DonutSeriesCalloutLine extends BaseProperties {
 }
 
 export class DonutSeriesProperties extends SeriesProperties<AgDonutSeriesOptions> {
+    override isValid(): boolean {
+        if (!super.isValid()) {
+            return false;
+        } else if (this.innerRadiusRatio == null && this.innerRadiusOffset == null) {
```
iMoses: nit: you don't need `else` if the previous `if` `return`s :)

jacobp100: I tend to do it because it saves a line - but can change if that’s the consensus

alantreadway: Hmmm... isn't the logic here wrong anyway, shouldn't we always perform the validations here?

jacobp100: Updated

alantreadway (approved):
LGTM

PR #963, opened 2024-02-07
AG-10409 TooltipManager cleanup

olegat:

Avoid caching event objects in `Chart` and `TooltipMeta`. The only things that's need is `offsetX` and `offsetY`


alantreadway (approved):
LGTM

PR #974, opened 2024-02-09
AG-10614 Move minichart to enterprise module

jacobp100:
https://ag-grid.atlassian.net/browse/AG-10614


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1435,10 +1435,13 @@ export abstract class Chart extends Observable implements AgChartInstance {
             this.zoomManager.updateZoom();
         }
 
-        const miniChartInstance = navigatorModule?.miniChartInstance;
-        if (miniChartInstance != null) {
-            const seriesStatus = this.applySeries(miniChartInstance, processedOptions);
-            this.applyAxes(miniChartInstance, processedOptions, seriesStatus);
+        const navigator = this.modules.get('navigator') as any;
```
iMoses: notice you already have the `navigatorModule` 8 lines up.

jacobp100: Updated

packages/ag-charts-community/src/chart/factory/expectedEnterpriseModules.ts
```diff
@@ -26,6 +26,7 @@ export const EXPECTED_ENTERPRISE_MODULES: EnterpriseModuleStub[] = [
         chartTypes: ['cartesian', 'polar', 'hierarchy'],
         identifier: 'gradient',
     },
+    { type: 'root', optionsKey: 'miniChart', chartTypes: ['cartesian'], identifier: 'mini-chart' },
```
alantreadway: Nit: Drop `identifier`, it's mostly for series and axes cases where a `type` discriminator is involved (see other `'root'` cases above where we don't specify this too).

alantreadway: I'm debating if we should add a `type: 'navigator'` case, as this isn't technically an (options) root module - could we do that relatively easily?

alantreadway (approved):
LGTM

PR #976, opened 2024-02-09
AG-10429 Add handling of parallel and stale lazy load requests

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10429

Also refactors it to use the other throttling method and an event model instead of awaiting (internally, exposed api is still a promise).


alantreadway (approved):
LGTM

PR #978, opened 2024-02-09
AG-10409 Add new region-based input handling

olegat:

Currently only the legend uses this new design as a PoC. The remaining interaction listeners will be migrated once we are satisfied with this.


packages/ag-charts-community/src/chart/interaction/interactionManager.ts
```diff
@@ -167,6 +190,37 @@ export class InteractionManager extends BaseManager<InteractionTypes, Interactio
...
     }
 
+    private makeRegion(name: RegionName): InteractionRegion {
+        return {
+            name,
+            listeners: new Listeners<InteractionTypes, (event: InteractionEvent<InteractionTypes>) => void>(),
+        };
+    }
+
+    public addRegion(name: RegionName, node: BBoxProvider): ObservableRegion {
```
alantreadway: I think this is complex enough to warrant living distinctly from `InteractionManager`? `RegionManager` as a concept seems good to me?

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
```diff
@@ -167,6 +190,37 @@ export class InteractionManager extends BaseManager<InteractionTypes, Interactio
...
+        const region = this.makeRegion(name);
+        const interactionManager = this;
+        this.regions.add(region, node);
```
alantreadway: How should region bboxes get updated - the legend position could change on every re-layout, so we need to be able to update the bbox positions frequently in common scenarios?

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
lsjroberts: Yeah I'd agree with this, it would be good to keep the interaction manager focused on just simple interaction events as much as possible.

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
olegat: I don't think that the region manager needs to worry about that because `node: BBoxProvider` is just an interface for a scene `Node`. It accesses the bounding box through `computeBBox()` and the scene-graph manages update/caching of these boxes based on the nodes' dirty states.

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
olegat: ACK.

olegat: I can move some functionality into a new `RegionManager` class, but the internal wrapper class in `InteractionManager.addRegion` needs to stay because it needs to know the exact type of the handler. Moving that wrapper class into `RegionManager` would require us to move types like `InteractionTypes` and `InteractionHandler<T>` into a new file to avoid circular imports.

olegat:
Note: This PR adds & reverts changes to several files such as _interactionManager.ts_. To preserve git-blame, I recommend squashing this PR instead of merging.

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
```diff
@@ -4,17 +4,20 @@ import { Logger } from '../../util/logger';
...
-    | 'dblclick'
-    | 'contextmenu'
-    | 'hover'
-    | 'drag-start'
-    | 'drag'
-    | 'drag-end'
-    | 'leave'
-    | 'page-left'
-    | 'wheel';
+export const InteractionTypesArray = [
```
alantreadway: Nit: As this isn't a type, maybe rename to `INTERACTION_TYPES` (`Array` seems redundant)?

packages/ag-charts-community/src/chart/legend.ts
```diff
@@ -231,10 +231,12 @@ export class Legend extends BaseProperties {
         this.item.marker.parent = this;
 
         const animationState = InteractionState.Default | InteractionState.Animation;
+        const region = ctx.regionManager.addRegion('legend', this.group);
```
alantreadway, line 234: Side-note: `Group`s don't have a defined `BBox`, it's aways derived by calculating the super-set `BBox` for all of its descendants - we may want to do something to optimise/cache the `BBox` calculation here.

alantreadway (approved):
This is starting to look good :tada: Let's get feedback from others too, but LGTM if you can address the one nit :+1:

lsjroberts (approved):
Looks great.

It's a lot nicer than the `pointerInsideLegend` stuff.

Would the intention in the future be to extract `Legend.getDatumForPoint()` out into nested regions for each legend item?

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
iMoses: these are the places where I wonder if we're not abusing the language a bit just do to what enums provide for free. any objections about using an enum here?

olegat: Generally I prefer strings than enums for several reasons:
1. Code readability: enums tend to be more verbose (`InteractionTypes.Leave` vs. `'leave'`)
2. Strings are more debuggable because the debugger just shows enums as numbers.
3. String unions typing is (apparently) more strictly enforced.

In fact, I just quickly tested out enums and it looks like the compiler does not distinguish between numbers and enums at all:

```typescript
enum MyTypeEnum {
    One = 1,
    Two = 2,
}

type RankedType<T extends MyTypeEnum> = {
    type: T;
    rank: number;
};

const myRank: RankedType<MyTypeEnum> = { type: 8, rank: MyTypeEnum.One };
```

This example replicates a realistic scenario where the programmer was accidentally swapped around the `type` and `rank` values. However, this code compiles just fine despite the fact that `8` is not a valid `MyTypeEnum`. This is definitely undesirable.

packages/ag-charts-community/src/chart/legend.ts
olegat: Hmm. I was under the impression that the `Group` class would only recalculate the `BBox` if the translation is dirty. If that's not the case, then we should certainly consider improving the scene-graph to cache this calculation.

In any case, I don't think that this is a blocker for the PR. The current implementation of the legend already relies on `Group.computeBBox()`, so we shouldn't see any significant performance regression.

olegat:
> Looks great.
> 
> It's a lot nicer than the `pointerInsideLegend` stuff.
> 
> Would the intention in the future be to extract `Legend.getDatumForPoint()` out into nested regions for each legend item?

Yes I agree. There was some discussion around using the scene-graph's hierarchy v.s. writing a new tree-structure for this. Until a decision is made, I've kept the region look-up code as simple as possible (just looping through an array of b-boxes). But nested regions would certainly be the next thing to consider 👍

packages/ag-charts-community/src/chart/legend.ts
alantreadway: Agree, no need to fix immediately - but once we get to the series-area case we will need it to start to address performance concerns :+1:

PR #984, opened 2024-02-12
AG-10409 Listen for legend click events during animation

olegat:
This resolves a bug where the 'dblclick' events would not get broadcasted to the legend's listener.


alantreadway (approved):
LGTM

PR #987, opened 2024-02-13
AG-9928 Fix legend cursor updates during animation

olegat:
https://ag-grid.atlassian.net/browse/AG-9928

Depends on https://github.com/ag-grid/ag-charts/pull/978


olegat:
Note: This PR includes a merge commit to resolve code-review conflicts from https://github.com/ag-grid/ag-charts/pull/978.

I recommend using the squash & merge mode for this PR

packages/ag-charts-community/src/chart/legend.ts
```diff
@@ -897,8 +898,24 @@ export class Legend extends BaseProperties {
 
     private handleLegendMouseExit(_event: InteractionEvent<'leave'>) {
         this.ctx.cursorManager.updateCursor(this.id);
-        this.ctx.highlightManager.updateHighlight(this.id);
         this.ctx.tooltipManager.removeTooltip(this.id);
+        // Updating the highlight can interrupt animations, so only clear the highlight if the chart
+        // is in a state when highlighting is possible.
+        if (this.ctx.interactionManager.getState() === InteractionState.Default) {
```
alantreadway, line 904: I think this might be the wrong way around to deal with this.

Specifically, if I move the pointer from one legend item to another during an animation (and the pointer is stationary at the end of the animation), I would expect that the highlight should be applied immediately after the animation.

The proposed approach loses the fact the highlight is changed, due to the events effectively being dropped.

We probably need to be able to buffer the most recent event, and once the `InteractionState` returns to `Default` then emit it.

PR #988, opened 2024-02-13
Fixes to minichart when updating data, disables legend events for minichart series

jacobp100:
https://ag-grid.atlassian.net/browse/AG-10675
https://ag-grid.atlassian.net/browse/AG-10653


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1647,14 +1659,30 @@ export abstract class Chart extends Observable implements AgChartInstance {
...
         }
 
+        const oldGrouping = target.seriesGrouping;
         if (seriesGrouping) {
             target.seriesGrouping = { ...target.seriesGrouping, ...(seriesGrouping as SeriesGrouping) };
         }
+
+        // This probably shouldn't be in this part of the flow
+        // However, the chart won't have access to the previous series grouping value to do this
+        if (applyLayers) {
```
jacobp100: I feel this definitely shouldn't exist here - but there's no way for a series to communicate to a chart that its grouping has changed. Maybe something like this:-

```ts
interface SeriesDelegate<T> {
    seriesGroupingDidChange: (series: Series<T>, oldValue?: SeriesGrouping) => void;
}

class Series<T> {
    delegate: SeriesDelegate?

      @ActionOnSet<Series<TDatum, TLabel>>({
        changeValue: function (newVal, oldVal) {
            this.onSeriesGroupingChange(oldVal, newVal);
            this.delegate?.seriesGroupingDidChange(this, oldVal)
        },
    })
    seriesGrouping?: SeriesGrouping = undefined;
}
```

Then the chart just has to implement `SeriesDelegate` and assign itself as the delegate to each series

Thoughts?

iMoses: From conversations I has with Alan it seems that we are heading towards using a centralized events-hub, which we'd like to use consistently as a mean of communication between modules as it promotes decoupling and provides easy access points to react to in future development.
So I'd go with managing it using events...

packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1405,10 +1405,13 @@ export abstract class Chart extends Observable implements AgChartInstance {
...
         let seriesStatus: SeriesChangeType = 'no-op';
         if (processedOptions.series && processedOptions.series.length > 0) {
-            seriesStatus = this.applySeries(this, processedOptions);
+            seriesStatus = this.applySeries(this, processedOptions, oldOpts, true);
             forceNodeDataRefresh = true;
         }
-        if (this.applyAxes(this, completeOptions, seriesStatus)) {
+        if (seriesStatus === 'replaced') {
+            this.resetAnimations();
+        }
```
alantreadway, line 1435: We can't do this as it impacts Integrated Charts use-cases.

packages/ag-charts-community/src/chart/chart.ts
alantreadway: Let sync on the need for this change, it's quite high risk and impacts correctness of stacking, grouping and Integrated Charts use cases?

packages/ag-charts-community/src/chart/chart.ts
alantreadway: Discussed offline: Moved from elsewhere, so this is a non-issue.

alantreadway (approved):
LGTM

PR #991, opened 2024-02-13
AG-9794 Fix listener cursor not clearing when moving outside series area

olegat:
https://ag-grid.atlassian.net/browse/AG-9794

Blocked by https://github.com/ag-grid/ag-charts/pull/978


alantreadway:
Looks risky regarding the use of  the `Group` of `seriesArea`, given the `BBox` of `Group`s aren't cached - let's consider this post-release.

PR #998, opened 2024-02-13
AG-10433 Add data source example to zoom docs

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10433
https://ag-grid.atlassian.net/browse/AG-10432

Adds a `dataSource` example to the zoom docs.

Also fixes a bug with lazy load erratically appearing, presumably due to an update conflict since it was attached to `layout-complete`, when it was effectively occuring at the `layout-start` stage previously.

Also adds a translucent background to the lazy load overlay after feedback (with a minimal overlay example on the overlays docs).

Also improves a few issues with the zoom docs.


alantreadway (approved):
LGTM

PR #999, opened 2024-02-13
AG-10630 - Fix grouped category axis with overlapping labels.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-10630

Fixes overlapping leaf-node label rendering when available band-width is smaller than the label height - we now render every N labels.

Bonus:
- Splits `grouped-category` axis tests into their own suite, adds a regression test-case for this bug.

Before fix:
Screenshot 2024-02-13 at 17 17 29

After fix:
Screenshot 2024-02-13 at 17 17 59


lsjroberts (approved):
LGTM

PR #1007, opened 2024-02-14
AG-10700: Add videos to docs

JamesSwinton:
https://ag-grid.atlassian.net/browse/AG-10700

Add VideoSection component to docs and add videos to Create a Basic Chart doc

### DIFF
--- a/packages/ag-charts-website/src/design-system/modules/VideoSection.module.scss
+++ b/packages/ag-charts-website/src/design-system/modules/VideoSection.module.scss
@@ -12,7 +12,7 @@
     }
 }
 
-.header > * {
+.videoSection .header > * {
     font-size: var(--text-fs-xl) !important;
 }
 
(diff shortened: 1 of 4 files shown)


packages/ag-charts-website/src/design-system/modules/VideoSection.module.scss
```diff
@@ -12,8 +12,8 @@
     }
 }
 
-.header {
-    font-size: var(--text-fs-lg);
+.header > * {
```
markdurrant: You can refactor this selector to get rid of the `!important`. 

```
.videoSection .header > * {
    font-size: var(--text-fs-xl);
}
```

markdurrant (commented):
Generally all looking good 👍 

Left a comment on one CSS niggle.

packages/ag-charts-website/src/design-system/modules/VideoSection.module.scss
JamesSwinton: Thanks, added this in

alantreadway (approved):
LGTM

PR #1011, opened 2024-02-14
AG-10647 Ensure showInMiniChart is only available for cartesian series

jacobp100:
https://ag-grid.atlassian.net/browse/AG-10647


packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts
```diff
@@ -237,6 +240,15 @@ export abstract class CartesianSeries<
         );
     }
 
+    override set visible(value) {
+        super.visible = value;
+    }
+
+    override get visible() {
+        const isInMiniChart = this.chart?.isMiniChart === true;
```
alantreadway: I don't like the asymmetry of setting `visible` vs. getting `visible` - could we just filter series options out just before calling `Chart.applySeries()`, so the series never exist in the first place?

jacobp100: Updated

alantreadway (approved):
LGTM

PR #1020, opened 2024-02-15
AG-10430 Update loading overlay style and dark theme

lsjroberts:

Also fixes a bug where the data window should not reload the data when the zoom hasn't changed. Previous logic was bust when refactored to the new api.

Also reverts the overlays processor back to using the `layout-complete` event. The original reason for the change was to fix the loading overlay sometimes not showing in rapid changes. But then the background of the loading overlay overlaps the axis tick labels, most notably visible when it stops at the end of the axis line and doesn't cover the full first or last label. The previous issue is hard to reproduce and isn't noticeable.


alantreadway (approved):
LGTM

PR #1022, opened 2024-02-15
AG-10426 Fix data window processor default window values

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10426


alantreadway (approved):
LGTM

PR #1025, opened 2024-02-15
AG-10677 Context menu not getting correct chart when multiple charts on page

olegat:
https://ag-grid.atlassian.net/browse/AG-10677

### DIFF
--- a/packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts
+++ b/packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts
@@ -13,18 +13,9 @@ type ContextMenuGroups = {
     extra: Array<ContextMenuItem>;
     extraNode: Array<ContextMenuItem>;
 };
+type ContextMenuAction = _ModuleSupport.ContextMenuAction;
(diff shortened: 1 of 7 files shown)


packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts
```diff
@@ -67,16 +59,14 @@ export class ContextMenu extends _ModuleSupport.BaseModuleInstance implements _M
 
     // Global shared state
     private static contextMenuDocuments: Document[] = [];
```
olegat, line 61: It's unclear to me whether we want to keep `contextMenuDocuments` as static or change it to non-static. AFAICT we want to keep it static because this variable is used to avoid adding CSS to the HTML document multiple times.

alantreadway: `static` is fine for this use-case for sure - we take same same approach for other CSS injection cases (infact we should probably factor those into one codified set of utilities) :+1:

alantreadway (approved):
LGTM

PR #1026, opened 2024-02-15
AG-10528 - Integrated animation fixes.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-10528

Improves the animation of series add/remove for (stacked) bar, area and line series - series now fade in rather than a) trying to animation incorrectly, or b) skip animations completely.

I've updated the animation regressions tests to cover these cases too.


lsjroberts (approved):
LGTM, great stuff

PR #1030, opened 2024-02-16
AG-10430 Update loading overlay example and add custom spinner example

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10429

Also removes extra debug info.


alantreadway (approved):
LGTM

PR #1033, opened 2024-02-16
AG-10431 Update data service to silently handle callback errors

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10431

Also adds a small delay on the loading overlay to prevent too much flashing

### DIFF
--- a/packages/ag-charts-community/src/chart/data/dataService.ts
+++ b/packages/ag-charts-community/src/chart/data/dataService.ts
@@ -1,6 +1,7 @@
 import { Debug } from '../../util/debug';
 import { throttle } from '../../util/function';
 import { Listeners } from '../../util/listeners';
+import { Logger } from '../../util/logger';
 import { ActionOnSet } from '../../util/proxy';
 import type { AnimationManager } from '../interaction/animationManager';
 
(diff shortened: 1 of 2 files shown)


packages/ag-charts-community/src/chart/data/dataService.ts
```diff
@@ -111,26 +111,25 @@ export class DataService<D extends object> extends Listeners<EventType, EventHan
...
         this.freshRequests.push(id);
 
+        let response;
         try {
-            const response = await this.dataSourceCallback(params);
+            response = await this.dataSourceCallback(params);
             this.debug(`DataService - response | ${performance.now() - start}ms | ${id}`);
+        } catch (error) {
+            this.debug(`DataService - error | ${id}`);
+            // Ignore errors in callback and keep chart alive
```
alantreadway, line 121: Nit: Maybe a `Logger.errorOnce()` is warranted in this case?

packages/ag-charts-community/src/chart/data/dataService.ts
```diff
@@ -87,7 +87,7 @@ export class DataService<D extends object> extends Listeners<EventType, EventHan
 
     private createThrottledDispatch(dispatchThrottle: number) {
         return throttle(
-            (id: number, data: D[]) => {
+            (id: number, data: D[] | any) => {
```
alantreadway, line 91: Needed?

alantreadway (approved):
LGTM, merge when ready

packages/ag-charts-community/src/chart/data/dataService.ts
lsjroberts: The response from the callback is `unknown`, in the future we'll handle this better I presume and only pass through valid data. But for now I added this to let the response always be dispatched.

PR #1041, opened 2024-02-19
CRT-223 Fix time interval to round fractional steps and warn

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-223


alantreadway (approved):
LGTM

PR #1044, opened 2024-02-19
Remove legacy code from HdpiCanvas

iMoses:
Talked to Sean and got the green light to remove the `measureText` legacy polyfill


packages/ag-charts-community/src/util/dom.ts
```diff
@@ -1,4 +1,48 @@
...
+    element.href = dataUrl;
+    element.download = fileName;
+    element.style.display = 'none';
+    body.appendChild(element);
+    element.click();
+    setTimeout(() => body.removeChild(element));
+}
+
+export function injectStyle(cssStyle: string) {
+    const document = getDocument();
```
alantreadway, line 45: This won't work for server-side rendering (e.g. thumbnail rendering), we need a non-global `Document` in those cases.

packages/ag-charts-community/src/util/draw-utils.ts
alantreadway, line 1: Duplicate?

alantreadway (changes requested):
Generally LGTM, but we need to allow `document` and `window` references to be passed in for server-side rendering cases.

packages/ag-charts-community/src/util/draw-utils.ts
iMoses: yeah, oops :)

packages/ag-charts-community/src/util/dom.ts
alantreadway: Discussed offline, we'll run with the changes and see the repurcussions.

alantreadway (approved):
LGTM

PR #1047, opened 2024-02-20
Introduce a StackTracker to replace repeated pattern in our code

iMoses:


packages/ag-charts-community/src/chart/interaction/cursorManager.ts
```diff
@@ -1,39 +1,17 @@
...
-
-        // Last added entry wins.
-        Object.entries(this.states)
-            .reverse()
-            .slice(0, 1)
-            .forEach(([_, { style }]) => (styleToApply = style));
-
-        this.element.style.cursor = styleToApply;
+        this.stateTracker.set(callerId, style);
+        this.element.style.cursor = this.stateTracker.stateValue()!;
```
lsjroberts, line 14: I don't know if it's possible in typescript, but would be nice to not have to put the `!` there and have the type of `stateValue()` not be `thing | undefined` if the `defaultValue` is not `undefined`?

The `!` always feels a unsafe when reading the code later and not knowing the details.

(this state tracker looks great though)

lsjroberts: oh checking again, I guess you could do `new StateTracker<string>('default')`

iMoses: I agree with you, tried to achieve something like that without any luck.
I can try again, maybe typing the constructor itself and what returns from it would work.

iMoses: So it seems that there no way to do this. The type must be declared *before* the constructor is called.

lsjroberts: doh, fair enough

alantreadway (approved):
LGTM

packages/ag-charts-community/src/chart/interaction/cursorManager.ts
alantreadway: Had a quick play, and I think it is possible - but you need to make the default value required, and add `| undefined` to the generic type parameter at call-sites for other cases.

PR #1049, opened 2024-02-20
CRT-237 Fix data source async clipping

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-237

The two changes here are to include the `clipSeries` status in the check for axis stability, and shifting the setting of that to after the axis layout is calculated as that ensures the normalised domain has been updated first.

It only adds 1 extra pass on the axis calculations when the data domain is clipped, otherwise it is as before.


alantreadway (approved):
LGTM

PR #1054, opened 2024-02-20
AG-7696 Enable animations for bar series with corner radii

jacobp100:
https://ag-grid.atlassian.net/browse/AG-7696


packages/ag-charts-community/src/util/interpolating.ts
```diff
@@ -0,0 +1,6 @@
+// FIXME: Use symbol (requires TS >=4.9)
+export const interpolate = '$interpolate';
+
+export interface Interpolating<T = any> {
+    $interpolate: (other: T, d: number) => T;
```
alantreadway, line 7: Why use a special symbol for this case, wouldn't a standard interface be sufficient?

jacobp100: I was going to make it a symbol, but the TS version we’re on doesn’t support it. I.e.

```ts
export const interpolate = Symbol('interpolate');

export interface Interpolating<T = any> {
    [interpolate]: (other: T, d: number) => T;
}
```

alantreadway (approved):
LGTM

PR #1056, opened 2024-02-20
AG-10733 Fix lazy load when chart layout comes in delayed

jacobp100:
https://ag-grid.atlassian.net/browse/AG-10733

### DIFF
--- a/packages/ag-charts-community/src/chart/update/dataWindowProcessor.ts
+++ b/packages/ag-charts-community/src/chart/update/dataWindowProcessor.ts
@@ -1,13 +1,21 @@
 import { ChartUpdateType } from '../chartUpdateType';
 import type { DataService } from '../data/dataService';
 import type { ZoomManager, ZoomState } from '../interaction/zoomManager';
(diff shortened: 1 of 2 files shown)


packages/ag-charts-community/src/chart/update/dataWindowProcessor.ts
```diff
@@ -1,27 +1,37 @@
...
+    min: Date;
+    max: Date;
+};
+
 export class DataWindowProcessor<D extends object> implements UpdateProcessor {
     private dirtyZoom = false;
     private dirtyDataSource = false;
-    private lastAxisZooms = new Map<string, ZoomState>();
+    private dirtyLayout = false;
+    private lastChartSize: { width: number; height: number } | undefined = undefined;
```
alantreadway, line 17: I feel like we maybe mixing too many concerns here, the `DataWindowProcessor` shouldn't be concerned about layout state, or embedding logic about how to determine if another layout is likely to happen?

PR #1058, opened 2024-02-20
AG-8255 - Allow fixing tooltip position to the chart

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-8255


packages/ag-charts-community/src/chart/tooltip/tooltip.ts
```diff
@@ -141,11 +142,23 @@ const defaultTooltipCss = `
...
+    | 'pointer'
+    | 'node'
+    | 'top'
+    | 'right'
+    | 'bottom'
+    | 'left'
+    | 'topLeft'
+    | 'topRight'
+    | 'bottomRight'
+    | 'bottomLeft';
```
alantreadway, line 155: Nit: Maybe we should use kebab-case here? Lets get Davids opinion.

alantreadway (approved):
LGTM

PR #1059, opened 2024-02-21
Allow conditional inclusion of .htaccess in built Astro site.

alantreadway:
Add support for conditional `.htaccess` inclusion in Astro build output, and only enable for production builds (the intent is that archive builds should NOT include this).


taktran (approved):
Looks good 👍🏼

seanlandsman (approved):

PR #1077, opened 2024-02-22
AG-8880 Add test to check that clicks are triggered on mousedown+mouseup only

olegat:
https://ag-grid.atlassian.net/browse/AG-8880


alantreadway (approved):
LGTM

PR #1078, opened 2024-02-22
AG-10766 Fix async zoom to only consider visible items for min rect

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10766

This adds `minVisibleRect` to the `update-complete` event. The `minRect` is still used by the axes.

The existing `minRect` is defined as the maximum unassociated x & y distance between two adjacent points.

The new `minVisibleRect` takes this but only considers points >= 0 and <= width or height.

This also removes the rounding on zoom ratios as they can get to 4 decimal points and smaller now with fine grained data. It was there to fix flickering, it seems that is not an issue anymore but something keep an eye out for.


lsjroberts:
Rebased badly, closed in favour of #1101

PR #1081, opened 2024-02-22
AG-8595 Performance improvements

olegat:
https://ag-grid.atlassian.net/browse/AG-8595

Tested on https://plnkr.co/edit/ir0o9n0pTV6lorTi?open=main.ts&preview

PR #1088, opened 2024-02-23
AG-8595 Performance improvements (part 1)

olegat:
https://ag-grid.atlassian.net/browse/AG-8595

Cache CartesianSeries.getMinRect - This is a very expensive operation if there is a lot of data, but the minRect is always the unless createNodeData() is called.

Split off PR https://github.com/ag-grid/ag-charts/pull/1081


alantreadway (approved):
LGTM

PR #1093, opened 2024-02-26
AG-8595 Performance improvements (part 2)

olegat:
https://ag-grid.atlassian.net/browse/AG-8595

Add a method `Node.getCachedBBox()` that can be called in placed the performance is crucial.

Split off PR https://github.com/ag-grid/ag-charts/pull/1081


alantreadway (approved):
LGTM

PR #1099, opened 2024-02-26
Topology

jacobp100:
image

https://plnkr.co/edit/BepR7tZtYrBZqawf?open=main.js


packages/ag-charts-community/src/options/chart/chartBuilderOptions.ts
alantreadway, line 1: Could we hold-back the options contract changes for now, as we won't likely be shipping until the next major release?

packages/ag-charts-community/src/chart/series/topology/geoJson.ts
alantreadway, line 1: Maybe we should add this as a devDependency from somewhere?

Maybe we could use https://www.npmjs.com/package/@types/geojson ?

packages/ag-charts-community/src/chart/series/topology/geoGeometry.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export class GeoGeometry extends Path {
+    @ScenePathChangeDetection()
+    geometry: Geometry | undefined = undefined;
+
+    @ScenePathChangeDetection()
+    scale: number = 1;
+
+    override updatePath(): void {
+        this.path.clear();
```
alantreadway, line 31: Nit: Use `trackChanges: true` to allow change detection on the shape?

packages/ag-charts-community/src/chart/series/topology/geoJson.ts
jacobp100: Done

packages/ag-charts-community/src/options/chart/chartBuilderOptions.ts
jacobp100: Updated

packages/ag-charts-community/src/chart/series/topology/geoGeometry.ts
jacobp100: What's the benefit here? The geometry can get quite large, so I'm hesitant to double memory usage here

alantreadway: Discussed offline - the geometry is immutable, so the decorator on the `geometry` field should be enough to determine if the path is dirty.

packages/ag-charts-community/src/scene/shape/path.ts
```diff
@@ -131,11 +130,15 @@ export class Path extends Shape {
...
-            this.path.draw(ctx);
-            this.fillStroke(ctx);
+            this.drawPath(ctx);
         }
 
         this.fillShadow?.markClean();
         super.render(renderCtx);
     }
+
+    protected drawPath(ctx: any) {
```
alantreadway, line 140: Nit: Can we use `CanvasRenderingContext2D` rather than `any`?

packages/ag-charts-enterprise/.dependency-cruiser.js
```diff
@@ -146,6 +146,7 @@ module.exports = {
             },
             to: {
                 dependencyTypes: ['npm-dev'],
+                dependencyTypesNot: ['type-only'],
```
alantreadway, line 149: Needed?

alantreadway (approved):
LGTM, I'll look at the depcruise config post merge.

PR #1101, opened 2024-02-27
AG-10766 Fix async zoom to only consider visible items for min rect

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10766

1. Adds `minVisibleRect` to the `update-complete` event. The `minRect` is still used by the axes.
	a. The existing `minRect` is defined as the maximum unassociated x & y distance between two adjacent points.
	b. The new `minVisibleRect` takes this but only considers points >= 0 and <= width or height.
2. Removes the rounding on zoom ratios as they can get to 4 decimal points and smaller now with fine grained data. It was there to fix flickering, it seems that is not an issue anymore but something keep an eye out for.
3. Changes the zoom to only discard zooms that violate the `minRatioX/Y` when zooming in, allowing zooming out and and panning.
4. Optimises the `CartesianSeries.getMinRects()` call to minimise looping of the data.
5. Recreates the caching from this change https://github.com/ag-grid/ag-charts/pull/1088 to take the scene width & height into account. I looked at using `memo()` but the stringify of the node data would probably be super slow and defeat the point.

### DIFF
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -672,7 +672,7 @@ export abstract class Chart extends Observable implements AgChartInstance {
         }
 
         if (!updateDeferred) {
-            this.updateService.dispatchUpdateComplete(this.getMinRect());
+            this.updateService.dispatchUpdateComplete(this.getMinRects());
         }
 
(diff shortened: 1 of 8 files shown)


alantreadway (approved):
LGTM

PR #1104, opened 2024-02-27
Normalize chart modules to use ModuleMap

iMoses:
Also converted `legend` into a standard module.
Next steps, change properties interface to the new implementation and remove modules from the root of the `chart` object.


alantreadway (approved):
LGTM

PR #1109, opened 2024-02-27
AG-10851 -Display at least one log tick value for small magnitude dat…

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-10850


alantreadway (approved):
LGTM

PR #1114, opened 2024-02-28
AG-10755 Fix zoom and navigator conflicts and add tests

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10775

This is focused on making zoom and navigator work better together and covering those conflicts with tests, while giving priority to zoom.

This is a significant rework of the navigator but I've added tests for it all and manually tested both sync and mini charts.

Other small tweaks include:
- Split the horizontal scroll panning into a separate behaviour class as it was unrelated to the normal panning, also boosted its speed as it was very slow and modified the speed with the `scrollingStep` to match normal scrolling.


packages/ag-charts-community/src/chart/navigator/navigator.ts
```diff
@@ -1,93 +1,101 @@
...
+            this._max = max;
+            this.updateZoom(this._min, max);
+        },
+    })
+    public max: number = 1;
 
     @Validate(BOOLEAN)
     @ObserveChanges<Navigator>((target) => target.updateGroupVisibility())
-    visible: boolean = true;
+    public visible: boolean = true; // what is this?
```
lsjroberts, line 53: This appears to start as `true`, was always updated to `true` in `performCartesianLayout()` and doesn't appear to be modified anywhere else.

I've kept it just in case, but can probably be removed? Unless it an integrated thing?

iMoses: I'm always in favour of getting rid of those cases

alantreadway: Lets remove it and see what happens - we have two weeks before code freeze to spot issues

alantreadway (approved):
LGTM

PR #1118, opened 2024-02-29
Add initial support for map series labels & markers

jacobp100:
Adds support for labels & markers. Currently doesn't do anything to stop labels overlapping - we will need to figure out what we want to do here

SCR-20240229-kfcq
SCR-20240229-kfek


alantreadway (approved):
Nice :tada: LGTM

PR #1119, opened 2024-02-29
AG-10409 Avoid unnecessary array allocations

olegat:

We only ever use the intersection values, just the count.


alantreadway (approved):
LGTM

PR #1123, opened 2024-02-29
AG-10789 - Improve series grouping change animations.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-10789

Changes the series-grouping change case so that we reset animations back to the initial load case, whilst matching the Integrated Charts series add/remove and data-update cases.

Bonus: Breaks some of the series options processing into distinct methods to reduce cyclomatic complexity of `Chart. applyOptions()`.


lsjroberts (approved):
LGTM

PR #1126, opened 2024-02-29
AG-10657 Fix bugs with enableSelecting=true and enablePanning=false

olegat:

depends on https://github.com/ag-grid/ag-charts/pull/1114

### DIFF
--- a/packages/ag-charts-enterprise/src/features/zoom/zoom.ts
+++ b/packages/ag-charts-enterprise/src/features/zoom/zoom.ts
@@ -291,14 +291,14 @@ export class Zoom extends _ModuleSupport.BaseModuleInstance implements _ModuleSu
 
         switch (this.dragState) {
             case DragState.Axis:
-                if (this.hoveredAxis) {
-                    const { id: axisId, direction } = this.hoveredAxis;
-                    const anchor =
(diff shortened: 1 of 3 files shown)


packages/ag-charts-enterprise/src/features/zoom/zoom.ts
```diff
@@ -249,89 +255,105 @@ export class Zoom extends _ModuleSupport.BaseModuleInstance implements _ModuleSu
...
-            this.updateAxisZoom(axisId, direction, newZoom);
-            return;
-        }
-
-        // Prevent the user from dragging outside the series rect (if not on an axis)
-        if (!this.paddedRect.containsPoint(event.offsetX, event.offsetY)) {
-            return;
+        switch (this.dragState) {
+            case DragState.Axis:
+                if (this.hoveredAxis) {
```
lsjroberts: nit: rather than indenting the whole block, `if (!this.hoveredAxis) break;`

lsjroberts (approved):
LGTM, definitely clearer, thanks.

packages/ag-charts-enterprise/src/features/zoom/zoom.ts
olegat: Done.

PR #1127, opened 2024-02-29
Implement animations, map backgrounds

jacobp100:
* Animates markers
* Adds `background` config for when your data is just markers and need more context

SCR-20240229-ohjx


packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts
```diff
@@ -1027,10 +1027,6 @@ export abstract class CartesianSeries<
         this.resetAllAnimation(data);
     }
 
-    protected animationTransitionClear() {
```
alantreadway, line 1030: If we're removing this, could you cleanup the state-machine config and typings too?

jacobp100:
GH isn't updating. Reopening PR

PR #1130, opened 2024-02-29
AG-10875 Fix navigator bounding box region

olegat:
https://ag-grid.atlassian.net/browse/AG-10875

When the min/max handles are on the edge of masks they stick out a bit, so take this offset into account. This fixes a bug where the cursor begin reset because the cursor left the region, and set back to 'ew-cursor' immediately after.


alantreadway:
Note: The branch name was causing conflicts on CI so I had to rename it to `AG-10875-qa_feedback_navigator_bugs`.

The error on CI was:
```
error: cannot lock ref 'refs/remotes/origin/AG-10875/qa_feedback_navigator_bugs': 'refs/remotes/origin/AG-10875' exists; cannot create 'refs/remotes/origin/AG-10875/qa_feedback_navigator_bugs'
```

olegat:
Oh... that's weird. Thanks for sorting it out 👍

PR #1131, opened 2024-03-01
AG-10875 Fix navigator bounding box region

olegat:
https://ag-grid.atlassian.net/browse/AG-10875

* When the min/max handles are on the edge of masks they stick out a bit, so take this offset into account. This fixes a bug where the cursor begin reset because the cursor left the region, and set back to 'ew-cursor' immediately after. (copy of https://github.com/ag-grid/ag-charts/pull/1130)
* Reset cursor in the navigator hover handler.
* Lock the `RegionManager.currentRegion` when dragging. This simplifies the navigator input handling logic.

### DIFF
--- a/packages/ag-charts-community/src/chart/navigator/shapes/rangeSelector.ts
+++ b/packages/ag-charts-community/src/chart/navigator/shapes/rangeSelector.ts
@@ -10,6 +10,8 @@ export class RangeSelector extends Group {
     private y = 0;
     private width = 200;
     private height = 30;
+    private lOffset = 0;
+    private rOffset = 0;
 
     constructor(children: Node[]) {
         super({ name: 'rangeSelectorGroup', layer: true, zIndex: Layers.NAVIGATOR_ZINDEX });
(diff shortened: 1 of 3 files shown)


packages/ag-charts-community/src/chart/navigator/shapes/rangeSelector.ts
```diff
@@ -44,6 +50,8 @@ export class RangeSelector extends Group {
 
     override computeBBox() {
         const { x, y, width, height } = this;
-        return new BBox(x, y, width, height);
+        const minOff = this.minHandle.width / 2;
+        const maxOff = this.maxHandle.width / 2;
+        return new BBox(x - minOff, y, width + (minOff + maxOff), height);
```
lsjroberts: The goal in my refactor was to make these scenes internally independent so it was more explicit in the `Navigator` class what was happening, as hiding this stuff in the scenes was causing lots of issues with the zoom.

So I'd prefer if in `Navigator.layoutNodes()` it passed in something like:

```
private layoutNodes(x: number, y: number, width: number, height: number) {
    // ...
    rangeSelector.layout(x - minOffset, y, width + (minOffset - maxOffset), height);
    // ...
}
```

So the `RangeSelector` doesn't need to know about the handles and mask directly, and instead the scenes are coordinated by the parent `Navigator`.

lsjroberts (changes requested):

packages/ag-charts-community/src/chart/navigator/shapes/rangeSelector.ts
olegat: Done. Moved the handle offset calculations to `Navigator.layoutNodes()`

lsjroberts (approved):
Nice, lgtm

PR #1133, opened 2024-03-01
AG-8218 Add warning when navigator initial values are rejected

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-8218

Also fixes the formatting on the zoom options comments for the default values.


alantreadway (approved):
LGTM

PR #1139, opened 2024-03-01
AG-10853 Make both the series and axes groups of the 'series' region

olegat:
https://ag-grid.atlassian.net/browse/AG-10853


alantreadway (approved):
LGTM

PR #1143, opened 2024-03-04
Add Charts benchmarks.

alantreadway:
Adds `nx benchmark ag-charts-community` which runs our release benchmarks in a automated manner.

One outstanding kink is that the stats report file is overwritten by each test suite and only contains data for one test-suite, despite use of a global variable to try and capture stats across all test-suites.


packages/ag-charts-community/benchmarks/benchmark.ts
```diff
@@ -0,0 +1,72 @@
...
+export function benchmark(name: string, ctx: BenchmarkContext, callback: () => Promise<void>) {
+    it(name, async () => {
+        const start = performance.now();
+        await callback();
+        const duration = performance.now() - start;
+
+        const newImageData = extractImageData(ctx.canvasCtx);
+        expect(newImageData).toMatchImageSnapshot(IMAGE_SNAPSHOT_DEFAULTS);
+
+        recordTiming(expect.getState().currentTestName ?? 'unknown', duration);
```
olegat: This may silently create duplicate timings all named `'unknown'`, which would overwrite existing results (i.e. only the last benchmark will have stats reported). It might be a good idea to ensure that only unique names are provided to `recordTiming`, such by appending UUIDs or sequenced numbers (e.g `'unknown-1'`, `'unknown-2'`, `'unknown-3'`, ...)

packages/ag-charts-community/benchmarks/benchmark.ts
```diff
@@ -0,0 +1,72 @@
...
+    });
+
+    afterEach(() => {
+        if (ctx.chart) {
+            ctx.chart.destroy();
+            (ctx.chart as unknown) = undefined;
+        }
+
+        expect(console.warn).not.toHaveBeenCalled();
+        expect(console.error).not.toHaveBeenCalled();
```
olegat: These assertions are redundant when `setupMockConsole()` is called
https://github.com/ag-grid/ag-charts/blob/778e2a2c825c7ea766fda9242caa89c1d165d46e/packages/ag-charts-community/src/chart/test/mockConsole.ts#L13

packages/ag-charts-community/benchmarks/timing.ts
```diff
@@ -0,0 +1,32 @@
+/* eslint-disable no-console */
```
olegat: optional: consider using `eslint-disable-next-line` because the console is only referenced once in this file

packages/ag-charts-community/benchmarks/benchmark.ts
alantreadway: From practical testing this seems to return the current test-case name; I will probably just make it a hard-fail if `currentTestName` is `undefined`.

packages/ag-charts-community/benchmarks/timing.ts
alantreadway: Done

packages/ag-charts-community/benchmarks/benchmark.ts
alantreadway: Good spot - done.

packages/ag-charts-community/benchmarks/benchmark.ts
alantreadway: Adjusted to a hard-fail in this case.

PR #1147, opened 2024-03-04
AG-10657 Always pop InteractionState.ZoomDrag

olegat:

There's no harm in popping ZoomDrag if it isn't pushed, because it just sets a flag to 0. This ensures that the chart remains interactable after the dragging when enablePanning=false.


alantreadway (approved):
LGTM

PR #1149, opened 2024-03-04
Add map marker series

jacobp100:
* Adds a map marker series for data that includes a lat/lon where it wouldn't make sense to have an associated topology file (e.g. crime, live data etc.)
* Improves default colours and themes
* Adds some more examples

Right now, the new map marker series is a lot of copy & paste from the map series (which is why sonarqube is complaining). For now I think it's a good idea leave them separate until they're more 'ready', then see what can be abstracted. There's additional complexities like the regular map series can have multiple markers per datum (as a side effect of using GeoJSON to get the marker positions), so it's not too straight forward

image


packages/ag-charts-community/src/chart/series/topology/mapSeriesProperties.ts
```diff
@@ -169,4 +170,10 @@ export class MapSeriesProperties extends SeriesProperties<AgMapSeriesOptions> {
 
     @Validate(OBJECT)
     override tooltip = new SeriesTooltip<AgMapSeriesTooltipRendererParams<any>>();
+
+    @Validate(STRING, { optional: true })
+    __POLYGON_STROKE?: string = undefined;
```
alantreadway, line 175: Is the intent here with the naming that we want to inject theme-based values, but not allow users to set these?  (and for `__LINE_STRING_STROKE` presumably)

jacobp100: If the user specifies a stroke, it'll use their stroke. Otherwise, full shapes use the background colour (i.e. country borders) and lines (i.e. roads, flight paths etc.) use the series colour

image

image

alantreadway (commented):
https://sonarcloud.io/component_measures?id=ag-charts-community-latest&pullRequest=1149&metric=new_duplicated_lines_density&view=list

Any sane way we can reduce the identified duplications?

jacobp100:
> Any sane way we can reduce the identified duplications?

I put a bit about this here - https://github.com/ag-grid/ag-charts/pull/1149#issue-2167018164. I think they might diverge a bit further than they are now, so I want to let that happen before trying to find common abstractions

alantreadway (approved):
LGTM

packages/ag-charts-community/src/chart/series/topology/mapSeriesProperties.ts
alantreadway: Discussed offline, we'll roll with this for now.

PR #1153, opened 2024-03-04
AG-10217 - Validate ordinal time scale data values during data proces…

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-10217


alantreadway (approved):
LGTM

PR #1154, opened 2024-03-04
AG-10782 Spike - Add range buttons with fixed options

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10782

### DIFF
--- a/packages/ag-charts-enterprise/src/features/range-buttons/rangeButtons.ts
+++ b/packages/ag-charts-enterprise/src/features/range-buttons/rangeButtons.ts
@@ -1,0 +1,17 @@
+import { _ModuleSupport, _Scene, _Util } from 'ag-charts-community';
+
+const { BOOLEAN, ActionOnSet, Validate } = _ModuleSupport;
+
+export class RangeButtons extends _ModuleSupport.BaseModuleInstance implements _ModuleSupport.ModuleInstance {
+    @ActionOnSet<RangeButtons>({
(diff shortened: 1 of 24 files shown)


packages/ag-charts-enterprise/src/features/range-buttons/rangeButtons.ts
```diff
@@ -0,0 +1,116 @@
...
+            this.layoutNodes(x, y, width, height);
+        }
+    }
+
+    private onHover({ offsetX, offsetY }: _ModuleSupport.InteractionEvent<'hover'>) {
+        this.activeButton = undefined;
+
+        if (!this.enabled) return;
+
+        for (const button of this.toolbar.children) {
```
alantreadway: The responsibilities seem a bit mixed here - I'm thinking perhaps `ToolbarManager` should be merged with this class in Community, but the only place we actually add buttons is from the Enterprise Zoom feature?

packages/ag-charts-enterprise/src/features/range-buttons/toolbar.ts
```diff
@@ -0,0 +1,9 @@
+import { _Scene } from 'ag-charts-community';
+
+const { Group } = _Scene;
+
+export class Toolbar extends Group {
```
alantreadway: Perhaps move into community?

alantreadway (approved):
LGTM

PR #1157, opened 2024-03-05
AG-10775 Change zoom to reset to initial range/ratio on dbl click

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10775


alantreadway (approved):
LGTM

PR #1161, opened 2024-03-05
AG-10657 Fix warnings when scrolling

olegat:

This restores https://github.com/ag-grid/ag-charts/pull/1139 and fixes warnings that were caused by it.

---

The `'series'` region from the `RegionManager` is the first listener to
receive `'wheel'` events. The Zoom module consumes `'wheel'` events which
prevents the Chart class from receiving these events and removing the
highlight.

Changing the Chart to listen for `'wheel'` events in the `'series'`
region ensures that it receives these events before the Zoom module.

Without this change, the following warning would be printing when
scrolling on a zoomable chart:

```
AG Charts - exceeded the maximum number of simultaneous updates (4), discarding changes and rendering
```

### DIFF
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -359,7 +359,6 @@ export abstract class Chart extends Observable implements AgChartInstance {
             seriesRegion.addListener('leave', (event) => this.onLeave(event)),
             this.interactionManager.addListener('page-left', () => this.destroy()),
 
-            seriesRegion.addListener('wheel', () => this.resetPointer()),
             this.interactionManager.addListener('drag', () => this.resetPointer()),
             this.interactionManager.addListener('contextmenu', (event) => this.onContextMenu(event), All),
 
@@ -367,9 +366,10 @@ export abstract class Chart extends Observable implements AgChartInstance {
                 this.update(ChartUpdateType.SCENE_RENDER);
             }),
             this.highlightManager.addListener('highlight-change', (event) => this.changeHighlightDatum(event)),
-            this.zoomManager.addListener('zoom-change', () =>
-                this.update(ChartUpdateType.PROCESS_DATA, { forceNodeDataRefresh: true, skipAnimations: true })
-            )
+            this.zoomManager.addListener('zoom-change', () => {
+                this.resetPointer();
+                this.update(ChartUpdateType.PROCESS_DATA, { forceNodeDataRefresh: true, skipAnimations: true });
+            })
         );
     }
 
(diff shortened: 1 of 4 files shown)


alantreadway:
Since this seems to be an order of registration issue, should we perhaps consider adding listener priorities to make the ordering concern a bit more explicit?

olegat:
> Since this seems to be an order of registration issue, should we perhaps consider adding listener priorities to make the ordering concern a bit more explicit?

Whilst I agree with the intent, I feel that adding listener priorities would be over-engineering. The core problem that the Zoom module implicitly depends on the Chart class clearing the tooltip/highlight when a wheel event is broadcast. This dependency shouldn't really exist though.

An easier solution would be to remove the `'wheel'` handler for the Chart and move the `resetPointer()` behaviour to the `'zoom-change'` handler.

```diff
diff --git a/packages/ag-charts-community/src/chart/chart.ts b/packages/ag-charts-community/src/chart/chart.ts
index 3904f00ba..519c5d5fa 100644
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -359,7 +359,6 @@ export abstract class Chart extends Observable implements AgChartInstance {
             seriesRegion.addListener('leave', (event) => this.onLeave(event)),
             this.interactionManager.addListener('page-left', () => this.destroy()),
 
-            seriesRegion.addListener('wheel', () => this.resetPointer()),
             this.interactionManager.addListener('drag', () => this.resetPointer()),
             this.interactionManager.addListener('contextmenu', (event) => this.onContextMenu(event), All),
 
@@ -367,9 +366,10 @@ export abstract class Chart extends Observable implements AgChartInstance {
                 this.update(ChartUpdateType.SCENE_RENDER);
             }),
             this.highlightManager.addListener('highlight-change', (event) => this.changeHighlightDatum(event)),
-            this.zoomManager.addListener('zoom-change', () =>
-                this.update(ChartUpdateType.PROCESS_DATA, { forceNodeDataRefresh: true, skipAnimations: true })
-            )
+            this.zoomManager.addListener('zoom-change', () => {
+                this.resetPointer();
+                this.update(ChartUpdateType.PROCESS_DATA, { forceNodeDataRefresh: true, skipAnimations: true });
+            })
         );
     }
 
```

I've tested this change and it seems to work as intended. This might be doable for the `'drag'` handler of Chart too. This is better approach because the Chart class does not actually care about `'wheel'` and `'drag'` events, it only cares whether the view is changing. It also makes more sense conceptually, because the Zoom module might receive `'wheel'` and realise that nothing needs to change (e.g. wheel-down when already fully zoomed out), so it wouldn't broadcast a `'zoom-change'` the Chart wouldn't need to clear the tooltip/highlight.

However, this change is also a bit risky and would require more testing to make sure that it doesn't have any unintended consequences to other modules like the navigator.

olegat:
Converted to draft because I've identified a bug with axis-zoom where the `'ew-resize'` and `'ns-resize'` cursor is not always reset when exiting the bounding box of the axes.

alantreadway (approved):
LGTM - thanks!

PR #1163, opened 2024-03-05
Move topology charts to enterprise, docs

jacobp100:


packages/ag-charts-community/src/chart/factory/expectedEnterpriseModules.ts
```diff
@@ -39,16 +39,18 @@ export const EXPECTED_ENTERPRISE_MODULES: EnterpriseModuleStub[] = [
     { type: 'series', optionsKey: 'series[]', chartTypes: ['cartesian'], identifier: 'candlestick' },
     { type: 'series', optionsKey: 'series[]', chartTypes: ['cartesian'], identifier: 'bullet' },
     { type: 'series', optionsKey: 'series[]', chartTypes: ['cartesian'], identifier: 'heatmap' },
+    { type: 'series', optionsKey: 'series[]', chartTypes: ['cartesian'], identifier: 'range-area' },
```
alantreadway, line 42: Good spot, thanks for this!

packages/ag-charts-community/src/options/series/topology/topologyOptions.ts
```diff
@@ -7,6 +7,8 @@ export type AgTopologySeriesOptions = AgMapSeriesOptions | AgMapMarkerSeriesOpti
 export interface AgBaseTopologyChartOptions {
     /** Series configurations. */
     series?: AgTopologySeriesOptions[];
+    /** Topology to use in all series. */
+    topology?: any;
```
alantreadway, line 11: Nit: Could we use an alias type called `GeoJson` to give some hint as to what is expected here?

packages/ag-charts-enterprise/tsconfig.lib.json
```diff
@@ -3,6 +3,10 @@
   "compilerOptions": {
     "types": ["node"]
   },
-  "include": ["src/**/*.ts"],
+  "include": [
+    "src/**/*.ts",
+    "../ag-charts-community/src/chart/series/topology/mercatorScale.ts",
+    "../ag-charts-community/src/chart/series/topology/lonLatBbox.ts"
```
alantreadway: This looks wrong, can you export these via the `main.ts` in `ag-charts-community` instead?

packages/ag-charts-enterprise/tsconfig.lib.json
```diff
@@ -3,6 +3,10 @@
...
+  ],
```
jacobp100:
```suggestion
  "include": ["src/**/*.ts"],
```

packages/ag-charts-community/src/options/series/topology/topologyOptions.ts
jacobp100: Without a peer dep on `geojson`, it's not easily possible. I do want to add some runtime validation

alantreadway:
(LGTM)

PR #1165, opened 2024-03-05
Data controller optimisations

iMoses:


packages/ag-charts-community/src/chart/data/dataController.ts
```diff
@@ -220,6 +219,52 @@ export class DataController {
     }
 
     private static mergeRequests(requests: RequestedProcessing<any, any, any>[]): MergedRequests<any, any, any> {
+        const skipKeys = new Set<string>(['id', 'ids', 'type', 'scopes', 'useScopedValues']);
+
+        function deepEqual<T>(a: T, b: T): boolean {
```
alantreadway: Nit: Maybe worth adding a comment to explain this is an optimisation (otherwise it's likely we'll revert this to a general diff function in the future)?

Also, maybe worth extracting to a top-level function in this TS module if there is no dependency on the context?

alantreadway (approved):
LGTM, merge on green and once addressing the one nit?

PR #1170, opened 2024-03-06
AG-10782 Fix zoom range buttons not triggering update after navigator drag

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10782

### DIFF
--- a/packages/ag-charts-enterprise/src/features/zoom/zoomRange.ts
+++ b/packages/ag-charts-enterprise/src/features/zoom/zoomRange.ts
@@ -35,6 +35,33 @@ export class ZoomRange {
         return this.getRangeWithValues(this.initialStart, this.initialEnd);
     }
 
+    public extendToEnd(extent: number) {
+        if (!this.domain) return;
+
+        const end = this.domain.at(-1);
+        if (end == null) return;
+
+        this.end = end;
+        this.start = Number(end) - extent;
+    }
+
+    public extendWith(fn: (end: Date | number) => Date | number) {
+        if (!this.domain) return;
+
+        const end = this.domain.at(-1);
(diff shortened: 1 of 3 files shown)


packages/ag-charts-enterprise/src/features/zoom/zoomRange.ts
```diff
@@ -35,6 +38,35 @@ export class ZoomRange {
...
+
+    public extendWith(fn: (end: Date | number) => Date | number) {
+        if (!this.domain) return;
+
+        const end = this.domain.at(-1);
+        if (end == null) return;
+
+        this.dirty = true;
+
+        this.end = end;
```
alantreadway, line 51: Discussed offline: Maybe add a comment that `this.dirty` maybe reset as a side-effect of setting these fields?

packages/ag-charts-enterprise/src/features/zoom/zoomRange.ts
```diff
@@ -35,6 +38,35 @@ export class ZoomRange {
...
+
+        if (this.dirty) this.onChange?.(this.getRange());
+    }
+
+    public extendAll() {
+        if (!this.domain) return;
+
+        this.dirty = true;
+
+        this.start = this.domain[0];
```
alantreadway: Ditto

packages/ag-charts-enterprise/src/features/zoom/zoomRange.ts
lsjroberts: Changed it to check the values have changed directly and not use side-effects.

alantreadway (approved):
LGTM

PR #1176, opened 2024-03-06
Navigator interaction optimisations

iMoses:


packages/ag-charts-community/src/scene/shape/text.ts
```diff
@@ -699,5 +716,5 @@ function getVerticalOffset(textBaseline: CanvasTextBaseline): number {
 }
 
 export function splitText(text?: string) {
-    return typeof text === 'string' ? text.split(/\r?\n/g) : [];
+    return typeof text === 'string' ? text.split('\n') : [];
```
alantreadway, line 719: Does this case deal with Windows carriage return cases correctly?

iMoses: it does, a return carriage without a line break is simply ignored.

iMoses: Since we can't really test this on a Linux machine, I suggest we merge and then ask David to test this on Windows.

alantreadway (approved):
LGTM

PR #1178, opened 2024-03-06
AG-10923 Fix regression where clicks reset the zoom

olegat:
https://ag-grid.atlassian.net/browse/AG-10923

If the Zoom class receives a 'drag-start' and 'drag-end' event but no 'drag' event, then no do not attempt to update the zoom.


alantreadway (approved):
LGTM

PR #1179, opened 2024-03-06
Topology updates

jacobp100:
* Renders labels on map markers the same as the bubble and scatter series
* Allows configuring the placement of labels (top, bottom, left, right)
* Adds `pin` and `star` markers
* Allows markers to specify their own centre (so map pins align)
* Hides `colorKey` from `map-marker` series
* Renames `topologyProperty` to `topologyIdKey`

image


packages/ag-charts-community/src/chart/series/polar/donutSeries.ts
```diff
@@ -584,6 +577,13 @@ export class DonutSeries extends PolarSeries<DonutNodeDatum, DonutSeriesProperti
             title.node.translationY = isFinite(dy) ? dy : 0;
         }
 
+        for (const circle of [this.zerosumInnerRing, this.zerosumOuterRing]) {
+            circle.fillOpacity = 0;
+            circle.stroke = this.properties.calloutLabel.color;
```
jacobp100, line 582: This property isn't available in the constructor, so was using the default fill (which I've removed)

packages/ag-charts-enterprise/src/series/waterfall/waterfallThemes.ts
```diff
@@ -1,27 +1,25 @@
 import { _Theme } from 'ag-charts-community';
 
+const itemTheme = {
+    strokeWidth: 0,
+    label: {
+        fontStyle: undefined,
+        fontWeight: _Theme.FONT_WEIGHT.NORMAL,
+        fontSize: 12,
+        fontFamily: _Theme.DEFAULT_FONT_FAMILY,
+        color: _Theme.DEFAULT_LABEL_COLOUR,
```
jacobp100, line 10: This colour wasn't being set by default, so was using the default fill (which I removed)

packages/ag-charts-community/src/chart/marker/marker.ts
```diff
@@ -15,9 +18,9 @@ export abstract class Marker extends Path {
 
     override computeBBox(): BBox {
         const { x, y, size } = this;
-        const half = size / 2;
+        const { center } = this.constructor as any;
```
alantreadway, line 21: This seems a bit of a weird way to do this - could you either just declare a non-static `center` (and assign a constant value), allowing the sub-class to update this property, or add a `getCenter(): Point` method that sub-classes can override?

packages/ag-charts-community/src/chart/label.ts
```diff
@@ -146,5 +146,7 @@ export function calculateLabelBBox(
     return {
         point: { x, y, size: 0 },
         label: { width, height, text },
+        marker: undefined,
+        placement: 'bottom',
```
alantreadway: Is this default consistent with all existing cases - i.e. is there any risk this is a breaking change?

packages/ag-charts-community/src/chart/marker/marker.ts
jacobp100: The reason for this was otherwise we'd have to pass actual marker instances to `labelPlacement` - but they wouldn't actually correspond with real markers in the scene (and wouldn't have their x/y set). We needed a way to get the centre without actually constructing the instances

packages/ag-charts-community/src/chart/label.ts
jacobp100: I've managed to untangle these types. It never needed `size` either

packages/ag-charts-community/src/chart/marker/marker.ts
alantreadway: Ok, let's roll with this for now.

alantreadway (approved):
LGTM

PR #1183, opened 2024-03-07
AG-10835 Fix waterfall connectors animation to not collapse

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10835


alantreadway (approved):
LGTM

PR #1184, opened 2024-03-07
Make zoom panning slippy, like in Google Maps

jacobp100:
https://ag-grid.atlassian.net/browse/AG-10941


lsjroberts (approved):
lgtm, thanks!

packages/ag-charts-enterprise/src/features/zoom/zoomPanner.ts
```diff
@@ -1,19 +1,136 @@
...
+        this.zoomCoordsHistoryIndex = 0;
+        this.coordsHistory.length = 0;
+
+        if (deltaT > 0 && this.deceleration < 1) {
+            const xVelocity = deltaX / deltaT;
+            const yVelocity = deltaY / deltaT;
+            const velocity = Math.hypot(xVelocity, yVelocity);
+            const angle = Math.atan2(yVelocity, xVelocity);
+            const t0 = performance.now();
+            this.inertiaHandle = requestAnimationFrame((t) => {
```
alantreadway, line 97: I'm not sure if we can use `requestAnimationFrame()` directly (here and below) - if chart rendering is slower than 60 FPS, I'm concerned that the updates this generates will not play well with an in-progress Chart.performUpdate() cycle.

Could we adjust the solution to use `AnimationManager` to execute the animation steps, as it already handles coordination with `Chart.performUpdate()`?

alantreadway: Discussed offline - as this uses `ZoomManager` to percolate state changes, we are already taking this concern indirectly into account.

alantreadway (approved):
LGTM

PR #1185, opened 2024-03-07
AG-10578 Remove scoped data even if not extracting

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10578


alantreadway (approved):
Change LGTM, but can we trivially add a unit-test for `DataController` to avoid regressing again in the future?

alantreadway (approved):
LGTM

PR #1188, opened 2024-03-07
Improve label placement in map series, add gallery exampels

jacobp100:
Takes into account label bbox for the placing of labels - rather than finding a good place for a single point, and seeing if a label can fit (see the small states like Connecticut)

###### Before
SCR-20240307-oisk

###### After
SCR-20240307-oirh


alantreadway (approved):
Nice, LGTM

PR #1198, opened 2024-03-08
AG-10904 - Prevent no-op animations.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-10904
https://ag-grid.atlassian.net/browse/AG-10861

Attempts to prevent redundant animations when the final rendering state is identical to the starting state.

Explicitly introduces a `no-op` animation phase to allow intermediate calculations to control the animation behaviour more explicitly. (Any animation request reaching the `AnimationManager` with this phase will be treated as-if it has been explicitly skipped).

Specifically fixes area and line series path animations being no-ops, typically causing markers and labels to fade in and out unecessarily.

### DIFF
--- a/packages/ag-charts-community/src/chart/data/processors.ts
+++ b/packages/ag-charts-community/src/chart/data/processors.ts
@@ -323,7 +323,7 @@ export function diff(
                 const prevId = prev ? createDatumId(prev.keys) : '';
                 const datumId = datum ? createDatumId(datum.keys) : '';
 
-                if (prevId === datumId) {
+                if (datum && prev && prevId === datumId) {
                     if (!arraysEqual(prev.values, datum.values)) {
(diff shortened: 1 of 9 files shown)


packages/ag-charts-community/src/chart/data/processors.ts
```diff
@@ -323,7 +323,7 @@ export function diff(
                 const prevId = prev ? createDatumId(prev.keys) : '';
                 const datumId = datum ? createDatumId(datum.keys) : '';
 
-                if (prevId === datumId) {
+                if (datum && prev && prevId === datumId) {
```
alantreadway, line 326: Fixes a bug when `diff` is used with continuous line series..

lsjroberts (approved):
Nice, lgtm

PR #1199, opened 2024-03-08
Ag 10217/ordinal time scale

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-10217


packages/ag-charts-community/src/scale/bandScale.ts
```diff
@@ -72,7 +72,7 @@ export class BandScale<D> implements Scale<D, number, number> {
     ticks(): D[] {
         this.refresh();
         const { interval = 1 } = this;
-        const step = Math.abs(Math.round(interval));
+        const step = Math.abs(Math.round(interval as number));
```
alantreadway: Nit: Can we avoid this forced cast?

alantreadway (approved):
LGTM other than one nit

packages/ag-charts-community/src/scale/bandScale.ts
alantreadway: Fixed

PR #1205, opened 2024-03-11
AG-10620 - Only support initial-load animation in mini-charts.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-10620

Adds a new `ChartAnimationPhase` of `disabled`, so that for the mini-chart case we can disable everything except the initial load animation.


lsjroberts (approved):
lgtm

PR #1207, opened 2024-03-11
Suppress expected warnings

olegat:
\[NO BUG\]

Use `mockConsole.ts` in `properties.test.ts` to suppress expected warnings from the console when running tests.

This moves `mockConsole.ts` into the same directory as `mockCanvas.ts`.

The main change is in `packages/ag-charts-community/src/util/properties.test.ts`. The rest is all just import changes due the move of `mockConsole.ts`.


alantreadway (approved):
LGTM

PR #1210, opened 2024-03-11
AG-10991 Fix label placement in map-marker series when sizeKey is not set, rename {lat,lon}{Key,Name}

jacobp100:
https://ag-grid.atlassian.net/browse/AG-10991


alantreadway (approved):
LGTM

PR #1212, opened 2024-03-11
AG-10943 Fix highlight z indices for maps

jacobp100:
https://ag-grid.atlassian.net/browse/AG-10943

Most series will show their highlight above every other series - but we don't want that for map series


alantreadway (approved):
LGTM

PR #1213, opened 2024-03-11
AG-10853 Remove obsolete RTI-1406 fix

olegat:
https://ag-grid.atlassian.net/browse/AG-10853
https://ag-grid.atlassian.net/browse/RTI-1406

The RegionManager will ensure that a 'leave' event is dispatched when the mouse pointer exits the seriesRect, so the tooltip will automatically get cleared anyway. No need for additional hit-testing in the tooltipManager.


alantreadway (approved):
LGTM - nice cleanup!

PR #1217, opened 2024-03-12
AG-10898 Fix dates in bandscales never matching multiple data per series

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10898


packages/ag-charts-community/src/chart/series/cartesian/lineSeries.test.ts
```diff
@@ -447,4 +447,90 @@ describe('LineSeries', () => {
...
+            chart = AgCharts.create(options) as Chart;
+
+            // TODO: replace with `compare()` with 0 percent threshold
+            await waitForChartStability(chart);
+
+            const imageData = extractImageData(ctx);
+            expect(imageData).toMatchImageSnapshot({
+                failureThreshold: 0,
+                failureThresholdType: 'percent',
+            });
```
lsjroberts, line 491: These tests didn't fail with the change reverted and using the default threshold. But changing the threshold for all the tests led to them all updating (just the removal of the markers so probably fine, but worth saving for another PR).

alantreadway (approved):
LGTM

PR #1218, opened 2024-03-12
AG-10780 Adding properties to mini chart series ignore list

AG-DavidG:
I missed these on AG-10780 originally.


alantreadway (approved):
LGTM

PR #1220, opened 2024-03-12
AG-10996 Fix tooltip defaults when using map shape accessory

jacobp100:
https://ag-grid.atlassian.net/browse/AG-10996


alantreadway (approved):
LGTM

PR #1221, opened 2024-03-12
AG-10958 Fix navigator and zoom start and end values validation

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10958


alantreadway (approved):
LGTM

PR #1225, opened 2024-03-12
Update default snapshot threshold to 0

lsjroberts:
Almost all of this is some more axis label jiggling.

Exceptions:
- Pie series snapshot with outdated borders
- Line series snapshot with outdated markers


alantreadway (approved):
LGTM

PR #1231, opened 2024-03-12
AG-10937 Fix data per series with multiple values on one prop

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10937


packages/ag-charts-community/src/chart/data/dataController.test.ts
```diff
@@ -352,6 +352,63 @@ describe('DataController', () => {
...
+                        processor: () => (value) => `key2 ${value}`,
+                    },
+                ],
+            });
+
+            await controller.execute();
+            const results = await Promise.all([promise1, promise2]);
+
+            expect(results[0].processedData.data[0].datum).toEqual({ valueProp1: 100 });
+            expect(results[1].processedData.data[0].datum).toEqual({ valueProp1: 40 });
```
lsjroberts, line 406: Specifically, these objects should not be `{ valueProp1: 'key2 100' }`

alantreadway (approved):
LGTM

PR #1232, opened 2024-03-13
AG-10553 - Fix animation NaN delay bug.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-10553?focusedCommentId=60876

Fixes a very specific bug that happens when zoom with user-specified range is used in conjunction with animations and a series which implements initial-load swipe-in animation for markers.

Root cause: The marker-swipe-in delay animation parameter was evaluating to `NaN`, which stalled all animations from running.

Remediation:
- Detect the `NaN` case in `markerSwipeScaleInAnimation()` and set `delay: 0` instead.
- Defensive changes to `AnimationBatch` so that over-running animations don't run infinitely.
- Defensive changes to `AnimationManager` to hard-stop if a `NaN` or value outside the range `0` and `2`  is provided as an animation duration/delay.

### DIFF
--- a/packages/ag-charts-community/src/chart/interaction/animationBatch.ts
+++ b/packages/ag-charts-community/src/chart/interaction/animationBatch.ts
@@ -19,7 +19,7 @@ export class AnimationBatch {
     /** Guard against premature animation execution. */
     private isReady = false;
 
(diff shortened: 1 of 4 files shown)


lsjroberts (approved):
Good find! lgtm

PR #1233, opened 2024-03-13
AG-11013 Implement tooltip position type node for map shape series

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11013

We need to defer this midpoint calculation because it's really expensive - possibly we could refactor here not rely on `midPoint` and always ask the series what the value is


alantreadway (approved):
LGTM

PR #1247, opened 2024-03-14
AG-10842 Update bar, range bar, histogram and boxplot corner radii

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10842


alantreadway (approved):
LGTM - thanks!

PR #1248, opened 2024-03-14
AG-11011 Update range area and range bar gallery examples to snap crosshairs

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-11011


alantreadway (approved):
LGTM

PR #1251, opened 2024-03-14
AG-8416 Per series tooltip range option

olegat:
https://ag-grid.atlassian.net/browse/AG-8416

This is required set the default to `'exact'` for bar series and `'nearest'` for line series.

Tested on this example:

```
const options = {
  series: [
    { type: "bar", xKey: "quarter", yKey: "iphone", yName: "iPhone", tooltip: { range: 5 } },
    { type: "bar", xKey: "quarter", yKey: "mac", yName: "Mac" },
    { type: "line", xKey: "quarter", yKey: "ipad", yName: "iPad" },
    { type: "bar", xKey: "quarter", yKey: "wearables", yName: "Wearables" },
    { type: "bar", xKey: "quarter", yKey: "services", yName: "Services", tooltip: { range: 'nearest' } },
  ],
};
```


packages/ag-charts-community/src/chart/series/cartesian/barSeriesModule.ts
```diff
@@ -31,6 +31,7 @@ export const BarSeriesModule: SeriesModule<'bar'> = {
             },
         ],
     },
+    tooltipDefaults: { range: 'exact' },
```
alantreadway, line 32: Do we need a distinct `tooltipDefaults` options, rather than just specifying this in `themeTemplate.series.tooltip`? I think this would also mean we don't need special handling for falling back to defaults?

olegat: Using `themeTemplate.series.tooltip` would require us to add `series[].tooltip.range` to the options contract and expose this option in the API. This is undesirable; because it would require us to support the API feature indefinitely or removing at some point (breaking change). Using a distinct `tooltipDefaults` option gives us flexibility to remove/change it as we want.

We could perhaps add it into something generalised like `internalThemeTemplate`, but it unclear right now at this whether that's necessary.

olegat:
Closing PR. We agreed to delay the `tooltip.range` feature for some series until the next major release. Therefore retaining the default behaviour in the next minor release is no longer needed.

PR #1253, opened 2024-03-14
AG-11006 Cache label placement for map shape series

jacobp100:
Computes the label placement on a fixed scale rather than using the current scale, then converts back to the current scale. This means we can cache the label placement values between resizes


alantreadway (approved):
LGTM

PR #1256, opened 2024-03-14
AG-10923 - Fix phantom highlight on zoom selection.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-10923

Fixes phantom highlight on zoom selection - this seems to have been a race condition between the `Chart.performUpdate()` getting to the `SERIES_UPDATE` step vs. `requestAnimationFrame()` callback for the pointer handling.

To mitigate this, I've added tracking of the `Chart.performUpdate()` cycle and update state so that we don't try to update the highlight in the same processing cycle if still running.


olegat:
LGTM for the next release.

It would be good to understand why never saw this race condition in the past to ensure that we do not instabilities elsewhere, but we may not have time for that.

PR #1260, opened 2024-03-15
AG-11006 Improve label placement performance

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11006

- Ensures `maxValue` bound is greater than all children bounds (a requirement for the algorithm, we did this in 99% of cases, now it's 100%)
- Fix ordering of candidates to try so the ones with the most potential get tried first - rather than the least. This actually fixes a possible infinite loop


alantreadway (approved):
LGTM

PR #1266, opened 2024-03-15
AG-10955 Use hierarchy colours for map backgrounds

jacobp100:
Spoke with design - they'd like to use the second fill and stroke we use for non-leaf nodes in the treemap series for consistency

image
image


alantreadway (approved):
LGTM

PR #1270, opened 2024-03-15
Add gallery thumbnail sanity checks.

alantreadway:
Since we've had several breaks of thumbnails in the past few releases, lets add some minimal coverage so we are aware of breaks earlier.

This change adds `nx test:thumbnails ag-charts-website`, which will also run as part of `nx test ag-charts-website` and `nx blt`, as well as on CI.


taktran (approved):
Cool. Not sure if it's in here, but what would the process/command be to generate a new thumbnail if there is expected to be a change?

alantreadway:
> Cool. Not sure if it's in here, but what would the process/command be to generate a new thumbnail if there is expected to be a change?

It is identical to all other cases, run with config `update` - this is the command to update all snapshots:
```
nx test -c update
```

alantreadway:
Re-opened as https://github.com/ag-grid/ag-charts/pull/1299 due to issues with the branch.

PR #1286, opened 2024-03-19
CRT-271 Use canvas width/height as defaults

olegat:
https://ag-grid.atlassian.net/browse/CRT-271

This is a regression caused by the fix for [CRT-218](https://ag-grid.atlassian.net/browse/CRT-218). To avoid this problem in the future, `width` and `height` are explicitly defined as `number` types so that the compiler can spot `undefined` values.


alantreadway (approved):
LGTM

PR #1288, opened 2024-03-19
CRT-262 Fix zoom on vertical category axis inverted range

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-262

This was caused by the flipping of the axis range for vertical axes on the left/right positions, but then not also flipping the visible range.


olegat:
Snapshots may need to be update to reflect this behavioural change. Otherwise, LGTM

lsjroberts:
Snapshots updated, thanks @olegat.

For horizontal bars the x-axis in zooming refers to the vertical one as they are flipped, and it defaults to `anchorPointX: end` which means the top of the axis in this case.

alantreadway (approved):

PR #1304, opened 2024-03-20
Reduce highlight and annotation layers.

alantreadway:
Updates all series highlight + annotation layers to use a single layer for each, and adds dynamic index support for the highlight layer case.


iMoses (approved):
LGTM

PR #1310, opened 2024-03-20
AG-11084 - Only force legend grid layout result if both width and hei…

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-11084


alantreadway (approved):
LGTM

PR #1312, opened 2024-03-21
AG-11080 Fix zoom range example with defined dates

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-11080


alantreadway (approved):
LGTM

PR #1319, opened 2024-03-21
AG-10196 Change cornerRadiusBBox to clipBBox

jacobp100:
Changes `cornerRadiusBBox` - where you passed a bigger box than the rect - to clipBBox - where you pass a smaller box. This aligns better with most conventions out there (we already use clipping a lot), and will more closely match how the sector will eventually handle this clipping

CC https://ag-grid.atlassian.net/browse/AG-10196


alantreadway (approved):
LGTM

PR #1323, opened 2024-03-22
Cleanup CartesianSeries subGroup support.

alantreadway:
Removes the entirely redundant `subGroup` support from `CartesianSeries` as a prerequisite to wider rendering changes being contemplated by @iMoses.


packages/ag-charts-community/src/chart/axis/__image_snapshots__/grouped-category-axis-test-ts-grouped-category-axis-examples-for-integrated-charts-grouped-category-axis-example-flip-it-should-render-to-canvas-as-expected-1-snap.png
olegat, line 1: It's unclear whether these two updated snapshots are truly rendering the expected result. Can you confirm that the old snapshot are incorrect and that the new ones are correct?

olegat:
LGTM assuming snapshots are correct.

packages/ag-charts-community/src/chart/axis/__image_snapshots__/grouped-category-axis-test-ts-grouped-category-axis-examples-for-integrated-charts-grouped-category-axis-example-flip-it-should-render-to-canvas-as-expected-1-snap.png
alantreadway: The difference is the bar rendering, which is a consequence of the base example not being quite correct WRT the expected Integrated Charts contract - irrespective, the test is really focusing on the axis rendering, which is identical in both cases

PR #1327, opened 2024-03-25
AG-6576 Collect canvas statistics in benchmark tests

timkendrick:
**Notes**

- This extends the existing `heapUsed` benchmark statistic with the following additional sibling statistics:
  - `canvasCount`: number of active canvases present after the test has run
  - `canvasBytes`: estimated combined memory usage of canvas instances (assumes canvas pixel data is stored as raw 32-bit bitmap data)
  - `memoryUsage`: Combined total of `heapUsed` and `canvasBytes`
- Benchmark charts have been updated to plot `memoryUsage` rather than `heapUsed`
- Previously the benchmark chart data only plotted the memory _change_ over the course of the test (i.e. heap size before/after diff), whereas now it shows the total 'after' heap size
  - This was done to prevent misleadingly small figures for 'incremental' interaction test charts (can easily be changed if necessary)
- The benchmarking code uses `WeakRef`s to track the allocated `Canvas` instances that are still referenced in memory after the test run has finished


alantreadway (approved):
LGTM

PR #1331, opened 2024-03-26
AG-11125 Fix highlight layer not being clipped

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11125

If you'd prefer, I can keep this logic in `BaseLayoutProcessor` by extending `ChartLike` to include these two layers - let me know


alantreadway (approved):
LGTM - thanks for fixing this.

PR #1333, opened 2024-03-26
AG-10154 Make react examples more idiomatic

jacobp100:
https://ag-grid.atlassian.net/browse/AG-10154

- Use the function naming format `setXY` rather than `useXY`
- Use `structuredClone(options)` rather than shallow cloning
- Simplify a few examples where it makes sense


alantreadway (approved):
LGTM

PR #1342, opened 2024-03-28
AG-11165 Fix nightingale with corner radius for negative values

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11165


packages/ag-charts-community/src/scene/shape/__image_snapshots__/sector-test-ts-sector-clipping-should-render-as-expected-1-snap.png
jacobp100, line 1: Four bugs in the bottom left - will fix separately

alantreadway (approved):
LGTM

PR #1346, opened 2024-03-28
AG-11142 Fix framework generator for financial charts examples

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11142

### DIFF
--- a/packages/ag-charts-website/src/content/docs/animation/_examples/data-updates/index.html
+++ b/packages/ag-charts-website/src/content/docs/animation/_examples/data-updates/index.html
@@ -22,5 +22,5 @@
             Toggle Ticking Updates
         </button>
     </div>
-    <div id="myChart" style="flex: 1"></div>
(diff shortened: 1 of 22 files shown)


packages/ag-charts-website/src/content/docs/animation/_examples/data-updates/index.html
```diff
@@ -22,5 +22,5 @@
             Toggle Ticking Updates
         </button>
     </div>
-    <div id="myChart" style="flex: 1"></div>
+    <div id="myChart" style="flex: 1; min-height: 0"></div>
```
jacobp100, line 25: Some of the frameworks are setting `height: 100%` on their divs, which is causing layouts to break (https://defensivecss.dev/tip/flexbox-min-content-size/#long-card-title). We have a ticket to change how the autosizing works, and when we do that, we can remove these min-heights

alantreadway (approved):
LGTM

PR #1348, opened 2024-03-28
AG-11149 Allow specifying range buttons

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11149

Introduces a concept of button groups, which are a pre-filled set of buttons. When we allow adding custom buttons to the toolbar, you'd place these button groups in amongst your buttons


lsjroberts (approved):
LGTM

PR #1358, opened 2024-04-03
AG-11149 Change options contract for zoom range buttons

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11149


lsjroberts (approved):
LGTM

PR #1388, opened 2024-04-08
AG-11130 Add annotations to toolbar and docs

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-11130


packages/ag-charts-enterprise/src/features/annotations/annotations.ts
```diff
@@ -186,7 +259,90 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
...
+
+        const editingStepFn = this.editingStepFns[editingType]?.[editingStep];
+        if (editingStepFn) {
+            editingStepFn(datum, node, point, offset);
+        }
+
+        updateService.update();
+    }
+
+    private editingStepFns: Partial<Record<AnnotationType, Partial<Record<EditingStep, EditingStepFn>>>> = {
```
lsjroberts: I'm not massively happy with this, but using a switch statement with `case AnnotationType.Line + EditingStep.End:` resulted in function complexity failure. I may refactor this more when we add more annotation types and have a better feel for it.

packages/ag-charts-community/src/chart/annotation/annotationManager.ts
```diff
@@ -11,4 +15,13 @@ export class AnnotationManager {
             return this;
         };
     }
+
+    public setAnnotationStyles(styles: any) {
```
alantreadway: Could we use the options type here instead of `any`?

packages/ag-charts-enterprise/src/features/annotations/annotations.ts
```diff
@@ -186,7 +264,90 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
...
+                });
+            },
+        },
+    };
+
+    private onClick(event: _ModuleSupport.PointerInteractionEvent<'click' | 'drag-start'>) {
+        if (this.addingStep == null) {
+            this.onClickSelecting();
+        } else if (event.type === 'click') {
+            this.onClickAdding(event as _ModuleSupport.PointerInteractionEvent<'click'>);
```
alantreadway: Nit: Is the cast needed?

packages/ag-charts-enterprise/src/features/annotations/annotations.ts
```diff
@@ -36,25 +52,48 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
...
+            this.annotations.nodes()[this.active].toggleActive(false);
+            this.active = undefined;
+        }
+
+        if (event.section !== 'annotations') return;
+
+        this.ctx.interactionManager.pushState(InteractionState.Annotations);
+
+        this.addingStep = AddingStep.Start;
+        this.addingType = event.value as AnnotationType;
```
alantreadway: Could/should we validate this rather than forcing a cast with `as`?

packages/ag-charts-enterprise/src/features/annotations/scenes/channel.ts
```diff
@@ -52,9 +52,20 @@ export class Channel extends Annotation {
...
+                topRight: show,
+                bottomLeft: show,
+                bottomMiddle: show,
+                bottomRight: show,
+            };
+        }
+
         for (const [handle, node] of Object.entries(this.handles)) {
-            node.visible = show;
+            node.visible = show[handle as ChannelHandle] ?? true;
```
alantreadway, line 68: Possible to avoid the forced cast here?

alantreadway (commented):
A few nits, but otherwise LGTM

packages/ag-charts-enterprise/src/features/annotations/annotations.ts
lsjroberts: Added a static `isPointerEvent()` method to the interaction manager which restricts the type.

packages/ag-charts-enterprise/src/features/annotations/scenes/channel.ts
lsjroberts: Not without creating lots of extra noise that doesn't really assert any extra safety.

alantreadway (approved):
LGTM

PR #1390, opened 2024-04-08
AG-11264 - Use `ag-website-shared` components

taktran:
* Migrate icons, search box and pricing page to https://github.com/ag-grid/ag-website-shared


packages/ag-charts-website/src/ag-website-shared/.gitrepo
alantreadway, line 1: I'm not really a fan of the nesting here - can we skip the top level folders and just pull in `src/` to:
- avoid all of the unecessary and potentially confusing configuration files in the root
- avoid repeated `src` folder in the path 

Alternatively/additionally, could we maybe pull this in somewhere more obvious in the repo structure, e.g.:
`{workspareRoot}/external/ag-website-shared`?

taktran: I can have a look at trying `{workspareRoot}/external/ag-website-shared` or similar. I'll need to see how imports work, or whether we'll need to do some extra npm/file linking. Hot reloading will also prob need an update, since it's outside of the website folder. The benefit is that it will only need to be done once in this setup phase.

> avoid all of the unecessary and potentially confusing configuration files in the root

I'm hoping no one apart from the website team needs to look in this folder, so it won't be as confusing, as they will know why it's there.

> avoid repeated src folder in the path

I tried to mitigate the problem by updating the import aliases in`tsconfig.json` to have `"@ag-website-shared/*": ["src/*"]`, so imports ignore the extra `src` path eg, `@ag-website-shared/components/icon/Icon`. The only place that requires the actual path is `markdoc.config.ts` ie, `component('./src/ag-website-shared/src/components/icon/EnterpriseIcon', 'EnterpriseIcon')`. Most consumers of the shared folder won't need to know about it.

taktran: Turns out, it was quite easy to move to `{workspareRoot}/external/ag-website-shared`. Just moved the files and updated the references to point to it. Changing the `tsconfig.json` appears to give us hot reloading for free (vite/astro probably knows how to handle it).

I do agree, it's a better home for the files.

alantreadway: Sounds great - thank you!

alantreadway (approved):

PR #1395, opened 2024-04-09
AG-11272 Move ModuleContext values to chartContext.ts

olegat:

This is prep-work for screen-readers, because we'll probably need a new service/manager for updating Aria tags.


alantreadway (approved):
LGTM

PR #1400, opened 2024-04-11
AG-11268 Update toolbar styles to use grid variables

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-11268


packages/ag-charts-community/src/styles/theme-default-dark.ts
alantreadway, line 1: Is this identical to the `.css` file adjacent? Should we allow the `.css` files to be imported directly?

packages/ag-charts-community/src/styles/theme-default.ts
alantreadway, line 1: Ditto

alantreadway (approved):
LGTM

PR #1401, opened 2024-04-11
AG-11251 - Improve default time format calculation.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-11251

Attempts to address default time axis label inconsistencies, and cleanup the logic to improve readability.


manapeirov (approved):
LGTM thanks Alan

PR #1402, opened 2024-04-11
AG-11272 Add aria announcement service

olegat:

This adds very basic aria screen reader announcement for focus elements.


alantreadway (approved):
LGTM

PR #1404, opened 2024-04-11
AG-11251 - Fix single date datapoint axis labels.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-11260

Attempts to fix the current bug of a single `Date` data-point causing ticks with labels of the form `.000` and `.001` to be displayed, rather than a single 'short month' label.


manapeirov (approved):
LGTM

PR #1406, opened 2024-04-12
docs tweaks plus placeholder accessibility page

AG-DavidG:


alantreadway (approved):
LGTM, assuming we're going to tailor the accessibility page content to the scope we can deliver for 9.3.0

PR #1408, opened 2024-04-12
Cleanup Matrix and use native DOMMatrix where possible

iMoses:
+ add DOMMatrix polyfill for server-side
+ cleanup document/window direct access on server-side


packages/ag-charts-enterprise/src/license/licenseManager.ts
alantreadway, line 1: We should avoid making changes here, as the code is shared with Grid - are these changes strictly needed?

packages/ag-charts-enterprise/src/setup.ts
```diff
@@ -80,9 +80,6 @@ export function setupEnterpriseModules() {
     );
 
     _ModuleSupport.enterpriseModule.isEnterprise = true;
-    _ModuleSupport.enterpriseModule.licenseManager = (options: AgChartOptions) =>
-        new LicenseManager(
-            options.container?.ownerDocument ?? (typeof document === 'undefined' ? undefined : document)
-        );
+    _ModuleSupport.enterpriseModule.licenseManager = new LicenseManager();
```
alantreadway: As per my other comment, I would prefer to keep the `document` resolution out of `LicenseManager` for now?

alantreadway (approved):
LGTM other than some nits.

alantreadway (approved):
LGTM

PR #1421, opened 2024-04-16
AG-11335 Assertive aria announcer

olegat:
https://ag-grid.atlassian.net/browse/AG-11335


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -591,7 +591,7 @@ export abstract class Chart extends Observable implements AgChartInstance {
 
                 const { seriesRect } = this;
                 await Promise.all(seriesToUpdate.map((series) => series.update({ seriesRect })));
-                setAriaLabel(this.ctx.scene.canvas.container, this.getAriaLabel());
+                setAttribute(this.ctx.scene.canvas.container, 'aria-label', this.getAriaLabel());
```
alantreadway: Nit: I think you can just use `this.element` as the first argument here, as it's the same element?

alantreadway (commented):
LGTM other than one nit

packages/ag-charts-community/src/chart/chart.ts
olegat: Done.

alantreadway (approved):
LGTM

PR #1423, opened 2024-04-16
AG-11315 - Add support for specifying tabIndex.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-11315

Adds `tabindex="0"` to the Chart `<canvas />` when `keyboard.enabled: true`, and allows user override via a new `tabIndex` root-level option.

### DIFF
--- a/packages/ag-charts-community/src/options/chart/chartOptions.ts
+++ b/packages/ag-charts-community/src/options/chart/chartOptions.ts
@@ -193,6 +193,4 @@ export interface AgBaseChartOptions<TDatum = any> extends AgBaseThemeableChartOp
      * __Important:__ Make sure to read the `autoSize` config description for information on how the container element affects the chart size (by default).
      */
     container?: HTMLElement | null;
-    /** HTML tabIndex to apply to chart canvas to allow keyboard navigation. Defaults to 0. */
-    tabIndex?: number;
 }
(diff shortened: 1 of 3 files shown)


packages/ag-charts-community/src/options/chart/chartOptions.ts
```diff
@@ -193,4 +193,6 @@ export interface AgBaseChartOptions<TDatum = any> extends AgBaseThemeableChartOp
      * __Important:__ Make sure to read the `autoSize` config description for information on how the container element affects the chart size (by default).
      */
     container?: HTMLElement | null;
+    /** HTML tabIndex to apply to chart canvas to allow keyboard navigation. Defaults to 0. */
+    tabIndex?: number;
```
olegat: I'm unsure whether we want to expose this in the options contract just yet. `keyboard.enabled` is undocumented but defaults to `false`, so this option could be confusing because by default it will set `tabindex: -1` and ignore keyboard events.

olegat (changes requested):
LGTM - apart from the added option in `chartOptions.ts`

packages/ag-charts-community/src/options/chart/chartOptions.ts
alantreadway: I've removed the options update, and moved the new (hidden) option to `keyboard.tabIndex` for the moment - this should hopefully make the relationship a little clear, subject to final options naming.

PR #1425, opened 2024-04-16
AG-11268 Toolbar styles, visibility and font icons

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-11268


packages/ag-charts-community/src/chart/toolbar/toolbar.ts
```diff
@@ -36,106 +45,113 @@ export class Toolbar extends BaseModuleInstance implements ModuleInstance {
         left: new Set(),
     };
 
+    private groupCallers: Record<ToolbarGroup, number> = {
+        annotations: 0,
+        ranges: 0,
+    };
```
lsjroberts, line 51: A count of how many modules have tried to enable each group. If `> 0` and `toolbar.[group].enabled === true` then it will show. This allows toolbars and groups to be toggled by the combined boolean of `toolbar.enabled`, each group enabled flag and at least one of the linked modules's enabled flags.

alantreadway (approved):
LGTM

PR #1429, opened 2024-04-17
AG-11315 - Align element where focus-related listeners and tabindex are applied.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-11315

Aligns the element where we set `tabindex="0"` and where we add focus-related event listeners, so that tab-handling behaviour is correct for Standalone AND Integrated cases.


olegat:
LGTM (assuming that it works as intended, I haven't test this change)

Has this change been tested on our examples Plunkers? https://ag-grid.atlassian.net/browse/AG-11190?focusedCommentId=62114

PR #1435, opened 2024-04-18
AG-8448 Fix zoom toolbar button disabling

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-8448

The zoom module can be enabled, and tries to enable the toolbar buttons, before the toolbar module is enabled and the buttons have been added to the dom. So this captures the pending events and iterates them when the toolbar module is enabled.


alantreadway (approved):
LGTM

PR #1436, opened 2024-04-18
AG-11338 - Implement CSS variables for focus indicator styling.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-11338

First steps to implementing CSS-based styling for the focus indicator.

### DIFF
--- a/packages/ag-charts-community/src/chart/interaction/regionManager.ts
+++ b/packages/ag-charts-community/src/chart/interaction/regionManager.ts
@@ -1,7 +1,8 @@
 import type { BBox } from '../../scene/bbox';
-import { getDocument } from '../../util/dom';
+import { getDocument, injectStyle } from '../../util/dom';
 import { Listeners } from '../../util/listeners';
 import { buildConsumable } from './consumableEvent';
+import * as focusStyles from './focusStyles';
 import type { InteractionManager, PointerInteractionEvent, PointerInteractionTypes } from './interactionManager';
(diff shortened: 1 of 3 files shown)


packages/ag-charts-community/src/chart/interaction/regionManager.ts
```diff
@@ -74,14 +75,12 @@ export class RegionManager {
...
         this.focusIndicator = getDocument()?.createElement('div');
         if (this.focusIndicator !== undefined && element !== undefined) {
-            this.focusIndicator.className = 'ag-charts-focusindicator';
-            this.focusIndicator.style.position = 'absolute';
-            this.focusIndicator.style.border = '2px solid red';
-            this.focusIndicator.style.display = 'none';
-            this.focusIndicator.style.pointerEvents = 'none';
-            this.focusIndicator.style.userSelect = 'none';
+            this.focusIndicator.classList.add(focusStyles.block);
+            this.focusIndicator.classList.add(focusStyles.elements.indicator);
```
lsjroberts: nit: under BEM you'd have elements as a child of the block, not have the block and element on the same node, doesn't really matter but you can probably ditch the element as it doesn't have any styles/behaviour as far as I can see?

lsjroberts (approved):
lgtm, with minor nit

packages/ag-charts-community/src/chart/interaction/regionManager.ts
alantreadway: Fair point, removed.

PR #1441, opened 2024-04-18
AG-11272 Screen reader changes

olegat:

### DIFF
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -387,7 +387,12 @@ export abstract class Chart extends Observable implements AgChartInstance {
     }
 
     getAriaLabel(): string {
-        return [this.title.text, this.subtitle.text, this.footnote.text].filter((v) => v).join('. ');
+        const captionText = [this.title, this.subtitle, this.footnote]
+            .filter((caption) => caption.enabled && caption.text)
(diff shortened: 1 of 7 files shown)


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -591,7 +596,8 @@ export abstract class Chart extends Observable implements AgChartInstance {
 
                 const { seriesRect } = this;
                 await Promise.all(seriesToUpdate.map((series) => series.update({ seriesRect })));
-                setAttribute(this.element, 'aria-label', this.getAriaLabel());
+                setAttribute(this.ctx.scene.canvas.element, 'aria-label', this.getAriaLabel());
```
alantreadway: Does this need to be on the canvas element itself, or could we add a more specialised element as a child of `this.element` for this purpose?

Also do we need to perform these updates here? Perhaps it would be better to follow a similar pattern as for `this.updateTabIndex();` earlier in the update pipeline here?

olegat: Moving this to `updateTabIndex()` sounds sensible. I've done this.

I don't fully understand why my screen reads out the label twice when it's added to `this.element`. We could add a separate child of `this.element`, but I do not know whether that will work.

alantreadway (approved):
LGTM

PR #1443, opened 2024-04-18
AG-11192 TC6 Ignore mouse event when browser focus is absent

olegat:


alantreadway (approved):
LGTM

PR #1444, opened 2024-04-18
Hide alpha features on website

alantreadway:
Allows explicit hiding of development-related site pages from the Algolia and external (Google) search crawling indexes.


taktran (approved):
Looks good 👍🏼

PR #1449, opened 2024-04-19
AG-11291 - Ensure Event.preventDefault() on consumed keyboard events.

alantreadway:
Fixes containing page scrolling when using up/down arrows to navigate a chart, for in-line docs examples.

The core of the change is to ensure that `ConsumableEvent`s percolate back a `Event.preventDefault()` on consumption.

### DIFF
--- a/packages/ag-charts-community/src/chart/interaction/consumableEvent.ts
+++ b/packages/ag-charts-community/src/chart/interaction/consumableEvent.ts
@@ -12,12 +12,17 @@ type TypedListener<TType extends string, TEvent extends TypedConsumableEvent<TTy
     (event: TEvent) => void
 >;
 
-export function buildConsumable<T>(obj: T): T & ConsumableEvent {
+export function buildConsumable<T>(obj: T & { sourceEvent?: Event | ConsumableEvent }): T & ConsumableEvent {
     const builtEvent = {
         ...obj,
         consumed: false,
         consume() {
             builtEvent.consumed = true;
+            if (obj.sourceEvent instanceof Event) {
+                obj.sourceEvent?.preventDefault();
+            } else {
+                obj.sourceEvent?.consume();
+            }
         },
     };
     return builtEvent;
--- a/packages/ag-charts-community/src/chart/interaction/regionManager.ts
+++ b/packages/ag-charts-community/src/chart/interaction/regionManager.ts
@@ -245,8 +245,12 @@ export class RegionManager {
     }
 
     private dispatchTabStart(event: KeyNavEvent<'tab'>): boolean {
-        const { delta, interactionEvent } = event;
-        const startEvent: KeyNavEvent<'tab-start'> = buildConsumable({ type: 'tab-start', delta, interactionEvent });
+        const { delta, sourceEvent } = event;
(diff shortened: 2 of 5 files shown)


packages/ag-charts-community/src/chart/interaction/consumableEvent.ts
```diff
@@ -12,12 +12,17 @@ type TypedListener<TType extends string, TEvent extends TypedConsumableEvent<TTy
     (event: TEvent) => void
 >;
 
-export function buildConsumable<T>(obj: T): T & ConsumableEvent {
+export function buildConsumable<T>(obj: T & { sourceEvent?: Event | ConsumableEvent }): T & ConsumableEvent {
```
olegat, line 15: The change to this function is highly likely to break something with interactions.

The `consume()` function is intended to stop other listeners from receiving events. There are almost certainly some scenarios where we want to both call `preventDefault()` and have multiple listeners to do handle the event. So mixing up the behaviour of `consume()` / `preventDefault()` is not a good idea.

Instead, I think it would be better to have a simple solution like calling `preventDefault()` directly when an arrow key is pressed.

packages/ag-charts-community/src/chart/interaction/regionManager.ts
```diff
@@ -307,7 +311,7 @@ export class RegionManager {
             if (newRegion === undefined || !newRegion.properties.canInteraction()) {
                 this.updateFocusIndicatorRect(undefined);
             } else {
-                event.interactionEvent.sourceEvent.preventDefault();
+                event.sourceEvent.consume();
```
olegat, line 314: See other GitHub comment. Instead of changing the behaviour of `consume()`, it would be better to keep this line as-is, but add something similar in `RegionManager.onNav()`:

```diff
     private onNav(event: KeyNavEvent<'blur' | 'nav-hori' | 'nav-vert' | 'submit'>) {
         const focusedRegion = this.getTabRegion(this.currentTabIndex);
+        if (event.type !== 'blur') {
+            event.interactionEvent.sourceEvent.preventDefault();
+        }
         this.dispatch(focusedRegion, event);
    }
```

olegat (changes requested):

packages/ag-charts-community/src/chart/interaction/consumableEvent.ts
alantreadway: I don't believe that we should have multiple listeners consuming the same consumable event - that is the whole point of it being a consumable event, it's only consumed once.

Calling `preventDefault()` for this very specific case (up/down arrows) is the wrong approach IMO, as it's a general principle that if we're consuming an event, the browsers default handling should not kick in.

olegat:
Sync'd on video call. We've agreed to accept the risk and split off the `consume()` / `preventDefault()` functionality on a case-by-case basis if need be.

packages/ag-charts-community/src/chart/interaction/consumableEvent.ts
alantreadway: Discussed offline.

olegat (approved):

PR #1451, opened 2024-04-19
AG-11291 - Fix focus indicator positioning with fixed width/height chart.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-11291

Fixes positioning of the focus indicator when `width` or `height` are specified on the chart options.

### DIFF
--- a/packages/ag-charts-community/src/chart/interaction/regionManager.ts
+++ b/packages/ag-charts-community/src/chart/interaction/regionManager.ts
@@ -49,7 +49,8 @@ export interface RegionProperties {
 export class RegionManager {
     private currentTabIndex = 0;
     public readonly keyNavManager: KeyNavManager;
-    private readonly focusIndicator?: HTMLDivElement;
+    private readonly focusWrapper: HTMLDivElement;
+    private readonly focusIndicator: HTMLDivElement;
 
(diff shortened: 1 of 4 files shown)


packages/ag-charts-community/src/chart/interaction/regionManager.ts
```diff
@@ -76,12 +78,14 @@ export class RegionManager {
         );
 
         injectStyle(focusStyles.css, focusStyles.block);
-        this.focusIndicator = getDocument()?.createElement('div');
```
olegat, line 79: Note: I vaguely recall having an `undefined` document in one of the tools (perhaps it was the thumbnail generator, but I do not remember). Just letting you know in case something in the CI fails.

olegat:
LGTM.

olegat (approved):

packages/ag-charts-community/src/chart/interaction/regionManager.ts
alantreadway: Tested locally, I can verify that the thumbnail generate successfully.

FYI: The intent is that in all of these non-browser cases that the tooling provide an appropriate substitute for these APIs, which is why `getDocument()` exists specifically.

PR #1454, opened 2024-04-19
AG-10879 - Add default theme palette colors for candlestick and ohlc …

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-10879


packages/ag-charts-community/src/chart/themes/chartTheme.ts
```diff
@@ -358,6 +359,13 @@ export class ChartTheme {
         return deepClone(themeInstance);
     }
 
+    protected static getDefaultColors(): { fills: { [key: string]: string }; strokes: { [key: string]: string } } {
```
alantreadway: Nit: We're repeating this (return) typing in a bunch of places, could we extract a `type` or `interface` and re-use it?

alantreadway (approved):
LGTM

PR #1457, opened 2024-04-19
AG-11367 Add zoom toolbar buttons proxy in zoom module

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-11367

Includes some refactoring of the zoom module to separate out context menu and toolbar actions.

See [17358ad](https://github.com/ag-grid/ag-charts/pull/1457/commits/17358adffaa564cf1a2aee04dcbaaab76bd36890) for just this ticket.

### DIFF
--- a/packages/ag-charts-community/src/chart/toolbar/toolbar.ts
+++ b/packages/ag-charts-community/src/chart/toolbar/toolbar.ts
@@ -5,7 +5,12 @@ import type { BBox } from '../../scene/bbox';
 import { createElement, injectStyle } from '../../util/dom';
 import { ObserveChanges } from '../../util/proxy';
 import { BOOLEAN, Validate } from '../../util/validation';
-import type { ToolbarButtonToggledEvent, ToolbarGroupToggledEvent } from '../interaction/toolbarManager';
+import type { PointerInteractionEvent } from '../interaction/interactionManager';
+import type {
+    ToolbarButtonToggledEvent,
+    ToolbarGroupToggledEvent,
+    ToolbarProxyGroupOptionsEvent,
+} from '../interaction/toolbarManager';
 import { ToolbarGroupProperties } from './toolbarProperties';
 import * as styles from './toolbarStyles';
 import {
@@ -14,7 +19,7 @@ import {
     TOOLBAR_POSITIONS,
     type ToolbarButton,
     type ToolbarGroup,
-    type ToolbarPosition,
+    ToolbarPosition,
 } from './toolbarTypes';
 
 export class Toolbar extends BaseModuleInstance implements ModuleInstance {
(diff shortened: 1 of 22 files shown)


packages/ag-charts-enterprise/src/features/zoom/zoom.ts
```diff
@@ -70,6 +104,18 @@ export class Zoom extends _ModuleSupport.BaseModuleInstance implements _ModuleSu
...
+        if (!buttons) return;
+        target.ctx.toolbarManager.proxyGroupOptions('zoom', buttons.toJson());
+    })
+    public buttons = new ZoomButtonsProperties(() => {
+        if (!this.buttons) return;
+        this.ctx.toolbarManager.proxyGroupOptions('zoom', {
+            enabled: this.buttons.enabled,
+            buttons: this.buttons.buttons,
+        });
+    });
```
lsjroberts, line 117: Zoom fires the `proxyGroupOptions()` event...

packages/ag-charts-community/src/chart/toolbar/toolbar.ts
```diff
@@ -145,13 +163,56 @@ export class Toolbar extends BaseModuleInstance implements ModuleInstance {
...
+        const { group, options } = event;
+
+        this.groupProxied.add(group);
+
+        this.createGroup(group);
+        this.createGroupButtons(group, options.buttons);
+        this.toggleGroup(group, options.enabled);
+
+        this[group].set(options);
+    }
```
lsjroberts, line 180: ... which is handled in toolbar here

packages/ag-charts-community/src/util/listeners.ts
```diff
@@ -45,6 +52,21 @@ export class Listeners<EventType extends string, EventHandler extends Handler> {
...
+    public dispatchAlwaysResolve(eventType: EventType, ...params: Parameters<EventHandler>): void {
+        if (this.getListenersByType(eventType).length === 0) {
+            this.pendingDispatches.push({ eventType, params });
+            return;
+        }
+
+        this.pendingDispatches = this.pendingDispatches.filter((pending) => pending.eventType !== eventType);
+
+        this.dispatch(eventType, ...params);
+    }
```
lsjroberts: And worth highlighting this change, which handles the situation when `zoom.buttons` is set before the toolbar has added it's listener to `proxy-group-options`.

packages/ag-charts-community/src/util/listeners.ts
```diff
@@ -19,6 +20,12 @@ export class Listeners<EventType extends string, EventHandler extends Handler> {
             this.registeredListeners.set(eventType, [record]);
         }
 
+        for (const pending of this.pendingDispatches) {
+            if (pending.eventType === eventType) {
+                this.dispatch(eventType, ...pending.params);
+            }
+        }
```
alantreadway: This feels a little sketchy, as if you have two listeners registering for the events, only one of them will receive the pending events.

Can we sync on Monday about the approach here, I think it might be better to defer some options propagation to the `ChartUpdateType.PERFORM_LAYOUT` rendering lifecycle phase, which might 🤞 mean we don't have to solve race conditions at options injection time?

lsjroberts: I haven't tested that situation, but it should send it to both listeners. It only clears the pending events when a new one of the same type is dispatched to `dispatchAlwaysResolve()`.

alantreadway (approved):
LGTM - thanks for the updates!

PR #1460, opened 2024-04-22
AG-11290 - Stop showing tooltip for invalid BarSeries datums.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-11290

Fixes display of invalid BarSeries datums, and plumbs in support generally for handling of invalid datums in all node search mechanisms.

Side-note: There seemed to be several spots where we were unnecessarily using a `Required<>` wrapper around `CartesianSeriesNodeDatum` when only `point` or `midPoint` were actually needing to be made required - I've removed that where necessary to avoid needing to always add a `valid` property to all cartesian series datums.

### DIFF
--- a/packages/ag-charts-community/src/chart/series/seriesTypes.ts
+++ b/packages/ag-charts-community/src/chart/series/seriesTypes.ts
@@ -45,6 +45,7 @@ export interface SeriesNodeDatum {
     readonly itemId?: any;
     readonly datum: any;
     readonly point?: Readonly<SizedPoint>;
+    readonly valid?: boolean;
     midPoint?: Readonly<Point>;
 }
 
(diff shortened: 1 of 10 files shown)


packages/ag-charts-community/src/chart/series/seriesTypes.ts
```diff
@@ -45,6 +45,7 @@ export interface SeriesNodeDatum {
     readonly itemId?: any;
     readonly datum: any;
     readonly point?: Readonly<SizedPoint>;
+    readonly valid?: boolean;
```
olegat: nit: [isNA](https://en.wikipedia.org/wiki/N/A) might be a more suitable name to indicate that this is a missing data point.

olegat (approved):
LGTM aside from optional nit.

packages/ag-charts-community/src/chart/series/seriesTypes.ts
alantreadway: Discussed offline, renamed to `missing` since `NA` isn't a term we use in the existing code.

PR #1462, opened 2024-04-22
AG-11362 - Aligns tooltip renderer callback parameters with options contract.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-11362

The original issue was around treemap tooltip renderer, but on making the tooltip callback parameter typings stricter to flush out other gaps, turns out there were actually quite a few gaps - this PR corrects those gaps.


manapeirov (approved):
LGTM

PR #1464, opened 2024-04-22
AG-10852 - Fix inconsistency between series when checking scale type …

manapeirov:
PR for a bug found when creating gallery example https://ag-grid.atlassian.net/browse/AG-10852


alantreadway (approved):
LGTM - thanks @manapeirov!

PR #1465, opened 2024-04-22
AG-11192 Bugfix: Do not focus on missing data points

olegat:


packages/ag-charts-community/src/chart/series/dataModelSeries.ts
alantreadway, line 1: This looks like the right approach, but `datum.missing` may not be set on all series types where it is needed - might be worth sanity checking that before we can call this done?

(Specifically: bar, line and area series all support missing datums; for animation purposes, all bar-like series do too).

alantreadway: I'll follow up on this point.

alantreadway (approved):
LGTM

PR #1468, opened 2024-04-23
AG-11191 Fix keyboard pagination logic

olegat:
https://ag-grid.atlassian.net/browse/AG-11191 
Do not assume that each legend page has the same number of legend items. Instead, count the number of items every time. Now, Legend.focus.item refers to the index of the legend on the current page, rather than an item of Legend.data.


alantreadway (approved):
LGTM

PR #1475, opened 2024-04-23
AG-11360 Fix keyboard node clicks on all series

olegat:
https://ag-grid.atlassian.net/browse/AG-11360

Instead, just fire the nodeClicks event directly. We do not need to check whether the keyboard "pointer click" is in range because by we already know with node the use wants to click on.
    
This fixes nodeClicks on series types where the centre of the focus indicator fails the hit-test on the node shape (e.g. radial bars).


alantreadway (approved):
LGTM

PR #1476, opened 2024-04-23
AG-11391 Bugfix: clamp input datumIndex in pickFocus

olegat:
https://ag-grid.atlassian.net/browse/AG-11391

This fixes a bug where a console error would sometimes be printed when changing focus from one series (with more data) or another series (with less data) due to datumIndex being >= nodeData.length.


alantreadway (approved):
LGTM

PR #1478, opened 2024-04-23
AG-11192 Find enable datum in reverse when changing series focus

olegat:

This fixed a bug where the tooltip can show an undefined datum tooltip when switching the focus from one bar series (with more data) to another bar series (with less data).


alantreadway (approved):
LGTM

PR #1485, opened 2024-04-24
CRT-317 Fix toolbar buttons hover color transparency inherited from grid

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-317


alantreadway (approved):
LGTM

PR #1495, opened 2024-04-24
CRT-328 Fix sync to use chart.ctx.updateService

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-328

`chart.updateService` was moved to `chart.ctx.updateService` by 32d9744


alantreadway (approved):
LGTM

PR #1503, opened 2024-04-25
CRT-319 - Blur keyboard focus on animation start.

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-319

Makes a quick-fix for release to blur the keyboard focus on animation start.

### DIFF
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -360,7 +360,7 @@ export abstract class Chart extends Observable implements AgChartInstance {
             ctx.interactionManager.addListener('dblclick', (event) => this.onDoubleClick(event)),
             seriesRegion.addListener('hover', (event) => this.onMouseMove(event)),
             seriesRegion.addListener('leave', (event) => this.onLeave(event)),
-            seriesRegion.addListener('blur', (event) => this.onBlur(event)),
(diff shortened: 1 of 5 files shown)


packages/ag-charts-community/src/chart/interaction/keyNavManager.ts
```diff
@@ -37,7 +38,9 @@ export class KeyNavManager extends BaseManager<KeyNavEventType, KeyNavEvent> {
 
             interactionManager.addListener('blur', (e) => this.onBlur(e), InteractionState.All),
             interactionManager.addListener('focus', (e) => this.onFocus(e), InteractionState.All),
-            interactionManager.addListener('keydown', (e) => this.onKeyDown(e), InteractionState.All)
+            interactionManager.addListener('keydown', (e) => this.onKeyDown(e), InteractionState.All),
+
+            animationManager.addListener('animation-start', (e) => this.animationBlur(e))
```
olegat: This listener might be better suited as part of the `Chart` to avoid adding too much dependency amongst managers. The `Chart.onBlur` method can probably be reused too (by removing the unused `_event` parameter).

olegat:
Thanks for the change. This new approach is much cleaner. LGTM.

olegat (approved):

PR #1518, opened 2024-04-26
AG-11439 Implement MapMarkerSeries.computeFocusBounds

olegat:
https://ag-grid.atlassian.net/browse/AG-11439

### DIFF
--- a/packages/ag-charts-enterprise/src/series/map-marker/mapMarkerSeries.ts
+++ b/packages/ag-charts-enterprise/src/series/map-marker/mapMarkerSeries.ts
@@ -26,6 +26,7 @@ const {
     SeriesNodePickMode,
     Layers,
     valueProperty,
+    computeMarkerFocusBounds,
 } = _ModuleSupport;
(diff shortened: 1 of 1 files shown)


packages/ag-charts-enterprise/src/series/map-marker/mapMarkerSeries.ts
```diff
@@ -842,7 +843,61 @@ export class MapMarkerSeries
         );
     }
 
-    protected override computeFocusBounds(_opts: _ModuleSupport.PickFocusInputs): _Scene.BBox | undefined {
-        return undefined; // TODO
+    public getFormattedMarkerStyle(markerDatum: MapMarkerNodeDatum) {
```
olegat, line 846: Note: a lot of the code in `getFormattedMarkerStyle` is taken from `updateMarkerNodes`. This should be tidied up, but for now I've just copy/pasted the code to avoid risky changes.

alantreadway (approved):
LGTM

PR #1523, opened 2024-04-29
Ag 11454 fix framework specific meta descriptions

JamesSwinton:
Updated logic for generating meta-descriptions so that it checks against the current framework. I also updated the descriptions for the Quick Start to add keywords for SEO.


alantreadway (approved):
LGTM

PR #1524, opened 2024-04-29
AG-8416 Change tooltip.range default for all series types

olegat:
https://ag-grid.atlassian.net/browse/AG-8416

The only expections are:

*   map-line-background
*   map-shape-background


alantreadway (approved):
LGTM

PR #1528, opened 2024-04-30
AG-11324 Range area focus

olegat:
https://ag-grid.atlassian.net/browse/AG-11324


alantreadway (approved):
LGTM, thanks for the update!

PR #1530, opened 2024-04-30
AG-11480 preventDefault when Space is pressed

olegat:
https://ag-grid.atlassian.net/browse/AG-11480

This fixes a bug when the page scrolls down in Firefox when firing keyboard clicks.


alantreadway (approved):
LGTM

PR #1533, opened 2024-05-01
AG-11491 - Remove deprecations.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-11491

Cleans up deprecations and removed some obviously redundant code as a result of this.


lsjroberts (approved):
lgtm, very satisfying

PR #1538, opened 2024-05-03
AG-11128 Annotations

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-11128

- Refactored the behaviour for each annotation type to be distinct, with options built up from a set of higher order properties classes / mixins.
  - e.g. `parallel-channel` now only takes a `start`, `end` and `size` so it is not possible to create non-parallel lines.
- Updated the `StateMachine` to be a hierarchical finite state machine. Added an `AnnotationsStateMachine` that can transition to child machines for each annotation type that then handle their own click and hover behaviour.


packages/ag-charts-community/src/chart/interaction/cursorManager.ts
```diff
@@ -1,6 +1,20 @@
 import { StateTracker } from '../../util/stateTracker';
 import type { DOMManager } from '../dom/domManager';
 
+export enum Cursor {
```
alantreadway, line 4: Contentious point: I'm actually against `enum` for the sake of `enum` - I think this would be better as a simple string-union type; TS `enum`s don't really have any strong benefit, in fact they add an extra layer of abstraction which reduces readability IMO.

You can achieve the same efficiency of declaration using `as const`:
```
export const CURSORS = ['auto', 'default'] as const;
export type Cursor = typeof CURSOR_TYPES[number];
```

packages/ag-charts-community/src/util/properties.ts
```diff
@@ -84,6 +84,38 @@ export class PropertiesArray<T extends BaseProperties> extends Array {
     }
 }
 
+export class TypedPropertiesArray<T extends BaseProperties> extends PropertiesArray<T> {
```
alantreadway, line 87: @iMoses could you review this part?

iMoses: It doesn't seem to be used. Am I missing something?

packages/ag-charts-community/src/chart/interaction/cursorManager.ts
iMoses: I'm of the opposite opinion. I think the greatest benefit comes when refactoring, when looking for usage + you always have the context of the value attached to it whereas seeing `'default'` or `'auto'` can be very confusing and out of place in case a few entities are involved. (and I'm not sure but I think it'll also force us to hack `as const` in many places)

lsjroberts: It was simply because sonarqube was complaining about repeated strings, and adding a `const CursorNotAllowed = 'not-allowed';` in my particular case seemed a bit meh since it is a shared component. But can just do that if you prefer?

packages/ag-charts-community/src/util/properties.ts
lsjroberts: It is used here https://github.com/ag-grid/ag-charts/pull/1538/files#diff-fde1cb072dc3708f277225348f863116c84ddf03dd09aad886b49be27fa0a65cR84 with this definition https://github.com/ag-grid/ag-charts/pull/1538/files#diff-fde1cb072dc3708f277225348f863116c84ddf03dd09aad886b49be27fa0a65cR33-R37

packages/ag-charts-community/src/chart/interaction/cursorManager.ts
lsjroberts: And yeh, I agree with Ido, `as const` is very hacky to my eyes

alantreadway: I'm not entirely convinced still, but let's roll with this for now, there's no silver bullet just trade-offs I would say.

alantreadway (approved):
LGTM

PR #1542, opened 2024-05-07
AG-10085 Refactor ticks fractionDigits to make this more explicit

jacobp100:
Other than this property being quite hidden, there are (and has always been) issues around this property being accidentally erased through `.map`, `.filter` etc.


alantreadway (approved):
Nice, LGTM

PR #1543, opened 2024-05-07
AG-9034 Fix bandscale padding with single element

jacobp100:
https://ag-grid.atlassian.net/browse/AG-9034


alantreadway (commented):
Code changes LGTM, but is there a change to the OOB defaults which would better approximate the current band-scale renderings?

jacobp100:
@alantreadway we can’t really match it - but we should see if design want to tweak these values after the fix is applied

alantreadway (approved):
Thanks, these look better to my eye. LGTM

PR #1546, opened 2024-05-07
AG-11328 Fix hierarchy chart keynev

olegat:
https://ag-grid.atlassian.net/browse/AG-11328

The pickFocus logic was written under the assumption that depth=0 means the root and depth=1 is its children. But this has been reverted to depth=undefined for the root and depth=0 for it children.


alantreadway (approved):
LGTM

PR #1550, opened 2024-05-07
AG-11450 - Fix tooltip hover events.

alantreadway:

Fixes hover of interactive tooltips; as a bonus, cleans up some of the digging through `sourceEvent` to find related `HTMLElement` instances.


lsjroberts (approved):
LGTM

PR #1553, opened 2024-05-08
Add PropertiesArray support for a factory function

iMoses:
currently only support a constructor as a "factory".
usage:
```
new PropertiesArray((params) => myTypes[param.type] ?? defaultType)
```


alantreadway (approved):
LGTM

PR #1556, opened 2024-05-09
AG-11504 Upgrade TS, drop Vue2

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11504


packages/ag-charts-angular/package.json
alantreadway, line 1: Should there be an Angular version bump here?

packages/ag-charts-website/src/stores/frameworkStore.ts
```diff
@@ -33,10 +33,6 @@ export const setInternalFramework = (internalFramework: InternalFramework) => {
         $frameworkContext.setKey('useTypescript', 'false');
     } else if (internalFramework === 'reactFunctionalTs') {
         $frameworkContext.setKey('useTypescript', 'true');
-    } else if (internalFramework === 'vue') {
-        $frameworkContext.setKey('useVue3', 'false');
-    } else if (internalFramework === 'vue3') {
-        $frameworkContext.setKey('useVue3', 'true');
```
alantreadway, line 39: Should this have been removed? (`useVue3: 'false'` above seems at odds with this too?)

packages/ag-charts-angular/package.json
jacobp100: I'm going to split this ticket out into a few PRs because of the amount of changes accumulated so far

packages/ag-charts-website/src/stores/frameworkStore.ts
jacobp100: Yes - updated to remove the `useVue3` you pointed out

taktran (approved):
The website side looks good. Good stuff 👍🏼

alantreadway (approved):
LGTM, nice work!

PR #1558, opened 2024-05-09
AG-7538 - Fix crosshair value when snap is false

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-7538


alantreadway (approved):
LGTM

PR #1560, opened 2024-05-09
AG-11148 Add saving and restoring of annotation mementos

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-11148


packages/ag-charts-community/.dependency-cruiser.js
```diff
@@ -209,7 +209,7 @@ module.exports = {
             comment: "Don't use top-level export bundles internally.",
             severity: 'error',
             from: { path: 'src/.*/' },
-            to: { path: 'src/[^/]*.ts' },
+            to: { path: 'src/[^/]*.ts', pathNot: 'src/version.ts' },
```
lsjroberts, line 212: @alantreadway does this seem ok, or is there a better way around this? The version is used in `src/chart/agChartV2.ts`.

packages/ag-charts-community/src/scale/ordinalTimeScale.ts
```diff
@@ -212,6 +213,13 @@ export class OrdinalTimeScale extends BandScale<Date, TimeInterval | number> {
             d = new Date(d);
         }
 
+        if (typeof d === 'string') {
+            const maybeDate = new Date(d);
+            if (isValidDate(maybeDate)) {
+                d = maybeDate;
+            }
+        }
```
lsjroberts: When a `Date` is put through `JSON.stringify()` it converts to a string. When `Annotations.convertPoint()` then tries to convert it using the scale, it needs to be turned back into a date (if valid).

Alternatively, the memento could carry the type information in addition to the value, and make the restoration of types be the originators responsibility?

packages/ag-charts-community/.dependency-cruiser.js
alantreadway: Ah, on balance this seems fine - thanks for flagging.

packages/ag-charts-community/src/chart/agChartV2.ts
```diff
@@ -142,9 +144,25 @@ export abstract class AgCharts {
         }
         return AgChartsInternal.getImageDataURL(chart, options);
     }
+
+    public static getAnnotations(chart: AgChartInstance) {
```
alantreadway: I think there maybe a discussion about whether this should be on `AgCharts` or the `AgChartInstance` itself - but let's roll with this for now.

Nit: I would probably try and keep the naming symetrical with the `restore` case - either `get` and `set` prefixes, or `save` and `restore`?

packages/ag-charts-community/src/chart/annotation/annotationManager.ts
```diff
@@ -1,11 +1,57 @@
...
 
-export class AnnotationManager {
+export interface AnnotationsRestoreEvent {
+    type: 'restore-annotations';
+    annotations: AnnotationsMemento['annotations'];
+}
+
+export class AnnotationsMemento implements Memento {
+    type = 'annotations';
+    version = 10;
```
alantreadway: Should this default to the version in `version.ts`?

packages/ag-charts-community/src/chart/memento.ts
```diff
@@ -0,0 +1,82 @@
...
+ * The caretaker manages the encoding and decoding of mementos from originators, ensuring they can be provided to and
+ * received from external systems. A memento encapsulates the state of an originator at a point in time. A memento
+ * is also versioned to ensure it can be migrated to newer versions of the originator.
+ */
+export class MementoCaretaker {
+    private readonly version: number;
+
+    constructor(version: string) {
+        // Only consider the major version when versioning mementos, there should be no breaking changes in minor or
+        // patch versions that require migrating.
```
alantreadway: We might actually want to proactively migrate 'deprecated' fields so that users don't see deprecation warnings - so this statement isn't strictly true?

packages/ag-charts-community/src/scale/ordinalTimeScale.ts
alantreadway: I'm thinking we should either:
- Include the original type in the persisted state
  - maybe even a convention of a `__date` suffix on the property name would be sufficient (stripped out on restore)?
  - use a more complex `object` structure for values with a `type`.
- Alternatively, keep a scheme of the expected property types for use during save/restore, which we can validate against and use to work out deserialised type?

lsjroberts: Agreed, I'll try both methods out and see what seems best.

packages/ag-charts-community/src/chart/annotation/annotationManager.ts
lsjroberts: My thinking was that if there is a change in v11, this would change to:

```ts
class AnnotationsMemento implements Memento {
    version = 11;
}

class AnnotationsMemento10 implements Memento {
    version = 10;
}
```

So as to be precise about the version it is for. But it may prove to be unnecessary to have different classes when we do migrations.

Either way, when the memento is created, it is overwritten to be the version passed through from the caretaker.

packages/ag-charts-community/src/chart/memento.ts
lsjroberts: Good point, I'll update this and note it on the ticket

packages/ag-charts-community/src/scale/ordinalTimeScale.ts
lsjroberts: I've updated it to use the replacer and receiver functions on `JSON.stringify()` and `JSON.parse()`. It transforms `Date` objects into `{ __type: 'date', value: ... }` and back to dates.

packages/ag-charts-community/src/chart/agChartV2.ts
lsjroberts: Agreed, updated to `save` and `restore` for both.

packages/ag-charts-community/src/chart/annotation/annotationManager.ts
```diff
@@ -1,11 +1,57 @@
...
 
-export class AnnotationManager {
+export interface AnnotationsRestoreEvent {
+    type: 'restore-annotations';
+    annotations: AnnotationsMemento['annotations'];
+}
+
+export class AnnotationsMemento implements Memento {
+    type = 'annotations';
+    version = '10.0';
```
alantreadway: Sorry I didn't spot this on the first pass - I would default to the `version.ts` version, so that we always know which version of the library a saved state came from - as there is a non-zero probability of accidental breaks, knowing which version produced the saved-state is always useful, even if we anticipate no changes?

packages/ag-charts-community/src/chart/annotation/annotationManager.ts
```diff
@@ -1,11 +1,57 @@
...
     private styles?: AgAnnotationsThemeableOptions;
 
-    constructor(private readonly annotationRoot: Group) {}
+    constructor(private readonly annotationRoot: Group) {
+        super();
+    }
+
+    public createMemento(_version: string) {
+        const memento = new AnnotationsMemento(this.annotations);
+        memento.version = '10.0';
```
alantreadway: Ditto?

alantreadway (approved):
Nice, LGTM!

PR #1563, opened 2024-05-09
AG-11504 Remove babel from systemjs in favour of TS compiler

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11504


alantreadway (approved):
Nice, LGTM

PR #1564, opened 2024-05-09
AG-11304 - Reduce html size

taktran:
by fetching resolved interfaces instead of injecting onto page

Get `resolved-interfaces.json` on client side instead of injecting it from the astro file into the react component. This caused the html to be very large.

On Grid: https://github.com/ag-grid/ag-grid/pull/7956


packages/ag-charts-community/project.json
```diff
@@ -127,6 +127,18 @@
         "output": "dist/packages/ag-charts-community/resolved-interfaces.AUTO.json"
       }
     },
+    "docs-resolved-interfaces-patched": {
```
alantreadway: Why do we need a new/distinct build target?

plugins/ag-charts-generate-code-reference-files/src/executors/generate/executor.ts
```diff
@@ -36,6 +37,14 @@ async function generateFile(options: ExecutorOptions) {
         case 'docs-interfaces':
             return await writeJSONFile(options.output, typeMapper.resolvedEntries());
 
+        // Patched docs-interface for front end to consume
```
alantreadway: I believe the front-end is the only consumer, so let's not cater for redundant cases?

packages/ag-charts-community/project.json
taktran: It was mainly for comparing, during debugging, but can remove

PR #1557, opened 2024-05-09
AG-11385 Proxy DOM elements

olegat:
https://ag-grid.atlassian.net/browse/AG-11385

PR #1566, opened 2024-05-10
AG-11385 Move Region.focusIndicator logic into separate class

olegat:

This is prep-work for https://github.com/ag-grid/ag-charts/pull/1557.

The hidden proxy-buttons will need update the focusIndicator bounds. But we don't want to need up with a circular dependency where RegionManager depends on DOMManager and vice-versa. So separate this out.


packages/ag-charts-community/src/chart/interaction/regionManager.ts
```diff
@@ -67,7 +67,7 @@ function addHandler<T extends RegionEvent['type']>(
 
 export class RegionManager {
     private currentTabIndex = 0;
-    private readonly focusIndicator: HTMLElement;
+    public readonly focusIndicator: FocusIndicator;
```
olegat, line 70: I am unsure where the ownership of the `FocusIndicator` object should be. Currently I'm thinking of these three possible options:

1) Keeping it here in `RegionManager`
2) Moving it to `ChartContext` passing the `FocusIndicator` object to `RegionManager`'s constructor
3) Moving it to `DOMManager` (and keeping `RegionManager`'s `domManager` constructor param as `private readonly`).

I'm leaning towards options 3), but it will likely change the `appendChild` ordering of the DOM elements.

alantreadway: 🤔 I would probably lean toward 2) as the FocusIndicator doesn't seem like a responsibility suitable for the `DOMManager`?

alantreadway (approved):
LGTM

packages/ag-charts-community/src/chart/interaction/regionManager.ts
olegat: ACK. I'll move it in a follow-up PR.

PR #1569, opened 2024-05-10
AG-11450 - More new DOM structure fixes.

alantreadway:

More refinements of the new DOM:
- Tooltips now respect `window` bounds correctly, and are no longer clipped due to removal of redundant `overflow: hidden` in several spots.
- Creating a chart into a DOM with ZERO CSS styles now renders a chart correctly.
- Setting chart `width` and/or `height` now correctly auto-centers the chart in the container DIV.
  - Additionally, tooltips, focus box and other overlaid elements now appear in the correct positions relative to the canvas.
- Reduction of number of inline styles in favour of CSS-based styles when feasible.

Example-runner cleanup:
- Stop polluting the container element:
  - Moved `data-ag-charts` onto `.ag-charts-wrapper` which we manage.

Use-cases tested - configs:
- Zero CSS / auto-sized chart ✅ 
- Zero CSS / fixed-sized chart ✅ 
- Container fixed/% / auto-sized chart ✅ 
- Container fixed/% / fixed-size chart ✅ 

Features tested:
- Tooltips ✅ 
- Keyboard Nav ✅ 
- Crosshairs ✅ 
- Context Menu ✅ 
- Zoom buttons ✅

### DIFF
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -172,12 +172,10 @@ export abstract class Chart extends Observable implements AgChartInstance {
         newValue(value: HTMLElement) {
             if (this.destroyed) return;
 
-            value.setAttribute('data-ag-charts', '');
             this.ctx.domManager.setContainer(value);
             Chart.chartsInstances.set(value, this);
         },
         oldValue(value: HTMLElement) {
-            value.removeAttribute('data-ag-charts');
             Chart.chartsInstances.delete(value);
         },
     })
--- a/packages/ag-charts-community/src/chart/dom/domManager.ts
+++ b/packages/ag-charts-community/src/chart/dom/domManager.ts
@@ -24,10 +24,6 @@ const STYLES = `
     display: flex !important;
     align-items: center !important;
     justify-content: center !important;
-    top: 0;
-    bottom: 0;
-    left: 0;
-    right: 0;
 }
 
 .ag-charts-canvas {
--- a/packages/ag-charts-community/src/chart/dom/domManager.ts
+++ b/packages/ag-charts-community/src/chart/dom/domManager.ts
@@ -3,7 +3,7 @@ import { GuardedElement } from '../../util/guardedElement';
 import { type Size, SizeMonitor } from '../../util/sizeMonitor';
 import { BaseManager } from '../baseManager';
 
-const domElementClasses = ['styles', 'canvas', 'canvas-overlay', 'hidden'] as const;
+const domElementClasses = ['styles', 'canvas', 'canvas-overlay'] as const;
 export type DOMElementClass = (typeof domElementClasses)[number];
 
 const domElementConfig: Record<
@@ -13,7 +13,6 @@ const domElementConfig: Record<
     styles: { childElementType: 'style' },
     canvas: { childElementType: 'canvas' },
     'canvas-overlay': { childElementType: 'div' },
-    hidden: { childElementType: 'div' },
 };
 
 const STYLES = `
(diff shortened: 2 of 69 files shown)


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -172,12 +172,10 @@ export abstract class Chart extends Observable implements AgChartInstance {
         newValue(value: HTMLElement) {
             if (this.destroyed) return;
 
-            value.setAttribute('data-ag-charts', '');
```
jacobp100, line 175: We use this for the dark mode logic (plugins/ag-charts-generate-example-files/src/executors/generate/generator/utils/getDarkModeSnippet.ts)

packages/ag-charts-community/src/chart/dom/domManager.ts
```diff
@@ -13,38 +14,48 @@ const domElementConfig: Record<
...
     touch-action: none;
     box-sizing: border-box;
-    overflow: hidden;
+    display: flex !important;
+    align-items: center !important;
+    justify-content: center !important;
+    top: 0;
+    bottom: 0;
+    left: 0;
+    right: 0;
```
jacobp100: This does nothing for position: relative

packages/ag-charts-community/src/chart/dom/domManager.ts
```diff
@@ -13,38 +14,48 @@ const domElementConfig: Record<
...
-.ag-charts-canvas, .ag-charts-canvas-overlay {
-    position: absolute;
-    display: block;
-    top: 0;
-    bottom: 0;
-    left: 0;
-    right: 0;
+.ag-charts-canvas-overlay {
+    position: relative;
+    display: contents;
```
jacobp100, line 39: What's display: contents meant to be doing here?

packages/ag-charts-community/src/scene/canvas/hdpiCanvas.ts
```diff
@@ -99,7 +99,7 @@ export class HdpiCanvas {
 
     private onEnabledChange() {
         if (this.element) {
-            this.element.style.display = this.enabled ? 'block' : 'none';
+            this.element.style.display = this.enabled ? '' : 'none';
```
jacobp100, line 102: We need the canvas to be `display: block` because it defaults to `inline-block`. Not setting it can mess up the alignment and outer height

packages/ag-charts-enterprise/src/features/crosshair/crosshairLabel.ts
```diff
@@ -17,7 +17,7 @@ export const defaultLabelCss = `
     font: 12px Verdana, sans-serif;
     overflow: hidden;
     white-space: nowrap;
-    z-index: 99999;
+    z-index: 99998; // Needs to be below tooltips!
```
jacobp100, line 20: Are these children of the new `ag-charts-canvas` div? If so, we may wish to do some follow up work to reduce these values, as all values will be relative to that div

packages/ag-charts-community/src/chart/chart.ts
alantreadway: Yes, I've moved this onto the part of the DOM we actually manage and updated the dark-mode logic.

packages/ag-charts-community/src/chart/dom/domManager.ts
alantreadway: I'll remove.

packages/ag-charts-community/src/chart/dom/domManager.ts
alantreadway: In one iteration it seemed that the `<div>` was adding some extra 3px to the canvas size (which caused an infinitely growing canvas feedback loop), this seemed to remove any extra contribution of this element from the impact on the parents size.

packages/ag-charts-community/src/scene/canvas/hdpiCanvas.ts
alantreadway: This is now being set in the CSS managed by the DOMManager by default.

packages/ag-charts-enterprise/src/features/crosshair/crosshairLabel.ts
alantreadway: Yup - further refinement needed, my intent with this change was just to get back to something working for all of the known use-cases.

packages/ag-charts-community/src/chart/dom/domManager.ts
jacobp100: The infinite feedback loop happens when you miss `display: block` from the canvas - it defaults to `inline-block`, which creates extra padding

PR #1572, opened 2024-05-13
AG-11504

jacobp100:
- Fix `className` not showing up for React examples
- Use `class="chart"` rather than targeting ID in CSS, as the ID is not passed on for React examples (I think we should leave it this way - having ID everywhere would be confusing)
- Fix styles for examples


alantreadway (approved):
LGTM, merged conflicts though

PR #1574, opened 2024-05-13
AG-11385 Legend proxy DOM elements

olegat:

This PR is a rebase & edited version of https://github.com/ag-grid/ag-charts/pull/1557

### DIFF
--- a/packages/ag-charts-community/src/chart/legend.ts
+++ b/packages/ag-charts-community/src/chart/legend.ts
@@ -392,13 +392,13 @@ export class Legend extends BaseProperties {
         const itemMaxWidthPercentage = 0.8;
         const maxItemWidth = maxWidth ?? width * itemMaxWidthPercentage;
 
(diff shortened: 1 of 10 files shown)


packages/ag-charts-community/src/chart/legend.ts
```diff
@@ -225,6 +222,8 @@ export class Legend extends BaseProperties {
 
     private readonly destroyFns: Function[] = [];
 
+    private readonly proxyLegendToolbar;
```
alantreadway, line 225: Nit: Could you add an explicit type here? Not sure why this is passing CI.

alantreadway (approved):
LGTM other than one nit.

packages/ag-charts-community/src/chart/legend.ts
alantreadway: Disregard, I guess this is a new TS feature to derive the field type from the constructor assigned value.

PR #1575, opened 2024-05-13
Misc cleanups

iMoses:


packages/ag-charts-community/src/util/textMeasurer.ts
```diff
@@ -88,6 +88,7 @@ export class TextMeasurer {
         return this.wrapLines(text, options).join('\n');
     }
 
+    // eslint-disable-next-line sonarjs/cognitive-complexity
```
alantreadway: Nit: Don't add these, they are just warning?

.gitignore
```diff
@@ -36,7 +36,7 @@ node_modules
 /.sass-cache
 /.scannerwork
 /connect.lock
-/coverage
+coverage
```
alantreadway: Nit: Why? If this needs changing, do the other entries here need changing too?

iMoses: I'm assuming you added coverage since lately I started getting these every time I executed tests.
image
I looked at the other entries and it didn't seem relevant to change any of them.

alantreadway (approved):
LGTM

PR #1576, opened 2024-05-13
AG-9528 Add sankey chart type

jacobp100:
https://ag-grid.atlassian.net/browse/AG-9528

The minimal code to get a Sankey chart registered so I can avoid merge conflicts going forward

Only thing to note is this is the first series type that has the same name as a chart. This will change if we're able to merge the Sankey and chord charts - I'm not certain on that yet

### DIFF
--- a/packages/ag-charts-community/src/chart/series/sankey/sankeySeriesModule.ts
+++ b/packages/ag-charts-community/src/chart/series/sankey/sankeySeriesModule.ts
@@ -1,0 +1,19 @@
+import type { SeriesModule } from '../../../module/coreModules';
+import { EXTENDS_SERIES_DEFAULTS } from '../../themes/symbols';
+import { SankeySeries } from './sankeySeries';
+
+export const SankeySeriesModule: SeriesModule<'sankey'> = {
+    type: 'series',
+    optionsKey: 'series[]',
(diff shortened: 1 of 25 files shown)


packages/ag-charts-community/src/chart/series/sankey/sankeySeriesModule.ts
```diff
@@ -0,0 +1,19 @@
+import type { SeriesModule } from '../../../module/coreModules';
+import { EXTENDS_SERIES_DEFAULTS } from '../../themes/symbols';
+import { SankeySeries } from './sankeySeries';
+
+export const SankeySeriesModule: SeriesModule<'sankey'> = {
+    type: 'series',
+    optionsKey: 'series[]',
+    packageType: 'community',
```
alantreadway, line 8: This should probably be an enterprise package type - so we might need to hoist some of these files into the enterprise package.

jacobp100: Yep - it’s easier to do development on community packages, so I’ll move it to enterprise when it’s more ready (like maps)

alantreadway (approved):
LGTM

PR #1582, opened 2024-05-15
Sankey 2

jacobp100:
Implements a first attempt Sankey and Chord diagrams

https://ag-grid.atlassian.net/browse/AG-11561
https://ag-grid.atlassian.net/browse/AG-9528


jacobp100:
SonarQube is failing due to duplication. I think we will be able to deduplicate `processData`, `updateSelections`, maybe a few others (more than we did in maps) - but I want to get these working 100% first

packages/ag-charts-community/src/chart/series/sankey/sankeySeries.ts
```diff
@@ -1,29 +1,391 @@
...
     override properties = new SankeySeriesProperties();
 
+    @Validate(ARRAY, { optional: true, property: 'nodes' })
+    private _chartNodes?: any[] = undefined;
+
+    private get nodes() {
+        return this.properties.nodes ?? this._chartNodes;
+    }
+
+    private readonly nodesDataController = new DataController('standalone');
```
alantreadway, line 82: I'm a bit unsure about the need for an entire `DataController`, as the responsibility of this class is coordination of multiple data requests, with an intent to optimise the processing of common data - could we just use a `DataModel` directly.

jacobp100: Might need help with this one. I get close then it can't resolve the properties. The scope cache appears to be empty - but the resolved data seems to be right
image

alantreadway: Discussed offline, something is awry so let's continue with the previous approach

alantreadway (approved):

PR #1583, opened 2024-05-15
AG-11622 - Allow page scrolling on mobile devices for canvas events.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-11622

Allows mobile devices to drag-to-scroll when zoom features are disabled.

### DIFF
--- a/packages/ag-charts-community/src/chart/interaction/gestureDetector.ts
+++ b/packages/ag-charts-community/src/chart/interaction/gestureDetector.ts
@@ -2,6 +2,7 @@ import { Logger } from '../../util/logger';
 import { partialAssign } from '../../util/object';
 import { BaseManager } from '../baseManager';
 import type { DOMManager } from '../dom/domManager';
+import { buildConsumable } from './consumableEvent';
 
 type PinchEventTypes = 'pinch-start' | 'pinch-move' | 'pinch-end';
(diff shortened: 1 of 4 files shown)


packages/ag-charts-community/src/chart/interaction/gestureDetector.ts
```diff
@@ -61,7 +63,7 @@ export class GestureDetector extends BaseManager<GestureEventTypes, GestureEvent
     public constructor(private readonly domManager: DOMManager) {
         super();
         this.domManager.addEventListener('touchstart', this.touchstart, { passive: true });
-        this.domManager.addEventListener('touchmove', this.touchmove, { passive: true });
+        this.domManager.addEventListener('touchmove', this.touchmove);
```
olegat: Can you confirm that the removal of this ` { passive: true } ` option does not regress [AG-10605](https://ag-grid.atlassian.net/browse/AG-10605)?

olegat (approved):
LGTM subject to regression testing

packages/ag-charts-community/src/chart/interaction/gestureDetector.ts
alantreadway: Good shout - I've added `{ passive: false }` to suppress the warning & reflect that we might interrupt the browser default handling.

PR #1586, opened 2024-05-16
AG-11450 - Cleanup autoSize option.

alantreadway:

Removes `autoSize` option, as `autoSize: false` was implicit by virtue of setting `width` + `height` options.

Adds `minWidth` and `minHeight` options to replace and allow finer control over the `autoSize: true` case.

Reworks the canvas centering so for better Safari compatibility.


packages/ag-charts-community/src/chart/dom/domManager.ts
```diff
@@ -161,13 +161,13 @@ export class DOMManager extends BaseManager<Events['type'], Events> {
...
-    setSize(autoSize?: boolean, optionsWidth?: number, optionsHeight?: number) {
+    setSize(minWidth: number = 0, minHeight: number = 300, optionsWidth?: number, optionsHeight?: number) {
         const { style } = this.parentElement.element;
 
         style.width = optionsWidth != null ? `${optionsWidth}px` : '100%';
-        style.minWidth = optionsWidth != null || autoSize === true ? '' : '300px';
+        style.minWidth = optionsWidth != null ? '' : `${minWidth}px`;
         style.height = optionsHeight != null ? `${optionsHeight}px` : '100%';
-        style.minHeight = optionsHeight != null || autoSize === true ? '' : '300px';
+        style.minHeight = optionsHeight != null ? '' : `${minHeight}px`;
```
jacobp100: What's the `100%` in relation to?

alantreadway: Discussed offline, we've addressed this by reworking the DOM and moving which elements play which roles.

packages/ag-charts-community/src/chart/dom/domManager.ts
```diff
@@ -161,24 +171,37 @@ export class DOMManager extends BaseManager<Events['type'], Events> {
...
-    setSize(autoSize?: boolean, optionsWidth?: number, optionsHeight?: number) {
+    setSizeOptions(minWidth: number = 0, minHeight: number = 300, optionsWidth?: number, optionsHeight?: number) {
         const { style } = this.parentElement.element;
 
-        style.width = optionsWidth != null ? `${optionsWidth}px` : '100%';
-        style.minWidth = optionsWidth != null || autoSize === true ? '' : '300px';
-        style.height = optionsHeight != null ? `${optionsHeight}px` : '100%';
-        style.minHeight = optionsHeight != null || autoSize === true ? '' : '300px';
+        style.minWidth = `${optionsWidth ?? minWidth}px`;
+        style.minHeight = `${optionsHeight ?? minHeight}px`;
```
jacobp100: This should be equivalent to

```suggestion
        style.width = `${optionsWidth ?? minWidth}px`;
        style.height = `${optionsHeight ?? minHeight}px`;
```

The advantage of explicit sizes are the browser can optimise its layout calculations better

packages/ag-charts-community/src/options/chart/chartOptions.ts
```diff
@@ -147,13 +147,15 @@ export interface AgBaseThemeableChartOptions<TDatum = any> {
...
     /**
-     * By default (unset) the chart will assume a minimum size of 300x300 px, and grow with the container element.
-     *
-     * Set to true to remove the minimums (e.g. for use with a CSS-managed layout).
-     *
-     * Set to false to disable automatic sizing.
+     * Sets the minimum height of the chart.
+     * Default: `300`
+     */
+    minHeight?: PixelSize;
```
jacobp100, line 153: Only thing about this naming is `height` overrides it. Maybe `preferredMinHeight` or something - or just adjust the documentation to mention height is an override

packages/ag-charts-community/src/chart/dom/domManager.ts
```diff
@@ -47,8 +54,10 @@ const STYLES = `
```
jacobp100, line 54: Is `ag-charts-canvas-overlay` still used? What does `display: contents` do here?

jacobp100 (approved):

packages/ag-charts-community/src/chart/dom/domManager.ts
alantreadway: Yes it's still in use - it just skips rendering/calculation/contribution of of margin/padding/border from that element, and just displays contents.

packages/ag-charts-community/src/chart/dom/domManager.ts
alantreadway: Tested it out and you are right - fixed.

packages/ag-charts-community/src/options/chart/chartOptions.ts
alantreadway: I've adjusted the documentation, thanks!

packages/ag-charts-community/src/chart/dom/domManager.ts
jacobp100: Probably need to see it in the browser to follow what's happening. Don't hold off merging, I'll look after it's merged

PR #1598, opened 2024-05-21
AG-11297 Make the navigator a vertical toolbar

olegat:
https://ag-grid.atlassian.net/browse/AG-11297

Every browser seems to treat a role="scrollbar" element differently. But more importantly, not of them announce changes to the scrollbar when moving back and forth. Also, most examples of role="scrollbar" that I've seen online assume that there's just a single thumb on track (which isn't the case for use, because we've also got min/max handles). The recommendation is to use native scrollbars, which won't work in our use case.


alantreadway (approved):
LGTM

PR #1599, opened 2024-05-21
AG-11450 - Adjust event handler bindings for shadow DOM case.

alantreadway:

Adjustments to allow Charts to execute inside a shadow-DOM.

### DIFF
--- a/packages/ag-charts-community/src/chart/dom/domManager.ts
+++ b/packages/ag-charts-community/src/chart/dom/domManager.ts
@@ -101,7 +101,7 @@ export class GuardedAgChartsWrapperElement extends GuardedElement {
     }
 }
 
-type Events = { type: 'hidden' } | { type: 'resize'; size: Size };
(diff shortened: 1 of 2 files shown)


packages/ag-charts-community/src/chart/dom/domManager.ts
```diff
@@ -245,6 +247,24 @@ export class DOMManager extends BaseManager<Events['type'], Events> {
...
+            }
+            if (parent.parentElement == null) {
+                return parent;
+            }
+
+            parent = parent.parentElement as HTMLElement;
+        }
+
+        return docRoot;
+    }
```
iMoses, line 266: nit: `this.container.ownerDocument.body`

alantreadway: Good spot

alantreadway: Unfortunately `ownerDocument` is not set to the right element in the shadow-DOM case, so we still need to traverse the parents to find the closest enclosing element that is part of the shadow-DOM root.

PR #1601, opened 2024-05-21
AG-10486 Context Menu Refactor

olegat:
https://ag-grid.atlassian.net/browse/AG-10486

TL;DR - The `ContextMenu` module now listens for `ContextMenuEvents` instead of `PointerInteractionEvents<'contextmenu'>`.


packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts
```diff
@@ -128,43 +124,46 @@ export class ContextMenu extends _ModuleSupport.BaseModuleInstance implements _M
...
     }
 
-    private onContextMenu(event: _ModuleSupport.PointerInteractionEvent<'contextmenu'>) {
+    private onContext(event: ContextMenuEvent) {
         if (!this.enabled) return;
+        event.consume();
+
+        this.showEvent = event.sourceEvent.sourceEvent as MouseEvent;
+        this.x = event.sourceEvent.pageX;
+        this.y = event.sourceEvent.pageY;
```
alantreadway: Nit: Can we avoid using the `sourceEvent` here? We're applying corrections and adjustments to original event coordinate space to align with the new DOM setup, I'd rather avoid bypassing those and/or necessary corrections there that might be thrown up by this case?

olegat: It's not clear to me what should be used instead of `event.sourceEvent`. The original logic used the `pageX, pageY` values from the `PointerInteractionEvent<'contextmenu'>` object, which are the same values being used here.

Do you mean that I want to hide our internal `PointerInteractionEvent` from the `ContextMenu` module? We could change the `ContextMenuEvent` type to include the exact x/y position if that's preferred.

alantreadway: Updating the `ContextMenuEvent` to have x/y position relative to the canvas would work well I think - in `InteractionManager` we attempt to make `offsetX` and `offsetY` be relative to the `<canvas>` (0,0) position, so that (x, y) is always in the canvas coordinate space - I think for events that makes the most sense (i.e. working in canvas coordinate space), as the new DOM layout also means that HTML elements should be positionable in the same coordinate space too?

https://github.com/ag-grid/ag-charts/blob/d85c8e6628aad4761cd540ceec87f977ea838b82/packages/ag-charts-community/src/chart/interaction/interactionManager.ts#L461-L470

olegat: OK. I've changed this part of the code so that the module can extract the x/y coordinates straight from the `ContextMenuEvent` instead.

alantreadway (approved):
Thanks for the updates - LGTM

PR #1614, opened 2024-05-23
AG-10486 Contextmenu callback type-safety and docs

olegat:
https://ag-grid.atlassian.net/browse/AG-10486

### DIFF
--- a/packages/ag-charts-community/src/chart/interaction/contextMenuRegistry.ts
+++ b/packages/ag-charts-community/src/chart/interaction/contextMenuRegistry.ts
@@ -1,3 +1,4 @@
+import type { AgContextMenuOptions } from '../../options/chart/contextMenuOptions';
 import { Listeners } from '../../util/listeners';
 import type { CategoryLegendDatum } from '../legendDatum';
 import type { SeriesNodeDatum } from '../series/seriesTypes';
@@ -22,22 +23,25 @@ type ContextEventProperties<K extends ContextType = ContextType> = {
 export type ContextType = keyof ContextTypeMap;
 export type ContextMenuEvent<K extends ContextType = ContextType> = ContextEventProperties<K> & ConsumableEvent;
 
-export type ContextMenuAction = {
+// Extract the TEvent types from the AgContextMenuOptions contract:
+type TEventOfContextMenuActions<K extends keyof Omit<AgContextMenuOptions, 'enabled'>> = Parameters<
+    NonNullable<AgContextMenuOptions[K]>[number]['action']
+>[0];
+export type ContextMenuCallback<K extends ContextType> = {
+    all: (params: TEventOfContextMenuActions<'extraActions'>) => void;
+    series: (params: TEventOfContextMenuActions<'extraNodeActions'>) => void;
+    legend: (params: TEventOfContextMenuActions<'extraLegendItemActions'>) => void;
+}[K];
+
+export type ContextMenuAction<K extends ContextType> = {
     id?: string;
     label: string;
-    type: ContextType;
-    action: (params: ContextMenuActionParams) => void;
-};
-
-export type ContextMenuActionParams = {
-    datum?: any;
-    itemId?: string;
-    seriesId?: string;
-    event: MouseEvent;
+    type: K;
+    action: ContextMenuCallback<K>;
 };
 
 export class ContextMenuRegistry {
-    private readonly defaultActions: Array<ContextMenuAction> = [];
+    private readonly defaultActions: Array<ContextMenuAction<ContextType>> = [];
     private readonly disabledActions: Set<string> = new Set();
     private readonly hiddenActions: Set<string> = new Set();
     private readonly listeners: Listeners<'', (e: ContextMenuEvent) => void> = new Listeners();
@@ -70,6 +74,14 @@ export class ContextMenuRegistry {
         return event.type === type;
     }
 
+    public static checkCallback<T extends ContextType>(
+        desiredType: T,
+        type: ContextType,
+        _callback: ContextMenuCallback<ContextType>
+    ): _callback is ContextMenuCallback<T> {
+        return desiredType === type;
+    }
+
     public dispatchContext<T extends ContextType>(
         type: T,
         pointerEvent: PointerInteractionEvent<'contextmenu'>,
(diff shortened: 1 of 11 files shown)


packages/ag-charts-community/src/chart/interaction/contextMenuRegistry.ts
```diff
@@ -22,22 +23,25 @@ type ContextEventProperties<K extends ContextType = ContextType> = {
...
-export type ContextMenuAction = {
+// Extract the TEvent types from the AgContextMenuOptions contract:
+type TEventOfContextMenuActions<K extends keyof Omit<AgContextMenuOptions, 'enabled'>> = Parameters<
+    NonNullable<AgContextMenuOptions[K]>[number]['action']
+>[0];
+export type ContextMenuCallback<K extends ContextType> = {
+    all: (params: TEventOfContextMenuActions<'extraActions'>) => void;
+    series: (params: TEventOfContextMenuActions<'extraNodeActions'>) => void;
+    legend: (params: TEventOfContextMenuActions<'extraLegendItemActions'>) => void;
+}[K];
```
alantreadway: These types seem overly complex, could you simplify it to something like this to improve readability?
```
type ContextMenuActionMapping = {
    all: AgChartContextMenuEvent;
    legend: AgChartLegendContextMenuEvent;
    series: AgNodeContextMenuActionEvent;
};

export type ContextMenuAction<K extends ContextType> = {
    id?: string;
    label: string;
    type: K;
    action: (params: ContextMenuActionMapping[K]) => void;
};
```

If you wanted to keep the stricter references to `AgContextMenuOptions` then although more verbose, it is easier to follow:
```
type ContextMenuActionMapping = {
    all: Parameters<NonNullable<AgContextMenuOptions['extraActions']>[number]['action']>[0];
    legend: Parameters<NonNullable<AgContextMenuOptions['extraLegendItemActions']>[number]['action']>[0];
    series: Parameters<NonNullable<AgContextMenuOptions['extraNodeActions']>[number]['action']>[0];
};
```

packages/ag-charts-enterprise/src/features/zoom/zoomContextMenu.ts
```diff
@@ -66,11 +64,11 @@ export class ZoomContextMenu {
         }
     }
 
-    private onZoomToHere({ event }: ContextMenuActionParams, props: ZoomProperties) {
+    private onZoomToHere({ event }: AgNodeContextMenuActionEvent, props: ZoomProperties) {
         const { rect } = this;
         const { enabled, isScalingX, isScalingY, minRatioX, minRatioY } = props;
 
-        if (!enabled || !rect || !event || !event.target) return;
+        if (!enabled || !rect || !event || !event.target || !(event instanceof MouseEvent)) return;
```
alantreadway, line 71: Nit: Why do we need to check for `MouseEvent` here and below now?

packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts
```diff
@@ -33,17 +32,17 @@ export class ContextMenu extends _ModuleSupport.BaseModuleInstance implements _M
     /**
      * Extra menu actions with a label and callback.
      */
-    public extraActions: Array<ContextMenuAction> = [];
+    public extraActions: NonNullable<AgContextMenuOptions['extraActions']> = [];
```
lsjroberts, line 35: nit: I think we're generally trying to separate options types and modules/features types, even if it produces a bit of duplication?

packages/ag-charts-community/src/chart/interaction/contextMenuRegistry.ts
olegat: Yes, the intent was to use stricter references so that the TS compiler can check that we are running callbacks that matches the documentation. I don't have strong opinions whether with use the `ContextMenuActionMapping` or `TEventOfContextMenuActions` approach to achieve this, so happy to make this proposed change.

packages/ag-charts-enterprise/src/features/zoom/zoomContextMenu.ts
olegat: It is needed because `ContextMenuActionParams['event']` is of type `MouseEvent`, whereas `AgNodeContextMenuActionEvent['event']` is of type `Event`.

https://github.com/ag-grid/ag-charts/blob/1a7fe5c9c9e1af099fddf2a33d5215e3b51b98f3/packages/ag-charts-community/src/chart/interaction/contextMenuRegistry.ts#L36

https://github.com/ag-grid/ag-charts/blob/1a7fe5c9c9e1af099fddf2a33d5215e3b51b98f3/packages/ag-charts-community/src/options/chart/eventOptions.ts#L3

Therefore we need to check that the event is a `MouseEvent` so that we can read the `clientX, clientY` properties a few lines below.

We could (should?) perhaps change the `AgChartEvent` interface so that we can specify an event subtype:
```
interface AgChartEvent<T extends string, E extends Event> {
    type: T;
    event: E;
}
```

packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts
olegat: Correct.

We still do this: The internal state is stored in `ContextMenu.groups`, whereas `ContextMenu.extraActions` stores the input user options.

https://github.com/ag-grid/ag-charts/blob/546335e8cd1d67aa650974ee3e87095fc5cf4c81/packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts#L53

https://github.com/ag-grid/ag-charts/blob/546335e8cd1d67aa650974ee3e87095fc5cf4c81/packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts#L10

The `ContextMenu.groups` property contains some duplicated values like `label` and `action`, but it also contains internal meta-data like `id?` and `type`:

https://github.com/ag-grid/ag-charts/blob/1a7fe5c9c9e1af099fddf2a33d5215e3b51b98f3/packages/ag-charts-community/src/chart/interaction/contextMenuRegistry.ts#L25

packages/ag-charts-enterprise/src/features/zoom/zoomContextMenu.ts
alantreadway: Discussed offline, as this is using a public API, let's roll with this for now.

alantreadway (approved):
LGTM

PR #1623, opened 2024-05-28
AG-8701 Anti-alias bar series on a continuous axis when the spacing i…

jacobp100:
…s between 0 and 1 px (non-inclusive)

https://ag-grid.atlassian.net/browse/AG-8701


alantreadway (approved):
LGTM

PR #1625, opened 2024-05-28
Normalize key:value formatting on tooltips

iMoses:
waiting for David's input


packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
```diff
@@ -476,43 +476,39 @@ export class HistogramSeries extends CartesianSeries<Rect, HistogramSeriesProper
         if (!this.properties.isValid() || !xAxis || !yAxis) {
             return EMPTY_TOOLTIP_CONTENT;
         }
-
+        ``;
```
iMoses:
```suggestion

```

PR #1626, opened 2024-05-29
AG-11133 Add annotation delete on floating toolbar

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-11133


packages/ag-charts-community/src/chart/mapping/__snapshots__/prepare.test.ts.snap
alantreadway, line 1: NOTE: This file has moved.

alantreadway (approved):
LGTM

PR #1628, opened 2024-05-29
Fixes guarantees around {Chart,ChartOptions}.userOptions being immutable.

alantreadway:
Ensures that we don't mutate `userOptions`, so that subsequent `Chart.updateDelta()` calls are working against the previously provided options, not our mutated/cleaned copy.

Additional fixes:
- Fixes the cleaning of `Chart.queuedIserOptions` that was obscuring the main behaviour.
- Consolidates `optionsModule.ts` tests into `optionsModule.test.ts`.


packages/ag-charts-community/src/module/optionsModule.ts
```diff
@@ -114,7 +111,7 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
             this.processedOptions.legend?.enabled == null
         ) {
             this.processedOptions.legend ??= {};
-            this.processedOptions.legend.enabled = this.processedOptions.series!.length > 1;
+            this.processedOptions.legend.enabled = (this.processedOptions.series?.length ?? 0) > 1;
```
iMoses: IIRC at this stage `series` is guaranteed to be an array thanks to `sanityCheckAndCleanup`

packages/ag-charts-community/src/module/optionsModule.ts
```diff
@@ -214,7 +211,7 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
             userPalette,
         };
 
-        const processedSeries = options.series!.map((series) => {
+        const processedSeries = options.series?.map((series) => {
```
iMoses: ditto

iMoses (approved):
LGTM 👍 
just minor notes about probably unnecessary type changes.

packages/ag-charts-community/src/module/optionsModule.ts
alantreadway: Reverted, there was a need for this during testing, but rolling back I don't see a problem anymore.

packages/ag-charts-community/src/module/optionsModule.ts
alantreadway: Reverted

PR #1630, opened 2024-05-29
AG-11591 Add series-specific series[].tooltip.range option

olegat:
https://ag-grid.atlassian.net/browse/AG-11591


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1304,7 +1328,7 @@ export abstract class Chart extends Observable {
             return;
         }
 
-        const pick = this.pickSeriesNode({ x: offsetX, y: offsetY }, range === 'exact', pixelRange);
+        const pick = this.pickTooltip({ x: offsetX, y: offsetY });
```
olegat, line 1335: Note: This is the key change of this PR.

packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1941,6 +1963,17 @@ export abstract class Chart extends Observable {
...
+    // global option (`chart.tooltip.range`) that could overriden the theme. But now, the tooltip range
+    // option is series-specific.
+    //
+    // To preserve backward compatiblity, the `chart.tooltip.range` theme default has been changed from
+    // 'nearest' to undefined.
+    private applySeriesTooltipDefaults(target: Series<SeriesNodeDatum, SeriesProperties<never>>) {
+        target.properties.tooltip.range ??= this.tooltip.range;
+        target.properties.tooltip.range ??= target.defaultTooltipRange;
+    }
+
```
olegat, line 1980: Note: As explained in the comment above this method, this chunk of code is responsible for compatibility. It ensure that the old (now deprecated) option `tooltip.range` continues to work.

We discussed perhaps removing the `tooltip.range` option entirely rather than deprecating it. If we decide to remove the option, then this chunk (and other parts of the PR) will become redundant and can be removed.

packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1047,26 +1048,46 @@ export abstract class Chart extends Observable {
...
+            if (useSeriesTooltipRange) {
+                if (tooltipRange === 'exact') {
+                    pickModes = [SeriesNodePickMode.EXACT_SHAPE_MATCH];
+                } else {
+                    pickModes = undefined;
+                }
+            } else {
+                pickModes = sharedPickModes;
+            }
+
```
alantreadway: Nit: Just wondering, since we have distinct entry points for different values of `useSeriesTooltipRange`, whether we could move some of this logic here up to the callers of this function, rather than passing the boolean in?

alantreadway (commented):
LGTM generally just a couple of nits - let's sync with David on the backwards-compatibility today.

packages/ag-charts-community/src/chart/chart.ts
olegat: Good idea. I'll try that.

alantreadway (approved):
LGTM

PR #1632, opened 2024-05-29
Ag 11136/cross line annotation

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-11136


packages/ag-charts-community/src/chart/interaction/interactionManager.ts
```diff
@@ -73,9 +73,25 @@ const EVENT_HANDLERS = [
...
+    NAVIGATOR: 'navigator',
+    PAGINATION: 'pagination',
+    ROOT: 'root',
+    SERIES: 'series',
+    HORIZONTAL_AXES: 'horizontal-axes',
+    VERTICAL_AXES: 'vertical-axes',
+    TOOLBAR: 'toolbar',
+} as const;
+
+export type RegionName = (typeof REGIONS)[keyof typeof REGIONS];
```
alantreadway: Could you move this to a distinct file as regions don't fit into the responsibility of InteractionManager?

(Oli and I agreed yesterday that we should remove `region` from `BaseInteractionEvent` here in a follow-up PR)

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
```diff
@@ -73,9 +73,25 @@ const EVENT_HANDLERS = [
     'keyup',
 ] as const;
 
+export const REGIONS = {
```
alantreadway: This is kinda a poor-mans `enum` - could you make this a simple array of the values and drop the keys?

E.g.:
```
export const REGIONS = ['title', 'subtitle', ....] as const;

export type Region = (typeof REGIONS)[number];
```

packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1886,6 +1893,17 @@ export abstract class Chart extends Observable {
...
+
+        const axisGroups: { [Key in ChartAxisDirection]: Group[] } = {
+            [ChartAxisDirection.X]: [],
+            [ChartAxisDirection.Y]: [],
+        };
+
+        chart.axes.forEach((axis) => axisGroups[axis.direction].push(axis.getAxisGroup()));
+
+        this.ctx.regionManager.updateRegion(REGIONS.HORIZONTAL_AXES, ...axisGroups[ChartAxisDirection.X]);
+        this.ctx.regionManager.updateRegion(REGIONS.VERTICAL_AXES, ...axisGroups[ChartAxisDirection.Y]);
```
alantreadway, line 1907: Side-note: There is a lot of axis-related functionality being added to `Chart`, I think we maybe missing an abstraction around the axes (naively an `AxisGroup` at the least) which could have some of these responsibilities.

lsjroberts (commented):
Annotations changes lgtm! Nice one. Agreed on Alan's points.

I might look at moving the validation down into the annotation types now we've got some different variations 🤔

packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -352,7 +355,11 @@ export abstract class Chart extends Observable {
             ctx.regionManager.listenAll('click', (event) => this.onClick(event)),
             ctx.regionManager.listenAll('dblclick', (event) => this.onDoubleClick(event)),
             seriesRegion.addListener('hover', (event) => this.onMouseMove(event)),
-            seriesRegion.addListener('leave', (event) => this.onLeave(event)),
+            horizontalAxesRegion.addListener('hover', (event) => this.onMouseMove(event)),
+            verticalAxesRegion.addListener('hover', (event) => this.onMouseMove(event)),
+            seriesRegion.addListener('leave', (event) => this.onLeave(event), All),
```
olegat: There's a risk that the `All` parameter might interrupt animations when the mouse-pointer leaves this bounding box. I would advice checking for a regression using these repro steps:

1. Load any example with animation. The [Marker vs. Path Animation](https://charts-staging.ag-grid.com/javascript/line-series-test/#marker-vs-path-animation) example is a good test case because the animation is very slow there.
2. Move the mouse pointer over the series rect.
3. Move the mouse pointer out of the chart.

packages/ag-charts-community/src/chart/chart.ts
olegat: I'm not persuaded that an `AxisGroup` class is necessary. However, it might make more sense to move this logic into the `CartesianChart` sub-class because the concept of horizontal & vertical axes is not applicable to all series types.

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
olegat: Yes I think this can be simplified. `BaseInteractionEvent.region` and `KeyNavEvent.region` can stay as-is for this PR. I'll rebased https://github.com/ag-grid/ag-charts/pull/1640 & resolve conflicts once this PR is merged.

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
manapeirov: I did this so I could use the strings when I'm getting the regions

lsjroberts: Yeah I've had this issue with repeated strings getting flagged by sonarqube. The solution I used was a proper enum like `enum Regions { Title = 'title', Subtitle = 'subtitle' }`.

The regions array type method helps constrain the type, but it leaves that sonarqube error. The enum is perhaps a little unnecessary in reality, but its better than having lots of `const title = 'title'` style things.

packages/ag-charts-community/src/chart/chart.ts
manapeirov: tried moving that stuff to cartesian chart but it caused some intergrated charts cross-filtering snapshot failures and it wasn't obvious to me what was causing that so undid it. Leaving as is for now?

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
manapeirov: Converted it to an enum

packages/ag-charts-community/src/chart/chart.ts
manapeirov: Tested, seems like no regression but removed it anyway

packages/ag-charts-community/src/chart/interaction/chartRegions.ts
alantreadway, line 1: Nit: Could we name the file and the main export consistently?

packages/ag-charts-enterprise/src/features/annotations/annotations.ts
```diff
@@ -532,6 +607,37 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
...
+        if (datum.value == null) return;
+
+        const { scaleX, scaleY } = this;
+
+        if (!scaleX || !scaleY) return;
+
+        let x1,
+            x2,
+            y1,
+            y2 = 0;
```
alantreadway: Nit: Given this is across lines anyway, maybe just a single `let` per variable?

alantreadway (approved):
LGTM

PR #1637, opened 2024-05-30
AG-NONE Update config files to mutes TS compiler errors

olegat:


alantreadway:
Just to note - I'd prefer we went the other way, enforced type checking on tests and fix up the tests.

olegat:
> Just to note - I'd prefer we went the other way, enforced type checking on tests and fix up the tests.

I agree. There are a lot of errors in the tests right now. I'll have a look at that

libraries/ag-charts-test/src/canvas/mock-canvas.ts
```diff
@@ -23,6 +27,7 @@ export class MockContext {
         this.realCreateElement = realCreateElement;
         this.ctx = {
             nodeCanvas,
+            getRenderContext2D: () => this.ctx.nodeCanvas.getContext('2d') as unknown as CanvasRenderingContext2D,
```
olegat, line 30: Note: The TS compiler is reporting that the real and mock canvases are not compatible (which isn't what I would expect from a full mock). This isn't an ideal solution, but it's better that adding `as` in every `.test.ts` file that uses the `CanvasRenderingContext2D` object.

packages/ag-charts-community/src/options/series/cartesian/cartesianSeriesTooltipOptions.ts
```diff
@@ -15,6 +15,7 @@ export interface AgCartesianSeriesTooltipRendererParams<TDatum = any> extends Ag
 export interface AgErrorBoundSeriesTooltipRendererParams {
     /** The key to use to retrieve lower bound error values from the x axis data. */
     xLowerKey?: string;
+    // TODO(olegat) shouldn't these values be removed??
```
alantreadway: Nit: As this is the public API, it might be better to avoid adding these types of internal note here in case they leak.

packages/ag-charts-community/tsconfig.json
```diff
@@ -11,7 +11,7 @@
     "noUnusedLocals": true
   },
   "files": [],
-  "include": [],
+  "include": ["jest.config.ts", "jest.setup.ts", "src/**/*.ts"],
```
alantreadway: Can you check the impact of this on the output in `dist/types/` - I'm just concerned this could lead to test-typings being emitted there?

packages/ag-charts-community/tsconfig.types.json
```diff
@@ -5,6 +5,6 @@
         "emitDeclarationOnly": true,
         "types": ["node"]
     },
-    "include": ["src/*.ts"],
+    "include": ["jest.config.ts", "jest.setup.ts", "src/**/*.ts"],
```
alantreadway: Ditto

packages/ag-charts-community/src/options/series/cartesian/cartesianSeriesTooltipOptions.ts
olegat: Well spotted! Done.

packages/ag-charts-community/tsconfig.json
olegat: I'm seeing `*.d.test.ts` files in the `dist/types/` directory. However I'm also seeing that on `latest`, so it's unclear whether these changes are actually impacting the output.

Shall we just merge this PR and clean up the `dist/types` directory in a follow-up?

PR #1638, opened 2024-05-30
AG-11680 Area series line style

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11680


packages/ag-charts-community/src/chart/cartesianChart.test.ts
```diff
@@ -240,7 +241,9 @@ describe('CartesianChart', () => {
         it.each(YKEYS)(`should render series with yKey [%s] appropriately`, async (yKey) => {
             const { chart: testee, highlightManager, nodeData } = await buildChart({ ...tcOptions }, yKey);
             highlightManager.updateHighlight(testee.id, nodeData?.nodeData[3]);
-            await compare(testee);
+            await compare(testee, {
+                failureThreshold: 2,
```
alantreadway, line 245: These threshold changes shouldn't be needed, let me sync the branch and see what might be causing you to need this.

packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
```diff
@@ -504,42 +577,38 @@ export class AreaSeries extends CartesianSeries<
...
-        }
-        if (lastPoint) {
-            fillPath.lineTo(lastPoint.x, lastPoint.y);
-        }
-        fillPath.closePath();
+        const { line } = this.properties;
+
+        let start = 0;
+        fill.path.clear(true);
+        splitPartialPaths(points.map((point) => point.point)).forEach((topPath) => {
```
alantreadway: I'd have a small concern about performance regression here, as `for ()` loops and iterators seems to scale much better than `.map()` and `.forEach()` - could you validate that for large-scale data-sets (e.g. maybe try switching one of these series in this example to `area`?)
https://charts-staging.ag-grid.com/javascript/benchmarks/#large-scale-dataset-test-4x-100k-series

packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
```diff
@@ -504,42 +577,38 @@ export class AreaSeries extends CartesianSeries<
...
+
         strokePath.clear(true);
-        for (const { point } of strokeData.points) {
-            if (point.moveTo) {
-                strokePath.moveTo(point.x, point.y);
-            } else {
-                strokePath.lineTo(point.x, point.y);
-            }
-        }
+        splitPartialPaths(strokeData.points.map((point) => point.point)).forEach((path) => {
```
alantreadway: Ditto

packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
jacobp100: Updated. We still have array construction with `splitPairData` - but that's only used for animation. Assume that's ok?

packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
jacobp100: Updated

packages/ag-charts-community/src/chart/cartesianChart.test.ts
jacobp100: Updated these to ensure it's a pixel threshold - it's only 2 pixels. Assume that's ok?

PR #1640, opened 2024-05-30
AG-11136 Move event.region? property into regionManager.ts

olegat:
https://ag-grid.atlassian.net/browse/AG-11136


alantreadway (approved):
LGTM, thanks for the update!

PR #1644, opened 2024-05-31
Rework dynamic function use

alantreadway:
Removes use of dynamic change-detection functions to reduce issues when running in a CSP-constrained environment, at the trade-off of a little performance and repeated code.

Also adds version-controlled CSP regression test cases (these were previously manually ran during regression testing).


packages/ag-charts-website/src/content/docs/benchmarks/_examples/summary/data.ts
```diff
@@ -554,107 +554,107 @@ export function getData() {
...
+                    memoryUsage: 260344176,
+                    heapUsed: 193144176,
                     canvasCount: 35,
                     canvasBytes: 67200000,
                 },
                 'large-dataset benchmark initial load': {
-                    timeMs: 5530.321125,
-                    memoryUsage: 787808480,
-                    heapUsed: 772448480,
+                    timeMs: 6658.801583000001,
```
jacobp100, line 578: That's a pretty big increase

jacobp100 (approved):
Seems fine other than the perf drop. Maybe later down the line we could find/make a babel plugin to do the code generation at compile time

packages/ag-charts-website/src/content/docs/benchmarks/_examples/summary/data.ts
alantreadway: Agreed - we're probably going to be looking at ways to optimise the number of change-detections needed in the future, specifically either:
- optimising the changing of properties in the first place, so the extra overhead here becomes less relevant.
- batching rendering, which will mean for this specific use case we'd only be performing O(1) update per scene-graph property rather than O(N).

PR #1665, opened 2024-06-03
AG-11133 Add annotation toolbar tooltip suppression

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-11133

Prevents the tooltip from showing while an annotation is selected.

### DIFF
--- a/packages/ag-charts-enterprise/src/features/annotations/annotations.ts
+++ b/packages/ag-charts-enterprise/src/features/annotations/annotations.ts
@@ -479,10 +479,13 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
         this.active = hovered;
         toolbarManager.toggleGroup('annotations', 'annotationOptions', this.active != null);
 
-        if (this.active != null) {
+        if (this.active == null) {
+            this.ctx.tooltipManager.unsuppressTooltip('annotations');
+        } else {
             const node = annotations.nodes()[this.active];
             node.toggleActive(true);
             this.ctx.toolbarManager.changeFloatingAnchor('annotationOptions', node.getAnchor());
+            this.ctx.tooltipManager.suppressTooltip('annotations');
         }
 
         updateService.update(ChartUpdateType.PERFORM_LAYOUT, { skipAnimations: true });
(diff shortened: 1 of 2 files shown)


packages/ag-charts-enterprise/src/features/annotations/annotations.ts
```diff
@@ -479,10 +479,13 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
...
         toolbarManager.toggleGroup('annotations', 'annotationOptions', this.active != null);
 
-        if (this.active != null) {
+        if (this.active == null) {
+            this.ctx.tooltipManager.unsuppressTooltip('annotations');
+        } else {
             const node = annotations.nodes()[this.active];
             node.toggleActive(true);
             this.ctx.toolbarManager.changeFloatingAnchor('annotationOptions', node.getAnchor());
+            this.ctx.tooltipManager.suppressTooltip('annotations');
```
lsjroberts, line 488: Using `updateTooltip('annotations')` with null content gets overriden immediately by the updated tooltip meta and content from `Chart.handlePointerTooltip()` when a new pointer event occurs.

We could alternatively change the tooltip state to keep the states in the order they are first created, but this would differ from the rest of our states where it is sorted by last updated.

lsjroberts: e.g.

```diff
+++ b/packages/ag-charts-community/src/util/stateTracker.ts
@@ -1,13 +1,14 @@
 export class StateTracker<T, K = string> extends Map<K, T> {
     constructor(
         protected readonly defaultValue?: T,
-        protected readonly defaultState?: K
+        protected readonly defaultState?: K,
+        protected readonly sort: 'created' | 'updated' = 'updated'
     ) {
         super();
     }

     override set(key: K, value?: T) {
-        this.delete(key); // removed even if re-set to make sure we're last
+        if (this.sort === 'updated') this.delete(key); // removed even if re-set to make sure we're last
```

alantreadway: Current approach seems reasonable for now - let's keep an eye out for any gotchas 👍

alantreadway (approved):
LGTM

PR #1670, opened 2024-06-04
AG-11757 Remove includeConsumedEvents property

olegat:
https://ag-grid.atlassian.net/browse/AG-11757

This property adds unnecessary complexity to the RegionManager class. So instead just call fireEvent everytime in the listener.


olegat:
TODO:
- [x] Check this change does not affect the annotations https://charts.ag-grid.com/javascript/annotations/#adding--editing
- [x] Check this change does not break page-scrolling when zoom is disabled.

jacobp100 (approved):
Just to check - events are propagated through everything even when `preventDefault` is called, and we just ignore events that are `defaultPrevented` in the handler?

olegat:
> Just to check - events are propagated through everything even when `preventDefault` is called, and we just ignore events that are `defaultPrevented` in the handler?

The former "events are propagated through everything even when `preventDefault` is called"

alantreadway (approved):
LGTM

PR #1671, opened 2024-06-04
AG-11653 Ensure labels show when pie/donut has a solid fill highlight

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11653


alantreadway (approved):
LGTM

PR #1674, opened 2024-06-04
AG-8192 - Update API Docs with new AgChartsInstance + Promise-based APIs

alantreadway:
https://ag-grid.atlassian.net/browse/AG-8192

Updates docs and examples to allow us to demonstrate the new APIs recently added:
- Move of `AgCharts` static methods into `AgChartsInstance`.
- Addition of `Promise`-based returns for update and other async APIs.

Bonus:
- Adds support for handling of `async` functions to all FW variants, although we're not using them yet.

### DIFF
--- a/packages/ag-charts-website/src/content/docs/api-create-update/_examples/wait-for-update/main.ts
+++ b/packages/ag-charts-website/src/content/docs/api-create-update/_examples/wait-for-update/main.ts
@@ -22,6 +22,7 @@ const chart = AgCharts.create(options as AgChartOptions);
 
 let running = false;
 async function start() {
+    if (running) return;
     running = true;
 
     while (running) {
--- a/packages/ag-charts-website/src/content/docs/api-create-update/_examples/wait-for-update/main.ts
+++ b/packages/ag-charts-website/src/content/docs/api-create-update/_examples/wait-for-update/main.ts
@@ -1,0 +1,35 @@
+// @ag-skip-fws
+import { AgChartOptions, AgCharts } from 'ag-charts-community';
+
+import { generateDatum, getData } from './data';
+
+const options: AgChartOptions = {
+    container: document.getElementById('myChart'),
+    title: {
+        text: 'Simple Promise-base example',
+    },
+    data: getData(),
+    series: [
+        {
+            type: 'line',
+            xKey: 'count',
+            yKey: 'value',
+        },
+    ],
+};
+
+const chart = AgCharts.create(options as AgChartOptions);
+
+let running = false;
+async function start() {
+    running = true;
+
+    while (running) {
+        options.data = [...options.data!, generateDatum()];
+        await chart.update(options);
+    }
+}
+
+function stop() {
+    running = false;
+}
--- a/packages/ag-charts-website/src/content/docs/api-create-update/index.mdoc
+++ b/packages/ag-charts-website/src/content/docs/api-create-update/index.mdoc
@@ -8,15 +8,15 @@ Learn about creating and updating charts in more detail.
 ## Creating and Updating Charts
 
 {% if isFramework("javascript") %}
-`AgCharts` exposes `create()` and `update()` static methods to perform chart initialisation and update based upon
-the `AgChartOptions` configuration structure.
+`AgCharts` exposes `create()` static methods to perform chart initialisation, and the resulting `AgChartInstance` has methods
+such as `AgChartInstance.update()` to allow updating configuration. The `AgChartOptions` type defines the configuration structure.
 
 Mutations to the previously used options object are not automatically picked up by the chart implementation,
-`AgCharts.update()` should be called for changes to be applied.
+`AgChartInstance.update()` should be called for changes to be applied.
 
 {% note %}
-We expect the options supplied to `AgCharts.update()` to be the full configuration state to update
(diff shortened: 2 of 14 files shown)


packages/ag-charts-website/src/content/docs/api-create-update/index.mdoc
```diff
@@ -8,25 +8,33 @@ Learn about creating and updating charts in more detail.
...
 
 Mutations to the previously used options object are not automatically picked up by the chart implementation,
-`AgCharts.update()` should be called for changes to be applied.
+`AgChartInstance.update()` should be called for changes to be applied.
 
 {% note %}
-We expect the options supplied to `AgCharts.update()` to be the full configuration state to update
-to, not a partial configuration. Use `AgCharts.updateDelta()` to apply partial updates.
+We expect the options supplied to `AgChartInstance.update()` to be the full configuration state to update
+to, not a partial configuration. Use `AgChartInstance.updateDelta()` to apply partial updates.
```
lsjroberts, line 19: This may cause some people to expect to be able to call `AgChartInstance.update(chart)` instead of `chart.update()`? But may be fine.

lsjroberts (approved):
lgtm

packages/ag-charts-website/src/content/docs/api-create-update/index.mdoc
alantreadway: Maybe, of the two style options `chart.update()` and `AgChartInstance.update()` I'm leaning towards the later as it has a stronger binding to actual types, over the convention of `chart`.

We talk about getting an `AgChartInstance` from the `AgCharts.create()` API, and the `AgChartsInstance` API section is just before this too, so I'm hoping it's not much of a leap 🤞

packages/ag-charts-website/src/content/docs/api-create-update/_examples/wait-for-update/main.ts
```diff
@@ -0,0 +1,35 @@
...
+            yKey: 'value',
+        },
+    ],
+};
+
+const chart = AgCharts.create(options as AgChartOptions);
+
+let running = false;
+async function start() {
+    running = true;
```
jacobp100, line 26:
```suggestion
    if (running) return;
    running = true;
```

plugins/ag-charts-generate-example-files/src/executors/generate/generator/examplesGenerator.ts
```diff
@@ -94,10 +94,14 @@ export const getGeneratedContents = async (params: GeneratedContentParams): Prom
     let indexHtml = await readFile(path.join(folderPath, 'index.html'));
     extractOptions ||= entryFile.includes('@ag-options-extract');
 
-    if (entryFile.includes('@ag-skip-fws') && internalFramework !== 'vanilla') {
-        entryFile = PLACEHOLDER_MAIN_TS;
-        indexHtml = `<div id="myChart"></div>`;
-        extractOptions = false;
+    if (entryFile.includes('@ag-skip-fws')) {
```
jacobp100, line 97: Can we use `provided/modules` here?

plugins/ag-charts-generate-example-files/src/executors/generate/generator/transformation-scripts/parser-utils.ts
```diff
@@ -81,7 +81,7 @@ export function modulesProcessor(modules: string[]) {
 }
 
 export function removeFunctionKeyword(code: string): string {
-    return code.replace(/^function /, '').replace(/\n\s?function /, '\n ');
+    return code.replace(/^(async ){0,1}function /, '$1').replace(/\n\s?function /, '\n ');
```
jacobp100:
```suggestion
    return code.replace(/^(async )?function /, '$1').replace(/\n\s?function /, '\n ');
```

plugins/ag-charts-generate-example-files/src/executors/generate/generator/transformation-scripts/parser-utils.ts
```diff
@@ -90,12 +90,12 @@ export function getFunctionName(code: string): string {
 }
 
 export const convertFunctionToProperty = (code: string) =>
-    code.replace(/function\s+([^(\s]+)\s*\(([^)]*)\)/, '$1 = ($2) =>');
+    code.replace(/(async ){0,1}function\s+([^(\s]+)\s*\(([^)]*)\)/, '$2 = $1($3) =>');
```
jacobp100:
```suggestion
    code.replace(/(async )?function\s+([^(\s]+)\s*\(([^)]*)\)/, '$2 = $1($3) =>');
```

jacobp100 (approved):

plugins/ag-charts-generate-example-files/src/executors/generate/generator/examplesGenerator.ts
alantreadway: We could - however after some discussion David and I agreed that we would keep this example for the JS/TS case as an advanced use-case.

I strongly dislike provided examples, and would prefer we either wrote examples that work with the example generator OR improve the example generator over adding provided examples, which are longer-term harder to maintain consistently.

We're using `@ag-skip-fws` in a few places where we have complex examples (mostly `-test` pages) where we don't care about FW generation cases (and they are currently broken).

PR #1675, opened 2024-06-04
AG-10318 Localisation MVP

jacobp100:


packages/ag-charts-community/src/options/chart/localeOptions.ts
```diff
@@ -0,0 +1,10 @@
+export interface AgLocaleThemeableOptions<MessageFormat = any> {
+    messages?: Record<string, MessageFormat> | undefined;
+    formatMessage?: MessageFormatter<MessageFormat> | undefined;
```
alantreadway: I'm thinking we should avoid this for now, due to the overlap with theme/formatters in the options contract.

alantreadway: Discussed offline, I've misunderstood.

PR #1676, opened 2024-06-04
AG-11502 Add annotations keyboard cancel and delete actions

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-11502

### DIFF
--- a/packages/ag-charts-community/src/chart/interaction/keyNavManager.ts
+++ b/packages/ag-charts-community/src/chart/interaction/keyNavManager.ts
@@ -10,7 +10,15 @@ import type {
 } from './interactionManager';
 import { InteractionState } from './interactionManager';
 
-export type KeyNavEventType = 'blur' | 'browserfocus' | 'tab' | 'nav-hori' | 'nav-vert' | 'submit';
+export type KeyNavEventType =
+    | 'blur'
+    | 'browserfocus'
+    | 'tab'
+    | 'nav-hori'
+    | 'nav-vert'
+    | 'submit'
+    | 'cancel'
+    | 'delete';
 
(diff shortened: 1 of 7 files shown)


packages/ag-charts-community/src/chart/interaction/keyNavManager.ts
```diff
@@ -108,6 +116,11 @@ export class KeyNavManager extends BaseManager<KeyNavEventType, KeyNavEvent> {
             case 'Space':
             case 'Enter':
                 return this.dispatch('submit', 0, event);
+            case 'Escape':
+                return this.dispatch('cancel', 0, event);
+            case 'Backspace':
+            case 'Delete':
+                return this.dispatch('delete', 0, event);
```
lsjroberts, line 123: Is the `KeyNavManager` the correct place for these events? They are not specifically about navigation, but may affect it. And the abstraction between key code and event is good.

olegat: This is a good place for this, because this is where we convert key-codes to actions. I don't think that it will have a negative effect on navigation, because most listeners will only care about blur, focus, tab, nav-hori and nav-vert.

It does feel like the class could have a better name than "`KeyNavManager`" but we can live with that.

olegat (approved):

packages/ag-charts-community/src/chart/interaction/keyNavManager.ts
alantreadway: I think so, the responsibility of this class is to abstract keyboard interactions, these seem to fit within that scope.

alantreadway (approved):
LGTM

PR #1684, opened 2024-06-05
AG-10318 Localisation docs

jacobp100:


packages/ag-charts-website/src/content/docs/localisation/index.mdoc
```diff
@@ -0,0 +1,304 @@
...
+
+{% /tabItem %}
+
+{% /if %}
+
+{% /tabs %}
+
+## All Built-In Translations
+
+```js
```
alantreadway: Any way for us to just inline the `en.ts` file here, so we can reduce maintenance overhead and potential for drift?

jacobp100: Updated

alantreadway (approved):
LGTM

PR #1693, opened 2024-06-05
Add basic AxisManager

alantreadway:
To support annotations; allow axis interrogation without direct coupling between modules and `Axis`.


manapeirov (approved):
LGTM

PR #1702, opened 2024-06-06
AG-11778 Flow proportion a11y

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11778

### DIFF
--- a/packages/ag-charts-community/src/chart/series/polar/pieUtil.ts
+++ b/packages/ag-charts-community/src/chart/series/polar/pieUtil.ts
@@ -1,9 +1,8 @@
-import { sectorBox } from 'packages/ag-charts-community/src/scene/util/sector';
-
 import type { FromToMotionPropFn, FromToMotionPropFnContext, NodeUpdateState } from '../../../motion/fromToMotion';
 import { BBox } from '../../../scene/bbox';
 import type { Point } from '../../../scene/point';
 import type { Sector } from '../../../scene/shape/sector';
+import { sectorBox } from '../../../scene/util/sector';
 import { isBetweenAngles, toRadians } from '../../../util/angle';
 import type { Circle } from '../../marker/circle';
 import type { PickFocusInputs, SeriesNodePickMatch } from '../series';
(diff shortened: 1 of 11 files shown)


packages/ag-charts-community/src/scene/util/sector.ts
```diff
@@ -14,6 +14,40 @@ interface LineCoordinates {
     end: { x: number; y: number };
 }
 
+export function sectorBox({ startAngle, endAngle, innerRadius, outerRadius }: SectorBoundaries) {
```
olegat, line 17: This new function seems to do the same thing as `computeSectorFocusBounds`:
https://github.com/ag-grid/ag-charts/blob/dbd45c5f9b8ad7ec11c413482e232ab87cc6ef83/packages/ag-charts-community/src/chart/series/polar/pieUtil.ts#L168

If so, it would be worth joining these two functions to avoid code duplication.

olegat (changes requested):

packages/ag-charts-enterprise/src/series/sankey/sankeySeries.ts
```diff
@@ -524,4 +524,22 @@ export class SankeySeries extends FlowProportionSeries<
     override getLabelData(): _Util.PointLabelDatum[] {
         return [];
     }
+
```
olegat, line 527: I think that these two `computeFocusBounds` methods are fine, but apparently SonarCloud is upset about code duplication. Consider adding abstract like `computeNodeFocusBounds` and `computeLinkFocusBounds` that the derived classes can override.

jacobp100: If we updated all the shapes to have a definite bounding box, could we remove this method entirely from all series?

olegat: All shapes already have a definite bounding box. But the keyboard navigation operates on datums not shapes. There could perhaps be a method like `getShapeOfDatum(datumIndex): Shape` that retrieves the shape of datum.

packages/ag-charts-community/src/chart/series/polar/pieUtil.ts
```diff
@@ -1,8 +1,10 @@
+import { sectorBox } from 'packages/ag-charts-community/src/scene/util/sector';
```
olegat: fix this import

olegat (approved):
LGTM subject to CI pass

alantreadway (approved):

PR #1705, opened 2024-06-07
Add support for CSS + HTML module imports.

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11825


alantreadway:
Looking good, could you address the two issues flagged by SonarCloud please?
https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&pullRequest=1705&id=ag-charts-community-latest

packages/ag-charts-community/src/chart/toolbar/toolbarStyles.ts
```diff
@@ -2,31 +2,31 @@ import { ToolbarPosition } from './toolbarTypes';
 
 export const block = 'ag-charts-toolbar';
 export const elements = {
-    align: `${block}__align`,
-    button: `${block}__button`,
-    icon: `${block}__icon`,
-    label: `${block}__label`,
+    align: 'ag-charts-toolbar__align',
```
lsjroberts, line 5: I'd be inclined to keep the variables here even if we can't use them in the `.css` files themselves. Saves typos.

lsjroberts (commented):
Excellent stuff!

packages/ag-charts-community/src/chart/toolbar/toolbarStyles.ts
jacobp100: I went this way so it's easier to validate the JS and CSS class names match exactly, and stuff like ctrl + F will work. Additionally too, in the future, we may be able to add tools that detect unused classes, and class names that don't point to anything in the CSS - but they only work if you spell out your classes in full without any abstraction

packages/ag-charts-community/src/styles/styles.css
```diff
@@ -1,7 +1,101 @@
...
+    --ag-charts-toolbar-background-color: var(--ag-header-background-color, color-mix(in srgb, #fff, #182230 93%));
+}
+
+/********
+* ICONS *
+********/
 [class^='ag-charts-icon'] {
     display: block;
-    /* speak: none; */
+    speak: none;
```
lsjroberts, line 98: This is not an official css property, so should probably review it. It came from the grid but I commented it out. https://caniuse.com/?search=speak

lsjroberts: Or rather, it's not on MDN and isn't propertly implemented.

jacobp100: Chatted offline, just for posterity,

- It's a real & official spec (https://www.w3.org/TR/css-speech-1/#speaking-props-speak)
- Chrome isn't standards compliant, and only supports `none` (https://issues.chromium.org/issues/40813740)
- We'll add the spec compliant version and Chrome's

alantreadway (approved):
LGTM

PR #1706, opened 2024-06-07
AG-11385 Use aria-disabled="true" instead of disabled

olegat:

The guideline is to make buttons of toolbars/menus focusable even when they are disabled. This is so that the buttons are discoverable to screenreader users.

Reference:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/roles/toolbar_role

### DIFF
--- a/packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts
+++ b/packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts
@@ -18,7 +18,8 @@ type ContextMenuEvent = _ModuleSupport.ContextMenuEvent;
 type ContextMenuAction<T extends ContextType = ContextType> = _ModuleSupport.ContextMenuAction<T>;
 type ContextMenuCallback<T extends ContextType> = _ModuleSupport.ContextMenuCallback<T>;
 
-const { BOOLEAN, Validate, createElement, initMenuKeyNav, ContextMenuRegistry } = _ModuleSupport;
+const { BOOLEAN, Validate, createElement, initMenuKeyNav, makeAccessibleClickListener, ContextMenuRegistry } =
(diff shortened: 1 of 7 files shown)


lsjroberts (commented):
If I understand this correctly - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled

I think we'll need to add a check on the `onclick` for the button's `ariaDisabled === 'true'` to stop it from firing?

olegat:
I've added `onclick` listeners to `aria-disabled="true"` buttons.

packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts
```diff
@@ -276,10 +277,14 @@ export class ContextMenu extends _ModuleSupport.BaseModuleInstance implements _M
     }
 
     private createActionElement({ id, label, type, action }: ContextMenuAction): HTMLElement {
+        let button: HTMLElement;
         if (id && this.registry.isDisabled(id)) {
-            return this.createDisabledElement(label);
+            button = this.createDisabledElement(label);
+        } else {
+            button = this.createButtonElement(label);
```
lsjroberts: nit: could merge `createButtonElement` and `createDisabledElement` as they're so similar? Then add the `ariaDisabled` in here if disabled?

lsjroberts (approved):
LGTM

packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts
olegat: Done.

PR #1711, opened 2024-06-07
AG-11707 - Allow finer control over axis layout for Financial Charts.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-11707

Investigate if we can achieve some of the 10.0.0 requirements by allowing finer-grained control of axis layout.


lsjroberts (approved):
LGTM. Great stuff, axis looks clearer.

PR #1715, opened 2024-06-10
Fix zoom button locale updates to apply to correct button

lsjroberts:
Fixes a bug with locale updates being applied to the wrong zoom buttons. For example, the last button here is the wrong one, should be reset not pan-start (due to mistmatching indices from the proxied zoom options).

Screenshot 2024-06-10 at 11 03 33


alantreadway (approved):
LGTM

PR #1721, opened 2024-06-10
ReDoS fixes

jacobp100:
Fixes SonarQube issue

All validated against https://devina.io/redos-checker


packages/ag-charts-community/src/scale/linearScale.test.ts
```diff
@@ -150,6 +150,12 @@ describe('LinearScale', () => {
             const f = scale.tickFormat({ specifier: '~s' });
             expect(f(43000000)).toBe('43M');
         }
+        {
```
jacobp100, line 153: This was always passing, but thought it was good to double check

alantreadway (approved):
LGTM

PR #1723, opened 2024-06-10
Fix annotations dragging on toolbar buttons and color picker

lsjroberts:
Fixes issues where clicking and accidentally dragging on the floating toolbar buttons would close the toolbar. Replaces a previous (more hacky) attempt at fixing this for the color picker.

### DIFF
--- a/packages/ag-charts-enterprise/src/features/color-picker/colorPicker.ts
+++ b/packages/ag-charts-enterprise/src/features/color-picker/colorPicker.ts
@@ -8,6 +8,7 @@ const { createElement } = _ModuleSupport;
 const { Color } = _Util;
 
 const moduleId = 'color-picker';
+const canvasOverlay = 'canvas-overlay';
 
 const getHsva = (input: string) => {
     try {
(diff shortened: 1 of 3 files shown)


packages/ag-charts-enterprise/src/features/color-picker/colorPicker.ts
```diff
@@ -24,14 +25,18 @@ const getHsva = (input: string) => {
 export class ColorPicker extends _ModuleSupport.BaseModuleInstance implements _ModuleSupport.ModuleInstance {
     private readonly element: HTMLElement;
 
+    static isManagedChildDOMElement(domManager: _ModuleSupport.DOMManager, element: HTMLElement) {
```
alantreadway: This seems like an odd spot to do this - I would go with either:
- Make `moduleId` public so you can do this check elsewhere.
- Make it an instance method (since `ColorPicker` has a reference to `DOMManager` anyway).

?

lsjroberts: I felt like making `moduleId` public would be wrong since it's more of an internal detail. Like with the `ToolbarManager` version which has multiple parent elements to check.

And similarly for making it static to match the `ToolbarManager`, and presumably if we add it to other managers in the future. Since `ColorPicker` is the odd one out being instantiated by and accessible to the `Annotations` module.

alantreadway: Perhaps make it an instance method `ColorPicker.isChildElement()` perhaps? It just seems really odd to have a static method where we're passing references as parameters that are already referenceable from the instance?

alantreadway (approved):
LGTM

PR #1725, opened 2024-06-10
AG-11370 Canvas focus refactor spike

olegat:
https://ag-grid.atlassian.net/browse/AG-11370

### DIFF
--- a/packages/ag-charts-community/src/chart/chartContext.ts
+++ b/packages/ag-charts-community/src/chart/chartContext.ts
@@ -76,17 +76,13 @@ export class ChartContext implements ModuleContext {
         }
     ) {
         const { scene, root, syncManager, container, updateCallback, updateMutex, overrideDevicePixelRatio } = vars;
-
-        const transferredCanvasElement: HTMLCanvasElement | undefined = scene?.canvas.element;
         this.chartService = chart;
         this.syncManager = syncManager;
         this.zoomManager = chart.zoomManager;
-        this.domManager = new DOMManager(container, transferredCanvasElement);
-
-        const canvasElement: HTMLCanvasElement = this.domManager.getCanvasElement();
-        this.scene = scene ?? new Scene({ pixelRatio: overrideDevicePixelRatio, canvasElement });
+        this.domManager = new DOMManager(container);
+        scene?.setContainer(this.domManager);
+        this.scene = scene ?? new Scene({ pixelRatio: overrideDevicePixelRatio, domManager: this.domManager });
         this.scene.setRoot(root);
-
         this.axisManager = new AxisManager(root);
 
         this.localeManager = new LocaleManager();
--- a/packages/ag-charts-community/src/chart/chartContext.ts
+++ b/packages/ag-charts-community/src/chart/chartContext.ts
@@ -76,15 +76,15 @@ export class ChartContext implements ModuleContext {
         }
     ) {
         const { scene, root, syncManager, container, updateCallback, updateMutex, overrideDevicePixelRatio } = vars;
+
+        const transferredCanvasElement: HTMLCanvasElement | undefined = scene?.canvas.element;
         this.chartService = chart;
         this.syncManager = syncManager;
         this.zoomManager = chart.zoomManager;
-        this.domManager = new DOMManager(container);
+        this.domManager = new DOMManager(container, transferredCanvasElement);
 
-        this.scene = scene ?? new Scene({ pixelRatio: overrideDevicePixelRatio });
-        const parent = this.domManager.addChild('canvas', 'scene-canvas', this.scene.canvas.element).parentElement;
-        if (!parent) throw new Error('AG Charts - error get canvas parent');
-        this.scene.setContainer(parent);
+        const canvasElement: HTMLCanvasElement = this.domManager.getCanvasElement();
+        this.scene = scene ?? new Scene({ pixelRatio: overrideDevicePixelRatio, canvasElement });
         this.scene.setRoot(root);
 
         this.axisManager = new AxisManager(root);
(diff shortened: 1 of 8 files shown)


olegat:
Note: This seems to fix the tabbing regressions that we've been seeing. However, the change to the `BASE_DOM` in `domManager.ts` is breaking the positioning of the tooltip. I'll have a look at that.

packages/ag-charts-community/src/scene/scene.ts
```diff
@@ -59,18 +47,9 @@ export class Scene {
...
-            const { element } = this.canvas;
-            element.parentElement?.removeChild(element);
-            value.appendChild(element);
-        } else {
-            this.domManager = value;
-            this.domManager.addChild('canvas', 'scene-canvas', this.canvas.element);
-        }
+    setContainer(container: HTMLElement) {
+        this.canvas.element.remove();
+        container.appendChild(this.canvas.element);
```
alantreadway: I'd prefer if `DOMManager` was still responsible for creation + destruction of the canvas element.

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
```diff
@@ -159,7 +155,8 @@ export class InteractionManager extends BaseManager<InteractionTypes, Interactio
 
     public constructor(
         private readonly keyboardOptions: { readonly enabled: boolean },
-        private readonly domManager: DOMManager
+        private readonly domManager: DOMManager,
+        private readonly canvasElement: HTMLCanvasElement
```
alantreadway: I'd prefer we didn't pass the `canvasElement` in here as it's implicitly hardcoding the expected DOM structure and relations within that here.

Can't we have the `DOMManager` understand where the appropriate places are to attach listeners to the DOM structure its responsible for?

packages/ag-charts-community/src/chart/chartContext.ts
```diff
@@ -80,9 +80,13 @@ export class ChartContext implements ModuleContext {
         this.syncManager = syncManager;
         this.zoomManager = chart.zoomManager;
         this.domManager = new DOMManager(container);
-        scene?.setContainer(this.domManager);
```
alantreadway: The code here is slightly convoluted due to the 're-use resources from a destroyed chart instance' case, but I don't think we should be dealing with the raw elements here. The code here shouldn't care about specific elements are involved in the composition of the chart.

packages/ag-charts-community/src/scene/scene.ts
olegat: Do we still have a use-case for `Scene.setContainer` (e.g. integrated charts)?

The `Scene.domManager` is only used in the constructor, so the coordination between the constructor and `setContainer` doesn't make a lot of sense. It would be cleaner if the `Scene` constructor just took a non-nullable and `HTMLCanvasElement`, and leave it up to the caller to add that element in the correct place.

*Edit:* It seems like we use `setContainer` in sparklines. We could remove this usage to simply things (https://github.com/ag-grid/ag-grid/pull/8078). Or we could just not use the `setContainer` method in the ag-charts and add a `// comment` to clarify that this method is only intended to be used by sparklines.

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
olegat: ACK. It should be feasible to update the logic of `DOMManager.addEventListener`

packages/ag-charts-community/src/chart/chartContext.ts
olegat: > The code here shouldn't care about specific elements are involved in the composition of the chart.
Neither should the `Scene` object - it only needs an `HTMLCanvasElement`. I think that this is a good opportunity to this tidy up.

The initialisation of the chart-context shouldn't be so complex. The  `'re-use resources from a destroyed chart instance'` case isn't any different to a create-or-update logic.

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
olegat: ✅ Done.
⚠️ Note: I had to tweak the logic in `DOMManager.addEventListener` a little bit to avoid firing the keyboard events twice.

packages/ag-charts-community/src/scene/scene.ts
alantreadway: So yes the Integrated Charts case is real and exists - without this change, switching chart types yield a janky experience because we remove the old Chart instance and it's Canvas, then there maybe several frames before a new canvas is put back in place and the Chart rendered.

packages/ag-charts-community/src/chart/dom/domManager.ts
```diff
@@ -222,46 +218,53 @@ export class DOMManager extends BaseManager<Events['type'], Events> {
...
+        this.guardedElement ??= this.createTabGuards();
+        this.guardedElement.tabIndex = tabIndex;
     }
 
     addEventListener<K extends keyof HTMLElementEventMap>(
         type: K,
         listener: (this: HTMLElement, ev: HTMLElementEventMap[K]) => any,
         options?: boolean | AddEventListenerOptions
     ) {
-        this.element.addEventListener(type, listener, options);
```
alantreadway: Can we leave this as a general-purpose mechanism please, there is no need to hardcode the behaviour to `canvas` cases?

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
```diff
@@ -142,7 +142,6 @@ export class InteractionManager extends BaseManager<InteractionTypes, Interactio
     private readonly debug = Debug.create(true, 'interaction');
 
     private rootElement: HTMLElement;
-
```
alantreadway: Nit: Revert unecessary change please.

packages/ag-charts-community/src/scene/scene.ts
olegat: Done. Change reverted.

packages/ag-charts-community/src/chart/chartContext.ts
olegat: Done. Change reverted.

alantreadway (approved):
Thanks for persevering on this one - LGTM!

PR #1730, opened 2024-06-11
AG-11699 Update crosshairs to show when drawing annotations

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-11699

https://github.com/ag-grid/ag-charts/assets/3817697/162f8ecb-a38a-4d8b-b9e6-9c1a8ae9f209

### DIFF
--- a/packages/ag-charts-enterprise/src/features/crosshair/crosshair.ts
+++ b/packages/ag-charts-enterprise/src/features/crosshair/crosshair.ts
@@ -7,7 +7,8 @@ type AgCrosshairLabelRendererResult = any;
 
 const { Group, Line, BBox } = _Scene;
 const { createId } = _Util;
-const { POSITIVE_NUMBER, RATIO, BOOLEAN, COLOR_STRING, LINE_DASH, OBJECT, Validate, Layers } = _ModuleSupport;
+const { POSITIVE_NUMBER, RATIO, BOOLEAN, COLOR_STRING, LINE_DASH, OBJECT, InteractionState, Validate, Layers } =
+    _ModuleSupport;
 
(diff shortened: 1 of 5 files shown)


packages/ag-charts-enterprise/src/features/crosshair/crosshair.ts
```diff
@@ -64,19 +65,19 @@ export class Crosshair extends _ModuleSupport.BaseModuleInstance implements _Mod
...
+            region.addListener('drag', (event) => this.onMouseMove(event), mouseMoveStates),
+            region.addListener('leave', () => this.onMouseOut(), mouseMoveStates),
             ctx.highlightManager.addListener('highlight-change', (event) => this.onHighlightChange(event)),
             ctx.layoutService.addListener('layout-complete', (event) => this.layout(event)),
             () => Object.entries(this.labels).forEach(([_, label]) => label.destroy())
         );
     }
 
     private layout({ series: { rect, paddedRect, visible }, axes }: _ModuleSupport.LayoutCompleteEvent) {
-        this.hideCrosshairs();
```
lsjroberts, line 78: @manapeirov This didn't seem necessary, and it appears to work without it. Having it there causes flickering on the crosshair when dragging. And as far as I can tell the crosshairs are hidden by specific events. Just checking it wasn't required for some specific reason?

alantreadway: Let's remove and see where we land - I would suspect that maybe it might be better in one of the short-circuit cases below only, if needed.

alantreadway (approved):
LGTM

PR #1731, opened 2024-06-11
AG-11370 Fix CSS bounding rects of legend pagination buttons

olegat:
This change set the CSS bounds on the proxyLegendPagination container and then sets the CSS bounds on the proxy buttons to be relative to that container.

This is necessary, because I've tried rearranging the 'ag-charts-canvas' and 'ag-charts-canvas-overlay' elements in the DOMManager's BASE_DOM, but weird things start to happen when the legend pagination buttons come into focus.


alantreadway (approved):
LGTM

PR #1732, opened 2024-06-11
AG-11853 Add separate react components for each chart type

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11853


packages/ag-charts-react/package.json
```diff
@@ -33,22 +33,16 @@
...
     "tslib": "^1.10.0",
     "typescript": "^5.4.5",
-    "vinyl-source-stream": "^2.0.0",
     "rimraf": "3.0.2"
   },
   "peerDependencies": {
     "ag-charts-community": "9.3.1-beta.20240609",
-    "react": "^16.3.0 || ^17.0.0 || ^18.0.0",
-    "react-dom": "^16.3.0 || ^17.0.0 || ^18.0.0"
+    "react": "^16.8.0 || ^17.0.0 || ^18.0.0 || ^19.0.0"
```
alantreadway, line 45: We need to stay in alignment with Grid on these:
https://github.com/ag-grid/ag-grid/blob/latest/packages/ag-grid-react/package.json#L46

alantreadway (commented):
Generally looks good (see my one blocker).

Can you verify that `nx pack:verify ag-charts-react` passes, and that the NPM package contents then found in `dist/packages/contents/ag-charts-react/package` match what we generated previously?

alantreadway:
The benchmark failure looks unrelated to this change. There was a PR from @iMoses yesterday that had a similar failure which disappeard on re-run, so I suspect there has been a recent change which has introduced some flakiness :-(

PR #1739, opened 2024-06-12
Update vue component

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11853

* Rewrite vue3 component to avoid class component package
* Updates vue3 to use esbuild (same as React)
* Removes now unused vue packages
* Fixes styles passed to vue and angular components
* Moves React to index.ts rather than having extra files


alantreadway (approved):
LGTM

PR #1746, opened 2024-06-12
AG-11136 - Add annotations axis button

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-11136

### DIFF
--- a/packages/ag-charts-enterprise/src/features/annotations/annotations.test.ts
+++ b/packages/ag-charts-enterprise/src/features/annotations/annotations.test.ts
@@ -127,7 +127,7 @@ describe('Annotations', () => {
         });
     });
 
-    describe('cross-line', () => {
+    describe.skip('cross-line', () => {
         it('when y-axis is clicked it should create a horizontal cross-line', async () => {
             await prepareChart();
             await clickAction(40, height / 3)(chart);
--- a/packages/ag-charts-enterprise/src/features/annotations/annotations.test.ts
+++ b/packages/ag-charts-enterprise/src/features/annotations/annotations.test.ts
@@ -1,7 +1,7 @@
 import { afterEach, describe, expect, it } from '@jest/globals';
 
 import { type AgCartesianChartOptions, AgCharts } from 'ag-charts-community';
-import { clickAction, extractImageData, setupMockCanvas, waitForChartStability } from 'ag-charts-community-test';
+import { extractImageData, setupMockCanvas, waitForChartStability } from 'ag-charts-community-test';
 
 import { prepareEnterpriseTestOptions } from '../../test/utils';
 
@@ -22,9 +22,6 @@ describe('Annotations', () => {
         },
     };
 
-    let width = 0;
-    let height = 0;
-
     async function prepareChart(
         annotationsOptions?: AgCartesianChartOptions['annotations'],
         baseOptions = EXAMPLE_OPTIONS
@@ -34,8 +31,6 @@ describe('Annotations', () => {
             annotations: { ...baseOptions.annotations, ...(annotationsOptions ?? {}) },
         };
         prepareEnterpriseTestOptions(options);
-        width = options.width!;
-        height = options.height!;
         chart = AgCharts.create(options);
         await waitForChartStability(chart);
     }
@@ -123,20 +118,6 @@ describe('Annotations', () => {
                     },
                 ],
             });
-            await compare();
-        });
-    });
-
-    describe.skip('cross-line', () => {
-        it('when y-axis is clicked it should create a horizontal cross-line', async () => {
-            await prepareChart();
-            await clickAction(40, height / 3)(chart);
-            await compare();
-        });
-
-        it('when x-axis is clicked it should create a vertical cross-line', async () => {
-            await prepareChart();
-            await clickAction(width / 3, height - 30)(chart);
             await compare();
         });
     });
--- a/packages/ag-charts-enterprise/src/features/annotations/annotations.ts
+++ b/packages/ag-charts-enterprise/src/features/annotations/annotations.ts
@@ -12,6 +12,8 @@ import type {
 } from './annotationTypes';
 import { ANNOTATION_BUTTONS, AnnotationType, stringToAnnotationType } from './annotationTypes';
 import { calculateAxisLabelPadding, invertCoords, validateDatumPoint } from './annotationUtils';
+import { AxisButton, DEFAULT_ANNOTATION_AXIS_BUTTON_CLASS } from './axisButton';
+import axisButtonCss from './axisButton.css';
 import { CrossLineAnnotation } from './cross-line/crossLineProperties';
 import { CrossLine } from './cross-line/crossLineScene';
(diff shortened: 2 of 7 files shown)


packages/ag-charts-enterprise/src/features/annotations/annotations.test.ts
```diff
@@ -127,7 +127,7 @@ describe('Annotations', () => {
         });
     });
 
-    describe('cross-line', () => {
+    describe.skip('cross-line', () => {
```
alantreadway: Nit: Remove

packages/ag-charts-enterprise/src/features/annotations/axisButton.ts
```diff
@@ -0,0 +1,134 @@
...
+
+        return {
+            wrapper,
+            button,
+        };
+    }
+
+    destroyElements(domManager: _ModuleSupport.DOMManager) {
+        domManager.removeChild('canvas-overlay', DEFAULT_ANNOTATION_AXIS_BUTTON_CLASS);
+    }
```
alantreadway: Nit: Make these `setup()` & `destroyElements` methods:
- `private`
- Use `this.ctx.domManager` rather than passing it in as a parameter?
- Appear immediately AFTER the constructor?

packages/ag-charts-enterprise/src/features/annotations/annotations.ts
```diff
@@ -499,21 +511,48 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
...
         this.onCancel();
-        this.onAxisClickAdding(event, region);
+        this.onAxisClickAdding(coords, region);
     }
 
-    private onAxisClickAdding(
-        event: _ModuleSupport.PointerInteractionEvent<'click'>,
-        region: _ModuleSupport.RegionName
-    ) {
+    private onAxisClickAdding(coords?: Coords, region?: _ModuleSupport.RegionName) {
```
lsjroberts: Tbh, may as well just have the one function here `onAxisButtonClick()`, rather than calling out to `onAxisClickAdding()`. If/when we do other stuff we can break it up as appropriate.

packages/ag-charts-enterprise/src/features/annotations/annotations.ts
```diff
@@ -499,21 +511,48 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
...
         this.state.transition(AnnotationType.CrossLine);
 
-        this.onClickAdding(event, region);
+        const {
+            active,
+            annotationData,
+            annotations,
+            state,
+            ctx: { toolbarManager },
+        } = this;
```
lsjroberts, line 533: Move this above the `this.ctx...` line above, and add it to the destructure

packages/ag-charts-enterprise/src/features/annotations/axisButton.css
```diff
@@ -0,0 +1,31 @@
+.annotations-axis-button {
```
lsjroberts: Please use bem css naming - https://getbem.com/naming/

And it's best to avoid referencing the element type like `button` directly.

e.g.

```css
.ag-charts-annotations__axis-button-wrapper {}
.ag-charts-annotations__axis-button {}
.ag-charts-annotations__axis-button--hidden {}
```

and https://github.com/ag-grid/ag-charts/blob/latest/packages/ag-charts-community/src/chart/toolbar/toolbarStyles.css

packages/ag-charts-enterprise/src/features/annotations/axisButton.ts
lsjroberts: +1

I find a good ordering on class functions is: static fns -> constructor -> public fns -> private fns in the order they are referenced by constructor and other fns

packages/ag-charts-enterprise/src/features/annotations/axisButton.css
```diff
@@ -0,0 +1,31 @@
...
+    color: rgb(71, 71, 71);
+}
+
+.annotations-axis-button button:hover {
+    background: rgba(71, 71, 71, 0.8);
+    color: rgb(255, 255, 255);
+}
+
+.annotations-axis-button-hidden {
+    top: -10000px !important;
```
lsjroberts: I think `display: none` or `visibility: hidden` would be better, if either of those work for your situation.

packages/ag-charts-enterprise/src/features/annotations/axisButton.css
```diff
@@ -0,0 +1,31 @@
...
+    position: absolute;
+    left: 0px;
+    top: 0px;
+    user-select: none;
+    font:
+        15px Verdana,
+        sans-serif;
+    overflow: hidden;
+    white-space: nowrap;
+    z-index: 99998;
```
lsjroberts: Given the new chart-overlay container, I don't think these massive z-indices are needed any more.

And if we need to order elements within the chart-overlay we should use variables, e.g. `--ag-charts-layer-tooltip: 1;`, so we can coordinate them in the theme -> https://github.com/ag-grid/ag-charts/blob/latest/packages/ag-charts-community/src/styles/styles.css

packages/ag-charts-enterprise/src/features/annotations/axisButton.ts
```diff
@@ -0,0 +1,134 @@
...
+            return;
+        }
+
+        this.wrapper.style.transform = `translate(${Math.round(x)}px, ${Math.round(y)}px)`;
+    }
+
+    private updateButtonElement() {
+        const { button } = this;
+        button.onclick = _ModuleSupport.makeAccessibleClickListener(button, () => this.onButtonClick(this.coords));
+        button.innerHTML = `<span>+</span>`;
```
lsjroberts: We should use an icon instead of text:

e.g.

```css
/* community/src/styles/styles.css */

.ag-charts-icon-plus {
    mask-image: url(data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHdpZHRoPSIyNCIgaGVpZ2h0PSIyNCIgdmlld0JveD0iMCAwIDI0IDI0IiBmaWxsPSJub25lIiBzdHJva2U9ImN1cnJlbnRDb2xvciIgc3Ryb2tlLXdpZHRoPSIyIiBzdHJva2UtbGluZWNhcD0icm91bmQiIHN0cm9rZS1saW5lam9pbj0icm91bmQiIGNsYXNzPSJsdWNpZGUgbHVjaWRlLXBsdXMiPjxwYXRoIGQ9Ik01IDEyaDE0Ii8+PHBhdGggZD0iTTEyIDV2MTQiLz48L3N2Zz4=);
}
```

```ts
button.classList.add('ag-charts-icon-plus');
```

lsjroberts (changes requested):

lsjroberts (approved):
Excellent, lgtm

alantreadway (approved):
LGTM

PR #1748, opened 2024-06-13
AG-11700 Add dragging annotations from any line

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-11700

Bonus: fixes the code smells in annotations, adds some vector2 math utils that will help clean up various duplications and util fns dotted about annotations and zoom


alantreadway (approved):
Nice, LGTM

PR #1749, opened 2024-06-13
AG-11591 QA: Fix tooltip.range chart update

olegat:
https://ag-grid.atlassian.net/browse/AG-11591

This hooks the tooltip.range option into the same mechanism as the tooltip.position, which is more consistent, stable and maintainable.


alantreadway (approved):
LGTM

PR #1750, opened 2024-06-13
AG-11297 QA: Swap pan/min slider tab ordering around

olegat:
https://ag-grid.atlassian.net/browse/AG-11297


alantreadway (approved):
LGTM

PR #1752, opened 2024-06-13
AG-11527 - Rework presets into AgFinancialChartOptions.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-11527

Based upon feedback earlier this week, breaks the inheritance between `AgFinancialChartOptions` and `AgChartOptions` and moves `preset` up a level in the option heirarchy.

### DIFF
--- a/packages/ag-charts-community/src/api/preset/presets.ts
+++ b/packages/ag-charts-community/src/api/preset/presets.ts
@@ -7,5 +7,6 @@ export const PRESETS: { [K in Preset['type']]: (p: Preset & { type: K }) => AgCh
 };
 
 export function isAgFinancialChartOptions(x: any): x is AgFinancialChartOptions {
-    return x.type != null && PRESETS[x.type] != null;
+    const { type } = x;
+    return type != null && typeof type === 'string' && PRESETS[type as Preset['type']] != null;
 }
--- a/packages/ag-charts-community/src/api/preset/presets.ts
+++ b/packages/ag-charts-community/src/api/preset/presets.ts
@@ -7,5 +7,5 @@ export const PRESETS: { [K in Preset['type']]: (p: Preset & { type: K }) => AgCh
 };
 
 export function isAgFinancialChartOptions(x: any): x is AgFinancialChartOptions {
-    return x.type != null && Object.keys(PRESETS).includes(x.type);
+    return x.type != null && PRESETS[x.type] != null;
 }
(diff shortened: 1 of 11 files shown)


lsjroberts (approved):
LGTM (other than build break)

packages/ag-charts-react/src/index.ts
```diff
@@ -26,7 +26,7 @@ interface BaseChartProps {
 function getOptions(options: AgChartOptions, containerRef: RefObject<HTMLElement | null>): AgChartOptions {
     return {
         ...options,
-        container: containerRef.current,
+        container: containerRef.current ?? undefined,
```
jacobp100: Why was this null?

packages/ag-charts-community/src/api/preset/presets.ts
```diff
@@ -1,7 +1,11 @@
...
+import type { AgChartOptions, AgFinancialChartOptions, Preset } from 'ag-charts-types';
 
 import { candlestickVolumePreset } from './candlestickVolumePreset';
 
 export const PRESETS: { [K in Preset['type']]: (p: Preset & { type: K }) => AgChartOptions } = {
     'candlestick-volume': candlestickVolumePreset,
 };
+
+export function isAgFinancialChartOptions(x: any): x is AgFinancialChartOptions {
+    return x.type != null && Object.keys(PRESETS).includes(x.type);
```
jacobp100:
```suggestion
    return x.type != null && PRESETS[x.type] != null;
```

packages/ag-charts-react/src/index.ts
alantreadway: We can probably revert this part tbh, I was trying to get rid of the support for `null` in the options contract, but it would mean changing ALL examples, so needs a bit more thought.

jacobp100: I think it should never be null at this point - it should always be a HTMLDivElement

alantreadway: Reverted

packages/ag-charts-community/src/api/preset/presets.ts
```diff
@@ -7,5 +7,6 @@ export const PRESETS: { [K in Preset['type']]: (p: Preset & { type: K }) => AgCh
 };
 
 export function isAgFinancialChartOptions(x: any): x is AgFinancialChartOptions {
-    return x.type != null && PRESETS[x.type] != null;
+    const { type } = x;
+    return type != null && typeof type === 'string' && PRESETS[type as Preset['type']] != null;
```
jacobp100, line 11:
```suggestion
    return typeof type === 'string' && PRESETS[type as Preset['type']] != null;
```

PR #1754, opened 2024-06-14
AG-11876 Colour logic cleanup

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11876


packages/ag-charts-enterprise/src/series/box-plot/boxPlotThemes.ts
```diff
@@ -3,6 +3,7 @@ import { _Theme } from 'ag-charts-community';
 export const BOX_PLOT_SERIES_THEME = {
     series: {
         direction: 'vertical' as const,
+        fillOpacity: 0.3,
```
alantreadway: We deliberately made a choice NOT to use `fillOpacity` to create colour offsets previously as it is non-intuitive for users to adjust, and leads to incorrect rendering with overlapping series types - please revert.

jacobp100: Reverted - but left the new code in because I think we should switch to it

alantreadway (approved):
LGTM

PR #1757, opened 2024-06-14
AG-11879 Partial fix of imports getting replaced in strings

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11879

Partial fix - we should do this at the AST level. E.g. if you import `time`, it'll get replaced in `text: 'something time something'`. We don't have anything that operates like this at the moment, so we'd need to add it


alantreadway (approved):
LGTM

PR #1763, opened 2024-06-14
AG-11148 Add initial state to replace annotations initial

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-11148


packages/ag-charts-community/src/api/state/initialStateModule.ts
```diff
@@ -0,0 +1,11 @@
+import type { Module } from '../../module/module';
+import { InitialState } from './initialState';
+
+export const InitialStateModule: Module = {
+    type: 'root',
+    optionsKey: 'initialState',
+    dependencies: ['annotations'],
```
lsjroberts: This depends on annotations, which in turn depends on toolbar. The previous depedency ordering logic could not handle this, so required the change in this PR.

alantreadway: I'm wondering if this should be the other way around though - it doesn't make sense for a community feature to depend on an enterprise feature?

Could we work the other way around, and have the `InitialStateModule` a dependency of `AnnotationModule` - and the later registers with the former at runtime to percolate `get`/`set` handling for the registered sub-tree?

If a user is only using `ag-charts-community`, then we would only support the null-set of root properties right no? (The typings would still reflect the enterprise super-set though)?

lsjroberts: Added `StateManager` onto the module context. On `initialState.annotations` change it updates the state on the manager. Then `Annotations.constructor()` calls `StateManager.requestRestore()` to trigger a restoration if there is any state available.

packages/ag-charts-community/src/module/module.ts
```diff
@@ -56,48 +56,50 @@ class ModuleRegistry {
...
-            }
-        }
+        const yielded = new Set();
+
+        let count = 0;
+        const maxCount = 3;
+
+        const modulesByType = this.modules.filter((module) => types.includes(module.type));
+
+        do {
```
alantreadway, line 70: Might be worth adding a comment here to explain what is going on - my understanding is this is trying to take an iterative approach to resolving dependencies rather than a recursive approach, with a max iteration limit of 3?

Alternatively a recursive approach is fine, the only thing you need to do is pass down context of which modules have already been considered so you can detect loops and fail-fast?

packages/ag-charts-enterprise/src/features/annotations/__image_snapshots__/annotations-test-ts-annotations-cross-line-when-x-axis-is-clicked-it-should-create-a-vertical-cross-line-1-snap.png
alantreadway, line 1: These snapshot changes look suspect?

alantreadway (approved):
LGTM

PR #1771, opened 2024-06-17
AG-11737 Restore 9.3.0 Zoom double-click behaviour

olegat:
https://ag-grid.atlassian.net/browse/AG-11737


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1446,6 +1446,17 @@ export abstract class Chart extends Observable {
...
+        if (pickedNode) {
+            // See: AG-11737#TC3, AG-11676
+            //
+            // The Zoom module's double-click handler resets the zoom, but only if there isn't an
+            // exact match on a node. This is counter-intuitive, and there's no built-in mechanism
+            // in the InteractionManager / RegionManager for the Zoom module to listen to non-exact
+            // series-rect double-clicks. As a workaround, we'll set this boolean to tell the Zoom
+            // double-click handler to ignore the event whenever we are double-clicking exactly on
+            // a node.
+            event.preventZoomDblClick = true;
```
olegat, line 1458: Note: I acknowledge that this is a Hacky solution. However I believe that it's the best approach at this moment in time.

The double-click behaviour of the Zoom handler in 9.3.0 is inconsistent: it is only ignored when the double-click is exactly on a node, even if the `nodeClickRange` is set to `'nearest'`. IMHO, it would make more sense for the zoom dblclick handle to either be always or never fired when a node is double-clicked. But the desired requirement is to keep the same behaviour as 9.3.0.

I would be possible to avoid this hack by rewriting parts of the interaction backbone, but it's not worth it for this small edge-case.

packages/ag-charts-enterprise/src/features/zoom/zoom.ts
```diff
@@ -269,7 +271,7 @@ export class Zoom extends _ModuleSupport.BaseModuleInstance implements _ModuleSu
             const { direction } = hoveredAxis;
             const axisZoom = direction === ChartAxisDirection.X ? x : y;
             this.updateAxisZoom(direction, axisZoom);
-        } else if (paddedRect?.containsPoint(event.offsetX, event.offsetY)) {
+        } else if (paddedRect?.containsPoint(event.offsetX, event.offsetY) && !event.preventZoomDblClick) {
```
olegat, line 274: Note: This logic replicates a behaviour that was lost in https://github.com/ag-grid/ag-charts/commit/e9e97720c93c667308dce54dda80b615846d60ea

packages/ag-charts-community/src/chart/chart.ts
alantreadway: Let's sync on this tomorrow, I think we're still dancing around the intrinsic behaviour that only a single action should result from a given interactions (certainly in the case of `click`, `drag`, maybe not in the case of `hover`), and I think we need to include that capability into the core of the InteractionService.

alantreadway: Discussed offline, we'll roll with this for now.

alantreadway (approved):
LGTM

PR #1776, opened 2024-06-17
AG-11509 Add toolbar screenreader support

olegat:

### DIFF
--- a/packages/ag-charts-community/src/chart/locale/localeManager.ts
+++ b/packages/ag-charts-community/src/chart/locale/localeManager.ts
@@ -23,8 +23,16 @@ export class LocaleManager extends Listeners<'locale-changed', () => void> {
         }
     }
 
-    t(key: string, variables: Record<string, any> = {}): string {
+    t(key: string, variables?: Record<string, any>): string;
+    t(key: undefined, variables?: Record<string, any>): undefined;
+    t(key: string | undefined, variables?: Record<string, any>): string | undefined;
+
+    t(key: string | undefined, variables?: Record<string, any>): string | undefined {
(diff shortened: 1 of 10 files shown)


packages/ag-charts-community/src/chart/locale/localeManager.ts
```diff
@@ -23,8 +23,16 @@ export class LocaleManager extends Listeners<'locale-changed', () => void> {
         }
     }
 
-    t(key: string, variables: Record<string, any> = {}): string {
```
alantreadway: I think we should leave the strict signature and handle the exception (no `key`) to the caller - `variables` is already defaulted inline previously too.

jacobp100 (approved):
Looks good. Agree about leaving undefined to the caller just to simplify the API

packages/ag-charts-community/src/chart/locale/localeManager.ts
olegat: The signature is strict. It always returns `undefined` if the input is `undefined` and always returns a `string` if the input is a `string`.

It is preferable to handle the no-key case inside a shared function, because it's likely that we're going to encounter more optional text strings. For example, we're probably going to see more `ariaSomething?: string` properties the API options contract at some point. We don't want to duplicate the `if(undefined) return undefined` logic everywhere.

Whether or not the `LocaleManager.t` method should be a one to handle this no-key case is debate however. If preferred, we could leave `LocaleManager.t` as-is and add a second method `LocaleManager.tOptional`

alantreadway (approved):
LGTM

PR #1777, opened 2024-06-17
AG-11523 Make context menu for friendly to screen-readers

olegat:
https://ag-grid.atlassian.net/browse/AG-11523

### DIFF
--- a/packages/ag-charts-community/src/util/keynavUtil.ts
+++ b/packages/ag-charts-community/src/util/keynavUtil.ts
@@ -11,15 +11,13 @@ function addRemovableEventListener<K extends keyof HTMLElementEventMap>(
 function addEscapeEventListener(
     destroyFns: (() => void)[],
     elem: HTMLElement,
-    onEscape?: (event: KeyboardEvent) => void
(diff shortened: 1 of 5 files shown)


packages/ag-charts-community/src/util/keynavUtil.ts
```diff
@@ -8,6 +8,20 @@ function addRemovableEventListener<K extends keyof HTMLElementEventMap>(
     destroyFns.push(() => button.removeEventListener(type, listener));
 }
 
+function addEscapeEventListener(
+    destroyFns: (() => void)[],
+    elem: HTMLElement,
+    onEscape?: (event: KeyboardEvent) => void
+) {
+    if (onEscape) {
```
alantreadway: I don't think we should be pusing `undefined` handling down like this - there is no point invoking this method if `onEscape` is undefined in the first place, and makes the calling code less readable because it looks like an event handler is always added?

olegat: ACK.

alantreadway (approved):
LGTM

PR #1778, opened 2024-06-17
AG-10318 Localisation exports

jacobp100:

* Adds new ag-charts-locale bundle (with UMD build)
* Updates SJS to handle imports to the new package
* Note not all translations are available yet - I'll add in a follow up PR


jacobp100:
SonarQube doesn't like our translation files - think we can ignore

packages/ag-charts-website/public/example-runner/charts-angular-boilerplate/systemjs.config.dev.js
alantreadway, line 1: I'd prefer to avoid any changes is the SystemJS files, as the config here is very fragile and we don't have a convenient way to test the production configs

jacobp100: It’s necessary for the examples. I agree it’s not ideal, and tbh I don’t have high confidence this won’t break staging when it’s merged - but I will come back to check that

alantreadway (approved):
LGTM

alantreadway:
^ conflicts, also can you check for a required update to `bump-versions.sh`?

PR #1790, opened 2024-06-18
AG-11433, AG-11436 Use outline instead of border for focus indicator

olegat:
https://ag-grid.atlassian.net/browse/AG-11433
https://ag-grid.atlassian.net/browse/AG-11436


alantreadway (approved):
LGTM

PR #1792, opened 2024-06-19
AG-11898 Update annotation click distance fuzziness and add vec2 utils

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-11898


alantreadway (approved):
LGTM

PR #1793, opened 2024-06-19
AG-11895 Update annotation floating buttons to disable not hide on lock

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-11895


alantreadway (approved):
LGTM

PR #1808, opened 2024-06-20
AG-11917 AG-11899 Set pointer-events: none on overlay

olegat:
https://ag-grid.atlassian.net/browse/AG-11917
https://ag-grid.atlassian.net/browse/AG-11899

There is a recent bug regression where the KeyNavManager is not correctly receiving keyboard events.

This happening because we’ve swapped the DOM ordering of the canvas and overlay elements. This was done because we need keep the correct tab-ordering (the canvas, i.e. series, is the first region when pressing TAB).

Setting `pointer-events: none` on the .ag-charts-canvas-overlay element allow mouse event to pass-through and hit the <canvas> element, which ensures that it also receives keyboard input.

### DIFF
--- a/packages/ag-charts-community/src/chart/dom/domManager.ts
+++ b/packages/ag-charts-community/src/chart/dom/domManager.ts
@@ -53,6 +53,7 @@ const STYLES = `
 }
 
 .ag-charts-canvas-overlay {
+    pointer-events: none;
     position: absolute;
     top: 0;
     left: 0;


jacobp100 (approved):
Might also be worth seeing if the HTML attribute `inert` works better

olegat:
> Might also be worth seeing if the HTML attribute `inert` works better

The `.ag-charts-canvas-overlay` element contains interactive elements like buttons and sliders which can be focused by assistive technologies. However, nesting interactive components inside of an `inert` element is apparently not recommended.

> Note: Do not nest interactive elements or important content within elements with the inert attribute as inert content and their descendants are neither focusable nor perceivable by all users.
>
> -- https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert

alantreadway (approved):
LGTM

PR #1819, opened 2024-06-20
Fix pointer events

jacobp100:


alantreadway (approved):
LGTM

packages/ag-charts-community/src/chart/dom/domManager.ts
```diff
@@ -63,6 +63,7 @@ const STYLES = `
 
 .ag-charts-canvas-overlay > * {
     position: absolute;
+    pointer-events: auto;
```
olegat, line 66: Please re-test mouse events on the navigator and legend buttons before merging.

I had also written this line this morning but removed it because it broke legend toggling.

olegat (changes requested):

packages/ag-charts-community/src/chart/dom/domManager.ts
olegat: Correction: I've double-checked my local git history and found that I accidentally wrote `pointer-event: auto !important;`.

I've tested clicks on the legend with this PR and they work. LGTM

PR #1823, opened 2024-06-21
AG-11891 Focus styles for colour picker

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11891


packages/ag-charts-community/src/chart/dom/domStyles.css
```diff
@@ -1,14 +1,67 @@
...
-[class^='ag-charts']:after,
-[class^='ag-charts']:before {
+.ag-charts-root,
+.ag-charts-root:focus,
+.ag-charts-root:after,
+.ag-charts-root:before,
+.ag-charts-root *,
+.ag-charts-root *:focus,
+.ag-charts-root *:after,
+.ag-charts-root *:before {
```
jacobp100: These have the same specificity, but this way is a lot better for performance. Modern CSS engines have things like bloom filters for class names so avoids a lot of substring checks

packages/ag-charts-community/src/chart/dom/domStyles.css
```diff
@@ -1,14 +1,67 @@
 /********
 * RESET *
 ********/
-[class^='ag-charts'],
-[class^='ag-charts']:focus,
-[class^='ag-charts']:after,
-[class^='ag-charts']:before {
+.ag-charts-root,
```
alantreadway: Can we not just use `.ag-charts-wrapper` rather than adding another class?

packages/ag-charts-community/src/chart/dom/domStyles.css
alantreadway: Presuming that all of the focus boxes still appear as before?

packages/ag-charts-enterprise/src/features/color-picker/colorPickerStyles.css
```diff
@@ -48,6 +48,12 @@
     filter: drop-shadow(0 0 5px #00000038);
 }
 
+.ag-charts-color-picker__palette:active,
+.ag-charts-color-picker__palette:focus {
+    outline: 1px solid var(--ag-charts-active-color);
```
olegat: It feels like the right value to use here would be `outline: var(--ag-charts-focus-border);` so that we're consistent with the focus indicator used for scene nodes:

https://github.com/ag-grid/ag-charts/blob/3b828781e57c8ae9e424118d8c92a7c784ca263b/packages/ag-charts-community/src/chart/dom/focusStyles.css#L12

We may even want to consider adding a shared CSS focus-box class for the color-pick and scene-graph focus-indicator

olegat:
LGTM subject to Alan's comments

packages/ag-charts-community/src/chart/dom/domStyles.css
```diff
@@ -72,7 +125,7 @@
 
     /* Focus Indicator */
     --ag-charts-focus-border-color: var(--ag-input-focus-border-color, var(--ag-charts-active-color));
-    --ag-charts-focus-border: solid 2px var(--ag-charts-focus-border-color);
+    --ag-charts-focus-border: solid 1px var(--ag-charts-focus-border-color);
```
jacobp100, line 129: I changed this because this is what we'll use in all the form elements etc., and just override the width in the focus indicator for chart elements

packages/ag-charts-community/src/chart/dom/focusStyles.css
```diff
@@ -10,16 +10,14 @@
...
 .ag-charts-focus.ag-charts-focus__indicator > svg {
     width: 100%;
     height: 100%;
-}
-
-.ag-charts-focus.ag-charts-focus__indicator > svg > * {
-    fill: rgba(0, 0, 0, 0);
+    fill: none;
     stroke: var(--ag-charts-focus-border-color);
     stroke-width: 2px;
```
jacobp100, line 22: SVG inherits fills, strokes etc., so we don't need to target elements individually

alantreadway (approved):
LGTM

PR #1827, opened 2024-06-21
Add basic keyboard-nav tests.

alantreadway:


packages/ag-charts-website/e2e/keyboard-nav.spec.ts-snapshots/1st-datum-focus-chromium-darwin.png
olegat, line 1: This snapshot is incorrect; Onshore Wind - 2017 (first blue bar) should be highlighted.
This might be happening due to animations, because keyboard input is ignored during animation.

PR #1828, opened 2024-06-21
AG-11915 Skip mouse events in chart overlays

jacobp100:
* Fires enter/leave events in the interaction manager when the mouse enters a chart overlay
* Refactors `dispatchPointerEvent` to improve typing, and remove unnecessary arrays


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -374,6 +374,8 @@ export abstract class Chart extends Observable {
             seriesRegion.addListener('contextmenu', (event) => this.onContextMenu(event), All),
             ctx.keyNavManager.addListener('browserfocus', (event) => this.onBrowserFocus(event)),
             ctx.interactionManager.addListener('page-left', () => this.destroy()),
+            ctx.interactionManager.addListener('enter', () => this.update(ChartUpdateType.SCENE_RENDER)),
+            ctx.interactionManager.addListener('leave', (event) => this.onLeave(event, true)),
```
jacobp100: This might not be necessary - I think this might be able to be handled in the regionManager, but I wasn't too sure

alantreadway: These do look a bit redundant to my eye - can you try removing these and seeing if things behave as expected?

packages/ag-charts-community/src/chart/dom/domManager.ts
```diff
@@ -336,16 +349,8 @@ export class DOMManager extends BaseManager<Events['type'], Events> {
...
-        if (element == null) return false;
-
-        while (element !== this.element) {
-            element = (element as HTMLElement).parentElement;
-            if (element == null) return false;
-        }
-
-        return true;
+        const element = event.target as HTMLElement | null;
+        return element != null && this.element.contains(element);
```
alantreadway, line 366: I'd prefer to keep the previous implementation, unless we have strong guarantees that `Element.contains()` is going to be faster then traversing parents, which is typically going to be faster than a top-down tree traversal?

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
```diff
@@ -9,22 +8,22 @@ import type { DOMManager } from '../dom/domManager';
...
-    'leave',
-    'enter',
-    'page-left',
-    'wheel',
-] as const;
-
-const FOCUS_INTERACTION_TYPES = ['blur', 'focus'] as const;
-
-const KEY_INTERACTION_TYPES = ['keydown', 'keyup'] as const;
+    'click' as const,
```
alantreadway: A single `as const` on the array should be sufficient to keep type safety? (Here and below)

alantreadway (changes requested):
These changes will break other features, can we sync on this?

packages/ag-charts-community/src/chart/dom/domManager.ts
jacobp100: Element.contains does the same thing we did - it just checks the parents. If you were interested - https://github.com/WebKit/WebKit/blob/1b86aba5013c21c198a3182424ec5c10b585aeb6/Source/WebCore/dom/Node.cpp#L1198-L1201

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
jacobp100: `as const` on the array gives a tuple, and `as const` on each element gives an array with the union of all the types

packages/ag-charts-community/src/chart/chart.ts
jacobp100: It definitely didn't work without this - but I suspect it's possible to handle this somewhere better

alantreadway (approved):
LGTM

PR #1833, opened 2024-06-21
AG-11956 Ignore key-presses with modifiers in keynavUtil.ts

olegat:
https://ag-grid.atlassian.net/browse/AG-11956


alantreadway (approved):
LGTM

PR #1834, opened 2024-06-21
AG-11915 Fix z-index of floating toolbars

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11915

### DIFF
--- a/packages/ag-charts-community/src/chart/toolbar/toolbarStyles.css
+++ b/packages/ag-charts-community/src/chart/toolbar/toolbarStyles.css
@@ -53,6 +53,8 @@
 .ag-charts-toolbar--floating {
     padding: 0;
     width: auto;
+    /* Needs to beat .ag-crosshair-label */
+    z-index: 99999;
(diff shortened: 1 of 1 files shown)


alantreadway (approved):
LGTM

packages/ag-charts-community/src/chart/toolbar/toolbarStyles.css
```diff
@@ -53,6 +53,8 @@
 .ag-charts-toolbar--floating {
     padding: 0;
     width: auto;
+    /* Needs to beat .ag-crosshair-label */
+    z-index: 99999;
```
lsjroberts, line 57: css variable in `domStyles.css`? like `--ag-charts-layer-tooltip` (if this isn't a temp fix)

jacobp100: I’m hoping a temp fix because I want to go through and fix all our z indices

PR #1835, opened 2024-06-21
AG-11908 - Add vertical and horizontal line annotation buttons to too…

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-11908


packages/ag-charts-locale/src/main.test.ts
```diff
@@ -8,7 +8,7 @@ const formatters = ['number', 'percent', 'date', 'time', 'datetime'];
 const formatterRegExp = /\[([^\]]*)\]/g;
 const variableRegExp = /(\$\{\w+\})/g;
 
-describe('translations', () => {
+describe.skip('translations', () => {
```
alantreadway, line 12: Nit: Can you add a comment and link to the follow-up JIRA to re-enable this?

PR #1842, opened 2024-06-24
AG-NONE Do not use getComputedStyle()

olegat:
This is because the function is causing errors in our utilities (thumbnail generator). We cannot use alignElement.style.flexDirection directly either though, because that value expands to ''. So just create a simple ancestor traversor to find the nearest none-empty `flex-direction` property.

This is not a perfect solution, because it does not properly compute some cases (e.g. if an ancestor has an !important flex-direction set), but it is good enough for our use case.


alantreadway (approved):
LGTM

PR #1844, opened 2024-06-24
AG-11892 Update color picker to reposition within canvas bounds

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-11892

Screenshot 2024-06-24 at 11 50 48

### DIFF
--- a/packages/ag-charts-enterprise/src/features/color-picker/colorPicker.ts
+++ b/packages/ag-charts-enterprise/src/features/color-picker/colorPicker.ts
@@ -3,7 +3,7 @@ import { _ModuleSupport, _Scene, _Util } from 'ag-charts-community';
 import colorPickerStyles from './colorPickerStyles.css';
 import colorPickerTemplate from './colorPickerTemplate.html';
 
-const { createElement } = _ModuleSupport;
+const { clamp, createElement } = _ModuleSupport;
 
 const { Color } = _Util;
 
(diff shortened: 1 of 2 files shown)


packages/ag-charts-enterprise/src/features/color-picker/colorPicker.ts
```diff
@@ -22,6 +22,8 @@ const getHsva = (input: string) => {
 
 export class ColorPicker extends _ModuleSupport.BaseModuleInstance implements _ModuleSupport.ModuleInstance {
     private readonly element: HTMLElement;
+    private anchor?: { x: number; y: number };
```
jacobp100, line 25: Maybe the anchor should be a rect, and we position ourself around that rect?

jacobp100 (approved):
It'd be good if we followed as close to this as possible so we're able to easily switch when browser support improves - https://developer.chrome.com/blog/anchor-positioning-api

But also appreciate we're close to release, and this won't have a big knock-on effect after

packages/ag-charts-enterprise/src/features/color-picker/colorPicker.ts
lsjroberts: The reasoning for using the fallback with optional `x` and `y` is because in this instance, we do not want it to jump to the left of the buttons. The behaviour differs between x & y.

If you provide an x and not a y, it slides along the y axis, but jumps on the x. In the case where we have a vertical list of buttons, we will want this behaviour.

In the annotations case it slides along the x axis, but jumps on the y.

Sliding x:
Screenshot 2024-06-25 at 12 33 25

Jumping y:
Screenshot 2024-06-25 at 12 33 32

lsjroberts:
> It'd be good if we followed as close to this as possible so we're able to easily switch when browser support improves - https://developer.chrome.com/blog/anchor-positioning-api
> 
> But also appreciate we're close to release, and this won't have a big knock-on effect after

Thanks, looks very useful. It's entirely internal so lets look at it post-release.

alantreadway (approved):
LGTM

PR #1848, opened 2024-06-24
AG-11917 Remove defaultPrevent() on drag event

olegat:
https://ag-grid.atlassian.net/browse/AG-11917

This ensures that the canvas receives focus when it is clicked on.

Also, the defaultPrevent() call no longer seems necessary, is might be because the canvas is receiving the MouseEvents, and therefore nothing happens if defaultPrevent() isn't called.

The defaultPrevent() call is still necessary for the wheel event. I'm not sure whether it's necessray for the +/- keyboard presses, but I would assume yes.


alantreadway (approved):
LGTM

PR #1851, opened 2024-06-24
AG-11424 Announce changes to the chart overlay

olegat:
https://ag-grid.atlassian.net/browse/AG-11424

This will cause screenreaders to announce "No visible series" when the last legend item is toggled off. A polite aria-live is used to use interrupting the announcement of the tickbox/button's state change


olegat:
**Discussed in video-call:**
Hold-off on merge this PR until AG-11395 is Done.

alantreadway (approved):
LGTM

PR #1855, opened 2024-06-25
AG-11924 Ensure changing chart type preserves container size

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11924

Issue is when the chart changes, a new dom manager is created, but we don't read the container size


packages/ag-charts-community/src/chart/baseManager.ts
```diff
@@ -8,10 +8,6 @@ export abstract class BaseManager<EventType extends string = never, Event extend
         return this.listeners.addListener(type, handler);
     }
 
-    public removeListener(listenerSymbol: symbol) {
```
jacobp100, line 11: These weren't used

alantreadway (approved):
LGTM

PR #1859, opened 2024-06-25
AG-11395 Restore the same screen-reader behaviour on legend items as in production

olegat:
https://ag-grid.atlassian.net/browse/AG-11395


alantreadway (approved):
LGTM

PR #1860, opened 2024-06-25
AG-11194 Fix vertical toolbars using hardcoded aria-orientations

olegat:
https://ag-grid.atlassian.net/browse/AG-11194

This is a simplified & hard-coded solution that does not rely on the getComputedStyle() global function.


alantreadway (approved):
LGTM

PR #1861, opened 2024-06-25
AG-11701 Fix toolbar border styles

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-11701

Also changes z-indices to use layer vars.


alantreadway (approved):
LGTM

PR #1862, opened 2024-06-25
Add details_urls to GHA status checks.

alantreadway:


alantreadway:
Probably needs to use a different plugin such as: https://github.com/marketplace/actions/commit-status-updater

PR #1863, opened 2024-06-25
AG-11509 Add aria labels for toolbars

olegat:


packages/ag-charts-community/src/chart/toolbar/toolbar.ts
```diff
@@ -538,6 +540,22 @@ export class Toolbar extends BaseModuleInstance implements ModuleInstance {
...
+            alignElement = this.positionAlignments[position][align];
+            if (!alignElement) return;
+        }
+        const map = {
+            annotations: 'ariaLabelAnnotationsToolbar',
+            annotationOptions: 'ariaLabelAnnotationOptionsToolbar',
+            ranges: 'ariaLabelRangesToolbar',
+            zoom: 'ariaLabelZoomToolbar',
+        } as const;
+        alignElement.ariaLabel = this.ctx.localeManager.t(map[group]);
```
lsjroberts, line 556: I'm happy to go with this for now. But we could put it in the `AgToolbarGroup` options and `ToolbarGroupProperties` and access it here as `this[group].ariaLabel` if we want users to be able to change it?

lsjroberts: Also, a `.ag-charts-toolbar__align` is not limited to just one group of buttons (annotations/zoom etc).

If a user sets the same position and alignment for two button groups it will put them next to each other in this same element. And so will set the aria-label to be the latest updated.

The other possibility is the `.ag-charts-toolbar__section` element, but a group can be split across multiple sections, e.g. the annotations has the delete button in its own section:

Screenshot 2024-06-26 at 09 41 50

But that would end up with multiple elements having the same aria-label.

I think the current solution is fine, as by default all the groups are separate, but its worth noting.

lsjroberts (approved):
LGTM, but a couple of things to consider, if not for this PR then another ticket.

packages/ag-charts-community/src/chart/toolbar/toolbar.ts
olegat: The `this[group].ariaLabel` sounds like a good idea, but the user can already change this value simply by updating the localised string. So we can keep it as it for now.

The solution for multiple toolbars in a one `.ag-charts-toolbar__align` will require a bit more. I think that we'll need an intermediate container between `.ag-charts-toolbar__align` and `.ag-charts-toolbar__section`. Let's do this in a follow up PR.

PR #1870, opened 2024-06-25
AG-11395 Make legend items role="switch" elements

olegat:
https://ag-grid.atlassian.net/browse/AG-11395


olegat:
Abandoned. It would be preferable to use a toggle button using `aria-pressed`
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#associated_aria_roles_states_and_properties

PR #1871, opened 2024-06-25
Remove redundant Chart.processedOptions, fix stale options state bugs.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-11506

Removes redundant copy of options kept in `Chart.processedOptions`, and consequently tries to straighten out the set of options used in `Chart.applyOptions()` to reduce complexity and fix a couple of bugs due to mixed up options.


iMoses (approved):
LGTM

PR #1885, opened 2024-06-26
Fix ThemeApi docs page by replacing type-literals with interfaces

iMoses:


packages/ag-charts-types/src/chart/themeOptions.ts
```diff
@@ -51,6 +51,11 @@ export type AgChartThemeName =
     | 'ag-material'
     | 'ag-material-dark';
 
+interface AgPaletteColors {
+    fill?: CssColor;
```
jacobp100, line 55: Do these need comments?

alantreadway (approved):
LGTM

packages/ag-charts-types/src/chart/themeOptions.ts
iMoses: Yes, actually.. Yes they do :)

PR #1886, opened 2024-06-26
AG-11509 Add intermediate container for role="toolbar"

olegat:
This fixes a bug where two toolbars in the same position are seen by AT as a single role="toolbar".


alantreadway (approved):
LGTM

alantreadway (approved):
LGTM

PR #1887, opened 2024-06-26
AG-11993 Add zoom to initialState

lsjroberts:

This also fixes some issues with independent axis zooming (see ticket for details).


alantreadway (approved):
LGTM

lsjroberts:
Leaving as draft until we are ready to put this into a release.

packages/ag-charts-types/src/chart/chartBuilderOptions.ts
```diff
@@ -137,9 +137,15 @@ export interface ImageDataUrlOptions {
     fileFormat?: string;
 }
 
+// TODO: this is a breaking change as previously it was permissive of any type for each key
+// export interface AgChartState extends AgInitialStateOptions {
+//     version: string;
+// }
+
```
lsjroberts: We should have been doing this previously, but can't introduce it just yet as it is a breaking change.

packages/ag-charts-community/src/sparklines-util.ts
```diff
@@ -12,7 +12,7 @@ export * from './util/vector';
 export * from './util/zip';
 
 export { isValidDate as isDate, isFiniteNumber as isNumber, isString } from './util/type-guards';
-export { bifurcate, extent, normalisedExtent, normalisedExtentWithMetadata } from './util/array';
```
alantreadway, line 15: Can you confirm this isn't being used in the Grid repo?

lsjroberts: Yep, not used (I introduced it specifically for this use-case) - https://github.com/search?q=repo%3Aag-grid%2Fag-grid%20birfurcate&type=code

packages/ag-charts-types/src/chart/chartBuilderOptions.ts
lsjroberts: Added deprecation notice

alantreadway (approved):
LGTM

PR #1888, opened 2024-06-26
CRT-344, AG-11966 Fix aria announcements in VoiceOver/Safari

olegat:
https://ag-grid.atlassian.net/browse/CRT-344
https://ag-grid.atlassian.net/browse/AG-11966

This moves the aria-live element outside of the <canvas> element


alantreadway (approved):
LGTM

PR #1898, opened 2024-06-27
CRT-366 Fix async zoom example failing on date formatting

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-366

### DIFF
--- a/packages/ag-charts-website/src/content/docs/zoom/_examples/zoom-async/main.ts
+++ b/packages/ag-charts-website/src/content/docs/zoom/_examples/zoom-async/main.ts
@@ -44,12 +44,17 @@ const options: AgCartesianChartOptions = {
                 maxSpacing: 200,
             },
             label: {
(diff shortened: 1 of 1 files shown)


packages/ag-charts-website/src/content/docs/zoom/_examples/zoom-async/main.ts
```diff
@@ -44,12 +44,17 @@ const options: AgCartesianChartOptions = {
                 maxSpacing: 200,
             },
             label: {
-                formatter: ({ value }) =>
-                    Intl.DateTimeFormat('en-GB', {
+                formatter: (event) => {
+                    let date = new Date(event.value);
+                    if (isNaN(date as any)) {
+                        date = new Date(parseInt(event.value));
+                    }
```
alantreadway, line 51: Why is value sometimes different?

PR #1901, opened 2024-06-27
CRT-361 Fix toolbar button label text to not wrap

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-361


alantreadway (approved):
LGTM

PR #1904, opened 2024-06-27
CRT-363 Fix formatting of toolbar ranges button value type in api docs

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-363


alantreadway (approved):
LGTM

PR #1909, opened 2024-06-27
CRT-346 Fix annotation options toolbar moving when toggle lock

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-346


alantreadway (approved):
LGTM

PR #1912, opened 2024-06-27
CRT-337 Remove pointer events from elements that should not capture them

jacobp100:
https://ag-grid.atlassian.net/browse/CRT-337
https://ag-grid.atlassian.net/browse/CRT-388


alantreadway (approved):
Discussed offline, LGTM

PR #1913, opened 2024-06-27
CRT-365 Fix annotation handles strokeWidth when toggling lock

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-365
https://ag-grid.atlassian.net/browse/CRT-345


alantreadway (approved):
LGTM

PR #1915, opened 2024-06-27
CRT-347 Fix annotation axis labels to use default color

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-347


alantreadway (approved):
LGTM

PR #1916, opened 2024-06-27
CRT-340 Fix candlesticks not rendered on narrow charts

olegat:
https://ag-grid.atlassian.net/browse/CRT-340


packages/ag-charts-enterprise/src/series/ohlc/ohlcSeriesBase.ts
```diff
@@ -289,7 +289,8 @@ export abstract class OhlcSeriesBase<
                 closeValue,
                 highValue,
                 lowValue,
-                bandwidth: Math.round(barWidth),
+                // CRT-340 Use atleast 1px width to prevent nothing being drawn.
+                bandwidth: Math.max(1, Math.round(barWidth)),
```
alantreadway: Could we match `BarSeries` behaviour?
https://github.com/ag-grid/ag-charts/blob/609d8f6b08cf88cf6e04b989659fe3af6a205bb7/packages/ag-charts-community/src/chart/series/cartesian/abstractBarSeries.ts#L99-L104

olegat: Done.

PR #1935, opened 2024-06-28
CRT-337 - Click focus fix

alantreadway:


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1154,7 +1154,7 @@ export abstract class Chart extends Observable {
     }
 
     private onBrowserFocus(event: KeyNavEvent<'browserfocus'>): void {
-        if (event.delta > 0) {
+        if (event.delta >= 0) {
```
olegat, line 1157: Keep `> 0`.
The delta can be `0` when changing browser tab. When this happens, we do not want to reset the focus to the 1st datum / 1st series, but just reset the focus indicator on the current series/datum.

alantreadway: Let's accept this as a known issue for now, clicking on the canvas to focus is a more common scenario that needs this change?

alantreadway: Added a comment to this block so future readers are aware of the two use-cases.

olegat: Thank you for adding the comment to clarify this change. However I still cannot approve this change. `event.delta: 0` means "draw focus indicator", so we should honour that assumption. `focus: { datumIndex, seriesIndex }` is already initialised the 0, 0 by the Chart class, so we do not need to rely on `onBrowserFocus` to initialise these variables.

I've tested this change to understand what the problem is. AFAICT, it looks like the bug is in the `else` clause. It moves the current focus to the end if `event.delta` is 0 or -1. That is incorrect, it should only move the focus if `event.delta` is negative. This should be:

```
if (event.delta > 0) {
    this.focus.datum = undefined;
    this.focus.series = undefined;
    this.focus.datumIndex = 0;
    this.focus.seriesIndex = 0;
} else if  (event.delta < 0) {
    this.focus.datum = undefined;
    this.focus.series = undefined;
    this.focus.datumIndex = Infinity;
    this.focus.seriesIndex = Infinity;
}
```

alantreadway: Suggestion made post-merge.

PR #1938, opened 2024-06-28
Improve Hebrew locale translations

iMoses:


packages/ag-charts-types/src/chart/localeOptions.ts
```diff
@@ -1,8 +1,10 @@
+import type { Formatter } from './callbackOptions';
+
 export interface AgLocaleThemeableOptions {
     /** A record of locale texts keyed by id. */
     localeText?: Record<string, string>;
     /** Formatter that generates the text displayed to the user. */
-    getLocaleText?: (params: MessageFormatterParams) => string | undefined;
+    getLocaleText?: MessageFormatter;
```
jacobp100, line 7: Do the docs still look the same after this?

iMoses: Yes, it should. the function resolving was fixed to support it.

jacobp100 (approved):

alantreadway (approved):
LGTM

PR #1945, opened 2024-06-28
Fix mini-chart breaking drag-to-zoom-axes.

alantreadway:
Caused by mini-chart registering it's axes with the `RegionManager`, overwriting the main charts axes regions.


lsjroberts (approved):
LGTM

PR #1955, opened 2024-07-01
AG-11995 Cleanup toolPanel css

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11995

Summary:
- Use `.toolbar` class - it's automatically styled
- Buttons, labels etc. can go directly in the toolbar without extra wrappers, and are laid out horizontally
- Use `hr` in the toolbar if you need multiple rows


alantreadway (approved):
LGTM. For the test that is failing, the `Save` button doesn't update the chart state (just logs some JSON to the console) - it would be good to tweak the test executor to press the button and await a console log perhaps?

PR #1961, opened 2024-07-02
AG-12010 Fix zoom range to only update the related axis

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12010


alantreadway (approved):
LGTM, thanks for the fix!

PR #1963, opened 2024-07-03
AG-12042 Fix zoom to here going to wrong position

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12042

The `minRatioX/Y` values passed to `ZoomContextMenu` were `0` as they had not yet been calculated in the `onUpdateComplete()` listener. And the `clientX` was slightly off compared to the correct `offsetX` value.


packages/ag-charts-website/e2e/context-menu.spec.ts-snapshots/zoom-to-here-chromium-darwin.png
lsjroberts, line 1: I changed the target of these tests so the x-axis labels would be more useful in determining the centre point of the zoom.

alantreadway (approved):
LGTM

PR #1966, opened 2024-07-03
Normalise anything to do with text measurement

iMoses:


alantreadway (approved):
LGTM - lets just run this past David before merge though.

PR #1971, opened 2024-07-04
AG-11990 - Fix Financial Charts volume + price series hover conflicts.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-11990

Two changes to address Financial Charts volume + price series conflicting on hover/zoom:
- Allows node highlight to be disabled on a per-series basis, and the volume series has highlight disabled to avoid conflicts with selecting the closest price series datum.
- Updates category-axis ‘closest’ node logic to stop excluding nodes that are outside the other axis range (X-axis and Y-axis respectively for Financial Charts cases).


lsjroberts (approved):
LGTM

manapeirov (approved):
LGTM thanks Alan

PR #1980, opened 2024-07-05
AG-11997: A particular label is not shown when there is enough space

iMoses:
https://ag-grid.atlassian.net/browse/AG-11997


alantreadway (approved):
LGTM

PR #1987, opened 2024-07-08
AG-12088 Rename annotation classes and add stub text annotation

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12088


alantreadway (approved):
LGTM

PR #1989, opened 2024-07-08
Ag 11205/fix context menu download

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-11205 and https://ag-grid.atlassian.net/browse/AG-12036


packages/ag-charts-community/src/chart/noopAgChartInstance.ts
```diff
@@ -0,0 +1,38 @@
+import type { AgChartOptions, AgChartState } from 'ag-charts-types';
+
+const errorMessage = 'AG Charts - Method not implemented.';
+export class NoopAgChartInstance {
```
alantreadway: I'm actually wondering if we should remove this, and just make the `Chart.publicApi` property optional?

The call in the `ContextMenu` type then becomes:
```
await this.ctx.chartService.publicApi?.download({ fileName });
```

packages/ag-charts-community/src/scene/image.ts
```diff
@@ -2,7 +2,7 @@ import type { RenderContext } from './node';
 import { Node, RedrawType, SceneChangeDetection } from './node';
 
 export class Image extends Node {
-    private readonly sourceImage: HTMLImageElement;
```
alantreadway: Can this be kept as `private` now?

alantreadway (approved):
LGTM

PR #2000, opened 2024-07-09
AG-11947 Fix nested CSS in examples getting altered

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11947

We have snippets like:-

```js
        html`
            <div
                style="${css`
                    display: grid;
                    grid: 'a b' auto / 1fr auto;
                `}"
            >
                <div id="myChart"></div>
                <textarea></textarea>
            </div>
        `
```

The issue is running prettier in the example generator was changing the single quotes to double quotes, breaking the HTML. I've turned off prettifying nested CSS in the example generator, and it'll still run in VSCode etc. This is only used for our internal tests


alantreadway (approved):
LGTM

PR #2001, opened 2024-07-09
AG-11951 Render Legend symbols as sprites

olegat:
https://ag-grid.atlassian.net/browse/AG-11951


packages/ag-charts-community/src/scene/node.ts
```diff
@@ -165,12 +167,10 @@ export abstract class Node extends ChangeDetectable {
      * one should use the {@link insertBefore} method instead.
      * @param nodes A node or nodes to append.
      */
-    append(nodes: Node[] | Node) {
+    append(nodes: Iterable<Node> | Node) {
         // Passing a single parameter to an open-ended version of `append`
         // would be 30-35% slower than this.
```
alantreadway: Is this comment still relevant, and/or something that maybe relevant to the change here?

packages/ag-charts-community/src/scene/spriteRenderer.ts
```diff
@@ -0,0 +1,56 @@
...
+            renderCtx.postMatrix ??= new Matrix();
+            renderCtx.postMatrix.setElements([1, 0, 0, 1, 0, translateY]);
+        }
+    }
+
+    renderSprite(nodes: Node | Iterable<Node>, opts?: RenderSpriteOptions): ImageBitmap {
+        nodes = toIterable(nodes);
+        const { renderCtx, offCanvas } = this;
+
+        SpriteRenderer.setTransform(renderCtx, opts?.translateY);
```
alantreadway: From previous performance analysis, we've identified matrix calculations as a performance hotspot - do we need to involve them here, rather than taking a simpler approach like just rendering the sprites at a specific x/y offset in the `drawImage()` call?

packages/ag-charts-community/src/scene/spriteRenderer.ts
```diff
@@ -0,0 +1,56 @@
...
+            ctx,
+            devicePixelRatio: 1,
+            forceRender: true,
+            resized: false,
+            debugNodes: {},
+        };
+    }
+
+    resize(width: number, height: number) {
+        if (!isStrictPositive(width) || !isStrictPositive(height)) {
```
alantreadway: Nit: I think it might be more readable to just write `width > 0 && height > 0`?

packages/ag-charts-community/src/chart/markerLabel.ts
```diff
@@ -62,40 +67,35 @@ export class MarkerLabel extends Group {
...
+    setOpacity(opacity: number) {
+        this.opacity = opacity;
+        this.bitmap.opacity = opacity;
+        this.label.opacity = opacity;
+    }
+
+    update(spriteRenderer: SpriteRenderer, dimensionProps: { length: number; spacing: number }[]) {
         const { markers, lines } = this;
 
+        let [spriteX, spriteY] = [0, 0];
```
alantreadway: Nit: Could we just go with a simpler `let spriteX = 0;`, as it's more consistent with the rest of the code-style, and doesn't have a nebulous runtime performance/memory churn overhead?

packages/ag-charts-community/src/chart/markerLabel.ts
```diff
@@ -104,11 +104,14 @@ export class MarkerLabel extends Group {
 
             const size = marker?.size ?? 0;
 
+            let [lineTop, markerTop, markerLeft] = [Infinity, Infinity, Infinity];
```
alantreadway: Ditto

libraries/ag-charts-test/src/canvas/mock-canvas.ts
```diff
@@ -1,8 +1,19 @@
...
+// However, the Canvas.drawImage(img,...) method does accept a Canvas-type img parameter. So use as new Canvas
+// as an ImageBitmap.
+(Canvas.prototype as any).transferToImageBitmap = function () {
+    const that: Canvas = this;
+    const { width, height } = that;
+    const bitmap = new Canvas(width, height);
+    bitmap.getContext('2d').drawImage(that, 0, 0, width, height);
+    return bitmap;
+};
+
```
olegat, line 16: Note: This implementation is not perfect. Some of the snapshots have bizarre rendering problems on faded legend items. However, I think that this implement is good enough for the purposes of moving ahead the PR.

packages/ag-charts-website/src/content/docs/benchmarks/_examples/summary/data.ts
olegat, line 1: Note: I do not know why this file has changed. I noticed this update after running `nx run-many -t benchmark -c update`

packages/ag-charts-community/benchmarks/__image_snapshots__/large-dataset-test-ts-large-dataset-benchmark-after-load-1-x-legend-toggle-1-snap.png
alantreadway, line 1: Looks like the `bar` series legend item has the `line` series legend item behind it?

packages/ag-charts-website/src/content/docs/benchmarks/_examples/summary/data.ts
alantreadway: There's a special `all:benchmark` which writes the updated results into this example as a short-cut - you can just run `nx benchmark` to run benchmarks and generate this update too.

Here's what the example renders like: https://charts-staging.ag-grid.com/vanilla/benchmarks/examples/summary/

packages/ag-charts-community/benchmarks/__image_snapshots__/large-dataset-test-ts-large-dataset-benchmark-after-load-1-x-legend-toggle-1-snap.png
olegat: It looks like it, but it doesn't have a line. It's a bug with node-canvas implementation of `transferToBitmap()`. The legend items render fine on browsers:

Screenshot 2024-07-12 at 15 27 52

packages/ag-charts-community/src/scene/spriteRenderer.ts
olegat: This comment seems outdated. Please review the `SpriteRenderer` class at the tip of the branch or provide more detail on the performance analysis. Note that the `SpriteRenderer.renderSprite` method just calls `CanvasRenderingContext2D.setTransform` directly without intermediate allocations or calculations.

packages/ag-charts-community/src/scene/node.ts
olegat: Unknown.

I do not know what `open-ended version of 'append'` means, but I believe that comments is trying to say that `append` with 1 node would be slower if the signature was `append(...nodes: Node[])`. If so, then yes the comment should still be relevant in theory.

packages/ag-charts-community/src/scene/spriteRenderer.ts
olegat: Done.

packages/ag-charts-community/src/chart/markerLabel.ts
olegat: Done.

packages/ag-charts-community/src/chart/markerLabel.ts
olegat: Done.

packages/ag-charts-community/benchmarks/__image_snapshots__/large-dataset-test-ts-large-dataset-benchmark-after-load-1-x-legend-toggle-1-snap.png
olegat: Marking as resolved.

We should fix the rendering bug with the test, but I think that this can be done in a follow up PR.

packages/ag-charts-website/src/content/docs/benchmarks/_examples/summary/data.ts
olegat: Done. I've run this command and committed the changes: `nx run all:benchmark`

packages/ag-charts-community/src/scene/node.ts
alantreadway: I've reviewed the uses and I don't believe any are performance sensitive - I'm going to remove the comment.

libraries/ag-charts-test/src/canvas/mock-canvas.ts
alantreadway: There was a bug in the sprite rendering code which meant that the sprite canvas wasn't consistently cleared, which may have been the root cause of this - now fixed.

PR #2002, opened 2024-07-09
AG-12088 Add initial text annotation

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12088

This is a spike of the basic text annotation.

There's some known issues with the alignment and dragging the annotation, and you can't edit them, but this change was already getting too large for a spike.

https://github.com/ag-grid/ag-charts/assets/3817697/5b8155fe-9204-4efe-b151-4829fd399523


packages/ag-charts-enterprise/src/features/annotations/annotations.ts
```diff
@@ -183,30 +202,63 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
...
+                ctx.toolbarManager.toggleButton('annotations', annotationType, { active: false });
+            }
+            this.toggleAnnotationOptionsButtons();
+        };
+
+        const onExitSingleClick = () => {
+            this.active = this.annotationData.length - 1;
+        };
+
+        return new AnnotationsStateMachine(
```
lsjroberts, line 234: The setup of this state machine is getting a bit unwieldy. I'll look into improving it when working on the editing of text annotations.

packages/ag-charts-enterprise/src/features/annotations/annotations.ts
```diff
@@ -862,6 +936,24 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
...
+    private onKeyDown(event: _ModuleSupport.KeyInteractionEvent<'keydown'>) {
+        const { annotationData, state } = this;
+
+        const context = this.getAnnotationContext();
+        if (!context) return;
+
+        const datum = annotationData.at(-1);
+        const { key } = event.sourceEvent;
+
+        if (key === 'Tab') {
```
lsjroberts, line 948: There are other events that should trigger the saving of the text, will also look into this when doing the editing feature as it may be better to move some of this into the state machine instead.

packages/ag-charts-types/src/chart/toolbarOptions.ts
```diff
@@ -74,7 +74,6 @@ export interface AgToolbarAnnotationsGroup extends AgToolbarGroup {
 }
 
 export interface AgToolbarAnnotationsButton extends AgToolbarButton {
-    section?: 'create' | 'tools';
```
lsjroberts, line 77: This is the expanding of the type I mentioned yesterday. It is now the default `section?: string` instead.

packages/ag-charts-enterprise/src/features/annotations/annotations.ts
alantreadway: Might be worth capturing this choice on the JIRA so David can expand on the conditions?

alantreadway (approved):
LGTM - nice work!

PR #2003, opened 2024-07-09
AG-12176 - check highlighted datum is in range when snapping to cross…

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-12176


alantreadway (approved):
LGTM

PR #2005, opened 2024-07-11
AG-12088 Refactor annotations state and add text editing

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12159

Apologies for the size of this PR.

- Refactors the hover, click and drag actions out of `annotations.ts` and into the `AnnotationsStateMachine`, simplifying a lot of the logic and removing much of the need for `state.is('idle')` checks etc.
- Adds text annotation editing. This is the feature that needed the above refactoring.

There are a few `TODO` comments remaining, largely around removing leaky design of the state machine. These are held up by refactoring out this behaviour from annotation toolbar buttons. But this PR is already big enough.


alantreadway (approved):
LGTM

PR #2010, opened 2024-07-12
AG-11645 - Allow tooltips to render beyond canvas extent.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-11645

Allow tooltips to render outside the canvas bounds, falling back to either the first parent with `overflow` , any enclosing shadow-DOM root, and finally window viewport bounds (pre-10.0.0 behaviour).


packages/ag-charts-community/src/chart/dom/domManager.ts
```diff
@@ -241,10 +255,49 @@ export class DOMManager extends BaseManager<Events['type'], Events> {
...
+        return windowBBox.intersection(containerBBox)?.toDOMRect() ?? NULL_DOMRECT;
+    }
+
+    private getRawOverlayClientRect() {
+        let element: HTMLElement | null = this.element;
+
+        // Try and find a parent which will clip rendering of children - if found we should restrict
+        // to that elements bounding box.
+        while (element != null) {
+            const overflow = element.computedStyleMap?.().get('overflow')?.toString();
```
jacobp100: We need to check `overflow-x` and `overflow-y` (but can skip checking `overflow`)

packages/ag-charts-community/src/chart/dom/domManager.ts
```diff
@@ -241,10 +255,49 @@ export class DOMManager extends BaseManager<Events['type'], Events> {
...
+
+            if (overflow === 'clip' || overflow === 'hidden' || overflow === 'scroll' || overflow === 'auto') {
```
jacobp100: It'll be easier to check if it's not `visible` - just in case they do add new ones. It's unlikely they'll find any more forms of `visible`, so it's a safe bet that the new ones will clip

jacobp100 (approved):
Only other comment is it'd be nice if we had a way to turn off this behaviour and use what we currently have - just in case there's some other cases like items moving around while a tooltip is being placed; or things like `contain: paint`, which does clip, but checking those sort of properties would probably be a bridge too far

packages/ag-charts-community/src/chart/dom/domManager.ts
alantreadway: Updated - thanks!

packages/ag-charts-community/src/chart/dom/domManager.ts
alantreadway: Done

alantreadway:
> Only other comment is it'd be nice if we had a way to turn off this behaviour and use what we currently have

I've added `tooltip.bounds = 'canvas'` as an (undocumented) escape hatch for now.

packages/ag-charts-community/src/chart/dom/domManager.ts
```diff
@@ -241,10 +255,51 @@ export class DOMManager extends BaseManager<Events['type'], Events> {
...
+        return windowBBox.intersection(containerBBox)?.toDOMRect() ?? NULL_DOMRECT;
+    }
+
+    private getRawOverlayClientRect() {
+        let element: HTMLElement | null = this.element;
+
+        // Try and find a parent which will clip rendering of children - if found we should restrict
+        // to that elements bounding box.
+        while (element != null) {
+            const styleMap = element.computedStyleMap?.();
```
jacobp100, line 282: This isn't supported by Safari <16.4 and Firefox

jacobp100: https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle

packages/ag-charts-community/src/chart/dom/domManager.ts
```diff
@@ -241,10 +255,51 @@ export class DOMManager extends BaseManager<Events['type'], Events> {
...
+
+            if (overflowX !== 'visible' || overflowY !== 'visible') {
+                return element.getBoundingClientRect();
+            }
+
+            element = element.parentElement;
+        }
+
+        // If in a shadow-DOM case, use the shadow-DOMs bounding-box, intersected with the window
+        // viewport.
```
jacobp100, line 294: What is the expected behaviour for a shadow dom? Should the tooltip be able to be outside?

packages/ag-charts-community/src/chart/interaction/tooltipManager.ts
```diff
@@ -67,12 +67,13 @@ export class TooltipManager {
         }
 
         const canvasRect = this.domManager.getBoundingClientRect();
+        const boundingRect = this.tooltip.bounds === 'extended' ? this.domManager.getOverlayClientRect() : canvasRect;
```
jacobp100, line 70: Minor - but maybe this could be `auto` - as in automatically determine the correct element. Potentially you'd later add support for specifying an explicit HTMLElement

packages/ag-charts-community/src/chart/dom/domManager.ts
alantreadway: Right now I would expect that the tooltip cannot escape the bounds of the shadow DOM.

jacobp100: Does this check need to be moved before the while loop then?

packages/ag-charts-community/src/chart/dom/domManager.ts
alantreadway: I've improved the fallback handling to switch back to the canvas bounds if we can't use this.

packages/ag-charts-community/src/chart/dom/domManager.ts
alantreadway: No, we still want to respect `overflow` on parent DOM nodes inside the shadow DOM - it's just if we reach the shadow-DOM root that we want to behave differently.

PR #2011, opened 2024-07-12
AG-12213 Update sector to not render computed outer radius is less th…

jacobp100:
…an inner

https://ag-grid.atlassian.net/browse/AG-12213


alantreadway (approved):
LGTM

PR #2017, opened 2024-07-15
AG-12169 - Add close line to financial charts range-area chart type

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-12169

### DIFF
--- a/packages/ag-charts-community/src/api/preset/priceVolumePreset.ts
+++ b/packages/ag-charts-community/src/api/preset/priceVolumePreset.ts
@@ -18,6 +18,7 @@ import type {
 } from 'ag-charts-types';
 
 import type { ChartTheme } from '../../chart/themes/chartTheme';
+import { DEFAULT_STROKES } from '../../chart/themes/defaultColors';
 import {
     PALETTE_DOWN_FILL,
     PALETTE_DOWN_STROKE,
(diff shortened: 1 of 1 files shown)


packages/ag-charts-community/src/api/preset/priceVolumePreset.ts
```diff
@@ -313,6 +314,12 @@ function createPriceSeries(
                     fill: fill ?? PALETTE_DOWN_STROKE,
                     stroke: stoke ?? PALETTE_DOWN_STROKE,
                 } satisfies AgRangeAreaSeriesOptions,
+                {
+                    type: 'line',
+                    ...singleKeys,
+                    stroke: fromTheme(theme, (t) => t.overrides?.line?.series?.stroke) ?? DEFAULT_STROKES.GRAY,
```
alantreadway, line 327: Could we use `PALETTE_NEUTRAL_STROKE` here for consistency? I'd rather we didn't hardcode these colours.

manapeirov: Ok but that's blue, jira says it should be gray is that ok?

alantreadway: Discussed offline, let's roll with the original proposal

PR #2022, opened 2024-07-16
Refactor series-area handling out of Chart.

alantreadway:
Move all series-area related interaction handling out into a new `SeriesAreaManager`.

The intent in this change is just to lift-and-shift, not rework (or at least to minimise this).

### DIFF
--- a/packages/ag-charts-community/src/chart/flowProportionChart.ts
+++ b/packages/ag-charts-community/src/chart/flowProportionChart.ts
@@ -12,6 +12,10 @@ export class FlowProportionChart extends Chart {
     static readonly className = 'FlowProportionChart';
     static readonly type = 'flow-proportion' as const;
 
+    override getChartType() {
+        return 'flow-proportion' as const;
+    }
+
     override async updateData() {
         await super.updateData();
 
@@ -42,7 +46,6 @@ export class FlowProportionChart extends Chart {
 
         this.seriesRect = shrinkRect;
         this.animationRect = shrinkRect;
-        this.hoverRect = shrinkRect;
 
         const seriesVisible = this.series.some((s) => s.visible);
         seriesRoot.visible = seriesVisible;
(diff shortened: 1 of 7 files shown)


packages/ag-charts-community/src/chart/flowProportionChart.ts
```diff
@@ -12,6 +12,10 @@ export class FlowProportionChart extends Chart {
     static readonly className = 'FlowProportionChart';
     static readonly type = 'flow-proportion' as const;
 
+    override getChartType() {
+        return 'flow-proportion' as const;
```
jacobp100, line 16: Could this not on the chart base-class as `return this.constructor.type`?

packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
```diff
@@ -0,0 +1,630 @@
...
+    }
+
+    private handleSeriesFocus(otherIndexDelta: number, datumIndexDelta: number) {
+        if (this.chartType === 'hierarchy') {
+            this.handleHierarchySeriesFocus(otherIndexDelta, datumIndexDelta);
+            return;
+        }
+        const { focus, seriesRect, series } = this;
+        const visibleSeries = series.filter((s) => s.visible);
+        if (visibleSeries.length === 0) return;
```
jacobp100, line 534:
```suggestion
        const hasVisibleSeries = series.some((s) => s.visible);
        if (!hasVisibleSeries) return;
```

jacobp100 (approved):

packages/ag-charts-community/src/chart/flowProportionChart.ts
alantreadway: Yes, but let's do it in a follow-up as this is a big change.

alantreadway:
Superceeded by https://github.com/ag-grid/ag-charts/pull/2025

PR #2028, opened 2024-07-16
Refactor out common textual annotation behaviour

lsjroberts:
Also adds stubs for callout, comment and note annotation types.


alantreadway (approved):
LGTM

PR #2031, opened 2024-07-17
AG-11998 Area series fixes

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11998

### DIFF
--- a/packages/ag-charts-community/src/chart/data/dataModel.ts
+++ b/packages/ag-charts-community/src/chart/data/dataModel.ts
@@ -916,7 +916,7 @@ export class DataModel<
                 initDataDomain();
             }
 
-            if (valueInDatum && !(def.validation?.(value, datum) ?? true)) {
+            if (valueInDatum && def.validation?.(value, datum) === false) {
                 if ('invalidValue' in def) {
                     value = def.invalidValue;
                 } else {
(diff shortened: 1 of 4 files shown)


packages/ag-charts-community/src/util/type-guards.ts
```diff
@@ -49,7 +49,7 @@ export function isNumber(value: unknown): value is number {
 }
 
 export function isFiniteNumber(value: unknown): value is number {
-    return isNumber(value) && Number.isFinite(value);
+    return Number.isFinite(value);
```
jacobp100, line 52: Unrelated but it saves an additional check (`Number.isFinite` does not cast to a number, unlike `window.isFinite`)

packages/ag-charts-community/src/chart/data/dataModel.ts
```diff
@@ -916,7 +916,7 @@ export class DataModel<
                 initDataDomain();
             }
 
-            if (valueInDatum && !(def.validation?.(value, datum) ?? true)) {
+            if (valueInDatum && def.validation?.(value, datum) === false) {
```
jacobp100, line 919: Unrelated too but should be marginally faster

alantreadway (approved):
LGTM

PR #2032, opened 2024-07-17
AG-12294 Update icons

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12294
https://ag-grid.atlassian.net/browse/AG-12144


jacobp100 (approved):

packages/ag-charts-types/src/chart/toolbarOptions.ts
```diff
@@ -45,28 +45,65 @@ export interface AgToolbarButton {
     value: any;
 }
 
-export type AgIconName = IconNameAnnotation | IconNameZoom;
```
alantreadway, line 48: These look like breaking changes - were these types used in publicly documented options?

alantreadway: Discussed offline, there are two cases that we need to deprecate as they are not in the new icon set:
- `zoom-in-alt`
- `zoom-out-alt`

alantreadway (approved):
LGTM

PR #2038, opened 2024-07-19
AG-12039 Button focus style fixes

jacobp100:
https://ag-grid.atlassian.net/browse/AG-12039


markdurrant (approved):
LGTM 👍 

Would like to wait for Bernie to review before merging. 

Use of `:is` is nice. Not see it used in anger before.

alantreadway (approved):
LGTM, let's wait for feedback from @BernieSumption before merge.

BernieSumption (approved):
LGTM, with one bit of feedback.

I know this is preserved from the previous state not added here so consider this an FYI, but in general I try to avoid things like `.ag-charts-toolbar__button:nth-child(1 of :not(.ag-charts-toolbar__button--hidden-toggled))` and instead add styles in code like `ag-charts-toolbar__button--first` that are easier to target. The reason is that it's also easier for customers to understand and override if necessary. And for future grid devs to understand as well :)

jacobp100:
@BernieSumption we can revisit this post-release. Only concern is doing this now will introduce bugs, because there's a lot of manual dom manipulation

PR #2040, opened 2024-07-19
AG-12025 Show active state for toolbar zoom

jacobp100:
https://ag-grid.atlassian.net/browse/AG-12025

### DIFF
--- a/packages/ag-charts-community/src/chart/interaction/toolbarManager.ts
+++ b/packages/ag-charts-community/src/chart/interaction/toolbarManager.ts
@@ -50,7 +50,7 @@ export interface ToolbarButtonPressedEvent<T = any> extends Event<'button-presse
     rect: BBox;
 }
 
-export interface ToolbarButtonToggledEvent extends Event<'button-toggled'> {
+export interface ToolbarButtonToggledEvent<_T = any> extends Event<'button-toggled'> {
     id: string;
     active: boolean;
     enabled: boolean;
--- a/packages/ag-charts-community/src/chart/interaction/toolbarManager.ts
+++ b/packages/ag-charts-community/src/chart/interaction/toolbarManager.ts
@@ -34,7 +34,8 @@ interface Event<T extends EventTypes> {
 
 export interface ToolbarGroupToggledEvent extends Event<'group-toggled'> {
     caller: string;
-    visible: boolean;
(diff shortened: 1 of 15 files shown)


packages/ag-charts-enterprise/src/features/zoom/zoomToolbar.ts
```diff
@@ -78,6 +89,30 @@ export class ZoomToolbar {
...
+            const { selectedZoom } = this;
+            const { x } = e;
+
+            this.toolbarManager.toggleGroup('zoom-toolbar', 'ranges', { active: false });
+
+            if (
+                selectedZoom != null &&
+                x != null &&
+                Math.abs(selectedZoom.min - x.min) < 1e-9 &&
+                Math.abs(selectedZoom.max - x.max) < 1e-9
```
lsjroberts: nit: Could we change this to `1e-10` to match `isZoomEqual()` in `zoomUtils.ts`?

I've had bugs with zooms not matching and refining the granularity of the floating point of zoom to be the same everywhere.

packages/ag-charts-community/src/chart/interaction/toolbarManager.ts
```diff
@@ -44,19 +45,20 @@ export interface ToolbarFloatingAnchorChangedEvent extends Event<'floating-ancho
...
 
 export interface ToolbarButtonPressedEvent<T = any> extends Event<'button-pressed'> {
+    id: string;
     value: T;
     rect: BBox;
 }
 
-export interface ToolbarButtonToggledEvent<T = any> extends Event<'button-toggled'> {
-    value: T;
+export interface ToolbarButtonToggledEvent extends Event<'button-toggled'> {
```
lsjroberts: Can we keep this `value: T` typing? It is used in annotations https://github.com/ag-grid/ag-charts/blob/latest/packages/ag-charts-enterprise/src/features/annotations/annotations.ts#L358

lsjroberts (approved):
Nice! LGTM

alantreadway (approved):

packages/ag-charts-community/src/chart/interaction/toolbarManager.ts
lsjroberts: ignore this, sorry

PR #2045, opened 2024-07-19
AG-11990 - Only pick non-visible datums for Financial Charts cases.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-11990

Follow-up to only allow special category main axis node pick logic to apply for Financial Charts cases.


lsjroberts (approved):
LGTM

PR #2048, opened 2024-07-19
AG-12333 Update font styles

jacobp100:
https://ag-grid.atlassian.net/browse/AG-12333


alantreadway (approved):
LGTM, looks like this is just applying previous design feedback for the most part.

packages/ag-charts-community/src/chart/dom/theme.css
```diff
@@ -58,6 +61,7 @@
         var(--ag-charts-toolbar-foreground-color) 6%
     );
     --ag-charts-toolbar-gap: var(--ag-charts-size);
+    --ag-charts-toolbar-font-family: var(--ag-charts-font-family);
```
markdurrant: I would remove this as a separate variable, and have a single place to control font-family across all charts. 

(I can't see a use case for having a different font-family for toolbars to the rest of charts).

packages/ag-charts-community/src/chart/dom/theme.css
```diff
@@ -70,6 +74,7 @@
     --ag-charts-popover-focus-color: var(--ag-charts-toolbar-focus-color);
     --ag-charts-popover-hover-color: var(--ag-charts-toolbar-hover-color);
     --ag-charts-popover-active-background-color: var(--ag-charts-toolbar-active-background-color);
+    --ag-charts-popover-font-family: var(--ag-charts-toolbar-font-family);
```
markdurrant: I would remove this as a separate variable, and have a single place to control font-family across all charts. 

(I can't see a use case for having a different font-family for toolbars to the rest of charts).

markdurrant (changes requested):
I would remove the specific variables for setting a font family for toolbar or popover. Can't see a use case for this. 

Otherwise everything looking good.

packages/ag-charts-community/src/chart/dom/theme.css
jacobp100: Updated

markdurrant (approved):
LGTM 👍

PR #2051, opened 2024-07-22
AG-12350 - Restrict series nodeClick handling to series-area.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-12350

Fixes series `nodeClick` and `nodeDoubleClick` events firing from clicks outside the series area.


packages/ag-charts-community/src/chart/chart.test.ts
alantreadway, line 1: Fixes here are to tests that were clicking outside the series area and expecting series nodeClicks to be triggered - the change means that offset clicks are now offset towards the center of the series area to avoid this.

lsjroberts (approved):
LGTM

PR #2055, opened 2024-07-23
AG-12302 - Update toolbar & popover active/focus/hover state styles

markdurrant:
Update active/hover/focus styles for toolbar buttons and menu rows. 

- Increase contrast between foreground and background for active states 
- Refactor popover active states to use correct variables
- Make toolbar active colour the same as charts active colour


packages/ag-charts-community/src/chart/toolbar/toolbarStyles.css
```diff
@@ -186,7 +187,9 @@
 
 .ag-charts-toolbar__button.ag-charts-toolbar__button--active {
     background: var(--ag-charts-toolbar-active-background-color);
+    border: 1px solid var(--ag-charts-toolbar-active-color);
```
lsjroberts: `border-color: var(--ag-charts-toolbar-active-color)` in case we change the other border styles

lsjroberts (approved):
LGTM, just one minor thing

PR #2056, opened 2024-07-23
AG-12186 - Improve status bar chart type handling.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-12186

Corrects the status bar presentation for line/step-line/HLC/high-low financial chart types.

Additionally fixes the color of status bar text to follow the volume series logic for these series types (previously these were always shown in negative colours).


lsjroberts (approved):
LGTM

PR #2059, opened 2024-07-23
AG-12267 - Remove axis dependency on series datum rendering.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-12267


jacobp100 (approved):
Will this draw datums outside of the zoom range?

alantreadway:
> Will this draw datums outside of the zoom range?

As far as I can tell this has no impact in that respect, minRects was being used to work out how to space the ticks, nothing more.

PR #2061, opened 2024-07-24
AG-12302 - Tweak toolbar hover background colour for darkmode

markdurrant:
Tweaks darkmode toolbar hover colours for better visibility. 

Refactor toolbar active background colour to reference the hover colour so we're not repeating ourselves.


lsjroberts (approved):
LGTM

PR #2069, opened 2024-07-25
AG-11926 Update border for fill palette in toolbar

jacobp100:
https://ag-grid.atlassian.net/browse/AG-11926


markdurrant (approved):
Styling stuff LGTM 👍 

Using the foreground colour instead of light/dark values is a nice approach.

PR #2073, opened 2024-07-25
AG-12330 Add toolbar font size picker

jacobp100:
https://ag-grid.atlassian.net/browse/AG-12330


packages/ag-charts-enterprise/src/features/annotations/annotationsStateMachine.ts
```diff
@@ -229,6 +230,22 @@ export class AnnotationsStateMachine extends StateMachine<States, AnnotationType
...
 
+                fontSize: {
+                    guard: () => this.active != null,
+                    target: States.Idle,
+                    action: (fontSize: number) => {
+                        const datum = ctx.datum(this.active!);
+                        if (!datum) return;
+
+                        if (isTextType(datum)) {
+                            // @ts-ignore TS bug
```
alantreadway: Nit: Could we use a `@ts-expect-error` instead, so if the issue is fixed we'll get a compile failure here?

alantreadway (commented):
One nit

packages/ag-charts-enterprise/src/features/annotations/annotationsStateMachine.ts
jacobp100: It's annoying because VS code has an upgraded version, so you'll see an error there if I do that - but it won't show up on `blt` etc.

alantreadway: `blt` is the source of truth, please adjust your VSCode TS version to match.

jacobp100: Updated

alantreadway (approved):
LGTM

PR #2079, opened 2024-07-26
AG-12428 Documentation for chart sizing

jacobp100:
https://ag-grid.atlassian.net/browse/AG-12428

> The chart will auto-size to fill the size of the container element. The class and style attributes are preserved on the container element, each of which can be adjust the container's size. Additionally, the container can be placed as a child of a grid layout, and the grid will adjust the container's size.
> 
> The chart has a default minimum width and height of 300px, which can in some circumstances cause the container to grow in size. The minHeight and/or minWidth options can be used to change these values, or set them to zero.
> 
> If auto-sizing is not required, the width and/or height options will render the chart at a fixed size, ignoring minWidth and minHeight.
> 
> Below is an example of a chart with no CSS applied, followed by two different styling methods to change its size.

(Wording is slightly tweaked for FW examples)


packages/ag-charts-website/src/content/docs/layout/index.mdoc
```diff
@@ -7,26 +7,41 @@ This section explains how the chart and its components are sized and laid out wi
 
 ## Chart Size
 
-The chart will auto-size by default, taking the size of the `container` element and auto-sizing the chart dynamically.
-Use the `width` and/or `height` options if a fixed size is required.
+{% if isFramework("javascript") %}
+The chart will auto-size to fill the size of the `container` element. The `class` and `style` attributes are preserved on the `container` element, each of which can be adjust the container's size. Additionally, the container can be placed as a child of a `grid` layout, and the grid will adjust the container's size.
```
alantreadway: I feel that:
> The `class` and `style` attributes are preserved on the `container` element, each of which can be adjust the container's size.

...is misleading for JS/TS cases as it implies we do adjust the DOM of the `container`, which is not the case.

alantreadway (approved):
LGTM

PR #2082, opened 2024-07-26
Ag 12160/callout text annotation

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-12160


packages/ag-charts-enterprise/src/features/annotations/annotationsConfig.ts
```diff
@@ -139,7 +139,5 @@ export function colorDatum(datum: AnnotationProperties, color: string) {
         datum.fill = color;
     } else if ('color' in datum) {
         datum.color = color;
-    }
-
-    if ('stroke' in datum) datum.stroke = color;
+    } else if ('stroke' in datum) datum.stroke = color;
```
lsjroberts: This will now not update the stroke of the callout, is that intentional? Seems like it should update since it uses the same colour as the text by default?

packages/ag-charts-enterprise/src/features/annotations/callout/calloutScene.ts
```diff
@@ -1,23 +1,360 @@
...
+        const right = x + width;
+
+        const placement = this.calculateCalloutPlacement({ x: tailX, y: tailY }, bodyBounds);
+        const cornerRadius = Math.min(12, Math.min(width, height) / 2);
+
+        const { path } = this.shape;
+        path.clear();
+
+        path.moveTo(x, top + cornerRadius);
+        this.drawPath(
```
lsjroberts: nit: I think a cleaner approach here would be to create the basic points first as an array then iterate through them, calling small draw functions for each side and corner. Then draw the "callout" version of the side or corner based on the index and placement.

Something like:

```ts
const points = [
    { x0: ... },
    // ...
]

for (const [index, point] of points.entries()) {
    if (index === 0 && placement === 'topLeft') {
        this.drawCalloutCorner(path, { ...point, cx: tailX, cy: tailY, }, cornerRadius);
    } else if (index === 1 && placement === 'top') {
        this.drawCalloutSide(path, point, cornerRadius);
    // else etc...
    } else if (index % 2 === 0) {
        this.drawCorner(path, point, cornerRadius);
    } else {
        this.drawSide(path, point);
    }
}
```

Thoughts?

There may be a better final approach but in principle:
- iterating rather than sequentially calling the same function multiple times
- having small functions for each case rather than a long switch statement

lsjroberts: Agreed to come back to this later

lsjroberts (approved):
LGTM!

PR #2084, opened 2024-07-29
AG-12162 Add note annotation

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12162 

Also:

- Adds `_Scene.SvgPath` node e.g. `new SvgPath('M22 1.83333C22 0.820811...')`
- Fixes which color icons show for each annotation


packages/ag-charts-types/src/chart/toolbarOptions.ts
```diff
@@ -60,7 +60,6 @@ export type AgIconName =
     | 'hollow-candlestick-series'
     | 'horizontal-line-drawing'
     | 'line-color'
-    | 'fill-color'
```
lsjroberts, line 63: Duplicate of one above

alantreadway (approved):
LGTM

PR #2085, opened 2024-07-29
AG-12277: Hide deprecated union members from options reference

iMoses:
https://ag-grid.atlassian.net/browse/AG-12277


alantreadway (approved):
LGTM

PR #2086, opened 2024-07-29
Ag 12413 range button styles

markdurrant:
- Reduce font size for toolbar buttons 
- Increase horizontal padding for range toolbar buttons


packages/ag-charts-community/src/chart/toolbar/toolbarStyles.css
```diff
@@ -135,6 +135,10 @@
     white-space: nowrap;
 }
 
+.ag-charts-toolbar__button[data-toolbar-group='ranges'] {
```
jacobp100, line 138: Seems like a good way to target this. What do you think of moving this to the bottom and have a section like `specific overrides`, so it's clear where we're doing this sort of stuff

jacobp100 (approved):

alantreadway:
Please update the e2e snapshots locally by using:
```
nx test:e2e ag-charts-website -c update
```

(Linux/CI snapshots already updated)

PR #2090, opened 2024-07-29
AG-12428 CSS docs

jacobp100:
https://ag-grid.atlassian.net/browse/AG-12428

### DIFF
--- a/packages/ag-charts-website/src/content/docs/layout/_examples/chart-class/styles.css
+++ b/packages/ag-charts-website/src/content/docs/layout/_examples/chart-class/styles.css
@@ -1,0 +1,3 @@
+.chart {
+    outline: 1px solid red;
+}
--- a/packages/ag-charts-website/src/content/docs/layout/index.mdoc
+++ b/packages/ag-charts-website/src/content/docs/layout/index.mdoc
@@ -7,41 +7,131 @@ This section explains how the chart and its components are sized and laid out wi
 
 ## Chart Size
 
-{% if isFramework("javascript") %}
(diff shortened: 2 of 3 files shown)


packages/ag-charts-website/src/content/docs/layout/_examples/chart-class/styles.css
```diff
@@ -0,0 +1,3 @@
+.chart {
```
markdurrant, line 1: I don't think this class is actually referenced by the example. When I run the example locally I don't see this file in the files list, and no red border.

markdurrant (changes requested):
Generally looking good, I think the page is much clearer without the tabs. 

Only issue is the CSS class example not picking up the new `styles.css` file.

packages/ag-charts-website/src/content/docs/layout/_examples/chart-class/styles.css
jacobp100: I think you have to save this file to force it to rebuild the example

jacobp100: Definitely worked locally for me - maybe it's easier to just revisit if it doesn't show up on staging

markdurrant: Sorry my bad. Wasn't aware about the save thing. Can confirm all works locally for me. Everything looking great 👍

markdurrant (approved):
LGTM 👍

alantreadway (approved):

PR #2095, opened 2024-07-30
AG-12126 Fix annotations popover not closing

jacobp100:
https://ag-grid.atlassian.net/browse/AG-12126


jacobp100:
Will be fixed in https://github.com/ag-grid/ag-charts/pull/2096

PR #2099, opened 2024-07-30
AG-12009: FC axis labels clipped when range toolbar hidden

iMoses:
https://ag-grid.atlassian.net/browse/AG-12009


packages/ag-charts-community/src/chart/themes/financialLight.ts
```diff
@@ -49,7 +49,7 @@ export class FinancialLight extends ChartTheme {
         params.set(DEFAULT_ANNOTATION_BACKGROUND_FILL, FINANCIAL_LIGHT_FILLS.BLUE);
         params.set(DEFAULT_AXIS_GRID_COLOUR, '#F2F3F3');
 
-        params.set(DEFAULT_PADDING, 0);
+        params.set(DEFAULT_PADDING, 5);
```
alantreadway: Is it possible to only add the padding on the side that is affected?

(We can also just set this in `priceVoluePreset.ts` directly, to make life easier)

alantreadway (approved):
LGTM

alantreadway (approved):
LGTM

PR #2101, opened 2024-07-30
AG-12381 Fix pressing esc key in textual annotations

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12381


alantreadway (approved):
LGTM

PR #2102, opened 2024-07-30
AG-12330 Text size picker styling

jacobp100:
https://ag-grid.atlassian.net/browse/AG-12330

### DIFF
--- a/packages/ag-charts-enterprise/src/features/annotations/annotations.ts
+++ b/packages/ag-charts-enterprise/src/features/annotations/annotations.ts
@@ -530,7 +530,7 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
                     items: TEXT_SIZE_ITEMS,
                     onPress: this.onTextSizePopoverPress.bind(this),
                     onClose: this.onTextSizePopoverClose.bind(this),
-                    minWidth: 60,
+                    class: 'annotations__text-size',
                 });
                 break;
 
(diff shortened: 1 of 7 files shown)


packages/ag-charts-enterprise/src/features/annotations/annotations.ts
```diff
@@ -568,7 +568,9 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
         const anchor = Vec2.add(rect, Vec2.from(0, rect.height + 4));
         const fallback = { y: rect.y - 4 };
         this.colorPicker.setAnchor(anchor, fallback);
-        this.textSizePopover.setAnchor(anchor);
+
+        const textAnchor = { x: rect.x + 34 - 1, y: rect.y + rect.height - 1 };
```
alantreadway, line 572: This `34` seems to be shared with the `toobar.css`, is there a way we can ensure the consistency of this JS code vs. the CSS definition?

jacobp100: It's a hack for sure. There's not really a good sharing mechanism between CSS and JS that'll work for what we want. It not the only issue too - if the toolbar items change order, this won't reflect that either. I think it will be worth revisiting this once we've implemented more UI throughout charts and see if we can tidy up this implementation, or we find a better design that doesn't have these issues

markdurrant (approved):
Generally CSS LGTM. I've also tested locally and looks good visually too. 

One minor stylistic nitpick. Action or not at your discretion.

packages/ag-charts-enterprise/src/features/annotations/toolbar.css
```diff
@@ -0,0 +1,7 @@
+.annotations__text-size {
+    width: 34px;
+    min-width: 0;
+    border-top-left-radius: 0;
+    border-top-right-radius: 0;
+    --item-padding: 4px 8px;
+}
```
markdurrant, line 7: Generally I think it's good to have any CSS variables defined at the top of the ruleset, with a line break below. 

To me it makes it much easier to parse what variables are declared per selector. For example...


```
.annotations__text-size {
     --item-padding: 4px 8px;

     width: 34px;
     min-width: 0;
     border-top-left-radius: 0;
     border-top-right-radius: 0;
 }
 ```

jacobp100: I don’t mind going this way, but we should do it in one go because right now they’re always at the bottom

alantreadway (approved):
LGTM

PR #2104, opened 2024-07-30
AG-12383 Fix text canvas wrapping after overflow chart area

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12383


alantreadway (approved):
LGTM

PR #2108, opened 2024-07-30
AG-12380 Fix clear all annotations and toolbar while drawing

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12380


alantreadway (approved):
LGTM

PR #2110, opened 2024-07-30
AG-12377 Fix deleting text annotation while editing

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12377


alantreadway (approved):
LGTM

PR #2113, opened 2024-07-30
AG-12341 Remove text align from annotations memento

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12341

The other approach would be to remove the `textAlign` property from here https://github.com/ag-grid/ag-charts/blob/latest/packages/ag-charts-enterprise/src/features/annotations/annotationProperties.ts#L139-L140 - but then you have to handle a lot of exceptions in various places we use `datum.set()`.

This seemed cleaner for the actual end goal.


alantreadway (approved):
LGTM

PR #2119, opened 2024-07-31
AG-12301 - Rework Annotation Updates + Series Highlighting

alantreadway:
https://ag-grid.atlassian.net/browse/AG-12301

Reverts https://github.com/ag-grid/ag-charts/commit/676d15f2ee4b10626f941fe6664384cfc2ebdf43 which introduced a bug with infinite looping series highlight processing on hover.

Reworks the Annotation update cycle to avoid needing a full chart (re)layout, and consequently no longer conflicting with the highlight processing.


lsjroberts (approved):
LGTM

PR #2127, opened 2024-07-31
AG-12404 - Fix options update resetting annotations

alantreadway:
https://ag-grid.atlassian.net/browse/AG-12404

Ignores `initialState` on options update, if there is no change in the `initialState` configuration.


lsjroberts (approved):
LGTM. This is something we'll need to consider when doing the zoom initial state in https://github.com/ag-grid/ag-charts/pull/1887 - i.e. don't update the annotations when the zoom initial state changes perhaps

PR #2128, opened 2024-07-31
AG-12400 Fix changing color of text while creating

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12400

This will fail the code duplication for same reason as the last textual state change. To be addressed later.


alantreadway (approved):
LGTM

PR #2131, opened 2024-07-31
AG-12358 Fix annotation toolbar flashing when selecting diff types

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12358


lsjroberts:
This appears to cause an issue with note, one second

alantreadway (approved):
LGTM

PR #2135, opened 2024-08-01
AG-12366 Fix deleting text then clicking to deselect hide text input

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12366

**Targeting the https://github.com/ag-grid/ag-charts/pull/2134 branch, as it uses a change from there.**


alantreadway (approved):
LGTM

PR #2139, opened 2024-08-01
AG-12468 - Validate scales before animation to/from calculation.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-12468


packages/ag-charts-community/src/chart/series/cartesian/scaling.ts
```diff
@@ -45,3 +45,13 @@ export function areScalingEqual(a: Scaling | undefined, b: Scaling | undefined):
     }
     return false;
 }
+
+export function isScaleValid(scale?: Scaling) {
+    if (scale == null) return false;
+    if (!scale.range.every((v) => isFinite(v))) return false;
```
jacobp100: Minor - use `Number.isFinite` as `window.isFinite` will try to convert strings and such to numbers first, and returns `true` for booleans

jacobp100:
```suggestion
    if (!scale.range.every((v) => Number.isFinite(v))) return false;
```

packages/ag-charts-community/src/chart/series/cartesian/scaling.ts
```diff
@@ -45,3 +45,13 @@ export function areScalingEqual(a: Scaling | undefined, b: Scaling | undefined):
...
+
+    if (scale.type === 'category') {
+        return scale.domain.every((v) => v != null);
+    }
+    return scale.domain.every((v: any) => (typeof v === 'number' && isFinite(v)) || v instanceof Date);
```
jacobp100:
```suggestion
    return scale.domain.every((v: any) => (Number.isFinite(v)) || v instanceof Date);
```

Do we need to check the date is valid too?

```suggestion
    return scale.domain.every((v: any) => Number.isFinite(v) || (v instanceof Date && Number.isFinite(v.getTime()));
```

jacobp100 (approved):
Is `scale` a `Scale` object here? Would we be able to move this check onto the class?

alantreadway:
> Is `scale` a `Scale` object here? Would we be able to move this check onto the class?

No, it's a minimal snapshot of the scale state that lives with the nodeData generated by `createNodeData()`.

PR #2147, opened 2024-08-01
AG-12471 Fix escape key handling on annotations

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12471

This shifts responsibility for deleting while creating fully into the child state machines, allowing the removal of the leaky `state.is('some state')` check.


alantreadway (approved):
LGTM

PR #2148, opened 2024-08-01
AG-12301 - Fix sticky highlight + console warnings.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-12301


packages/ag-charts-community/src/chart/interaction/highlightManager.ts
```diff
@@ -34,6 +35,7 @@ export class HighlightManager extends BaseManager<'highlight-change', HighlightC
                 type: 'highlight-change',
                 currentHighlight: this.activeHighlight,
                 previousHighlight,
+                callerId,
```
jacobp100, line 38: Is this used?

jacobp100 (approved):

packages/ag-charts-community/src/chart/interaction/highlightManager.ts
alantreadway: Only for debugging.

PR #2149, opened 2024-08-01
AG-12400 Fix change font size when creating text annotations

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12400


alantreadway (approved):
LGTM

PR #2150, opened 2024-08-01
AG-12417 Add default comment padding which changes with font size

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12417


alantreadway (approved):
LGTM

PR #2151, opened 2024-08-01
AG-12369, AG-12466 Fix up/down keyboard navigation for Financial Chart toolbar

olegat:
This PR puts the `seriesTypes` and `annotations` toolbar-groups into the same aria-toolbar so that up/down works as intended.


alantreadway (approved):
LGTM

PR #2153, opened 2024-08-02
AG-12369 Fix keynav using initMenuKeyNav and correct aria roles

olegat:
https://ag-grid.atlassian.net/browse/AG-12369

### DIFF
--- a/packages/ag-charts-enterprise/src/features/popover/popover.ts
+++ b/packages/ag-charts-enterprise/src/features/popover/popover.ts
@@ -32,6 +32,7 @@ export class Popover extends _ModuleSupport.BaseModuleInstance implements _Modul
 
     show<Value = any>(opts: {
         items: MenuItem<Value>[];
+        sourceEvent: Event;
         value?: Value;
         onPress?: (item: MenuItem<Value>) => void;
         onClose: () => void;
--- a/packages/ag-charts-enterprise/src/features/popover/popover.ts
+++ b/packages/ag-charts-enterprise/src/features/popover/popover.ts
@@ -1,6 +1,6 @@
 import { type AgIconName, _ModuleSupport, _Scene, _Util } from 'ag-charts-community';
 
-const { createElement } = _ModuleSupport;
+const { createElement, initMenuKeyNav } = _ModuleSupport;
 
 const canvasOverlay = 'canvas-overlay';
 
@@ -13,7 +13,6 @@ export interface MenuItem<Value = any> {
 export class Popover extends _ModuleSupport.BaseModuleInstance implements _ModuleSupport.ModuleInstance {
     private readonly element: HTMLElement;
     private anchor?: { x: number; y: number } = undefined;
-    private removeGlobalEventListeners?: () => void = undefined;
 
     private readonly moduleId: string;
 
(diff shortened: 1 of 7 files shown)


packages/ag-charts-enterprise/src/features/popover/popover.ts
```diff
@@ -101,63 +98,15 @@ export class Popover extends _ModuleSupport.BaseModuleInstance implements _Modul
...
-            switch (e.key) {
-                case 'ArrowUp':
-                    focusIndexDelta(-1);
-                    break;
-                case 'ArrowDown':
-                    focusIndexDelta(1);
-                    break;
-                case 'Escape':
-                case 'Enter':
-                    opts.onClose();
```
alantreadway, line 132: Is `Escape` handling still maintained? I see there is an `onEscape` option to `initMenuKeyNav()` that might need to be used?

packages/ag-charts-enterprise/src/features/popover/popover.ts
```diff
@@ -101,63 +98,15 @@ export class Popover extends _ModuleSupport.BaseModuleInstance implements _Modul
...
-            if (popover.contains(e.target as any)) return;
-            popover.focus();
-        };
-        window.addEventListener('mousemove', windowMouseMoveEvent);
-
-        const windowMouseMoveDown = (e: MouseEvent) => {
-            if (e.button !== 0 || popover.contains(e.target as any)) return;
-
-            opts.onClose();
-        };
```
alantreadway, line 152: Is the behaviour of 'clicks outside the popup close the popup' still maintained?

olegat: It's working with the Trend Line and Text Annotations menus, but not with the Chart Type menu. I suspect that the Chart Type menu has some other logic that's clashing with.

packages/ag-charts-enterprise/src/features/popover/popover.ts
olegat: We could easily add an `onEscape: () => this.hide` option, but this wouldn't be a complete solution. The `contextMenu.ts` module keeps a reference to HTMLElement that triggered the popup:

https://github.com/ag-grid/ag-charts/blob/c811b8f9c052098c4dec9b94d76ed508a7583a3f/packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts#L240

`popover.ts` will need to do something similar. I was going to move the `ContextMenu.lastFocus` feature to be part of the `initMenuKeyNav` method, but on second thoughts I could do this in the PR too.

packages/ag-charts-enterprise/src/features/popover/popover.ts
olegat: Fixed: Clicks outside the popup will close the popup.

Notes:
- Right-clicks will also close the popup (previously, only left-clicks closed the pop-up).
- This change has also been to context menus.

packages/ag-charts-enterprise/src/features/popover/popover.ts
olegat: Fixed. The `ContextMenu.lastFocus` logic is now part of the `initMenuKeyNav` function and therefore is used by both context menu and financial charts popovers.

alantreadway (approved):
LGTM

PR #2156, opened 2024-08-02
AG-12449 - Add better opacity support for swatches / annotations.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-12449

### DIFF
--- a/packages/ag-charts-enterprise/src/features/annotations/annotations.ts
+++ b/packages/ag-charts-enterprise/src/features/annotations/annotations.ts
@@ -627,8 +627,8 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
         opacity?: number
     ) {
         if (color != null && opacity != null) {
-            const [r, g, b] = _Util.Color.parseHex(color) ?? [0, 0, 0];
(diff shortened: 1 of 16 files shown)


packages/ag-charts-enterprise/src/features/annotations/annotations.ts
```diff
@@ -611,15 +611,25 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
...
     }
 
-    private updateToolbarColorPickerFill(colorPickerType: AnnotationOptionsColorPickerType, color?: string) {
+    private updateToolbarColorPickerFill(
+        colorPickerType: AnnotationOptionsColorPickerType,
+        color?: string,
+        opacity?: number
+    ) {
+        if (color != null && opacity != null) {
+            const [r, g, b] = _Util.Color.parseHex(color) ?? [0, 0, 0];
```
manapeirov: Are we sure it will always be hex?

manapeirov (approved):
LGTM

packages/ag-charts-enterprise/src/features/annotations/annotations.ts
alantreadway: Good catch, switched to `Color.parseString()`

PR #2159, opened 2024-08-02
Cleanup layout page

AG-DavidG:


packages/ag-charts-website/src/content/docs/layout/index.mdoc
```diff
@@ -66,34 +60,54 @@ The `className` and `style` props can be used to style the chart.
...
-<ag-charts [options]="options" class="chart" style="width: 400px; height: 400px"></ag-charts>
 ```
 
+In this configuration:
+
+-   The element containing the chart is sized using the `style` attribute.
+-   It could also be sized by styling the class specified in the `class` attribute.
+
+{% note %}
+Due to the way that Angular works, the chart container should be styled as a `block` element in these scenarios.
```
alantreadway: Maybe:
```suggestion
As the `<ag-chars>`container element is a custom HTML element without implicit browser styling, it should be styled as a `block` element by either `style` or `class` attributes.
```

packages/ag-charts-website/src/content/docs/layout/index.mdoc
```diff
@@ -7,40 +7,30 @@ This section explains how the chart and its components are sized and laid out wi
 
 ## Chart Size
 
-The chart out of the box uses the minimal styling to ensure the chart is visible, and has a default minimum width and height of 300px.
+The chart will auto-size by default, taking the size of the container element and auto-sizing the chart dynamically. Use the `width` and/or `height` options if a fixed size is required.
 
-Below is an example of a chart with no CSS applied. In this case, the default for CSS is to make the width as large as possible and the height as small as possible (the minimum of 300px).
+If auto-sizing is used for one or both dimensions, the chart will have a minimum width and height of 300px. Set `minHeight: 0` and/or `minWidth: 0` to effectively remove this constraint.
```
alantreadway:
```suggestion
When auto-sizing, by default the chart will have a minimum width and height of 300px. Set `minHeight: 0` and/or `minWidth: 0` to remove this constraint.
```

packages/ag-charts-website/src/content/docs/layout/index.mdoc
```diff
@@ -7,40 +7,30 @@ This section explains how the chart and its components are sized and laid out wi
...
+The chart monitors the size of the `container` element, and resizes dynamically.
 
-```js
-{
-    minWidth: 100,
-    minHeight: 100
-}
-```
+When the chart size changes, the layout process is performed for the new available canvas space.
+Most components are typically a fixed size, so usually it is the series-area which shrinks or grows, with the other components moving around or adjusting alignment.
```
alantreadway:
```suggestion
When the container size changes, a chart layout is performed.
As most components are fixed sizes, any space increase or decrease usually impacts the series-area directly..
```

packages/ag-charts-website/src/content/docs/layout/index.mdoc
```diff
@@ -7,40 +7,30 @@ This section explains how the chart and its components are sized and laid out wi
...
-An explicit size can be set using the `width` and `height` options. In this case, the `minWidth` and `minHeight` are ignored.
+## Sizing the Chart Container
 
-```js
-{
-    width: 400,
-    height: 400
-}
-```
+The chart never affects the size of the `container` element.
```
alantreadway: This isn't true, the `minWidth` and `minHeight` affect the container size by default for `<div>`-based containers (they put a minimum size constraint in place).

packages/ag-charts-website/src/content/docs/layout/index.mdoc
```diff
@@ -53,46 +43,71 @@ const options = {
...
+
 ```
 
+In this configuration:
+
+-   The element containing the chart is sized using the `style` attribute.
+-   It could also be sized by styling the class specified in the `class` attribute.
+
+{% note %}
+As the `ag-charts` container element is a custom HTML element without implicit browser styling, it should be styled as a `block` element.
```
alantreadway, line 86:
```suggestion
As the `ag-charts` container element is a custom HTML element without implicit browser styling, you should style it as a `block` element unless your layout requires more specific configuration.
```

alantreadway (approved):
LGTM

PR #2161, opened 2024-08-02
AG-12380 - Fix annotation cancel when pressing toolbar button.

alantreadway:
TC2 of https://ag-grid.atlassian.net/browse/AG-12380


manapeirov (approved):
LGTM

PR #2176, opened 2024-08-05
CRT-415 Fix note text input placement

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-415

### DIFF
--- a/packages/ag-charts-enterprise/src/features/annotations/note/noteScene.ts
+++ b/packages/ag-charts-enterprise/src/features/annotations/note/noteScene.ts
@@ -53,12 +53,10 @@ export class NoteScene extends TextualPointScene<NoteProperties> {
     }
 
     override getTextBBox(datum: NoteProperties, coords: _Util.Vec2, context: AnnotationContext) {
+        const { seriesRect } = context;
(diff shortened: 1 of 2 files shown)


packages/ag-charts-enterprise/src/features/annotations/note/noteScene.ts
```diff
@@ -53,12 +53,10 @@ export class NoteScene extends TextualPointScene<NoteProperties> {
     }
 
     override getTextBBox(datum: NoteProperties, coords: _Util.Vec2, context: AnnotationContext) {
-        const bbox = super.getTextBBox(datum, coords, context);
-
         const { seriesRect } = context;
+        const bbox = super.getTextBBox(datum, coords, context);
 
-        bbox.y = clamp(seriesRect.y + bbox.height + LABEL_OFFSET + this.padding, bbox.y, seriesRect.height);
```
lsjroberts, line 60: I removed this vertical clamping to match other annotations (and because it isn't accessible in `noteProperties`). Probably should consider moving clamping logic to the text input as well in future.

alantreadway (approved):
LGTM

PR #2178, opened 2024-08-05
CRT-417 Fix range buttons z-index of hover < active < focused

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-417


alantreadway (approved):
LGTM

PR #2191, opened 2024-08-05
CRT-461 Use a moveTo SVG command when d-buffer is empty

olegat:
This avoid console errors with donuts and chords.


alantreadway (approved):
LGTM

PR #2202, opened 2024-08-07
CRT-433 Fix annotations floating buttons reposition delay on resize

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-433

Will look at better caching strategy, but might be a risky change for release.


alantreadway (approved):
LGTM

PR #2204, opened 2024-08-07
CRT-465 Fix change font size after create invalidates input bbox

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-465
https://ag-grid.atlassian.net/browse/CRT-460


alantreadway (approved):
LGTM

PR #2212, opened 2024-08-07
RTI-466 - Sort change log version numbers

taktran:
Intentionally make changelog.json out of order, so it can be tested


markdurrant (approved):
LGTM 👍

PR #2215, opened 2024-08-07
Revert "AG-12417 Fix comment annotation radius"

lsjroberts:
This reverts commit a8dc9fe30b97e6806d94d99fff2954ea82f33b5b.

It breaks on tall comments, the previous approach was a better solution.


alantreadway (approved):
LGTM

PR #2216, opened 2024-08-07
CRT-405 Fix sprite rendering on retina displays

olegat:
https://ag-grid.atlassian.net/browse/CRT-405

### DIFF
--- a/packages/ag-charts-community/src/chart/legend.ts
+++ b/packages/ag-charts-community/src/chart/legend.ts
@@ -19,8 +19,8 @@ import { RedrawType } from '../scene/node';
 import type { Scene } from '../scene/scene';
 import { Selection } from '../scene/selection';
 import { Line } from '../scene/shape/line';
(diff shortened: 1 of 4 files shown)


packages/ag-charts-community/src/chart/legend.ts
```diff
@@ -536,16 +536,17 @@ export class Legend extends BaseProperties {
         });
         spriteWidth += spriteAAPadding * 2;
         spriteHeight += spriteAAPadding * 2;
-        return { spriteAAPadding, spriteWidth, spriteHeight, markerWidth };
+        const spritePixelRatio = getWindow().devicePixelRatio;
```
olegat, line 539: This is the key change.

This is the only place where `devicePixelRation` is used. The `spritePixelRatio` value is used a scaling factor when rendering the sprite.

alantreadway (approved):
LGTM

PR #2221, opened 2024-08-07
CRT-464 update readme image links

JamesSwinton:
Update README image links to use direct links, so they'll work on external sites (e.g. NPM) post-release


alantreadway (approved):
LGTM

PR #2225, opened 2024-08-07
Fix toolbar button border flash

lsjroberts:
My previous assumption was false, it only appeared correct due to the `align-items` property getting caught in the `--transition` variable definition and blocking all transitions.

The only solution appears to be a hacky js one...


alantreadway (approved):
LGTM

PR #2231, opened 2024-08-08
AG-12524 Skip animation calculations when animations are skipped

jacobp100:
https://ag-grid.atlassian.net/browse/AG-12524


alantreadway (approved):
Nice find, LGTM!

PR #2233, opened 2024-08-08
AG-12519 - Reduce duplication of text measuring and wrapping

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-12519


alantreadway (commented):
LGTM

lsjroberts (approved):
LGTM. May just need a rebase to fix the broken snapshots, if it's not caused by this (which seems unlikely).

alantreadway (approved):

PR #2234, opened 2024-08-09
AG-12510 - Scene graph cleanup pt.1

alantreadway:


packages/ag-charts-community/src/scene/transformable.ts
```diff
@@ -0,0 +1,127 @@
...
+        private [TRANSFORM_MATRIX] = new Matrix();
+
+        private _dirtyTransform = false;
+        override markDirtyTransform() {
+            this._dirtyTransform = true;
+            super.markDirtyTransform();
+        }
+
+        protected updateMatrix(_matrix: Matrix) {
+            // For override by sub-classes.
```
iMoses, line 33: nit: if it needs to be defined then shouldn't it be an `abstract`?

packages/ag-charts-community/src/scene/transformable.ts
```diff
@@ -0,0 +1,127 @@
...
+    updateMatrix(matrix: Matrix): void;
+};
+
+export function MatrixTransform<N extends Node>(Parent: Constructor<N>) {
+    const ParentNode = Parent as Constructor<Node>;
+
+    // Make sure we don't mixin `MatrixTransformInternal` multiple times.
+    let nextType = Parent;
+    while (nextType != null) {
+        if (nextType.name === 'MatrixTransformInternal') {
```
iMoses, line 16: if you're counting on the builtin `name` then this could fail after obfuscation

alantreadway: Agreed and understood - I think lets proceed and see what comes out in the wash.

packages/ag-charts-community/src/scene/transformable.ts
alantreadway: I tried to make this `abstract`, but it looks like a limitation of the mixin approach that you can't, I end up with a runtime error of:
```
TypeError: (intermediate value).updateMatrix is not a function
```

PR #2240, opened 2024-08-09
AG-12175 Add chart type initial state

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12175

### DIFF
--- a/packages/ag-charts-community/src/api/preset/chartTypeOriginator.ts
+++ b/packages/ag-charts-community/src/api/preset/chartTypeOriginator.ts
@@ -1,0 +1,43 @@
+import type { AgFinancialChartOptions, AgInitialStateChartType } from 'ag-charts-types';
+
+import type { ChartService } from '../../chart/chartService';
+import { includes } from '../../util/array';
+import { isString } from '../../util/type-guards';
+import type { MementoOriginator } from '../state/memento';
+
+export type ChartTypeMemento = AgInitialStateChartType;
+
+const chartTypes: Array<AgInitialStateChartType> = [
+    'candlestick',
+    'hollow-candlestick',
+    'ohlc',
+    'line',
+    'step-line',
+    'hlc',
+    'high-low',
+];
+
+export class ChartTypeOriginator implements MementoOriginator<ChartTypeMemento> {
+    public mementoOriginatorKey = 'chartType' as const;
+
+    constructor(private readonly chartService: ChartService) {}
+
+    public createMemento() {
+        let chartType = (this.chartService.publicApi?.getOptions() as AgFinancialChartOptions)?.chartType;
+        if (chartType == null) chartType = 'candlestick';
+        return chartType;
+    }
+
+    public guardMemento(blob: unknown): blob is ChartTypeMemento | undefined {
+        return blob == null || (isString(blob) && includes(chartTypes, blob));
+    }
+
+    public restoreMemento(_version: string, _mementoVersion: string, memento: ChartTypeMemento | undefined) {
+        // Migration from older versions can be implemented here.
+
+        if (memento == null) return;
+
+        const options: AgFinancialChartOptions = { chartType: memento };
+        void this.chartService.publicApi?.updateDelta(options as any);
+    }
+}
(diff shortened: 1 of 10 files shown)


packages/ag-charts-community/src/api/preset/chartTypeOriginator.ts
```diff
@@ -0,0 +1,34 @@
...
+export type ChartTypeMemento = AgInitialStateChartType;
+
+export class ChartTypeOriginator implements MementoOriginator<ChartTypeMemento> {
+    public mementoOriginatorKey = 'chartType' as const;
+
+    constructor(private readonly chartService: ChartService) {}
+
+    public createMemento() {
+        let chartType = (this.chartService.publicApi?.getOptions() as AgFinancialChartOptions)?.chartType;
+        if (chartType === 'range-area') chartType = 'candlestick';
```
lsjroberts: Since `'range-area'` is deprecated, would we want to support it in this new api?

packages/ag-charts-types/src/api/initialStateOptions.ts
```diff
@@ -4,10 +4,21 @@ import type { AgStateSerializableDate } from './stateTypes';
...
 }
 
+export type AgInitialStateChartType =
+    | 'candlestick'
+    | 'hollow-candlestick'
+    | 'ohlc'
+    | 'line'
+    | 'step-line'
+    | 'hlc'
+    | 'high-low';
```
lsjroberts: Ideally this would just be `AgInitialStateChartType = Preset['chartType']`, but the docs would just show `Preset['chartType']`, so it needs to be recreated in full.

packages/ag-charts-community/src/api/preset/chartTypeOriginator.ts
alantreadway: Is there any programatic way to reach this state? (Guessing no?) I agree I don't think we need to support this value strictly.

packages/ag-charts-types/src/api/initialStateOptions.ts
alantreadway: I think I would opt for one of (in order of precedence):
- Fixing the doc generation.
- Duplication, but with a compile-time check that the types match (see `__VERIFY_THEME_OVERRIDES`).

lsjroberts:
On hold pending further discussion, converting to draft for now.

packages/ag-charts-community/src/api/preset/chartTypeOriginator.ts
lsjroberts: `'range-area'` has been removed

packages/ag-charts-types/src/api/initialStateOptions.ts
lsjroberts: Fixed, it will now expand `type AgInitialStateChartType = AgPriceVolumeChartType` into:

Screenshot 2024-11-01 at 11 56 05

So if we add new presets to this state then it shouldn't be a breaking change.

packages/ag-charts-community/src/chart/chartProxy.ts
```diff
@@ -216,16 +216,23 @@ export class AgChartInstanceProxy implements AgChartProxy {
...
 
         const originators = [];
 
-        if ('annotations' in options && options.annotations?.enabled) {
+        if ('annotations' in processedOptions && processedOptions.annotations?.enabled) {
             originators.push(annotationManager);
         }
 
-        if (options.navigator?.enabled || options.zoom?.enabled) {
+        const isFinancialChart = optionMetadata.presetType === 'price-volume';
```
lsjroberts, line 242: There may be a better check we can do here?

Without it, this test fails as it tries to restore the unrecognised `chartType` property, not entirely sure why it tries that for this test, but it seems reasonable to only include this originator in FC - https://github.com/ag-grid/ag-charts/blob/latest/packages/ag-charts-community/src/chart/agChartV2.test.ts#L183-L191

alantreadway: I think this is ok for now.

alantreadway (approved):
LGTM

PR #2242, opened 2024-08-09
AG-12510 - Remove translation support from Node.ts.

alantreadway:


iMoses (approved):
LGTM 👍

packages/ag-charts-community/src/scene/matrix.ts
```diff
@@ -1,5 +1,7 @@
 import { BBox } from './bbox';
 
+export const IDENTITY_MATRIX_ELEMENTS = Object.freeze([1, 0, 0, 1, 0, 0]);
```
jacobp100, line 3: There's a performance cost to object.freeze - especially on arrays. You can mark this `as readonly number[]`

PR #2245, opened 2024-08-12
AG-12510 - Scene graph cleanup pt3

alantreadway:

- Removes `Node.matrix`.
- Cleans-up the discrepancy between`Group.getBBox()` and `Node.getBBox()`.
- Fixes `Node.getBBox()` caching - all dependencies are now correctly tracked, and there is no dependency on `parent` or other ancestors.
- Moves some responsibility for canvas->local coordinates to `RegionManager`.
- Cleans-up some spots where transforms were overcomplicating calculations.


packages/ag-charts-community/src/chart/__image_snapshots__/cartesian-chart-test-ts-cartesian-chart-stacked-bar-series-highlighting-should-correctly-change-highlighting-state-and-reset-1-snap.png
alantreadway, line 1: Snapshot changes from here on are due to slight shifts in the legend positioning.

packages/ag-charts-community/benchmarks/__image_snapshots__/large-dataset-test-ts-large-dataset-benchmark-after-load-1-x-datum-highlight-1-snap.png
alantreadway, line 1: Benchmark snapshot changes are due to the hover point having changed, it's not important to the test for a specific node to be picked out.

packages/ag-charts-community/src/scene/transformable.ts
```diff
@@ -155,3 +233,75 @@ export function Translatable<N extends Node>(Parent: Constructor<N>): Constructo
     }
     return ScalableInternal as unknown as Constructor<TranslatableType<N>>;
 }
+
+export class Transformable {
+    static fromCanvas(node: Node, bbox: BBox) {
```
lsjroberts, line 250: nit: Would be nice to have some short comments on these functions to define their purpose. What the bbox is the bbox of, for example. These are good function names, but just to be abundantly clear what they are for.

lsjroberts (approved):
LGTM from what I understand. Impressive stuff!

packages/ag-charts-community/src/scene/transformable.ts
alantreadway: JSDocs added, thanks for pointing out.

PR #2247, opened 2024-08-13
AG-11993 Deprecate zoom range/ratio and navigator min/max

lsjroberts:

Deprecate decorator is intended for renames within the same module, so used a custom warn message.

### DIFF
--- a/packages/ag-charts-enterprise/src/features/zoom/zoomRange.ts
+++ b/packages/ag-charts-enterprise/src/features/zoom/zoomRange.ts
@@ -5,7 +5,7 @@ const { Logger } = _Util;
 
 export class ZoomRange {
     @ObserveChanges<ZoomRange>((target, start) => {
-        if (target.initialStart == null) {
+        if (target.initialStart == null && !target.hasRestored) {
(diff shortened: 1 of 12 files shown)


packages/ag-charts-enterprise/src/features/zoom/zoomRange.ts
```diff
@@ -1,9 +1,13 @@
...
+import { _ModuleSupport, _Util } from 'ag-charts-community';
 
 const { AND, DATE, NUMBER, OR, ObserveChanges, Validate } = _ModuleSupport;
+const { Logger } = _Util;
 
 export class ZoomRange {
     @ObserveChanges<ZoomRange>((target, start) => {
+        if (target.initialStart == null) {
+            Logger.warnOnce('Property [zoom.rangeX] is deprecated. Use [initialState.zoom.rangeX] instead.');
+        }
```
lsjroberts: This check stops it from warning when the range is restored from `initialState.zoom` as `ZoomRange.initialStart` is set before `ZoomRange.start` in `ZoomRange.reset()`, which is called by `Zoom.onRestoreZoom()`.

alantreadway (approved):
LGTM

alantreadway (approved):
LGTM

PR #2250, opened 2024-08-13
AG-12498 Add extending lines and channels

lsjroberts:

### DIFF
--- a/packages/ag-charts-enterprise/src/features/annotations/annotationProperties.ts
+++ b/packages/ag-charts-enterprise/src/features/annotations/annotationProperties.ts
@@ -171,10 +171,10 @@ export function Cappable<T extends Constructor>(Parent: T) {
 export function Extendable<T extends Constructor>(Parent: T) {
     class ExtendableInternal extends Parent {
         @Validate(BOOLEAN, { optional: true })
-        extendLeft?: boolean;
+        extendStart?: boolean;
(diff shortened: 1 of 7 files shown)


packages/ag-charts-enterprise/src/features/annotations/annotationProperties.ts
```diff
@@ -171,10 +171,10 @@ export function Cappable<T extends Constructor>(Parent: T) {
 export function Extendable<T extends Constructor>(Parent: T) {
     class ExtendableInternal extends Parent {
         @Validate(BOOLEAN, { optional: true })
-        extendLeft?: boolean;
+        extendStart?: boolean;
 
         @Validate(BOOLEAN, { optional: true })
-        extendRight?: boolean;
+        extendEnd?: boolean;
```
lsjroberts, line 177: The properties `extendLeft` and `extendRight` were not part of the options api, so they were not previously accessible and this renaming shouldn't be a breaking change. I believe.

alantreadway: SGTM - I usually only consider the `ag-charts-types` options as our public contract, if we aren't fulfilling that contract here it's a bug not a breaking change.

alantreadway (approved):
LGTM

PR #2251, opened 2024-08-13
Refactor popovers and color picker into anchored popover and menu

lsjroberts:

Refactors popovers into reusable components:

- Popover
  - AnchoredPopover
    - ColorPicker
    - Menu (annotation type picker, font size)
  - Dialog (future PR)
    - SettingsDialog (draggable popover for annotations text, etc)


lsjroberts:
Don't merge just yet, it will likely cause conflicts for other devs. Will co-ordinate with them.

packages/ag-charts-enterprise/src/features/color-picker/colorPickerStyles.css
```diff
@@ -24,6 +16,12 @@
     --checker: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" width="4" height="4"><rect x="0" y="0" width="4" height="4" fill="%23fff"/><path d="M0 0H2V4H4V2H0Z" fill="%23b2b2b2"/></svg>');
 }
 
+.ag-charts-color-picker__content {
```
jacobp100, line 19: Did the layout change here?

packages/ag-charts-enterprise/src/features/color-picker/colorPickerStyles.css
```diff
@@ -24,6 +16,12 @@
...
+    display: flex;
+    flex-direction: column;
+    color: var(--ag-charts-toolbar-foreground-color);
```
jacobp100, line 22: This is already in the popover styles, I believe

```suggestion
```

jacobp100 (approved):
Nice!

packages/ag-charts-enterprise/src/features/color-picker/colorPickerStyles.css
lsjroberts: Yeah, it's been flattened a little, merging `.ag-charts-popover` and `.ag-charts-color-picker` onto the same element (removing the `colorPickerContainer` element which served the same purpose as the outer popover element).

Ideally the `__content` styles could be merged into `.ag-charts-color-picker` but I had some issues with the width of the sliders. Will double check that.

From:
Screenshot 2024-08-15 at 16 47 57

To:
Screenshot 2024-08-15 at 16 48 24

packages/ag-charts-enterprise/src/features/color-picker/colorPickerStyles.css
lsjroberts: Woops, good spot, thanks!

packages/ag-charts-enterprise/src/components/popover/popover.ts
```diff
@@ -0,0 +1,85 @@
...
+        this.moduleId = `popover-${id}`;
+
+        this.element = ctx.domManager.addChild(canvasOverlay, this.moduleId);
+        this.element.role = 'presentation';
+
+        this.destroyFns.push(() => ctx.domManager.removeChild(canvasOverlay, this.moduleId));
+    }
+
+    public show(options: Options) {
+        const popover = createElement('div');
```
jacobp100, line 39: Maybe we could return this so things like the colour picker can attach css custom properties directly?

alantreadway (approved):
LGTM, merging as I know the intent is to continue iterating on this for the actual feature.

PR #2258, opened 2024-08-14
AG-11958 Remove annotations enabled toggle hack

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-11958

### DIFF
--- a/packages/ag-charts-enterprise/src/features/annotations/annotations.ts
+++ b/packages/ag-charts-enterprise/src/features/annotations/annotations.ts
@@ -45,6 +45,7 @@ const {
     ChartUpdateType,
     Cursor,
     InteractionState,
+    ObserveChanges,
     PropertiesArray,
     ToolbarManager,
     Validate,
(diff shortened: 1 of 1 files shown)


packages/ag-charts-enterprise/src/features/annotations/annotations.ts
```diff
@@ -134,26 +135,19 @@ class AxesButtons {
...
-    private __hackWasDisabled = false;
-
-    @_ModuleSupport.ObserveChanges<Annotations, boolean>((target, enabled) => {
+    @ObserveChanges<Annotations>((target, newValue?: boolean, oldValue?: boolean) => {
         const {
             ctx: { annotationManager, stateManager, toolbarManager },
         } = target;
 
-        toolbarManager.toggleGroup('annotations', 'annotations', { visible: Boolean(enabled) });
+        if (newValue === oldValue) return;
```
lsjroberts, line 143: Turns out the issue was simply that `ObserveChanges()` gets called twice on initial set with `{ newValue: true, oldValue: undefined }` then `{ newValue: true, oldValue: true }`.

Skipping this second call fixes the issue, much like `ActionOnSet()` but that is a transformer which gets called _before_ the value has changed, while the observers get called _after_ the value has changed.

I thought about adding an `ActionAfterSet()` as an observer, but didn't seem necessary for just this one case. Though it may be worth analysing our uses of `ObserveChanges()` to see if they can be improved similarly to prevent excessive calls and then use something like `ActionAfterSet()`.

alantreadway (approved):
LGTM

PR #2259, opened 2024-08-14
Add radial gauge series

jacobp100:


packages/ag-charts-types/src/main.ts
alantreadway, line 1: I wouldn't expect to see any additions to `ag-charts-types`, as this package is our public API.

packages/ag-charts-types/src/series/gauge/radialGaugeOptions.ts
alantreadway, line 1: Ditto

alantreadway (approved):
LGTM

PR #2261, opened 2024-08-15
AG-NONE Move `chart/locale/` and `chart/dom/` directories

olegat:
These folders contain very general-use classes and functionality, so makes sense to move them out of the `chart/` dir.

Moves:

- `src/chart/locale` -> `src/locale`
- `src/chart/dom` -> `src/dom`


alantreadway:
Agree with the `locale` and `dom` moves, but `BaseManager` isn't really a util - can we leave that where it is for now?

olegat:
> Agree with the locale and dom moves, but BaseManager isn't really a util - can we leave that where it is for now?

I've dropped the commit that moves `BaseManager`

alantreadway (approved):
LGTM

PR #2266, opened 2024-08-15
AG-12369 Differentiate between mouse in keyboard click in popover and context menus.

olegat:
https://ag-grid.atlassian.net/browse/AG-12369


packages/ag-charts-enterprise/src/features/popover/popover.ts
lsjroberts, line 1: These changes should be able to just be copied over to `components/menu/menu.ts` by the looks of it.

olegat: Done.

alantreadway (approved):
LGTM

PR #2269, opened 2024-08-16
AG-12510 - Misc cleanup post node matrix removal.

alantreadway:


packages/ag-charts-enterprise/src/features/annotations/scenes/annotationScene.ts
```diff
@@ -28,19 +26,10 @@ export abstract class AnnotationScene extends _Scene.Group {
...
         return _Scene.Transformable.toCanvas(
             this,
             _Scene.Group.computeChildrenBBox(this.children.filter((node) => !(node instanceof Handle)))
         );
     }
-
-    override render(renderCtx: _Scene.RenderContext) {
-        super.render(renderCtx);
-        this.cachedBBoxWithoutHandles = this.computeBBoxWithoutHandles();
-    }
```
lsjroberts, line 45: Just to double check, this was added to stop the anchor point (and therefore the floating options buttons) from shifting after you first select an annotation and then the next render changed the bbox.

lsjroberts (approved):
LGTM

packages/ag-charts-enterprise/src/features/annotations/scenes/annotationScene.ts
alantreadway: I'll double check the use-case, but this code was not being used so I don't believe it is currently having any effect.

alantreadway: Specifically, `getCachedBBoxWithoutHandles()` is not called anywhere, so the caching is not being used?

packages/ag-charts-community/benchmarks/__image_snapshots__/large-dataset-test-ts-large-dataset-benchmark-after-load-1-x-datum-highlight-1-snap.png
alantreadway, line 1: Reverts snapshot changes from #2245 as the benchmarks have been updated to use `Transform.toCanvasPoint()`.

packages/ag-charts-enterprise/src/features/annotations/scenes/annotationScene.ts
alantreadway: Discussed offline - agreed this code was redundant.

PR #2272, opened 2024-08-16
AG-12510 - Fix legend pixel shift.

alantreadway:

- Re-adds support for `window.agChartsSceneDebug`.
- Fixes hit-testing for legend pagination buttons (broken by #2245).
- Fixes pixel shift in legend, due to `pagination.updateMarkers()` not being called after `pagination.update()`.


packages/ag-charts-community/src/scene/transformable.ts
```diff
@@ -101,8 +101,10 @@ function MatrixTransform<N extends Node>(Parent: Constructor<N>) {
             return this.toParent(bbox);
         }
 
-        override pickNode(x: number, y: number) {
-            ({ x, y } = this.fromParentPoint(x, y));
+        override pickNode(x: number, y: number, localCoords = false) {
+            if (!localCoords) {
+                ({ x, y } = this.fromParentPoint(x, y));
```
jacobp100, line 106: Is this the direct parent, or the root node?

jacobp100 (approved):

packages/ag-charts-community/src/scene/transformable.ts
alantreadway: Direct parent - we don't do to/from root except in specific cases now.

There is just a little disconnect with a couple of cases that we need to address:
- For the common cases, the coordinates are in the parents coordinates space, so we need to apply local transforms into the local coordinate space.
- For the less common case, the coordinates are already in the nodes local coordinate space, so we don't need to transform them.

PR #2274, opened 2024-08-19
AG-12601 Add dialog popover and annotation settings dialog

lsjroberts:

- Changes popover to abstract, then concrete implementations use `const popover = this.showWithChildren(children, options);` (see https://github.com/ag-grid/ag-charts/pull/2251#pullrequestreview-2240817928)
- Adds the generic dialog and annotation settings dialog ui elements.

Further styling changes will be made in a future PR pending finalised design.

Handling of the text changing and rendering on the line will be done in https://ag-grid.atlassian.net/browse/AG-12600


alantreadway (approved):
LGTM, @markdurrant could you review the CSS additions?

packages/ag-charts-enterprise/src/components/dialog/dialog.css
```diff
@@ -0,0 +1,49 @@
...
+}
+
+.ag-charts-dialog__textarea {
+    --line-height: 1.38;
+    background: white;
+    cursor: text;
+    line-height: var(--line-height);
+    height: calc(var(--line-height) * 4em);
+    overflow-y: scroll;
+    padding: 1em;
```
markdurrant, line 43: Suggest updating `em` units to `px` values for things like padding. 

If not `px` then `rems`. Would not suggest using `em` except for exceptional circumstances.

packages/ag-charts-enterprise/src/components/dialog/dialog.css
```diff
@@ -0,0 +1,49 @@
+.ag-charts-dialog {
+    box-shadow: 2px 2px 4px rgba(0, 0, 0, 0.1);
+    display: flex;
+    flex-direction: column;
+    font-size: 14px;
+    gap: 1em;
+    padding: 1em;
```
markdurrant, line 7: Suggest updating `em` units to `px` values for things like padding. 

If not `px` then `rems`. Would not suggest using `em` except for exceptional circumstances.

markdurrant (approved):
Generally CSS LGTM 👍 

I've left a couple comments about replacing `em` units with `px`. But I guess these values will change post design so can update then if you prefer.

PR #2275, opened 2024-08-19
Gauge API

jacobp100:


packages/ag-charts-types/src/chart/chartBuilderOptions.ts
```diff
@@ -51,6 +51,8 @@ export type AgBaseFinancialPresetOptions = AgBasePresetOptions & Pick<AgCartesia
 
 export type AgFinancialChartOptions = AgFinancialChartPresets & AgBaseFinancialPresetOptions;
 
+export type AgGaugeChartOptions2 = AgBasePresetOptions & AgGaugeSeriesOptions;
```
alantreadway, line 54: `AgBasePresetOptions` includes `data`, which probably isn't relevant given we have `value` too?

jacobp100:
https://github.com/ag-grid/ag-charts/pull/2278

PR #2293, opened 2024-08-21
AG-12600 Add text labels to linear drawings

lsjroberts:


lsjroberts:
https://github.com/ag-grid/ag-charts/actions/runs/10504714929

packages/ag-charts-website/e2e/toolbar.spec.ts-snapshots/line-4-complete-chromium-linux.png
lsjroberts, line 1: This movement of the floating options shouldn't have happened, will fix.

lsjroberts:
https://github.com/ag-grid/ag-charts/actions/runs/10505588276

packages/ag-charts-community/src/scene/group.ts
```diff
@@ -301,10 +305,26 @@ export class Group extends Node {
         }
         if (stats) stats.nodesSkipped += skipped;
 
+        if (clipShape && clipMode === 'outside') {
```
alantreadway: Do we need to add support for this behaviour to `Group`? I'd prefer not to add more concerns into the mix for `Group` if possible (also, we should implement support for layer-handing for non-BBox clips if we add it here); could we either implement this behaviour as a mixin for specific cases, or as part of `Shape` or another relevant `Node` subclass?

packages/ag-charts-enterprise/src/features/annotations/__image_snapshots__/annotations-test-ts-annotations-lines-with-text-should-render-disjoint-channel-annotations-with-text-1-snap.png
alantreadway, line 1: 👍

packages/ag-charts-community/src/scene/clip.ts
```diff
@@ -0,0 +1,52 @@
...
+
+type Constructor<T> = new (...args: any[]) => T;
+type Circle = { x: number; y: number; radius: number };
+
+export type ClippableOutsideType<T> = T & {
+    clipShape?: BBox | Circle;
+    setClipMask: (shape?: BBox | Circle) => void;
+};
+
+export function ClippableOutside<N extends Node>(Parent: Constructor<N>): Constructor<ClippableOutsideType<N>> {
```
lsjroberts, line 13: I looked into doing this as a general `Clippable` mixin with a mode of `inside | outside`. For inside, I tried overriding the render method with:

```ts
ctx.globalCompositeOperation = 'source-in'; // and others
super.render(renderCtx);
ctx.globalCompositeOperation = prevOperation;
```

But this impacts all the other nodes within the same canvas layer, erasing them.

If we end up with a use-case for this approach we can look into this further.

alantreadway: Ah, so maybe you need to use `ctx.save()` and `ctx.restore()` rather than manually managing the ctx state rollback here? I think you need to use this approach anyway to avoid state leakage?

Agree we can leave this use-case for now.

packages/ag-charts-community/src/scene/clip.ts
```diff
@@ -0,0 +1,52 @@
...
+    const ParentNode = Parent as Constructor<Node>;
+    class ClippableOutsideInternal extends ParentNode {
+        clipShape?: BBox | Circle;
+
+        setClipMask(shape?: BBox | Circle) {
+            if (jsonDiff(this.clipShape, shape) != null) {
+                this.markDirty(this, RedrawType.MINOR);
```
alantreadway: Nit: `MAJOR` seems more appropriate, as the shape/boundary of the rendering will change.

packages/ag-charts-community/src/scene/clip.ts
```diff
@@ -0,0 +1,52 @@
...
+        override postRender(renderCtx: RenderContext) {
+            const { ctx } = renderCtx;
+            const { clipShape } = this;
+
+            if (!clipShape) {
+                super.postRender(renderCtx);
+                return;
+            }
+
+            const prevOperation = ctx.globalCompositeOperation;
```
alantreadway: Try `ctx.save()` here, and `ctx.restore()` below - these are more bullet-proof ways of restoring `ctx` state, and what we are using elsewhere?

packages/ag-charts-community/src/scene/transformable.ts
```diff
@@ -227,7 +227,7 @@ export type TranslatableType<T> = MatrixTransformType<
 export function Translatable<N extends Node>(Parent: Constructor<N>): Constructor<TranslatableType<N>> {
     const ParentNode = Parent as Constructor<Node>;
     const TRANSLATABLE_MATRIX = Symbol('matrix_translation');
-    class ScalableInternal extends MatrixTransform(ParentNode) {
+    class TranslatableInternal extends MatrixTransform(ParentNode) {
```
alantreadway, line 230: 👍

packages/ag-charts-community/src/scene/clip.ts
lsjroberts: Oh of course, thanks!

alantreadway (approved):
LGTM

lsjroberts:
An issue with this approach, annotations cut out the ones below as they are all in the same layer.

Screenshot 2024-08-23 at 15 09 05

PR #2306, opened 2024-08-23
TS lib to use es2023, remove dead code

iMoses:


alantreadway (approved):
LGTM, compatibility looks good: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findLast#browser_compatibility

PR #2310, opened 2024-08-23
AG-12600 Fix clipping on stacked linear drawings with text labels

lsjroberts:

https://github.com/ag-grid/ag-charts/pull/2293#issuecomment-2307178696

Screenshot 2024-08-23 at 15 09 05


packages/ag-charts-enterprise/src/features/annotations/line/lineScene.ts
```diff
@@ -29,7 +29,7 @@ export class LineScene extends LinearScene<LineTypeProperties> {
     private endCap?: CapScene;
 
     constructor() {
-        super();
+        super({ layer: true });
```
lsjroberts, line 32: Not ideal, but needs to be a different canvas to prevent it from erasing other drawings on the same canvas. Considered doing this dynamically, based on if it requires a clip, but those changes in `Group` might impact other things more than desired.

lsjroberts:
Closing in favour of a different approach.

PR #2312, opened 2024-08-27
AG-12121 Add background to status bar module

jacobp100:
https://ag-grid.atlassian.net/browse/AG-12121


jacobp100:
Sign off

image

PR #2327, opened 2024-08-28
Refactor annotation config per type

lsjroberts:


manapeirov (approved):
Looks great thanks for this Laurence!

PR #2348, opened 2024-08-29
Ag 12568/annotation options toolbar drag

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-12568


packages/ag-charts-community/src/chart/toolbar/toolbar.ts
```diff
@@ -676,6 +727,12 @@ export class Toolbar extends BaseModuleInstance implements ModuleInstance {
         button.onclick = makeAccessibleClickListener(button, (event) =>
             this.onButtonPress(event, button, group, options.id, options.value)
         );
+
+        if (options.value === 'drag') {
+            button.onmousedown = makeAccessibleClickListener(button, (event) => this.onDragStart(event, group));
```
lsjroberts:
```suggestion
            button.addEventListener('mousedown', makeAccessibleClickListener(button, (event) => this.onDragStart(event, group)));
```

packages/ag-charts-community/src/chart/toolbar/toolbar.ts
```diff
@@ -770,6 +827,46 @@ export class Toolbar extends BaseModuleInstance implements ModuleInstance {
...
+                Number(element.style.getPropertyValue('left').replace('px', '')),
+                Number(element.style.getPropertyValue('top').replace('px', ''))
+            ),
+            detached: true,
+        };
+
+        const onDrag = (e: MouseEvent) => this.onDrag(e, group);
+        const window = getWindow();
+        window.addEventListener('mousemove', onDrag);
+        window.addEventListener('mouseup', () => window.removeEventListener('mousemove', onDrag), {
```
lsjroberts, line 840: I was having some issues with this with the dialog since some events are not propagated up beyond our container element. I suggest using the `domManager` instead of `window` as that contains all we care about and will always receive the events.

```suggestion
        const { ctx: { domManager } } = this;
        domManager.addEventListener('mousemove', onDrag);
        domManager.addEventListener('mouseup', () => domManager.removeEventListener('mousemove', onDrag), {
```

packages/ag-charts-community/src/chart/toolbar/toolbarTypes.ts
```diff
@@ -45,3 +45,9 @@ export interface ToolbarButton extends ToolbarButtonConfig {
     role?: 'button' | 'switch';
     checkedOverrides?: ToolbarButtonConfig;
 }
+
+export interface Anchor {
```
lsjroberts: To be more specific:

```suggestion
export interface ToolbarAnchor {
```

packages/ag-charts-community/src/chart/toolbar/toolbar.ts
```diff
@@ -285,16 +302,16 @@ export class Toolbar extends BaseModuleInstance implements ModuleInstance {
...
+
+        const { group, anchor, index } = event;
+
+        if (this.index == index && this.dragState.detached) {
+            return;
+        }
 
-        const { group, anchor } = event;
+        this.index = index;
+        this.dragState.detached = false;
```
lsjroberts: This `index` is very specific to annotations and a bit of a random number to save here in toolbar. It's a bit leaky to have it here.

I _think_ you could achieve the same behaviour of reattaching the floating toolbar when you select a new annotation by resetting `detached` in the `Toolbar.toggleVisibilities()` function, something like:

```diff
for (const position of TOOLBAR_POSITIONS) {
    const visible = this.enabled && Array.from(this.positions[position].values()).some(isGroupVisible);
    this.elements[position].classList.toggle(styles.modifiers.hidden, !visible);
+   if (position === ToolbarPosition.Floating && !visible) {
+       this.dragState.detached = false;
+   }
}
```

As when you select a new annotation it will deselect the previous one first, hide the overlays, toggling the visibility of the floating toolbar to false and reattaching it here, then show the new version of it.

packages/ag-charts-community/src/chart/toolbar/toolbar.ts
```diff
@@ -512,6 +557,12 @@ export class Toolbar extends BaseModuleInstance implements ModuleInstance {
                 this.setButtonActive(button, active);
             }
         }
+
+        if (this[group].draggable) {
+            this.ctx.toolbarManager.toggleButton(group, 'drag', {
+                enabled,
+            });
+        }
```
lsjroberts: Calling from `toolbar` out to the `toolbarManager` to trigger a listener inside `toolbar` is a bit circuitous.

What do you need from the `onButtonToggled()` listener to happen? Other than the `setButtonActive()` that's already called a few lines above?

packages/ag-charts-community/src/chart/toolbar/toolbar.ts
```diff
@@ -622,7 +673,7 @@ export class Toolbar extends BaseModuleInstance implements ModuleInstance {
             if (this[group] == null) continue;
             const groupVisible = isGroupVisible(group);
             for (const button of this.groupButtons[group]) {
-                const buttonVisible = groupVisible && this[group].buttonConfigurations().some(isButtonVisible(button));
+                const buttonVisible = groupVisible && isButtonVisible(button);
```
lsjroberts: Did you mean to commit this?

You'll need to revert this change, the `isButtonVisible(button)` call returns a function so will always be truthy. And we need the `buttonConfigurations()` bit to fix issues with locale changes resetting buttons I believe (see https://github.com/ag-grid/ag-charts/commit/d4e833f3c766837ce7898e83eea3603aeda9d0f5).

lsjroberts (changes requested):

packages/ag-charts-community/src/chart/toolbar/toolbar.ts
lsjroberts: Though, may not be needed now - https://github.com/ag-grid/ag-charts/pull/2347/commits/c17cc10d94298fc5b1e71c20dbd96fb5319f7017

manapeirov: The issue I found with using domManager is that dragging will continue if you release the mouse (mouseup) outside of the canvas container, I did window so it behaves like the navigator drag

packages/ag-charts-community/src/chart/toolbar/toolbar.ts
manapeirov: I think I left that in there by accident :)

packages/ag-charts-community/src/chart/toolbar/toolbar.ts
manapeirov: Discussed offline and decided to fix in refactor of toobar later

lsjroberts (approved):
LGTM, excellent stuff!

PR #2351, opened 2024-08-30
AG-12614 Fix legend symbol shape when changing series type in `updateDelta`

olegat:
https://ag-grid.atlassian.net/browse/AG-12614


packages/ag-charts-website/src/content/docs/benchmarks/_examples/summary/data.ts
```diff
@@ -922,5 +922,129 @@ export function getData() {
                 },
             },
         },
+        {
```
jacobp100: Is this expected to be added?

jacobp100 (approved):

packages/ag-charts-website/src/content/docs/benchmarks/_examples/summary/data.ts
olegat: I think no. This was added when running `nx benchmark -c update`

PR #2352, opened 2024-08-30
Improve focus path for keyboard navigation

jacobp100:
https://ag-grid.atlassian.net/browse/AG-12481

### DIFF
--- a/packages/ag-charts-community/src/scene/extendedPath2D.ts
+++ b/packages/ag-charts-community/src/scene/extendedPath2D.ts
@@ -1,7 +1,7 @@
-import { angleDiff } from '../util/angle';
 import { arcDistanceSquared, lineDistanceSquared } from '../util/distance';
 import { Logger } from '../util/logger';
 import { arcIntersections, cubicSegmentIntersections, segmentIntersection } from './intersection';
+import type { Matrix } from './matrix';
 
 enum Command {
     Move,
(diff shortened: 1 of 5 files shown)


packages/ag-charts-community/src/scene/extendedPath2D.ts
```diff
@@ -299,48 +300,72 @@ export class ExtendedPath2D {
     }
 
     // https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/d
-    computeSVGDataPath(ox: number, oy: number): string {
+    computeSVGDataPath(matrix: Matrix): string {
```
olegat, line 303: I do not know whether this is correct approach.

My understanding is that we want to avoid using transformation matrices unless it's absolutely necessary. The `ExtendedPath2D` object should in theory have the rotation and scale transformations pre-computed before invoking `computeSVGDataPath` and only require an `{ox, oy}` translation.

Could you please provide more information about the issue this PR is solving? (e.g. ticket number, steps-to-repro, screenshot, ...)

olegat (changes requested):

packages/ag-charts-community/src/scene/extendedPath2D.ts
olegat: Discussed on Slack:
It has to do with markers which are scaleable and rotatable: https://github.com/ag-grid/ag-charts/blob/b96abab48840caf08dec8eba981ab7ee24cb7bec/packages/ag-charts-community/src/chart/marker/marker.ts#L58

olegat (approved):
LGTM

PR #2353, opened 2024-08-30
Detach DOM Manager from Scene implementation

iMoses:


packages/ag-charts-community/src/scene/scene.ts
```diff
@@ -63,21 +44,6 @@ export class Scene {
         return this.pendingSize?.[1] ?? this.canvas.height;
     }
 
-    setContainer(value: HTMLElement | DOMManagerLike) {
```
olegat, line 66: @iMoses AG Grid Sparkline uses the `HTMLElement` overload of this method:
https://github.com/ag-grid/ag-grid/blob/e8dc3809826568907440b6f8d198261bdddb905b/enterprise-modules/sparklines/src/sparkline/sparkline.ts#L157

You can safely remove the `DOMManagerLike` branch of this logic, but the `HTMLElement` branch must remain to avoid an ag-grid breakage.

PR #2362, opened 2024-09-02
Ag 12568/annotation options toolbar drag

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-12568


packages/ag-charts-enterprise/src/features/annotations/annotations.ts
```diff
@@ -291,18 +291,19 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
                     tooltipManager.suppressTooltip('annotations');
                 }
 
-                if (index == null || (index != null && index !== previous)) {
-                    ctx.toolbarManager.updateButton('annotations', 'line-menu', { icon: undefined });
-                    ctx.toolbarManager.updateButton('annotations', 'text-menu', { icon: undefined });
-                }
-
-                if (node) {
+                if (node && index != null) {
```
lsjroberts, line 294: I may be missing something, but I don't _think_ you need the `index != null` check here as `node` will be null if `index` is null. From line 279.

packages/ag-charts-enterprise/src/features/annotations/annotations.ts
```diff
@@ -291,18 +291,19 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
...
                     this.toggleAnnotationOptionsButtons();
                     toolbarManager.toggleGroup('annotations', 'annotationOptions', { visible: true });
                 } else {
                     toolbarManager.toggleGroup('annotations', 'annotationOptions', { visible: false });
                 }
 
+                if (index == null || (index != null && index !== previous)) {
+                    ctx.toolbarManager.updateButton('annotations', 'line-menu', { icon: undefined });
+                    ctx.toolbarManager.updateButton('annotations', 'text-menu', { icon: undefined });
+                }
```
lsjroberts, line 305: I shifted this to before the group toggling to fix it flashing when you change the annotation type, see https://github.com/ag-grid/ag-charts/pull/2131/files#diff-fde1cb072dc3708f277225348f863116c84ddf03dd09aad886b49be27fa0a65c

Could you check if that is still fixed after your changes? (see steps to reproduce here https://ag-grid.atlassian.net/browse/AG-12358, but the example is out of date so use this example https://ag-grid.com/charts/react/financial-charts-toolbar/#lines)

lsjroberts (approved):
Looks good, just a couple of things to double check

packages/ag-charts-enterprise/src/features/annotations/annotations.ts
manapeirov: Yeah I just did that for ts it was complaining that index is undefined

packages/ag-charts-enterprise/src/features/annotations/annotations.ts
manapeirov: Still looks good!

PR #2363, opened 2024-09-02
AG-12625 - New styles for ApiReference component

taktran:
Sync with Grid


packages/ag-charts-website/src/features/api-documentation/components/ApiReference.module.scss
```diff
@@ -159,34 +174,62 @@
...
+    text-decoration: underline;
+    padding-top: 2px;
+    padding-bottom: 2px;
+    padding-left: 8px;
+    padding-right: 8px;
+    border-radius: var(--radius-sm);
+    background-color: color-mix(in srgb, var(--color-link), var(--color-bg-primary) 90%);
 
     &:hover {
         color: var(--color-link-hover);
```
markdurrant: Looks like a lot of shared styles between `.metaValue`, `.defaultValue`, `initialValue`. Wonder if we can refactor this to combine shared styles.

packages/ag-charts-website/src/features/api-documentation/components/ApiReference.module.scss
```diff
@@ -300,9 +379,16 @@ button.seeMore {
...
+        border: none;
+    }
 }
 
 .expandedContent > td {
-    padding: 0 0 $spacing-size-4;
+    padding: $spacing-size-2 0 $spacing-size-4;
+    display: flex;
+    flex-direction: column;
 }
```
markdurrant: Nest `> td` within `.expandedContent`?

packages/ag-charts-website/src/pages/[framework]/[pageName].astro
```diff
@@ -48,7 +48,7 @@ const descriptionWithSeoTagline = `${description} ${seoTagline}`;
 
 <link rel="preconnect" href="https://fonts.googleapis.com" />
 <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
-<link href="https://fonts.googleapis.com/css2?family=JetBrains+Mono:wght@400;500&display=swap" rel="stylesheet" />
+<link href="https://fonts.googleapis.com/css2?family=JetBrains+Mono:wght@400;500;700&display=swap" rel="stylesheet" />
```
Comment (markdurrant) [inline packages/ag-charts-website/src/pages/[framework]/[pageName].astro]: Be good to check with KP exactly what font weights we need. Do we have bold monospace text anywhere? Do we need 400 AND 500?

markdurrant (changes requested):
Some minor CSS refactoring suggestions. Otherwise looking good.

packages/ag-charts-website/src/features/api-documentation/components/ApiReference.module.scss
taktran: yea, that is probably not needed

packages/ag-charts-website/src/pages/[framework]/[pageName].astro
Comment (taktran) [reply packages/ag-charts-website/src/pages/[framework]/[pageName].astro]: So `500` is not used, so have removed it

packages/ag-charts-website/src/features/api-documentation/components/ApiReference.module.scss
taktran: 👍🏼 Have moved shared styles to `import styles from '@ag-website-shared/components/reference-documentation/ApiReference.module.scss';` - will need another pass to make sure it's properly aligned between grid/charts

PR #2364, opened 2024-09-02
AG-12489 Ensure styles are added before instantiating modules

jacobp100:
Ensures styles are added to the dom before instantiating modules

### DIFF
--- a/packages/ag-charts-community/src/api/agCharts.ts
+++ b/packages/ag-charts-community/src/api/agCharts.ts
@@ -1,7 +1,7 @@
 import type { AgChartInstance, AgChartOptions, AgFinancialChartOptions, AgGaugeOptions } from 'ag-charts-types';
 
 import { CartesianChart } from '../chart/cartesianChart';
-import { Chart, type ChartExtendedOptions } from '../chart/chart';
+import { Chart, type ChartExtendedOptions, type TransferableResources } from '../chart/chart';
 import { AgChartInstanceProxy, type FactoryApi } from '../chart/chartProxy';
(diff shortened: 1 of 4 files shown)


packages/ag-charts-community/src/api/agCharts.ts
```diff
@@ -225,7 +231,15 @@ class AgChartsInternal {
     }
 
     private static createChartInstance(options: ChartOptions, oldChart?: Chart): Chart {
-        const transferableResource = oldChart?.destroy({ keepTransferableResources: true });
+        let transferableResource: TransferableResources | undefined;
+        if (oldChart != null) {
+            transferableResource = oldChart.destroy({ keepTransferableResources: true });
+        } else {
+            transferableResource = {
```
alantreadway: Why do we need to have this case now?

packages/ag-charts-community/src/chart/toolbar/toolbar.ts
```diff
@@ -755,9 +755,9 @@ export class Toolbar extends BaseModuleInstance implements ModuleInstance {
         }
         this.updateButton(button, options);
 
-        setTimeout(() => {
+        requestAnimationFrame(() => {
```
alantreadway, line 758: Side-note: We need to handle these cases more elegantly, but agree `requestAnimationFrame()` is probably closer to the target than `setTimeout()`.

jacobp100: We need to ensure the styles were applied, and setTimeout might be called before that

alantreadway (approved):
LGTM

PR #2366, opened 2024-09-02
AG-12700 Fix decimal place calculation

jacobp100:
https://ag-grid.atlassian.net/browse/AG-12700


packages/ag-charts-community/src/util/ticks.ts
```diff
@@ -160,7 +160,7 @@ export function tickFormat(ticks: any[], format?: string): (n: number | { valueO
                     if (!Number.isFinite(x) || x === 0) return 0;
                     const l = Math.floor(Math.log10(Math.abs(x)));
                     const digits = options.type ? 6 : 12;
-                    const [_integer, decimal] = x.toExponential(digits - 1).split(/\.|e/g);
+                    const { 1: decimal } = x.toExponential(digits - 1).split(/\.|e/g);
```
olegat: Nit: I think this way is more idiomatic:
```
const decimal = x.toExponential(digits - 1).split(/\.|e/g)[1];
```

olegat (approved):
LGTM aside from nit

PR #2367, opened 2024-09-02
AG-12692 a11y floating toolbar annotation options

olegat:
https://ag-grid.atlassian.net/browse/AG-12692

### DIFF
--- a/packages/ag-charts-enterprise/src/components/dialog/dialog.ts
+++ b/packages/ag-charts-enterprise/src/components/dialog/dialog.ts
@@ -8,7 +8,7 @@ const { Vec2 } = _Util;
 
 export interface DialogOptions extends PopoverOptions {}
 
-interface ButtonGroupOptions {
+interface RadioGroupOptions {
     label: string;
     options: Array<{ icon: AgIconName; altText: string; value: string }>;
     value: string;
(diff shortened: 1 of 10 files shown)


packages/ag-charts-enterprise/src/components/dialog/dialog.ts
```diff
@@ -72,28 +72,42 @@ export abstract class Dialog<Options extends DialogOptions = DialogOptions> exte
...
 
         for (const button of options) {
             const buttonEl = createElement('button', `ag-charts-dialog__button`);
             const iconEl = createElement('span', this.ctx.domManager.getIconClassNames(button.icon));
+            buttonEl.role = 'radio';
+            buttonEl.ariaLabel = this.ctx.localeManager.t(button.altText ?? '');
+            buttonEl.ariaChecked = 'false';
+            iconEl.ariaHidden = 'true';
             if (button.value === value) {
                 buttonEl.classList.add(activeClass);
```
lsjroberts, line 91: I think this needs `buttonEl.ariaChecked = 'true';` in this if block, rather than the `buttons[0]` check below?

lsjroberts (approved):
Awesome, LGTM with one small issue

packages/ag-charts-enterprise/src/components/dialog/dialog.ts
olegat: Good suggestion. Done https://github.com/ag-grid/ag-charts/commit/8b172f2527ff39edc32a900605b0681e3b34de15

PR #2369, opened 2024-09-03
AG-12711 Fix download state updates not being applied

jacobp100:
https://ag-grid.atlassian.net/browse/AG-12711


lsjroberts (commented):
LGTM, good catch

PR #2371, opened 2024-09-03
AG-12078 Add history manager with undo/redo of annotations

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12078


lsjroberts:
post snapshot update build https://github.com/ag-grid/ag-charts/actions/runs/10684262739

packages/ag-charts-enterprise/src/components/popover/popover.ts
```diff
@@ -36,6 +36,7 @@ export abstract class Popover<Options extends PopoverOptions = PopoverOptions>
     }
 
     public hide() {
+        if (this.element.children.length === 0) return;
```
olegat, line 41: Is this intentional? It feels redundant because `this.hideFns.forEach((fn) => fn());` should theoretically have no side-effect if there are not children.

olegat (approved):
LGTM (just minor feedback). Thanks for adding an e2e test too. Well done! :-)

packages/ag-charts-enterprise/src/components/popover/popover.ts
lsjroberts: It's due to annotations eagerly calling `colorPicker.hide()` whenever an annotation is clicked.

It does this because when you mouse down it fires `drag-start` and goes into the dragging state, then when you mouse up it fires `drag-end` and goes back into the default idle state.

Whenever it enters the idle start, it resets the stage by hiding overlays like the color picker and clearing other things.

Possibly, it shouldn't call hide overlays so eagerly, but this was the simpler fix.

olegat: Okay that makes sense. SGTM.
It might be wise add this explanation in a comment.

PR #2376, opened 2024-09-03
AG-12501 - Don't allow dragging via midline if it's not visible

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-12501


lsjroberts (approved):
LGTM

PR #2377, opened 2024-09-03
AG-12690 Set tabindex=0 on first menu item

olegat:
https://ag-grid.atlassian.net/browse/AG-12690

Depends on https://github.com/ag-grid/ag-charts/pull/2367


lsjroberts (approved):
LGTM

PR #2379, opened 2024-09-03
AG-12693 Compute BBox of custom shape

olegat:
https://ag-grid.atlassian.net/browse/AG-12693


packages/ag-charts-community/src/util/bezier.ts
```diff
@@ -0,0 +1,46 @@
+// Find the extreme points where the derivative is zero
+export function calculateDerivativeExtrema(p0: number, p1: number, p2: number, p3: number): number[] {
```
jacobp100: Nice!

We have a similar bezier file at packages/ag-charts-community/src/scene/util/bezier.ts . Not too bothered where the file ends up, but would be good to merge them

olegat: Done.

packages/ag-charts-community/src/chart/marker/marker.ts
```diff
@@ -6,6 +6,20 @@ import { Rotatable, Scalable, Translatable } from '../../scene/transformable';
 
 export type MarkerPathMove = { x: number; y: number; t?: 'move' };
 
+const BUILTIN_MARKERS: readonly string[] = [
+    'ArrowDown',
```
jacobp100, line 10: Minor - but couldn't we store the actual constructors and do referential equality so it's impossible to fool the check?

packages/ag-charts-community/src/scene/extendedPath2D.ts
```diff
@@ -374,4 +377,80 @@ export class ExtendedPath2D {
...
+                    const criticalAngles = [0, Math.PI / 2, Math.PI, (3 * Math.PI) / 2];
+                    for (const crit of criticalAngles) {
+                        if ((A0 < A1 && A0 <= crit && crit <= A1) || (A0 > A1 && (A0 <= crit || crit <= A1))) {
+                            joinAngle(crit);
+                        }
+                    }
+                    break;
+                }
+                case Command.ClosePath:
+                    break;
```
jacobp100, line 455: Do you need to reset sx, sy here?

jacobp100 (approved):

packages/ag-charts-community/src/scene/extendedPath2D.ts
```diff
@@ -374,4 +377,80 @@ export class ExtendedPath2D {
...
+
+                    const joinAngle = (angle: number, updatestart?: boolean) => {
+                        const px = cx + r * Math.cos(angle);
+                        const py = cy + r * Math.sin(angle);
+                        joinPoint(px, py, updatestart);
+                    };
+                    joinAngle(A0);
+                    joinAngle(A1, true);
+                    const criticalAngles = [0, Math.PI / 2, Math.PI, (3 * Math.PI) / 2];
+                    for (const crit of criticalAngles) {
```
jacobp100, line 446: Does this handle a clockwise arc where `A0` is 5 and `A1` is 1?

packages/ag-charts-community/src/chart/marker/marker.ts
olegat: Technically the base class shouldn't know anything about the derived classes, so this code is sort of hacky anyway. It might be a better idea to create a `BuiltinMarker` subclass of `Marker` which our markers can extend.

packages/ag-charts-community/src/scene/extendedPath2D.ts
olegat: Correct. I'll fix that.

packages/ag-charts-community/src/scene/extendedPath2D.ts
olegat: I haven't tested it, but in theory yes is should add critical angle 0 because the 2nd condition of the OR logic would expand to `true`:

Deconstruction of the `if` logic:
```
(A0 < A1 && A0 <= crit && crit <= A1) || (A0 > A1 && (A0 <= crit || crit <= A1))
( 5 < 1  && A0 <= crit && crit <= A1) || ( 5 > 1  && (A0 <= crit || crit <= A1))
( false  && A0 <= crit && crit <= A1) || ( true   && (A0 <= crit || crit <= A1))
                                false || ( true   && (A0 <= crit || crit <= A1))
                                false || ( true   && ( 5 <= 0    ||    0 <= 1))
                                false || ( true   && ( false     ||    true  ))
                                false || ( true   &&  true )
                                false || true
                                true
```

packages/ag-charts-community/src/chart/marker/marker.ts
olegat: I've tried improving this logic, but it causes failures in the unit/e2e tests because the `BBox` computation break. In the interest of not diverting from `latest` too much, let's keep this as-is for now.

PR #2381, opened 2024-09-03
AG-12600 Change inset & offset of text on linear drawings

lsjroberts:


manapeirov (approved):
LGTM

PR #2383, opened 2024-09-03
Linear gauge

jacobp100:


jacobp100:
Merging with intent to cleanup along with radial gauge

PR #2384, opened 2024-09-03
Ag 12068/arrow shape annotations

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-12068


packages/ag-charts-enterprise/src/features/annotations/annotationsModule.ts
```diff
@@ -136,11 +136,11 @@ export const AnnotationsModule: _ModuleSupport.Module = {
             },
             'arrow-up': {
                 fill: _Theme.PALETTE_UP_FILL,
-                handle: { ...handle },
+                handle: { ...handle, stroke: _Theme.DEFAULT_ANNOTATION_COLOR },
```
lsjroberts, line 139: Ah yeh, the stroke is not in `handle` cos the other annotations take it from their line stroke and pass it into the handle. My bad, sorry!

lsjroberts (approved):
Great, LGTM

PR #2388, opened 2024-09-04
AG-12437 Extract out childPropertiesButton to new component

kylerphillips:
https://github.com/user-attachments/assets/5792d1a8-f5ff-4740-be14-e8a4c1fe26b3


packages/ag-charts-website/src/features/api-documentation/components/ApiReference.tsx
```diff
@@ -29,6 +29,7 @@ import {
 } from '../apiReferenceHelpers';
 import { SelectionContext } from './OptionsNavigation';
 import { type CollapsibleType, PropertyTitle, PropertyType } from './Properties';
+import { ChildPropertiesButton } from './Properties';
```
taktran: Since `ChildPropertiesButton` is only used in this file, you could move it in here, and rename `CollapsibleButton` in `Properties.tsx` to `CodeCollapsibleButton`

taktran (approved):
Looks good, just a minor comment

PR #2391, opened 2024-09-04
AG-12601 Line text settings dialog styles and color picker

lsjroberts:

Note: more comprehensive handling of keyboard control and a11y to come at a later date.


lsjroberts:
Post snapshot update build https://github.com/ag-grid/ag-charts/actions/runs/10719192632

packages/ag-charts-community/src/dom/theme.css
```diff
@@ -18,11 +18,11 @@
     --ag-charts-border-critical: var(--ag-borders-critical, solid 1px);
 
     /* Layout */
-    --ag-charts-layer-ui-overlay: 5;
-    --ag-charts-layer-tooltip: 4;
-    --ag-charts-layer-toolbar: 3;
-    --ag-charts-layer-crosshair: 2;
-    --ag-charts-layer-annotations: 1;
+    --ag-charts-layer-ui-overlay: 50;
```
jacobp100: What's the reason for the increase?

packages/ag-charts-enterprise/src/components/dialog/dialog.ts
```diff
@@ -28,29 +30,57 @@ interface TextAreaOptions {
...
         const popover = super.showWithChildren(children, options);
         popover.classList.add('ag-charts-dialog');
         popover.setAttribute('role', 'dialog');
 
-        // Give the dialog's dimensions a chance to be calculated before positioning
-        requestAnimationFrame(() => {
-            const bbox = this.ctx.domManager.getBoundingClientRect();
-            const position = Vec2.sub(Vec2.sub(Vec2.from(bbox), Vec2.from(popover)), Dialog.offset);
-            this.updatePosition(position);
+        popover.addEventListener('mousedown', (event) => {
```
jacobp100, line 64: The popover should probably stop propagation of events so you don't need to detect this here

jacobp100: Or is this just a button click?

packages/ag-charts-community/src/dom/theme.css
lsjroberts: I needed to nest a color picker within a dialog popover, so it adds 1 to the layer when nesting. Wanted to avoid conflicts with other layers.

packages/ag-charts-enterprise/src/components/popover/popover.css
```diff
@@ -1,10 +1,12 @@
...
+
     position: absolute;
     border: var(--ag-charts-popover-border);
     border-radius: var(--ag-charts-popover-border-radius);
     background: var(--ag-charts-popover-background-color);
     color: var(--ag-charts-popover-foreground-color);
     font-family: var(--ag-charts-font-family);
     font-size: var(--ag-charts-popover-font-size);
-    z-index: var(--ag-charts-layer-ui-overlay);
+    z-index: calc(var(--ag-charts-layer-ui-overlay) + var(--popover-layer));
```
jacobp100: If we allow providing a container to attach to the popover so the popover is a child of the dialog, we can avoid this z-index stuff

packages/ag-charts-enterprise/src/components/dialog/dialog.ts
lsjroberts: Yeh, the color picker button triggers opening the color picker, and I want to hide it again if you click anywhere within the dialog. Needed to block it from immediately hiding the color picker when you click the button the first time, but thinking about it, there may be a better approach.

lsjroberts: I looked into only attaching the listener when the color picker is opened, but that wasn't any cleaner really.

I think I'll come back to this when reviewing all the a11y and keyboard controls.

jacobp100 (approved):

PR #2395, opened 2024-09-05
AG-12740 Fix integrated area series animation

jacobp100:
https://ag-grid.atlassian.net/browse/AG-12740


packages/ag-charts-community/src/chart/series/cartesian/lineInterpolationUtil.ts
```diff
@@ -35,10 +36,18 @@ interface SpanIndices {
     datumIndex: number;
 }
 
+function axisValue(value: any) {
+    if (isPlainObject(value) && Object.hasOwn(value, 'id')) {
```
alantreadway: I have a concern this might come into play in non-integrated cases - also we have some similar handling here that we might want to consolidate?
https://github.com/ag-grid/ag-charts/blob/f214d67fb4d6496a400c40b278c6796e8ee679ba/packages/ag-charts-community/src/util/value.ts#L30

alantreadway (approved):
LGTM

PR #2398, opened 2024-09-05
Ag 12660/crosshair axis button

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-12660


packages/ag-charts-enterprise/src/features/crosshair/crosshair.ts
```diff
@@ -196,12 +193,12 @@ export class Crosshair extends _ModuleSupport.BaseModuleInstance implements _Mod
...
 
-        if (event.type === 'hover' && this.visible && hoverRect.containsPoint(offsetX, offsetY)) {
-            crosshairGroup.visible = true;
-
+        if (hoverRect.containsPoint(offsetX, offsetY)) {
             const lineData = this.getData(event);
 
             this.updatePositions(lineData);
+
+            crosshairGroup.visible = true;
```
lsjroberts, line 203: In `hideCrosshairs()` it also hides the labels, do they need to be made visible here too or is that handled elsewhere?

lsjroberts (approved):
LGTM, one question, but merge when you're happy

packages/ag-charts-enterprise/src/features/crosshair/crosshair.ts
manapeirov: so that's done in the updatePositions function

PR #2403, opened 2024-09-06
AG-12078 Add undo/redo to clear all and line text edits

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12078

### DIFF
--- a/packages/ag-charts-enterprise/src/features/annotations/annotationsStateMachine.ts
+++ b/packages/ag-charts-enterprise/src/features/annotations/annotationsStateMachine.ts
@@ -199,6 +199,18 @@ export class AnnotationsStateMachine extends StateMachine<States, AnnotationType
             ctx.update();
         };
 
+        const actionSaveText = ({ textInputValue }: { textInputValue?: string }) => {
+            const datum = ctx.datum(this.active!);
+            if (textInputValue != null && textInputValue.length > 0) {
+                datum?.set({ text: textInputValue });
+                ctx.update();
+                ctx.recordAction(`Change ${datum?.type} annotation text`);
+            } else {
+                ctx.delete(this.active!);
+                ctx.recordAction(`Delete ${datum?.type} annotation`);
+            }
+        };
+
         const guardActive = () => this.active != null;
         const guardActiveHasLineText = () => {
             if (this.active == null) return false;
@@ -291,6 +303,7 @@ export class AnnotationsStateMachine extends StateMachine<States, AnnotationType
                         }
                         datum.text.set(props);
                         ctx.update();
+                        ctx.recordAction(`Update ${datum.type} text`);
                     },
                 },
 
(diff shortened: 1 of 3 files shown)


packages/ag-charts-enterprise/src/features/color-picker/colorPicker.ts
```diff
@@ -91,10 +91,11 @@ export class ColorPicker extends AnchoredPopover<ColorPickerOptions> {
                 colorInput.value = colorString.toUpperCase();
             }
 
+            if (!trackChange) return;
+
+            this.hasChanged = true;
             const plainColor = Color.fromHSB(h, s, v, 1).toHexString();
             opts.onChange?.(colorString, plainColor, a);
-
-            if (trackChange) this.hasChanged = true;
```
lsjroberts, line 97: Previously, this would cause a ghost action to the history manager for undo/redo.

packages/ag-charts-enterprise/src/features/color-picker/colorPicker.ts
```diff
@@ -91,8 +91,10 @@ export class ColorPicker extends AnchoredPopover<ColorPickerOptions> {
                 colorInput.value = colorString.toUpperCase();
             }
 
-            const plainColor = Color.fromHSB(h, s, v, 1).toHexString();
-            opts.onChange?.(colorString, plainColor, a);
+            if (trackChange || opts.color == null) {
+                const plainColor = Color.fromHSB(h, s, v, 1).toHexString();
+                opts.onChange?.(colorString, plainColor, a);
+            }
```
lsjroberts, line 97: Previously, this would always fire the onChange when it immediately opened creating a ghost action in the history manager for undo/redo.

jacobp100 (approved):

packages/ag-charts-enterprise/src/features/annotations/annotationsStateMachine.ts
```diff
@@ -199,6 +199,18 @@ export class AnnotationsStateMachine extends StateMachine<States, AnnotationType
             ctx.update();
         };
 
+        const actionSaveText = ({ textInputValue }: { textInputValue?: string }) => {
+            const datum = ctx.datum(this.active!);
+            if (textInputValue != null && textInputValue.length > 0) {
```
olegat, line 204: Would clearing the text (i.e. empty string) delete the annotation entirely? This seems undesirable. I would expect the annotation to only be removed when the trashcan icon is clicked.

olegat (approved):
LGTM. Just one question but I think the risk is low.

packages/ag-charts-enterprise/src/features/annotations/annotationsStateMachine.ts
lsjroberts: Yeah, that is the requirement from product. Previously only pressing enter would delete it, this change makes this the same for clicking off as well.

PR #2404, opened 2024-09-06
AG-NONE update e2e darwin snapshots

olegat:


olegat:
These snapshots are incorrect (they used an outdated ag-charts library)

PR #2405, opened 2024-09-06
AG-12606 Translate label key

olegat:
https://ag-grid.atlassian.net/browse/AG-12606


olegat:
Trivial change. Skipping review.

PR #2407, opened 2024-09-06
Add support for conic gradients to node canvas

jacobp100:
Required for gallery examples


libraries/ag-charts-test/src/canvas/conicGradient.ts
```diff
@@ -0,0 +1,94 @@
...
+        const rgba = Array.from(input.matchAll(/[\d.]+/g), (match) => Number(match[0]));
+
+        if (rgba.length === 3) {
+            return [rgba[0], rgba[1], rgba[2]];
+        } else if (rgba.length === 4) {
+            return [rgba[0], rgba[1], rgba[2], rgba[3]];
+        }
+    }
+
+    throw new Error(`Failed to parse "${input}"`);
```
olegat, line 34: We may want to also include the 140 named colour:
https://www.w3schools.com/cssref/css_colors.php

packages/ag-charts-community/src/scene/gradient/gradient.ts
```diff
@@ -57,6 +57,11 @@ export abstract class Gradient {
             c0 = c1;
         }
 
+        if ('createPattern' in gradient) {
+            // Node-canvas stubs
+            gradient = (gradient as any).createPattern();
```
olegat: Nit: Can we avoid the `as any` by specifying an explicit type for `gradient`?

Example:
```
let gradient: { createPattern?: ()=>unknown } = this.createCanvasGradient(ctx, bbox);
```

olegat (approved):
LGTM with 2 nits

libraries/ag-charts-test/src/canvas/conicGradient.ts
jacobp100: Possibly. I've copied this from color.ts, since I can't import that file. Due to the way we do gradients internally, I only really need to handle what outputs `color.ts` generates

packages/ag-charts-community/src/scene/gradient/gradient.ts
jacobp100: Updated

PR #2416, opened 2024-09-09
Cleanup dead feature

iMoses:
Simplify interface + reduce coupling


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -256,7 +256,7 @@ export abstract class Chart extends Observable {
         const container = resources?.container;
 
         const root = new Group({ name: 'root' });
-        const titleGroup = new Group({ name: 'titles', layer: true, zIndex: Layers.SERIES_LABEL_ZINDEX });
```
alantreadway: Can we keep this, it's useful for debugging?

packages/ag-charts-enterprise/src/features/status-bar/statusBar.ts
```diff
@@ -252,7 +252,7 @@ export class StatusBar
         this.labelGroup.visible = false;
 
         this.destroyFns.push(
-            ctx.scene.attachNode(this.labelGroup, 'titles'),
+            ctx.scene.attachNode(this.labelGroup),
```
alantreadway, line 260: This was hacked in to fix https://ag-grid.atlassian.net/browse/CRT-394 - can you verify that this case still works after this change?

iMoses: Verified and works as expected

alantreadway (approved):

PR #2425, opened 2024-09-10
AG-12700 Fix infinite loops

jacobp100:
https://ag-grid.atlassian.net/browse/AG-12700

Infinite loop when running `range(start, end, step=0)`

Added some more robust checks

### DIFF
--- a/packages/ag-charts-community/src/util/ticks.ts
+++ b/packages/ag-charts-community/src/util/ticks.ts
@@ -114,7 +114,7 @@ export function getTickInterval(
 
 export function tickStep(start: number, end: number, count: number, minCount = 0, maxCount = Infinity): number {
     if (start === end) {
-        return Math.min(Math.max(1, minCount), maxCount);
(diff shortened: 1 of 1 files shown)


packages/ag-charts-community/src/util/ticks.ts
```diff
@@ -114,9 +114,8 @@ export function getTickInterval(
 
 export function tickStep(start: number, end: number, count: number, minCount = 0, maxCount = Infinity): number {
     if (start === end) {
-        return 1;
-    }
-    if (count < 1) {
+        return Math.min(Math.max(1, minCount), maxCount);
```
iMoses:
```suggestion
        return clamp(1, minCount, maxCount);
```
nit: use `clamp`, it improved readability

iMoses (approved):
LGTM, left one nit comment

alantreadway (approved):
LGTM

PR #2418, opened 2024-09-09
AG-12776 Use `:focus-visible` CSS selector for focus indicator

olegat:
https://ag-grid.atlassian.net/browse/AG-12776

Depends on https://github.com/ag-grid/ag-charts/pull/2427

PR #2427, opened 2024-09-10
AG-12810 Fix legend item CSS bounds

olegat:
https://ag-grid.atlassian.net/browse/AG-12810


packages/ag-charts-community/src/chart/legend.ts
```diff
@@ -988,6 +993,30 @@ export class Legend extends BaseProperties {
...
         return closestLeftTop.datum;
     }
+    private getLegendItemHitbox(legendItemNode: { getBBox(): BBox }): BBox {
+        const bbox = legendItemNode.getBBox().clone();
+        bbox.grow(this.item.paddingX / 2, 'horizontal');
+        bbox.grow(this.item.paddingY / 2, 'vertical');
+        return bbox;
+    }
+    private getLegendItemHitboxProvider(legendItemNode: LegendMarkerLabel): BBoxProvider<BBox> {
+        // FIXME: This method is required for the `focusable` property of `createProxyElement`, but the `focusable` property should eventually be removed because it's only required to update the FocusIndicator
```
olegat, line 1003: Note: This will happen when https://github.com/ag-grid/ag-charts/pull/2418 is ready

packages/ag-charts-community/src/chart/legend.ts
```diff
@@ -726,7 +727,10 @@ export class Legend extends BaseProperties {
     }
 
     private updateItemProxyButtons() {
-        this.itemSelection.each((l) => setElementBBox(l.proxyButton?.listitem, Transformable.toCanvas(l)));
+        this.itemSelection.each((l) => {
+            const bbox = Transformable.toCanvas(l.parent!, this.getLegendItemHitbox(l));
```
alantreadway, line 731: Nit: Maybe either pass `this.group` (which should be the same as `l.parent`), or `l` as the first parameter to avoid the need to use `!`?

packages/ag-charts-community/src/chart/legend.ts
```diff
@@ -988,6 +993,30 @@ export class Legend extends BaseProperties {
...
+            get visible() {
+                return legendItemNode.visible;
+            },
+            toCanvasBBox() {
+                return Transformable.toCanvas(legendItemNode.parent!, legend.getLegendItemHitbox(legendItemNode));
+            },
+            fromCanvasPoint(x: number, y: number) {
+                return Transformable.fromCanvasPoint(legendItemNode!, x, y);
+            },
+        };
```
alantreadway: Nit: Maybe it would be better to add another sub-class of `NodeRegionBBoxProvider` to `regions.ts` which allows an alternative nodes coordinate space to be used, as `legend.ts` is already quite bloated?

packages/ag-charts-community/src/chart/legend.ts
olegat: Sure I'll change it. But in all honesty, I feel like the behaviour of `Transformable.toCanvas` is counter-intuitive. I would have expected `Transformable.toCanvas(l, this.getLegendItemHitbox(l))` to compute the correct bounding box.

packages/ag-charts-community/src/chart/legend.ts
olegat: I don't think this nit matters. This code only exists to provide a `focusable` parameter to `ProxyInteractionManager` interface. This `focusable` parameter is used to update the `FocusIndicator` bounds. But we're going to phase out some of `FocusIndicator` in favour of the [:focus-visible](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) pseudo-class which we're already using in financial charts interactive elements.

packages/ag-charts-community/src/chart/legend.ts
alantreadway: There is unfortunately still a little nuance to this stuff, mostly because of the need for `Node.getBBox()` to be in the local coordinate space of the parent element to make sense.

alantreadway (approved):
LGTM

PR #2428, opened 2024-09-10
Ag 12114/copy paste annotations

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-12114


packages/ag-charts-community/src/util/stateMachine.ts
```diff
@@ -115,6 +115,13 @@ export class StateMachine<State extends string, Event extends string> {
         }
     }
 
+    transitionAsync(event: Event, data?: any) {
+        // TODO: consider merging this with transition
+        setTimeout(() => {
+            this.transition(event, data);
+        }, 0);
+    }
```
lsjroberts, line 123: @alantreadway I discussed this with Mana.

The use-case of this was to ensure the transition to the new state happens after the current tick, since both the original tick and the transition were manipulating the `postUpdateFns` array in `annotations.ts`.

It is perhaps prefereable that all transitions should be asynchronous (and queued), but that's wayyy too risky to do now.

packages/ag-charts-enterprise/src/features/annotations/annotations.ts
```diff
@@ -476,6 +485,34 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
...
+    private createAnnotationDatumCopy({
+        node,
+        datum,
+    }: {
+        node: AnnotationScene;
+        datum: AnnotationProperties;
+    }): AnnotationProperties | undefined {
+        const { type } = datum;
+        if (type in annotationConfigs) {
+            const config = annotationConfigs[type];
```
lsjroberts, line 497: You can save some nesting here by inverting the condition:

```suggestion
        if (!(type in annotationConfigs)) {
            throw new Error(`AG Charts - ...`);
        }
        const config = annotationConfigs[type];
```

packages/ag-charts-enterprise/src/features/annotations/annotations.ts
```diff
@@ -484,6 +521,24 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
         this.update();
     }
 
+    private createAnnotation(type: AnnotationType, datum: AnnotationProperties, applyDefaults: boolean = true) {
```
lsjroberts, line 524: nit: This should be above the `onRestoreAnnotations` function, to keep it next to the other `create...` functions.

packages/ag-charts-enterprise/src/features/annotations/cross-line/crossLineScene.ts
```diff
@@ -196,6 +196,29 @@ export class CrossLineScene extends AnnotationScene {
...
+        const yOffset = isHorizontal ? offset.y : 0;
+        const xOffset = isHorizontal ? 0 : offset.x;
+
+        const point = invertCoords({ x: coords.x1 + xOffset, y: coords.y1 + yOffset }, context);
+
+        copiedDatum.set({ value: isHorizontal ? point.y : point.x });
+
+        return copiedDatum;
+    }
     override getCursor() {
```
lsjroberts, line 222:
```suggestion
    }
    
    override getCursor() {
```

packages/ag-charts-enterprise/src/features/annotations/scenes/linearScene.ts
```diff
@@ -83,18 +83,40 @@ export abstract class LinearScene<
...
+        context,
+    }: {
+        datum: Datum;
+        start: Coords;
+        end: Coords;
+        translation: _Util.Vec2;
+        context: AnnotationContext;
+    }) {
+        const a = Vec2.add(start, translation);
+        const b = Vec2.add(end, translation);
```
lsjroberts, line 109: nit: I guess these could be `const translatedStart`/`const translatedEnd` if you want

packages/ag-charts-community/src/util/stateMachine.ts
alantreadway: Agree, we should probably do a sweep post release for `setTimeout()`, `requestAnimationFrame()` and try to work out if there is a common pattern/approach we should adopt.

packages/ag-charts-enterprise/src/features/annotations/annotations.ts
```diff
@@ -476,6 +485,34 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
...
+
+            const newDatum = new config.datum();
+            newDatum.set(datum.toJson());
+
+            const context = this.getAnnotationContext();
+            if (!context) {
+                return;
+            }
+
+            const offset = { x: -30, y: -30 }; // just above and to the left
```
lsjroberts, line 507: Since we're not using this offset on a bunch of these scenes, it might be better to not pass it down at all and instead calculate it inside each scene.

For some it will may just be the `-30,-30`, but I can see us wanting different numbers for different text annotation types and also different point scenes that have different sized shapes.

packages/ag-charts-enterprise/src/features/annotations/parallel-channel/parallelChannelScene.ts
```diff
@@ -150,6 +150,17 @@ export class ParallelChannelScene extends ChannelScene<ParallelChannelProperties
...
     }
 
+    override copy(
+        datum: ParallelChannelProperties,
+        copiedDatum: ParallelChannelProperties,
+        context: AnnotationContext,
+        offset: Coords
+    ) {
+        super.copy(datum, copiedDatum, context, offset);
+        copiedDatum.height = datum.height;
```
lsjroberts, line 160: I think this is covered by the `copiedDatum.set(datum.getJson()` call previously? In which case this whole override can be deleted?

alantreadway (approved):
Nice work - LGTM

PR #2433, opened 2024-09-10
AG-12688 - Force series region to have seriesArea bbox, fixing line + area tooltips.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-12688

Fixes series-area region bbox calculation, so that `Chart.seriesRect` is used as a fallback - this fixes series-area hover detection for path-based series when no markers or grid-lines are present.

### DIFF
--- a/packages/ag-charts-community/src/chart/interaction/regionManager.ts
+++ b/packages/ag-charts-community/src/chart/interaction/regionManager.ts
@@ -52,9 +52,9 @@ function addHandler<T extends RegionEvent['type']>(
     );
 }
 
-type RegionNodeType = Node | { id: string; node: Node };
+type RegionNodeType = NodeRegionBBoxProvider | Node | { id: string; node: Node };
 
-function nodeToBBoxProvider(node: RegionNodeType | NodeRegionBBoxProvider) {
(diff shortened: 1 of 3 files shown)


packages/ag-charts-community/src/chart/interaction/regionManager.ts
```diff
@@ -54,10 +54,13 @@ function addHandler<T extends RegionEvent['type']>(
 
 type RegionNodeType = Node | { id: string; node: Node };
 
-function nodeToBBoxProvider(node: RegionNodeType) {
+function nodeToBBoxProvider(node: RegionNodeType | NodeRegionBBoxProvider) {
```
olegat: nit: redeclare `| NodeRegionBBoxProvider` as part of the `type RegionNodeType`. This will avoid the need of retyping `| NodeRegionBBoxProvider` in `RegionManager` methods.

olegat (approved):
LGTM just 1 nit

packages/ag-charts-community/src/chart/interaction/regionManager.ts
alantreadway: Fixed - thanks!

PR #2435, opened 2024-09-10
AG-12498 Fix extending channels into corners

lsjroberts:

Fixes this situation by drawing extra points to extend the background rect to the appropriate corners:

Screenshot 2024-09-10 at 16 11 00


alantreadway (approved):
LGTM

PR #2442, opened 2024-09-11
AG-12498 Settings dialog fixes

lsjroberts:

### DIFF
--- a/packages/ag-charts-enterprise/src/components/dialog/dialog.ts
+++ b/packages/ag-charts-enterprise/src/components/dialog/dialog.ts
@@ -12,6 +12,7 @@ const {
     createSelect,
     createTextArea,
     initRovingTabIndex,
+    getWindow,
     mapValues,
 } = _ModuleSupport;
 const { Vec2 } = _Util;
@@ -80,7 +81,7 @@ export abstract class Dialog<Options extends DialogOptions = DialogOptions> exte
         popover.addEventListener('keydown', this.onKeyDown.bind(this));
 
         // Give the dialog's dimensions a chance to be calculated before positioning
-        _ModuleSupport.getWindow().requestAnimationFrame(() => this.reposition());
+        getWindow().requestAnimationFrame(() => this.reposition());
 
         this.colorPicker.attachTo(this);
 
--- a/packages/ag-charts-enterprise/src/components/dialog/dialog.ts
+++ b/packages/ag-charts-enterprise/src/components/dialog/dialog.ts
@@ -59,24 +59,11 @@ export abstract class Dialog<Options extends DialogOptions = DialogOptions> exte
     private colorPickerAnchorElement?: HTMLElement;
     private dragStartState?: { client: _Util.Vec2; position: _Util.Vec2 };
     private seriesRect?: _Scene.BBox;
-    private readonly escapeHandler: (event: KeyboardEvent) => unknown;
     private initialFocus?: HTMLElement;
 
     constructor(ctx: _ModuleSupport.ModuleContext, id: string) {
         super(ctx, id);
         this.destroyFns.push(ctx.layoutManager.addListener('layout:complete', this.onLayoutComplete.bind(this)));
-        this.escapeHandler = (event: KeyboardEvent) => {
-            if (
-                !event.altKey &&
-                !event.ctrlKey &&
-                !event.ctrlKey &&
-                !event.metaKey &&
-                !event.isComposing &&
-                event.key === 'Escape'
-            ) {
-                this.hide();
-            }
-        };
     }
 
     protected override showWithChildren(children: Array<HTMLElement>, options: Options) {
@@ -90,6 +77,7 @@ export abstract class Dialog<Options extends DialogOptions = DialogOptions> exte
             if ((event.target as any).classList?.contains('ag-charts-dialog__color-picker-button')) return;
             this.colorPicker.hide();
         });
+        popover.addEventListener('keydown', this.onKeyDown.bind(this));
 
         // Give the dialog's dimensions a chance to be calculated before positioning
         _ModuleSupport.getWindow().requestAnimationFrame(() => this.reposition());
@@ -201,8 +189,6 @@ export abstract class Dialog<Options extends DialogOptions = DialogOptions> exte
                 buttonEl.classList.add(activeClass);
             }
 
-            buttonEl.addEventListener('keydown', this.escapeHandler);
-
             group.appendChild(buttonEl);
             buttons.push(buttonEl);
         }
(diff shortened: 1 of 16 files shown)


packages/ag-charts-enterprise/src/components/dialog/dialog.ts
```diff
@@ -59,26 +61,16 @@ export abstract class Dialog<Options extends DialogOptions = DialogOptions> exte
...
-                !event.altKey &&
-                !event.ctrlKey &&
-                !event.ctrlKey &&
-                !event.metaKey &&
-                !event.isComposing &&
-                event.key === 'Escape'
-            ) {
-                this.hide();
-            }
-        };
```
lsjroberts, line 78: I needed to add escape handling to the `Line` & `Text` tab buttons, so added `this.escapeHandler` to those buttons.

But it looks like these can all be handled with a single listener on the popover rather than individual ones for each element. Previously it didn't work as it was on the interaction manager which has some propagation blocks, but this appears to be fine unless I'm missing something?

packages/ag-charts-enterprise/src/features/color-picker/colorPicker.ts
```diff
@@ -47,15 +42,9 @@ export class ColorPicker extends AnchoredPopover<ColorPickerOptions> {
...
-        const popover = this.showWithChildren([element], options);
+        const popover = this.showWithChildren([element], { initialFocus, ...options });
         popover.classList.add('ag-charts-color-picker');
         popover.setAttribute('role', 'dialog');
-
-        const { type, lastFocus } = this.ctx.focusIndicator.guessDevice(options.sourceEvent);
-        if (type === 'keyboard' && lastFocus !== undefined) {
-            initialFocus.focus();
-            this.lastFocus = lastFocus;
-        }
```
lsjroberts, line 58: These focus changes have been moved to the parent `Popover` class, and are only active if a `sourceEvent` and `initialFocus` options are passed in. The `Dialog` popover sets the `initialFocus` to the first tab if `Dialog.createTabs()` is called.

alantreadway (approved):
LGTM, let's get feedback from @olegat and @markdurrant before merge.

markdurrant (approved):
All CSS stuff LGTM 👍

packages/ag-charts-enterprise/src/components/dialog/dialog.ts
olegat: Due to some tech debt, the interaction manager will not always dispatch keyboard events. Example:
https://github.com/ag-grid/ag-charts/blob/5b8622c70a3e43284de6726a2af88af140ed0bb1/packages/ag-charts-community/src/chart/interaction/interactionManager.ts#L296

However I think that adding the keyboard handler to the popover ancestor is safe 👍

packages/ag-charts-enterprise/src/components/dialog/dialog.ts
```diff
@@ -142,6 +144,10 @@ export abstract class Dialog<Options extends DialogOptions = DialogOptions> exte
         tabs[initial].onShow?.();
         tabButtons[initial].classList.add('ag-charts-dialog__tab-button--active');
 
+        initRovingTabIndex({ orientation: 'horizontal', buttons: Object.values(tabButtons) });
+
+        this.initialFocus = tabButtons[initial];
```
olegat, line 157: Check with David & Kiril to confirm that the first tab is the desired initial focus. I vaguely remember that the text field was the desired initial focus.

packages/ag-charts-enterprise/src/components/dialog/dialog.ts
```diff
@@ -114,6 +107,10 @@ export abstract class Dialog<Options extends DialogOptions = DialogOptions> exte
     ) {
         const element = createElement('div', 'ag-charts-dialog__tabs');
 
+        for (const tab of Object.values(tabs)) {
+            tab.content.role = 'tabpanel';
```
olegat: Although not strictly required, the WAI-ARIA recommendation is to explicitly specify the association between the `tab` and `tabpanel` using `aria-controls` and `aria-labelledby` respectively.
https://www.w3.org/TR/wai-aria/#tabpanel

In HTML code it would look something like this:

```html
<div id="mytab" role="tab" aria-controls="mytabpanel" ...> ... </div>
<div id="mytabpanel" role="tabpanel" aria-labelledby="mytab" ...> ... </div>

```

Reference example: https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/

However, this requires `id` attributes on the HTML elements. Unfortunately we don't yet have a standardised way to generate and set IDs. Therefore use your best judgment on whether to follow this recommendation or not.

packages/ag-charts-enterprise/src/components/popover/popover.ts
```diff
@@ -43,6 +46,11 @@ export abstract class Popover<Options extends PopoverOptions = PopoverOptions>
         }
         this.element.setAttribute('role', 'presentation');
 
+        this.hideFns.push(() => {
```
olegat: Optional: this hide logic could just be moved directly into the `hide()` method rather than appending a lambda

olegat (approved):
LGTM. Some recommended WAI-ARIA changes could be made but not a blocker

packages/ag-charts-enterprise/src/components/dialog/dialog.ts
lsjroberts: There's an outstanding item to make it open the text tab if you double click the text label, but otherwise the line tab.

packages/ag-charts-enterprise/src/components/popover/popover.ts
lsjroberts: Yeah agreed, good point.

packages/ag-charts-enterprise/src/components/dialog/dialog.ts
```diff
@@ -108,45 +102,65 @@ export abstract class Dialog<Options extends DialogOptions = DialogOptions> exte
...
      **************/
-    protected createTabs<T extends Record<string, { label: string; content: HTMLElement; onShow?: () => void }>>(
+    protected createTabs<T extends Record<string, { label: string; panel: HTMLElement; onShow?: () => void }>>(
         initial: keyof T,
         tabs: T
     ) {
         const element = createElement('div', 'ag-charts-dialog__tabs');
 
+        const tabButtonIds = mapValues(tabs, () => createElementId('ag-charts-dialog__tab'));
+        const tabPanelIds = mapValues(tabs, () => createElementId('ag-charts-dialog__tab-panel'));
```
lsjroberts, line 112: These use a simple incrementing id for the element ids, to enable the `aria-controls` and `aria-labelledby` attributes. Seems good enough for now, it's internal so we can fiddle with this later if needed.

alantreadway (approved):
LGTM - merging, let's wrap any additional changes into follow-up PRs.

PR #2448, opened 2024-09-11
AG-12712 Fix flicker when mounting segments in radial gauge

jacobp100:
https://ag-grid.atlassian.net/browse/AG-12712


alantreadway (approved):
LGTM

PR #2451, opened 2024-09-11
AG-NONE Partially revert `computeSVGDataPath` for arcs

olegat:


olegat:
This fix is incorrect

PR #2453, opened 2024-09-11
AG-12640 - Hide maps example titles

markdurrant:


packages/ag-charts-website/src/pages-styles/homepage.module.scss
```diff
@@ -522,12 +522,26 @@
...
+
     :global(.example-runner-outer) {
-        margin-bottom: 0 !important;
+        margin-top: $spacing-size-2;
+        margin-bottom: $maps-title-offset !important;
     }
 
     :global(div[class*='content']) {
-        border: none;
+        overflow: hidden;
```
taktran, line 539: I think there's already another one above, where there is a `border: none;`

taktran (approved):
Just a minor comment

JamesSwinton:
LGTM

packages/ag-charts-website/src/pages-styles/homepage.module.scss
markdurrant: This is intentional to keep all the quick fix stuff separate from other things.

PR #2454, opened 2024-09-11
AG-12598 Fix escape to cancel creating annotation

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12598


alantreadway (approved):
LGTM

PR #2458, opened 2024-09-11
AG-12820: Radial Bar rendering in incorrect position with one data item

iMoses:
https://ag-grid.atlassian.net/browse/AG-12820


alantreadway (approved):
LGTM

PR #2463, opened 2024-09-11
AG-12487 Fix TC1.2 use `:focus` pseudo-class for the mouse

olegat:
https://ag-grid.atlassian.net/browse/AG-12487


alantreadway (approved):
LGTM

PR #2465, opened 2024-09-12
AG-12660 - Hide crosshairs and axis button on zoom events rather than…

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-12660

### DIFF
--- a/packages/ag-charts-enterprise/src/features/crosshair/crosshair.ts
+++ b/packages/ag-charts-enterprise/src/features/crosshair/crosshair.ts
@@ -69,12 +69,16 @@ export class Crosshair extends _ModuleSupport.BaseModuleInstance implements _Mod
         const seriesRegion = ctx.regionManager.getRegion('series')!;
         const mouseMoveStates = InteractionState.Default | InteractionState.Annotations;
 
+        this.hideCrosshairs();
+
         this.destroyFns.push(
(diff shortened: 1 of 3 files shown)


packages/ag-charts-enterprise/src/features/crosshair/crosshair.ts
```diff
@@ -158,10 +160,10 @@ export class Crosshair extends _ModuleSupport.BaseModuleInstance implements _Mod
...
             line.lineDashOffset = lineDashOffset;
 
-            line.y1 = 0;
-            line.y2 = isVertical ? bounds.height : 0;
-            line.x1 = 0;
-            line.x2 = isVertical ? 0 : bounds.width;
+            line.y1 ||= 0;
+            line.y2 ||= isVertical ? bounds.height : 0;
+            line.x1 ||= 0;
+            line.x2 ||= isVertical ? 0 : bounds.width;
```
alantreadway, line 166: Should this be `??` instead of `||`, or is the intent to always replace `0` values too?

manapeirov: we want to replace falsy values including 0 as they are initialized with 0

alantreadway (approved):
LGTM

PR #2471, opened 2024-09-12
AG-12498 Change linear scenes to use non-extended lines for anchor

lsjroberts:


alantreadway (approved):
LGTM

PR #2473, opened 2024-09-12
AG-12710 Redirect bullet series, gauge API updates

jacobp100:
https://ag-grid.atlassian.net/browse/AG-12710
https://ag-grid.atlassian.net/browse/AG-12712


packages/ag-charts-website/astro.config.mjs
```diff
@@ -98,4 +98,10 @@ export default defineConfig({
         },
     },
     integrations: [react(), markdoc(), sitemap(getSitemapConfig()), agHtaccessGen({ include: HTACCESS === 'true' })],
+    redirects: {
+        '/javascript/bullet-series': '/javascript/linear-gauge/#bullet-series',
+        '/angular/bullet-series': '/angular/linear-gauge/#bullet-series',
+        '/react/bullet-series': '/react/linear-gauge/#bullet-series',
+        '/vue/bullet-series': '/vue/linear-gauge/#bullet-series',
+    },
```
alantreadway, line 106: NOTE: Redirects are normally handled by `.htaccess` config - we need to sync with @taktran and @seanlandsman on the approach for this - for Grid I can see we have tooling for this, entry-point is here: https://github.com/ag-grid/ag-grid/blob/dae22a177bb43fa75ad6e50dab7d999f2d684095/documentation/ag-grid-docs/plugins/agHtaccessGen.ts

alantreadway (changes requested):
LGTM other than the redirect question

packages/ag-charts-website/astro.config.mjs
alantreadway: Discussed offline, this results in `.htaccess` changes.

alantreadway (approved):

packages/ag-charts-website/astro.config.mjs
taktran: This would generate a html redirect, but it'd be `packages/ag-charts-website/plugins/agHtaccessGen.ts` that updates the `.htaccess`

packages/ag-charts-website/plugins/agHtaccessGen.ts
```diff
@@ -24,16 +24,24 @@ export default function createPlugin(options: Options): AstroIntegration {
...
-            'astro:build:done': async ({ dir }) => {
+            'astro:build:done': async ({ dir, routes }) => {
                 if (!options.include) {
                     // eslint-disable-next-line no-console
                     console.info('[agHtaccessGen] .htaccess generation disabled, skipping');
                     return;
                 }
 
+                const redirects = routes
+                    .flatMap(({ route, redirect }) =>
```
taktran, line 35: `redirect` doesn't seem to be in the [astro docs](https://docs.astro.build/en/reference/integrations-reference/#routedata-type-reference), but as mentioned offline, it seems to be in the type definition, and does generate as expected on prod build.

taktran (approved):
A bit weary of the `redirect` property usage, but seems like it works fine, and might just be the Astro docs being out of date.

Nice bonus that there's the html fallback.

PR #2476, opened 2024-09-12
AG-12819: Fix angle number axis labels position when endAngle specified

iMoses:
https://ag-grid.atlassian.net/browse/AG-12819


alantreadway (approved):
LGTM

PR #2477, opened 2024-09-12
AG-11973: Fix axis title position calculation

iMoses:
https://ag-grid.atlassian.net/browse/AG-11973
Animation is still broken


packages/ag-charts-community/src/motion/fromToMotion.ts
```diff
@@ -120,6 +120,8 @@ export function fromToMotion<
                 status = calculateStatus(node, node.datum, getDatumId, diff);
             }
 
+            node.datum.fadeOut = status === 'removed';
```
alantreadway: Feels a bit dirty to be jamming this onto the datum - perhaps we should have a flag (or status) on `Node` to indicate that we're expecting this `Node` to be removed - and then in `Group.computeChildrenBBox()` we allow an optional set of flag/statuses to include?

PR #2478, opened 2024-09-12
AG-12838 Update Locales

JamesSwinton:
https://ag-grid.atlassian.net/browse/AG-12838


packages/ag-charts-locale/src/ar-EG.ts
```diff
@@ -1,16 +1,28 @@
+/**
+ * Please note:
+ *
+ * Translations are provided as an illustration only and are
+ * not guaranteed to be accurate or error-free.
+ *
+ * They are designed to show developers where to store t heir
```
alantreadway: Nit: `t heir` here and below?

JamesSwinton: Good catch, thanks. fixed.

alantreadway (approved):
LGTM, merge on green!

olegat:
FYI - There's two more strings incoming https://github.com/ag-grid/ag-charts/pull/2479.

alantreadway (approved):
LGTM

PR #2479, opened 2024-09-12
AG-12606 Screenreader fixes

olegat:
https://ag-grid.atlassian.net/browse/AG-12606

### DIFF
--- a/packages/ag-charts-community/src/dom/elements.ts
+++ b/packages/ag-charts-community/src/dom/elements.ts
@@ -8,13 +8,23 @@ interface MissingAriaAttrs {
     ariaControls?: string;
 }
 
-export interface ButtonOptions {
-    label: string | HTMLElement;
+// These types force a compilation error if the developer tries to add an icon-only
+// menu item without an accessible text alternative.
+type LabelAndIcon = { label: string; icon?: AgIconName; altText?: undefined };
+type IconOnly = { label?: undefined; icon: AgIconName; altText: string };
+export type LabelIcon = LabelAndIcon | IconOnly;
+
+export type ButtonOptions = LabelIcon & {
     onPress: (event: MouseEvent) => void;
-}
+};
 export function createButton(options: ButtonOptions, attrs?: Attrs<HTMLButtonElement>) {
     const button = createElement('button', getClassName('ag-charts-input ag-charts-button', attrs));
-    button.append(options.label);
+    if (options.label !== undefined) {
+        button.append(options.label);
+    } else {
+        button.append(createIcon(options.icon));
+        button.ariaLabel = options.altText;
+    }
     button.addEventListener('click', options.onPress);
     applyAttrs(button, attrs);
     return button;
--- a/packages/ag-charts-community/src/dom/elements.ts
+++ b/packages/ag-charts-community/src/dom/elements.ts
@@ -87,6 +87,6 @@ function applyAttrs(element: HTMLElement, attrs?: Attrs) {
 }
 
 function normaliseMissingAriaAttrs(key: string) {
-    if (key === 'ariaControls') return 'aria-controls';
-    return key;
+    // Convert camelCaseString to kebab-case-string
+    return key.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase();
 }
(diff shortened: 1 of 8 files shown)


packages/ag-charts-community/src/dom/elements.ts
```diff
@@ -8,13 +8,23 @@ interface MissingAriaAttrs {
     ariaControls?: string;
 }
 
-export interface ButtonOptions {
-    label: string | HTMLElement;
+// These types force a compilation error if the developer tries to add an icon-only
+// menu item without an accessible text alternative.
+type LabelAndIcon = { label: string; icon?: AgIconName; altText?: undefined };
+type IconOnly = { label?: undefined; icon: AgIconName; altText: string };
+export type LabelIcon = LabelAndIcon | IconOnly;
```
olegat, line 10: Note: This `LabelIcon` allows 3 syntax modes:

```ts
{ label }         // text only
{ label, icon }   // text & icon
{ icon, altText } // icon only (with altText for screenreaders)
```

Specifying an `icon` without `altText`, or specifying both a `label` and `altText` will raise a compiler error.

packages/ag-charts-community/src/dom/elements.ts
```diff
@@ -87,6 +97,6 @@ function applyAttrs(element: HTMLElement, attrs?: Attrs) {
 }
 
 function normaliseMissingAriaAttrs(key: string) {
-    if (key === 'ariaControls') return 'aria-controls';
-    return key;
+    // Convert camelCaseString to kebab-case-string
+    return key.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase();
```
olegat: This won't work for some property-attribute pairs like `ariaSetSize` / `aria-setsize` or `ariaPosInSet` / `aria-posinset`. We might want to have a more strict interface for the `Attrs` type.

Thoughts?

packages/ag-charts-locale/src/en-US.ts
```diff
@@ -67,6 +67,8 @@ export const AG_CHARTS_LOCALE_EN_US: Record<string, string> = {
     iconAltTextAlignCenter: 'Center',
     // Alt-text for the 'position-right' icon
     iconAltTextAlignRight: 'Right',
+    // Alt-text for the 'close' icon
+    iconAltTextClose: 'Close',
```
lsjroberts, line 73: Worth highlighting this with @JamesSwinton - https://github.com/ag-grid/ag-charts/pull/2478

packages/ag-charts-community/src/dom/elements.ts
lsjroberts: Agreed, the current `Attrs` type also allows function properties like `setEventListener` which is a bit leaky.

If we only care about targetting aria attributes, could we just replace the `aria` with `aria-` and not kebab the whole thing? Or are there other attrs we need to consider?

The reason I called it "missing" aria attrs is the typescript types for elements includes many attrs like `ariaLabel` but not `ariaControls`. Not sure if that is just our ts version being out of date or if it implies something else?

packages/ag-charts-locale/src/en-US.ts
olegat: Noted. Thanks.

alantreadway: We'll hold off merging #2478 until these new keys are merged.

packages/ag-charts-community/src/dom/elements.ts
olegat: AG-Grid uses wrappers to set aria attributes: 
https://github.com/ag-grid/ag-grid/blob/419770ca302ce87bc3c9a64b51a464c86132ab7c/community-modules/core/src/utils/aria.ts

We have something similar but it's not so widely used unfortunately:
https://github.com/ag-grid/ag-charts/blob/c2f6bb944c7dd8bce14803cdac6717a485e2537f/packages/ag-charts-community/src/util/attributeUtil.ts

It might be a good idea to start using `attributeUtil.ts` in more places sure as here.

alantreadway (approved):
LGTM

PR #2491, opened 2024-09-13
AG-12466 Set `aria-haspopup` and `aria-expanded` on menu buttons

olegat:
https://ag-grid.atlassian.net/browse/AG-12466


alantreadway (approved):
LGTM

PR #2492, opened 2024-09-13
AG-9591: Polar axes crosslines outside angle-number domain not handle…

iMoses:
…d correctly
https://ag-grid.atlassian.net/browse/AG-9591


alantreadway (approved):
LGTM

PR #2500, opened 2024-09-16
CRT-478 - Fix charts homepage automated example

taktran:
* Fix grid charts cdn url logic
* Add `archive` configuration to nx
* Add extra debugging information for the website


alantreadway (approved):
LGTM

PR #2502, opened 2024-09-16
CRT-513 Add Missing Locale Key

JamesSwinton:
Fixes duplicate comment (which caused the key to be missed initially)

Re-translates `iconAltTextAlignCenter` key with new context, and adds the missing `iconAltTextPositionCenter` key to all locales


alantreadway (approved):
LGTM

PR #2506, opened 2024-09-16
CRT-486 - Fix annotations text wrapping when at the edge of the chart

manapeirov:
PR for https://ag-grid.atlassian.net/browse/CRT-486


alantreadway (approved):
LGTM

PR #2509, opened 2024-09-16
CRT-471 Fix channels to use actual top/bottom line for text position

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-471


alantreadway (approved):
LGTM

PR #2513, opened 2024-09-16
CRT-491 Fix history for cut & paste, removes copy from history

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-491


alantreadway (approved):
LGTM

PR #2520, opened 2024-09-16
CRT-485 Use overlay HTML element instead of aria announcer for bounds

olegat:
https://ag-grid.atlassian.net/browse/CRT-485


alantreadway (approved):
LGTM

PR #2523, opened 2024-09-16
CRT-500 Enforce DPI when downloading chart with explicit width/height

jacobp100:
https://ag-grid.atlassian.net/browse/CRT-500


alantreadway (approved):
LGTM

PR #2531, opened 2024-09-16
CRT-475 Fix zoom range and zoom reset button clash

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-475

### DIFF
--- a/packages/ag-charts-enterprise/src/features/zoom/zoomRange.ts
+++ b/packages/ag-charts-enterprise/src/features/zoom/zoomRange.ts
@@ -44,6 +44,9 @@ export class ZoomRange {
 
     public updateDomain(domain: Array<Date | number>) {
         this.domain = domain;
+        if (this.initialStart == null && this.initialEnd == null) {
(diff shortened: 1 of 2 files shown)


packages/ag-charts-enterprise/src/features/zoom/zoomRange.ts
```diff
@@ -44,6 +44,9 @@ export class ZoomRange {
 
     public updateDomain(domain: Array<Date | number>) {
         this.domain = domain;
+        if (this.initialStart == null && this.initialEnd == null) {
+            this.updateWith((start, end) => [start, end]);
+        }
```
lsjroberts, line 49: `updateDomain` is called after the first layout, so once the user options of `rangeX: { start, end }` have been processed. If none were provided this initialises it with the domain values of start and end so they are used as the "initial" zoom in `Zoom.getInitialZoom()`.

lsjroberts:
This fix doesn't work

PR #2533, opened 2024-09-16
CRT-510 Fix series focus order sorting

olegat:
https://ag-grid.atlassian.net/browse/CRT-510

Use the index value from the .map() function rather than the _declarationOrder member. When testing this on Firefox, all the _declarationOrder values were -1. Returning 0 in the sort function solves this, but let's just avoid using this faulty property anyway.

### DIFF
--- a/packages/ag-charts-community/src/chart/series/seriesAreaFocusManager.ts
+++ b/packages/ag-charts-community/src/chart/series/seriesAreaFocusManager.ts
@@ -54,15 +54,21 @@ export class SeriesAreaFocusManager extends BaseManager {
     }
 
     public seriesChanged(series: Series<any, SeriesProperties<any>>[]) {
-        this.series = [...series].sort((a, b) => {
-            let fpA = a.properties.focusPriority ?? Infinity;
-            let fpB = b.properties.focusPriority ?? Infinity;
+        const unsortedSeries = series.map((series, index) => {
+            return {series, index}
+        });
+        const sortedSeries = unsortedSeries.sort((a, b) => {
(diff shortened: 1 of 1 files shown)


packages/ag-charts-community/src/chart/series/seriesAreaFocusManager.ts
```diff
@@ -54,15 +54,21 @@ export class SeriesAreaFocusManager extends BaseManager {
...
-            let fpA = a.properties.focusPriority ?? Infinity;
-            let fpB = b.properties.focusPriority ?? Infinity;
+        const unsortedSeries = series.map((series, index) => {
+            return { series, index };
+        });
+        const sortedSeries = unsortedSeries.sort((a, b) => {
+            let fpA = a.series.properties.focusPriority ?? Infinity;
+            let fpB = b.series.properties.focusPriority ?? Infinity;
             if (fpA === fpB) {
-                [fpA, fpB] = [a._declarationOrder, b._declarationOrder];
```
alantreadway, line 61: Hold up - we need to respect declaration order, or things can get very messy after chart option updates which add/remove series or re-order them.

olegat: In that case, would it not be better to throw an error if the `_declaractionOrders` are equal? (rather than returning 0)

alantreadway: Let's roll with this and revert to `_declarationOrder` on `latest` post release.

alantreadway (approved):
LGTM

PR #2535, opened 2024-09-16
CRT-523 Fix chart proxying not applying when changing chart type, def…

jacobp100:
https://ag-grid.atlassian.net/browse/CRT-523


alantreadway (approved):
LGTM

PR #2544, opened 2024-09-17
Fix linear gauge docs

lsjroberts:


packages/ag-charts-website/src/content/docs/linear-gauge/index.mdoc
```diff
@@ -50,6 +52,8 @@ return (
 
 {% if isFramework("angular") %}
 
+To create a Linear Gauge, use the `<ag-gauge />` template with the type `linear-gauge`.
```
lsjroberts: template / import? not sure what to call it in angular

alantreadway:
```suggestion
To create a Linear Gauge, use the `<ag-gauge />` component with the type `linear-gauge`.
```

alantreadway (approved):
Reviewed with @AG-DavidG

PR #2545, opened 2024-09-17
AG-12818 Update legend keynav when orientation changes

olegat:
https://ag-grid.atlassian.net/browse/AG-12818

### DIFF
--- a/packages/ag-charts-community/src/chart/legend.ts
+++ b/packages/ag-charts-community/src/chart/legend.ts
@@ -352,10 +352,9 @@ export class Legend extends BaseProperties {
             .nodes()
             .map((markerLabel) => markerLabel.proxyButton?.button)
             .filter((button): button is HTMLButtonElement => !!button);
-        initRovingTabIndex({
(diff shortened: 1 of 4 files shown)


packages/ag-charts-community/src/chart/legend.ts
```diff
@@ -243,6 +244,7 @@ export class Legend extends BaseProperties {
     private readonly proxyLegendToolbar: HTMLDivElement;
     private readonly proxyLegendPagination: HTMLDivElement;
     private readonly proxyLegendItemDescription: HTMLParagraphElement;
+    private readonly proxyLegendToolbarDestroyFns: DestroyFns = new DestroyFns();
```
olegat, line 247: Note: I suspect that this is what's causing the memory benchmark failure. When there are a lot of legend item, this array will be very big because they are many event listeners to remove.

olegat: Obsolete. CI now passes.

alantreadway (approved):
LGTM

PR #2548, opened 2024-09-17
AG-12887 - Split mouse-down and drag-start events to reduce false drag start on click cases.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-12887

Stops emission of `drag-start` interaction event type by default on `mousedown` or `touchdown`, so that simple click/touch can end with a `click` event, without an intermediate drag-state.

As a first pass, simply falls back to emitting `drag-start` on first `mousemove` with the mouse button down.


jacobp100 (approved):
Looks good. A while ago I refactored `decideInteractionEventTypes` from returning an array to an optional. Might have been a bit premature after this

yarn.lock
jacobp100, line 1: Should this have changed?

alantreadway: No, for some reason `/snapshots` are trigger this update, will investigate distinctly.

PR #2550, opened 2024-09-17
AG-12899 - Fix axis button showing when hovering outside of the serie…

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-12899


alantreadway (approved):
LGTM

PR #2557, opened 2024-09-18
AG-12768 Fix rotated line labels mouse collisions

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12768


packages/ag-charts-community/src/scene/transformable.ts
```diff
@@ -101,6 +102,10 @@ function MatrixTransform<N extends Node>(Parent: Constructor<N>) {
             return this.toParent(bbox);
         }
 
+        computeBBoxWithoutTransforms() {
+            return super.computeBBox();
+        }
```
lsjroberts, line 107: This is required to get the original dimensions of the bbox before it was transformed, since `super.computeBBox() !== this.fromParent(this.computeBBox())`.

alantreadway (approved):
LGTM

PR #2558, opened 2024-09-18
Isolate benchmark tests by running each suite in a separate process

iMoses:


iMoses:
I kept the `single-process` option because I believe that the difference between the two modes of execution is an indication that we have a memory leaks when "destroying" chart instances, something we should look into in the future.

alantreadway (approved):
LGTM

PR #2564, opened 2024-09-20
AG-11986 Refactor toolbar module into reusable widgets

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-11986

**BREAKING CHANGE - DO NOT MERGE**

**Goals**

The goal of this change is to align the way the toolbars work with the way they are actually used:

1. Give responsibility for creating and managing the lifecyle of toolbars to their parent modules.
2. Remove duplication of interaction handling (click to open child menus, key nav, focusing, returning focus to last element).

This will entirely replace the `Toolbar` module, removing it and changing where the user options are defined.

**Components**

There are two categories of component:

- `Widget` - _A component that is immediately attached to the dom and takes space from the layout._
- `Popover` - _A component that can be attached and removed from the dom and does not take space from the layout._

The reusable components follow the aria roles:

- Menu
  - `MenuWidget` - _A component widget that offers a list of choices to the user. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role_
  - `MenuPopover` - _An anchored popover variant of the menu widget that offers a list of choices to the user._
    - Used for child menus of the `MenuWidget`.
- Toolbar
  - `ToolbarWidget` - _(may not be needed)_ 
  - `ToolbarPopover` - _A collection of commonly used function buttons or controls. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/toolbar_role_

These can then be extended into `ColorPicker`, `AnnotationsMenu` (for creating annotations), `AnnotationOptionsToolbar` (for editing an annotation) etc.

**`LayoutManager`**

To handle adding the `ChartTypeMenu` and `AnnotationsMenu` inline with each other, the layout manager will need more control over registering elements.

The proposal here is to change the api of `LayoutManager.registerElement()` to take options that define the position, alignment and size of the element and whether it is displayed inline or as a block.

This takes direct responsibility for shrinking the layout box/series area away from the modules, so they can just request the space they want and be given an x/y position back at which to place themselves.

**`DOMHierarchyManager` (or `FocusManager`, name tbd)**

Currently, focus has to be handled for every popover/menu/toolbar we add.

The widgets & popovers would be registered with this manager into a tree, and the manager would understand the current state of the tree: which components are visible and which component opened other components.

So when a component is closed, focus can be automatically returned to the correct location.

And if any specific branch or leaf in the tree is set to active, it will know the sequence to close and open other branches and leaves, ensuring only one node is active at a time.


lsjroberts:
Closed in favour of new widget approach.

PR #2566, opened 2024-09-20
AG-NONE Use `user-select` instead of `preventDefault()` to avoid HTML text selection

olegat:


olegat:
Obsolete. See https://github.com/ag-grid/ag-charts/pull/2569

PR #2567, opened 2024-09-20
AG-12776 - Fix chart focus lost on (pointer over) canvas click.

alantreadway:
Partial revert of https://ag-grid.atlassian.net/browse/AG-12776 + test fixes.


packages/ag-charts-community/src/chart/interaction/ariaAnnouncementServices.ts
olegat, line 1: I would expect the `'navigator'` e2e test in `zoom.spec.ts` to fail without a substitute fix. Has this change been tested to ensure that dragging the mouse outside the canvas doesn't start highlighting other HTML text elements?

alantreadway: I tried a few different things and couldn't reproduce the selection issue, could you point me to some steps to reproduce?

PR #2572, opened 2024-09-23
AG-12832 - Refactor dev watch task to be shareable with grid

taktran:


packages/all/project.json
```diff
@@ -97,7 +100,7 @@
         "parallel": false,
         "commands": [
           "nx run ag-charts-website-gallery_simple-bar_main.ts:generate-thumbnail",
-          "node {projectRoot}/watch.js & nx run ag-charts-website:dev"
+          "node {projectRoot}/../../external/ag-website-shared/scripts/watch/watch.js charts & nx run ag-charts-website:dev"
```
alantreadway: Nit: Maybe `{workspareRoot}/external/ag-website-shared/scripts/watch/watch.js` would be slightly more robust, rather than using `../..`?

packages/all/project.json
```diff
@@ -76,7 +76,10 @@
       "outputs": [],
       "options": {
         "parallel": true,
-        "commands": ["node {projectRoot}/watch.js", "nx run ag-charts-website:dev"]
+        "commands": [
+          "node {projectRoot}/../../external/ag-website-shared/scripts/watch/watch.js charts",
```
alantreadway: Ditto regarding `{projectRoot}` vs. `{workspaceRoot}`.

packages/all/project.json
taktran: I did try this before, but I get this error

```
NX  The {workspaceRoot} token is only valid at the beginning of an option. (all:dev.commands)
```

taktran: Removing `{projectRoot}/../../` fixes it, as the `workspaceRoot` is implied.

As per https://github.com/nrwl/nx/issues/14436#issuecomment-1778089590

alantreadway (approved):
LGTM

PR #2573, opened 2024-09-23
AG-12738 Gauge themes

jacobp100:
https://ag-grid.atlassian.net/browse/AG-12738

### DIFF
--- a/packages/ag-charts-community/src/module/optionsModule.ts
+++ b/packages/ag-charts-community/src/module/optionsModule.ts
@@ -101,8 +101,6 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
 
         let options = deepClone(userOptions, cloneOptions);
 
-        this.activeTheme = getChartTheme(options.theme);
-
         if (this.presetType != null) {
             const presetConstructor: ((options: AgPresetOptions, activeTheme: () => ChartTheme) => T) | undefined = (
                 PRESETS as any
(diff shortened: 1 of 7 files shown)


packages/ag-charts-community/src/module/optionsModule.ts
```diff
@@ -100,9 +102,22 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
...
+            const presetOptions = presetConstructor?.(presetParams, () => this.activeTheme) ?? options;
             this.debug('>>> AgCharts.createOrUpdate() - applying preset', options, presetOptions);
-            options = presetOptions;
+
+            const activeTheme = getChartTheme(options.theme);
+            const presetType = this.optionsType(presetOptions) as any as keyof AgPresetOverrides;
+            const presetTheme = activeTheme.presets[presetType];
+            if (presetTheme != null) {
+                const themedOptions = mergeDefaults(presetParams, presetTheme);
+                options = presetConstructor?.(themedOptions, () => this.activeTheme) ?? themedOptions;
```
jacobp100: This is less than ideal - but to determine the theme, we need to know the preset type, so we need to run the preset constructor. Then if it has a theme, we need to run it again with the theme applied

packages/ag-charts-community/src/module/optionsModule.ts
```diff
@@ -100,9 +102,22 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
...
-            const presetOptions = (PRESETS as any)[this.presetType]?.(options, () => this.activeTheme) ?? options;
+            const presetConstructor: ((options: AgPresetOptions, activeTheme: () => ChartTheme) => T) | undefined = (
+                PRESETS as any
+            )[this.presetType];
+            const presetParams = options as any as AgPresetOptions;
+            const presetOptions = presetConstructor?.(presetParams, () => this.activeTheme) ?? options;
             this.debug('>>> AgCharts.createOrUpdate() - applying preset', options, presetOptions);
-            options = presetOptions;
+
+            const activeTheme = getChartTheme(options.theme);
```
alantreadway: Nit: This seems to be duplicating work on later lines, could we move the initialisation of `this.activeTheme` to before this `if` block and only do it once?

packages/ag-charts-community/src/module/optionsModule.ts
alantreadway: Do we really need to do this - for gauges it looks like `type` is required? For Financial Charts we could fallback to `presetType`?

packages/ag-charts-community/src/module/optionsModule.ts
jacobp100: Not quite - because financial charts defines `theme` in its returned options

alantreadway (approved):
LGTM

PR #2580, opened 2024-09-24
AG-12646 - Add Docker e2e test executor

alantreadway:
https://ag-grid.atlassian.net/browse/AG-12646

Adds a Docker container for executing Playwright e2e tests; the intent is to allow a local mechanism to get consistent snapshots vs. CI, so we can remove the `-macos.png` snapshot versions as a source of inconsistencies.

To keep things as lightweight as possible, this change adds a couple of Nx target configs which:
- Runs `npx astro dev` on the host; this avoids needing to NPM install Linux-flavoured modules in the Docker container, as well as adding more spin-up overhead and complexity.
- Runs `npx playwright` in the container, with the repo volume mounted (mostly read-only, aside from report output folder + `e2e` test folder, to allow snapshot updates).

As part of this change I've uplifted the `playwright` version, which seems to have also uplifted the Chrome version being used for testing - this seems to be the source of snapshot updates (all snapshots generated locally, not on CI).

New commands added:
- `nx test:e2e ag-charts-website -c docker`
- `nx test:e2e ag-charts-website -c docker:update`


iMoses (approved):
LGTM 👍

PR #2581, opened 2024-09-24
AG-12983 Add `<div class="ag-charts-series-area">` element

olegat:

Note PR adds a new element `.ag-charts-series-area` under the `.ag-charts-canvas-proxy` container.

Screenshot 2024-09-25 at 16 30 12

The class-name of the new element is `.ag-charts-series-area` because it conceptually only interacts with the series/datums. However, it has the same bounds as the canvas to minimise the risks of breakages in the short-term. In the long-term, we might change these bounds to match those of the scene-graph group.


packages/ag-charts-types/src/chart/chartOptions.ts
```diff
@@ -146,7 +146,7 @@ export interface AgKeyboardOptions {
      *
      * Default: `0`
      */
-    tabIndex?: number;
+    tabIndex?: 0 | -1;
```
alantreadway: We can't change this in a minor release.

olegat: > We can't change this in a minor release.

Acknowledged. However, can we deprecate tabIndex > 0 with a warning? I think that allow this was a mistake, the recommendation is to only use 0 and -1. https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

olegat: Reverted here: https://github.com/ag-grid/ag-charts/commit/6332783884ddc04b971008ec7f6bd929dfe2a16a

alantreadway (approved):
LGTM

PR #2582, opened 2024-09-25
Refactor scene graph

iMoses:


packages/ag-charts-community/src/chart/zIndexMap.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Constants to declare the expected nominal zIndex for all types of layer in chart rendering.
+ */
+export enum zIndexMap {
```
alantreadway: Nit: Upper-camel-case this name (`ZIndexMap`)?

packages/ag-charts-community/src/scene/group.ts
```diff
@@ -14,115 +14,46 @@ export class Group extends Node {
...
-    protected layer?: HdpiCanvas;
+    static computeChildrenBBox(nodes: Iterable<Node>, skipInvisible = true) {
+        function* visible() {
+            for (const n of nodes) {
+                if (!skipInvisible || (n.visible && !n.transitionOut)) {
+                    const bbox = n.getBBox();
+                    if (bbox) yield bbox;
+                }
+            }
+        }
```
alantreadway: Nit(micro-optimisation): Should we extract this as a top-level function to improve optimisability?

packages/ag-charts-community/src/scene/group.ts
```diff
@@ -131,261 +62,175 @@ export class Group extends Node {
...
 
     /**
      * Transforms bbox given in the canvas coordinate space to bbox in this group's coordinate space and
      * sets this group's clipRect to the transformed bbox.
      * @param bbox clipRect bbox in the canvas coordinate space.
+     * @param useGroupCoordinateSpace keep provided bbox coordinate space.
      */
-    setClipRectInGroupCoordinateSpace(bbox?: BBox) {
-        this.clipRect = bbox ? Transformable.fromCanvas(this, bbox) : undefined;
+    setClipRect(bbox?: BBox, useGroupCoordinateSpace = true) {
```
alantreadway: Nit: I think the naming is off here - `this.clipRect` should always be in local coordinate space, so `useGroupCoordinateSpace` is mis-named, it should actually be about the coordinate space of the `bbox` argument. E.g. `bboxInLocalCoordinateSpace` seems more representative?

packages/ag-charts-community/src/scene/layer.ts
```diff
@@ -0,0 +1,201 @@
...
+            zIndex?: number;
+            zIndexSubOrder?: ZIndexSubOrder;
+            deriveZIndexFromChildren?: boolean; // TODO remove feature
+        }
+    ) {
+        super(opts);
+    }
+
+    override markDirty(type = RedrawType.TRIVIAL) {
+        super.markDirty(type, this.layer != null ? RedrawType.TRIVIAL : type);
```
alantreadway: Is the logic here correct? `this.layer` being `undefined` is just because of lazy initialisation? If so, I would have thought that `markDirty()` should always percolate dirty status of `TRIVIAL`?

iMoses: That's the current logic, I'll change it

alantreadway (approved):
LGTM

PR #2588, opened 2024-09-26
Stacked bar download fix

jacobp100:


alantreadway (changes requested):
Discussed offline, we need a tracking JIRA for QA purposes, and the root cause is probably due to use of `processedOptions` rather than `userOptions` in the download handling when cloning the chart.

alantreadway:
Superceeded by #2699

PR #2589, opened 2024-09-26
AG-12983 a11y captions

olegat:
Depends on https://github.com/ag-grid/ag-charts/pull/2581

PR changes the positioning of the a11y captions (title/subtitle go before the series-area, footnote & axes go after)


alantreadway (approved):
LGTM

PR #2590, opened 2024-09-26
AG-12895 Ignore tooltip hover event during `:focus-visible`

olegat:
https://ag-grid.atlassian.net/browse/AG-12895


alantreadway (approved):
LGTM

PR #2591, opened 2024-09-26
AG-9235 Axis defaults for funnel

jacobp100:
https://ag-grid.atlassian.net/browse/AG-9235


packages/ag-charts-enterprise/src/series/cone-funnel/coneFunnelModule.ts
```diff
@@ -13,17 +13,16 @@ export const ConeFunnelModule: _ModuleSupport.SeriesModule<'cone-funnel'> = {
...
-            type: _Theme.CARTESIAN_AXIS_TYPE.CATEGORY,
-            position: _Theme.POSITION.BOTTOM,
-        },
-        {
-            type: _Theme.CARTESIAN_AXIS_TYPE.NUMBER,
-            position: _Theme.POSITION.LEFT,
-        },
-    ],
-    swapDefaultAxesCondition: (series) => series?.direction !== 'vertical',
+    defaultAxes: (series) =>
```
alantreadway: Nit: Could this not also use `swapAxisCondition`?

packages/ag-charts-community/src/chart/themes/util.ts
```diff
@@ -0,0 +1,16 @@
...
+export function swapAxisCondition(axes: [CartesianAxis, CartesianAxis], swap: (series: any) => boolean) {
+    return (series: any) => {
+        if (swap(series)) {
+            return [
+                { ...axes[0], position: axes[1].position },
+                { ...axes[1], position: axes[0].position },
+            ];
+        } else {
+            return axes;
+        }
```
alantreadway: Nit: `else` is redundant, just drop this block and `return axes;` after the `if` block close?

alantreadway (approved):
LGTM other than a couple of nits.

packages/ag-charts-enterprise/src/series/cone-funnel/coneFunnelModule.ts
jacobp100: This will almost definitely end up with more complicated configuration - like the funnel

jacobp100: Updated to use the function from the funnel series

PR #2594, opened 2024-09-27
AG-NONE Add guard for getComputedStyle

olegat:


iMoses:
I don't think we should add `getComputedStyle` to `verifiedGlobals` because it's part of `window`, making it a bit redundant. 
If there's only one place that may not have it in certain cases (I'm assuming tests or thumbnail creations?) you could use optional chaining. (e.g. `getComputedStyle?.(focusableParent).opacity === '1'`

olegat:
I've changed it to just `getComputedStyle?.(focusParent)`. I cannot remember how to reproduce the original error, so I do not know whether this will work correctly. It seems to work properly when I run it locally.

olegat:
I'm still seeing errors when running `nx dev`

olegat:
` getWindow().getComputedStyle(focusParent)` seems to work as intended.

PR #2597, opened 2024-09-27
Ag 12074/annotations with shift key

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-12074

### DIFF
--- a/packages/ag-charts-community/src/util/vector.ts
+++ b/packages/ag-charts-community/src/util/vector.ts
@@ -1,5 +1,3 @@
-import { toRadians } from './angle';
-
 export const Vec2 = {
     add,
     angle,
@@ -15,7 +13,6 @@ export const Vec2 = {
     length,
     lengthSquared,
     multiply,
-    snapToAngle,
     normalized,
     origin,
     required,
@@ -249,16 +246,4 @@ function required(a?: Partial<Vec2>): Vec2 {
  */
 function origin(): Vec2 {
     return { x: 0, y: 0 };
-}
-
-function snapToAngle(center: Vec2, a: Vec2, step: number, direction: number = 1) {
-    const stepRadians = toRadians(step);
-    const r = distance(center, a);
-    const theta = Math.atan2(a.y - center.y, a.x - center.x);
-    const snapTheta = Math.round(theta / stepRadians) * stepRadians;
-
-    return {
-        x: center.x + r * Math.cos(snapTheta),
-        y: center.y + r * Math.sin(snapTheta) * direction,
-    };
 }
--- a/packages/ag-charts-community/src/util/vector.ts
+++ b/packages/ag-charts-community/src/util/vector.ts
@@ -1,3 +1,5 @@
+import { toRadians } from './angle';
+
 export const Vec2 = {
     add,
     angle,
@@ -13,6 +15,7 @@ export const Vec2 = {
     length,
     lengthSquared,
     multiply,
+    snapToAngle,
     normalized,
     origin,
     required,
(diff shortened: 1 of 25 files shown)


packages/ag-charts-community/src/util/vector.ts
```diff
@@ -247,3 +250,15 @@ function required(a?: Partial<Vec2>): Vec2 {
...
+    const stepRadians = toRadians(step);
+    const r = distance(center, a);
+    const theta = Math.atan2(a.y - center.y, a.x - center.x);
+    const snapTheta = Math.round(theta / stepRadians) * stepRadians;
+
+    return {
+        x: center.x + r * Math.cos(snapTheta),
+        y: center.y + r * Math.sin(snapTheta) * direction,
+    };
+}
```
lsjroberts: This is a snapped version of the `rotate()` function, so I'd suggest composing that function with the snapping to create a `rotateSnapped()` function, using the same signature/order of parameters.

We should ideally only take in radians in these utils, and avoid mixing degrees and radians in the signatures.

I'd also suggest shifting the `direction` parameter out of this function as that seems more specific than desired for these vector utils. They are intended to be very simple and composable.

So something like:

```ts
function rotate(a: Vec2, theta: number, b: Vec2 = origin()) {
    // ...
}

function rotateSnapped(a: Vec2, theta: number, snapTo: number, b: Vec2 = origin()) {
    return rotate(a, Math.round(theta / snapTo) * snapTo, b);
}
```

And use like:

```ts
const center = convertPoint(origin, context) ?? Vec2.origin();
Vec2.rotateSnapped(offset, angle(offset, center), toRadians(snapToAngle), center);
```

packages/ag-charts-enterprise/src/features/annotations/scenes/linearScene.ts
```diff
@@ -66,17 +66,17 @@ export abstract class LinearScene<
         };
     }
 
-    public drag(datum: Datum, target: Coords, context: AnnotationContext) {
+    public drag(datum: Datum, target: Coords, context: AnnotationContext, shiftKey: boolean) {
```
lsjroberts: The state of whether the shift key is pressed down while clicking/dragging should be handled by the `AnnotationStateMachine`, not the scenes.

Then the `AnnotationStateMachine` could provide a util on the `AnnotationStateContext` like `ctx.isSnapping()` to the other state machines.

I believe this would also allow `point` to remain as a simple value instead of a function. Passing a function into the state transitions is really not ideal and should be avoided. It's creating a lot of mess with unrelated code.

packages/ag-charts-enterprise/src/features/annotations/parallel-channel/parallelChannelScene.ts
```diff
@@ -63,36 +63,60 @@ export class ParallelChannelScene extends ChannelScene<ParallelChannelProperties
...
+    ) {
         const { activeHandle, handles } = this;
         if (activeHandle == null) return;
 
         const { offset } = handles[activeHandle].drag(target);
         handles[activeHandle].toggleDragging(true);
 
         const prev = datum.toJson();
-        let moves: Array<ChannelHandle> = [];
+        let moves: Array<[ChannelHandle, ChannelHandle | undefined]> = [];
```
lsjroberts: Perhaps have a new variable rather than making this a nested array, it's a bit messy.

```
let moves: Array<ChannelHandle> = [];
let origins: Array<ChannelHandle> = [];

// ...

switch (activeHandle) {
    case 'topLeft':
    case 'bottomLeft':
        moves = ['topLeft', 'bottomLeft'];
        origins = ['topRight', 'bottomRight'];
        break;

// ...
```

lsjroberts (changes requested):

packages/ag-charts-enterprise/src/features/annotations/parallel-channel/parallelChannelScene.ts
manapeirov: done

packages/ag-charts-community/src/util/vector.ts
manapeirov: I moved the `snapToAngle` function to another utils file as it's a bit less generic/ more specific to this use case

packages/ag-charts-enterprise/src/features/annotations/scenes/linearScene.ts
manapeirov: Ok I've removed the shiftKey being passed down and now we get the value from keyDown and keyUp events rather than mouse events

lsjroberts (approved):
LGTM

PR #2160, opened 2024-08-02
Ag 12476/annotations floating toolbar delete functionality

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-12476

PR #2605, opened 2024-09-30
AG-13008, AG-13041 Fix Delete/Escape key deleting the annotation in text-edit mode

olegat:
https://ag-grid.atlassian.net/browse/AG-13008
https://ag-grid.atlassian.net/browse/AG-13041


packages/ag-charts-enterprise/src/features/annotations/annotationsStateMachine.ts
```diff
@@ -390,8 +390,12 @@ export class AnnotationsStateMachine extends StateMachine<States, AnnotationType
 
                 delete: () => {
                     if (this.active == null) return;
+                    this.transition('cancel');
```
lsjroberts: This is an anti-pattern, we should not transition inside these events. Partly as that could easily create an infinite loop, but also as it goes against the design of state machines.

lsjroberts (changes requested):
There's been a lot of back and forth on deleting annotations due to changes in interactions.

Please make sure any changes continue to work for both of these fixes:

- https://github.com/ag-grid/ag-charts/pull/2160
- https://github.com/ag-grid/ag-charts/pull/2454

packages/ag-charts-website/e2e/toolbar.spec.ts
```diff
@@ -126,4 +126,46 @@ test.describe('toolbar', () => {
         await page.click(SELECTORS.canvas, { position: { x: 100, y: 100 } });
         await expect(page).toHaveScreenshot('callout-6-change-fill-color.png', { animations: 'disabled' });
     });
+
+    test('AG-13008 delete annotation', async ({ page }) => {
```
lsjroberts, line 140: Also, can we just put some simple tests around this into the existing `test('text', ...` block? These tests are supposed to be just smoke tests.

packages/ag-charts-enterprise/src/features/annotations/annotationsStateMachine.ts
olegat: Fair. Another solution could be to store the `visible: boolean` state in the `TextInput` class and check that instead. I'll do that.

olegat: Done.

packages/ag-charts-website/e2e/toolbar.spec.ts
olegat: I've added tests to `test('text', ...)` for the <kbd>Backspace</kbd> and <kbd>Delete</kbd> keys

lsjroberts (approved):
LGTM, thanks!

alantreadway (approved):
LGTM

olegat:
> There's been a lot of back and forth on deleting annotations due to changes in interactions.
> 
> Please make sure any changes continue to work for both of these fixes:
> 
> * [Ag 12476/annotations floating toolbar delete functionality #2160](https://github.com/ag-grid/ag-charts/pull/2160)
> * [AG-12598 Fix escape to cancel creating annotation #2454](https://github.com/ag-grid/ag-charts/pull/2454)

I've checked these.

✅ AG-12476 - No regression
✅ AG-12598 - No regression

olegat:
Rebased and resolved conflict by dropping the e2e snapshot update to this file:
[packages/ag-charts-website/e2e/zoom.spec.ts-snapshots/zoom-crosshairs-after-wheel-zoom-chromium-linux.png](https://github.com/ag-grid/ag-charts/blob/28b713bb53099125953b647ef0827e89c9a39897/packages/ag-charts-website/e2e/zoom.spec.ts-snapshots/zoom-crosshairs-after-wheel-zoom-chromium-linux.png)

PR #2612, opened 2024-10-01
AG-12952 - Style <select> in example controls

markdurrant:


packages/ag-charts-website/src/features/example-runner/framework-templates/lib/ExampleStyle.tsx
```diff
@@ -50,26 +51,31 @@ export const ExampleStyle = ({
...
-            background: var(--button-hover-background-color);
+        .toolbar button {
+            text-align: center;
+        }
+
+        .toolbar select {
+            padding-right: 2.4em;
+            background-repeat: no-repeat;
+            background-position: center right 4px;
+            background-image: var(--select-chevron);
```
jacobp100: Minor - but you can use shorthands for the background

jacobp100 (approved):

packages/ag-charts-website/src/features/example-runner/framework-templates/lib/ExampleStyle.tsx
markdurrant: Good shout. Have refactored to use shorthand for background.

PR #2613, opened 2024-10-01
SVG export

jacobp100:
Spike of SVG export. Virtual children don't work

### DIFF
--- a/packages/ag-charts-community/src/scene/transformable.ts
+++ b/packages/ag-charts-community/src/scene/transformable.ts
@@ -139,12 +139,14 @@ function MatrixTransform<N extends Node>(Parent: Constructor<N>) {
 
         override toSVG(): { elements: SVGElement[]; defs?: SVGElement[] | undefined } | undefined {
             const svg = super.toSVG();
-            if (svg == null) return;
+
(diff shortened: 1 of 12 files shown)


packages/ag-charts-community/src/scene/transformable.ts
```diff
@@ -136,6 +136,22 @@ function MatrixTransform<N extends Node>(Parent: Constructor<N>) {
...
+
+        override toSVG(): { elements: SVGElement[]; defs?: SVGElement[] | undefined } | undefined {
+            const svg = super.toSVG();
+            if (svg == null) return;
+
+            const g = document.createElementNS('http://www.w3.org/2000/svg', 'g');
+            g.append(...svg.elements);
+
+            const [a, b, c, d, e, f] = this[TRANSFORM_MATRIX].e;
+            g.setAttribute('transform', `matrix(${a} ${b} ${c} ${d} ${e} ${f})`);
```
alantreadway: Nit: You probably want to check if the matrix is the identity matrix so you can avoid redundant matrix use in the SVG.

alantreadway (approved):
LGTM

PR #2626, opened 2024-10-03
AG-13021: Synchronised Axes don't update correctly

iMoses:
https://ag-grid.atlassian.net/browse/AG-13021


packages/ag-charts-community/src/chart/axis/axis.ts
```diff
@@ -1043,13 +1040,9 @@ export abstract class Axis<S extends Scale<D, number, TickInterval<S>> = Scale<a
...
 
         if (tickCount && (ContinuousScale.is(scale) || OrdinalTimeScale.is(scale))) {
-            if (typeof tickCount === 'number') {
-                scale.tickCount = tickCount;
-                scale.minTickCount = minTickCount ?? 0;
-                scale.maxTickCount = maxTickCount ?? Infinity;
-            } else if (scale instanceof TimeScale) {
-                this.setTickInterval(tickCount as TickInterval<S>);
-            }
+            scale.tickCount = tickCount;
```
alantreadway, line 1043: Was the `TimeScale` case really dead-code?

packages/ag-charts-enterprise/src/features/sync/chartSync.ts
```diff
@@ -52,15 +53,11 @@ export class ChartSync extends BaseProperties implements _ModuleSupport.ModuleIn
...
-    }
-
-    private updateSiblings(groupId?: string) {
+    updateSiblings(groupId?: string) {
         const { syncManager } = this.moduleContext;
 
         const updateFn = async () => {
-            for (const chart of syncManager.getGroupSiblings(groupId)) {
+            for (const chart of syncManager.getGroupSiblings(groupId ?? this.groupId)) {
                 await chart.waitForDataProcess(120);
```
alantreadway: Q: Does this need to be done serially, or should we be making these calls in (pseudo-)parallel?

packages/ag-charts-community/src/chart/cartesianChart.ts
```diff
@@ -442,6 +451,30 @@ export class CartesianChart extends Chart {
         return { clipSeries, axisThickness };
     }
 
+    private getSyncedDomain(axis: ChartAxis) {
+        const syncModule = this.modulesManager.getModule<SyncModule>('sync');
+        if (!syncModule?.enabled) return;
```
alantreadway, line 456: Just to confirm, if we short-circuit here, then the axis/domain behaviour for non-synchonised charts is as before?

packages/ag-charts-community/src/chart/axis/axis.ts
iMoses: I looked everywhere that `tickCount` could come from, they're all numbers, and typing wouldn't allow otherwise.

packages/ag-charts-community/src/chart/cartesianChart.ts
iMoses: Yes

packages/ag-charts-enterprise/src/features/sync/chartSync.ts
iMoses: We could. You're not worried we may put too much strain on the engine, blocking UI on a large scale of synced charts?

iMoses: Removed `waitForDataProcess` so responsiveness has improved. It isn't needed with the new approach.

alantreadway (approved):
LGTM

PR #2634, opened 2024-10-03
AG-12646 - Stabilise tests.

alantreadway:
Addresses stability of some tests that have been highlighted as part of introducing AG-12646.


packages/ag-charts-website/e2e/toolbar.spec.ts
```diff
@@ -139,14 +139,17 @@ test.describe('toolbar', () => {
...
         await gotoExample(page, url);
+        const { wrapper } = await locateCanvas(page);
         const point = await canvasToPageTransformer(page);
         const hover = point(200, 200);
         const leave = point(300, 400);
 
         // Test 1. Check that the Delete & Backspace keys work:
         await page.locator(SELECTORS.textAnnotationMenu).click();
         await page.locator(SELECTORS.commentMenuItem).click();
+        await page.mouse.move(hover.x, hover.y);
```
olegat, line 149: Is this `move` necessary? There are no other moves in the steps after this which is unusual.

packages/ag-charts-website/e2e/toolbar.spec.ts
```diff
@@ -155,31 +158,44 @@ test.describe('toolbar', () => {
         await page.keyboard.press('Backspace');
         await page.keyboard.press('Home');
         await page.keyboard.press('Delete');
-        await expect(page).toHaveScreenshot('delete-erased-text.png');
+        await expect(wrapper).toHaveAttribute('data-update-pending', 'false');
+        await expect(page).toHaveScreenshot('delete-erased-text.png', { animations: 'disabled' });
```
olegat, line 159: Is `delete-erased-text.png` the only file that is causing trouble? I'm wondering whether so many `data-update-pending` checks are necessary 🤔

packages/ag-charts-website/e2e/toolbar.spec.ts-snapshots/delete-erased-text-chromium-linux.png
olegat, line 1: ✅ The updated snapshot is correct.

olegat:
LGTM - but it would be helpful to understand why the AG-13008 test is the only one that's affected by the concurrency.

Could it be that `page.mouse.click` and `page.click(elem)` have different specifications regarding asynchronicity? 

The documentation seems to suggest that they do:
* `page.click` (_deprecated_) https://playwright.dev/docs/api/class-page#page-click
* `page.mouse.click` https://playwright.dev/docs/api/class-mouse#mouse-click

olegat (approved):

packages/ag-charts-website/e2e/toolbar.spec.ts
alantreadway: Due to the nature of the failure, it looks like some interaction overtook a hover earlier in the flow - so waiting only before the snapshot wouldn't address this.

packages/ag-charts-website/e2e/toolbar.spec.ts
alantreadway: Maybe not, but it reflects user behaviour better, since clicks don't just appear in the middle of the canvas without some hover events beforehand.

packages/ag-charts-website/e2e/fixture.ts
olegat, line 1: LGTM - This will make the `.spec.ts` files far more readable and reliable. Thanks.

packages/ag-charts-website/e2e/toolbar.spec.ts
```diff
@@ -155,21 +156,21 @@ test.describe('toolbar', () => {
...
-        await expect(page).toHaveScreenshot('delete-erased-text.png');
+        await expect(page).toHaveScreenshot('delete-erased-text.png', { animations: 'disabled' });
 
         // Test 2. Check that Backspace key deletes the annotation when in idle state:
         // (Click away from the annotation, then reclick it to go into idle state)
         await page.mouse.click(leave.x, leave.y, { button: 'left' });
         await page.mouse.click(hover.x, hover.y, { button: 'left' });
         await page.keyboard.press('Backspace');
-        await expect(page).toHaveScreenshot('delete-annotation-removed.png');
+        await expect(page).toHaveScreenshot('delete-annotation-removed.png', { animations: 'disabled' });
```
olegat, line 166: nit: We may want to add a `toHaveScreenshot` proxy to change the default to `{ animations: 'disable' }`. This can be done in a follow-up PR though.

PR #2643, opened 2024-10-04
AG-12836 - Open 'text' tab in options dialog when trend line label te…

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-12836

### DIFF
--- a/packages/ag-charts-enterprise/src/features/annotations/annotationsStateMachine.ts
+++ b/packages/ag-charts-enterprise/src/features/annotations/annotationsStateMachine.ts
@@ -340,8 +340,14 @@ export class AnnotationsStateMachine extends StateMachine<States, AnnotationType
 
                 dblclick: {
                     guard: guardActiveHasLineText,
-                    action: () => {
-                        ctx.showAnnotationSettings(this.active!, undefined);
+                    action: ({ offset }) => {
+                        const nodeAtCoords = ctx.getNodeAtCoords(offset, this.active!);
+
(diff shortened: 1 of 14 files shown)


packages/ag-charts-enterprise/src/features/annotations/annotationsStateMachine.ts
```diff
@@ -333,8 +333,14 @@ export class AnnotationsStateMachine extends StateMachine<States, AnnotationType
 
                 dblclick: {
                     guard: guardActiveHasLineText,
-                    action: () => {
-                        ctx.showAnnotationSettings(this.active!, undefined);
+                    action: ({ offset }) => {
+                        const nodeAtCoords = ctx.getNodeAtCoords(offset, this.active!);
+
+                        if (!(nodeAtCoords === 'line' || nodeAtCoords === 'text')) {
```
lsjroberts:
```suggestion
                        if (nodeAtCoords !== 'line' && nodeAtCoords !== 'text') {
```

Generally, it's a bit more readable like this, as the `(!(` can be missed when scanning the code.

lsjroberts: Also, I'd guess we will want to open on the "line" tab when the user double clicks a handle? So could do something like `const node = nodeAtCoords === 'text' ? 'text' : 'line';`

lsjroberts (approved):
LGTM, with one small comment.

alantreadway (approved):
LGTM

PR #2645, opened 2024-10-04
AG-12983 Move proxy HTML elements into `.ag-charts-canvas-proxy`

olegat:


alantreadway (approved):
LGTM

PR #2646, opened 2024-10-04
AG-12922 - Wrap text on save not on every update

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-12922

### DIFF
--- a/packages/ag-charts-enterprise/src/features/annotations/annotations.ts
+++ b/packages/ag-charts-enterprise/src/features/annotations/annotations.ts
@@ -1153,7 +1153,8 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
 
         const offset = Vec2.from(event);
         const point = () => invertCoords(offset, context);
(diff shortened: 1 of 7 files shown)


packages/ag-charts-enterprise/src/features/annotations/annotations.ts
```diff
@@ -1153,9 +1153,9 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
 
         const offset = Vec2.from(event);
         const point = () => invertCoords(offset, context);
-        const textInputValue = this.textInput.getValue();
+        const { textInputValue, bbox } = this.textInput.getValue() ?? {};
```
lsjroberts: It is a little odd to return the bbox from a function called `getValue()`.

Perhaps would be nicer to make `getBBox()` public and just use `this.textInput.getBBox()`?

lsjroberts (approved):
LGTM, one minor nit

lsjroberts (approved):

alantreadway (approved):

PR #2649, opened 2024-10-07
AG-12863 Fix waterfall labels when first item does not have labels

jacobp100:
https://ag-grid.atlassian.net/browse/AG-12863


alantreadway (approved):
LGTM

PR #2653, opened 2024-10-07
AG-12967 - Clamp note annotation text box shape using fixed width

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-12967


lsjroberts (approved):
LGTM

PR #2655, opened 2024-10-07
AG-13062 Sparklines initial cut

jacobp100:
https://ag-grid.atlassian.net/browse/AG-13062

### DIFF
--- a/packages/ag-charts-community/src/api/preset/sparkline.ts
+++ b/packages/ag-charts-community/src/api/preset/sparkline.ts
@@ -1,0 +1,141 @@
+import type {
+    AgBaseSparklinePresetOptions,
+    AgCartesianChartOptions,
+    AgCartesianSeriesOptions,
+    AgChartTheme,
+    AgChartThemeName,
+    AgSparklineOptions,
+} from 'ag-charts-types';
+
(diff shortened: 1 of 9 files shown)


packages/ag-charts-community/src/api/preset/sparkline.ts
```diff
@@ -0,0 +1,141 @@
...
+    AgCartesianChartOptions,
+    AgCartesianSeriesOptions,
+    AgChartTheme,
+    AgChartThemeName,
+    AgSparklineOptions,
+} from 'ag-charts-types';
+
+import { IGNORED_PROP, assertEmpty, pickProps } from './presetUtils';
+
+const SPARKLINE_THEME: AgChartTheme = {
```
alantreadway, line 12: Could we perhaps add this as an actual theme (not visible in the public API ofc), similar to `FinancialLight` and `FinancialDark`?

jacobp100: I'm not sure that works in this case. This doesn't define any colours - only overrides for the series defaults. As well, you need to handle cases like `theme: ag-vivid`, which won't be possible if we we had a theme subclass

alantreadway (approved):
LGTM

PR #2656, opened 2024-10-07
AG-13036 Fix invalid cachedBBox bug when adding/removing children

olegat:
https://ag-grid.atlassian.net/browse/AG-13036

When a scene nodes cached BBox is invalidated, then ALL of the ancestor cached BBoxes are also invalid because their computations is the union of the descendants.

The bug we were seeing was happening because if the user triggers hover events before the map chart has fully loaded, then the scene graph will compute a BBox for the root node that excludes the series area (i.e. the Geo nodes). The Node.append method unsets cachedBBox, but that does not matter because the RegionManager uses the getBBox() on an ancestor node, so it will just use an outdated BBox in its hit-testing.

Therefore, setting cachedBBox = undefined for all ancestors is the correct thing to do here.

### DIFF
--- a/packages/ag-charts-community/src/scene/node.ts
+++ b/packages/ag-charts-community/src/scene/node.ts
@@ -267,7 +267,7 @@ export abstract class Node {
             }
         }
 
-        this.cachedBBox = undefined;
+        this.invalidateCachedBBox();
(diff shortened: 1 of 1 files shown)


alantreadway (approved):
LGTM

PR #2658, opened 2024-10-08
Revert "AG-13040 Mimic mouse event with the keyboard when snap is disabled"

olegat:
https://ag-grid.atlassian.net/browse/AG-13040


alantreadway (approved):
LGTM

PR #2660, opened 2024-10-08
Annotations cleanup and move Vec2 to _ModuleSupport

lsjroberts:
Moving `Vec2` out of `_Util` as that is for sparklines and into `_ModuleSupport`, to keep it with the other enterprise utils.


alantreadway (approved):
LGTM

PR #2663, opened 2024-10-08
AG-12983 Mark navigator scene-nodes as dirty when they change

olegat:

### DIFF
--- a/packages/ag-charts-community/src/chart/navigator/shapes/rangeMask.ts
+++ b/packages/ag-charts-community/src/chart/navigator/shapes/rangeMask.ts
@@ -1,4 +1,5 @@
 import { BBox } from '../../../scene/bbox';
+import { RedrawType } from '../../../scene/changeDetectable';
 import { Path } from '../../../scene/shape/path';
 
 export class RangeMask extends Path {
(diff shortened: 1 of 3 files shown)


packages/ag-charts-community/src/chart/navigator/shapes/rangeSelector.ts
```diff
@@ -30,6 +31,7 @@ export class RangeSelector extends Layer {
 
         this.background.translationX = x;
         this.background.translationY = y;
+        this.markDirty(RedrawType.MAJOR);
```
olegat, line 34: nit: Technically this is redundant because `this` is an ancestor of `this.background` and therefore `this.background.translationY = y` will set `this._dirty = RedrawType`. However I've just added it as a safety measure because redundant `markDirty` calls do not have any side-effects.

packages/ag-charts-community/src/chart/navigator/shapes/rangeMask.ts
olegat, line 1: nit: The `layout` and `update` function could check whether the x, y, width, height, min, max values actually change to only mark the node as dirty when necessary.

olegat: Done.

alantreadway (approved):
LGTM

PR #2676, opened 2024-10-08
AG-13098 Use `opacity: 0` instead of `appearance`

olegat:
https://ag-grid.atlassian.net/browse/AG-13098

This is to ensure that ag-grid does not show parts of the navigator proxy.

Also hide other experiemental pseudo-elemements define used by ag-grid: https://github.com/ag-grid/ag-grid/blob/b42ae20f8eca2336c5b553e98d0b2efcaf4227b9/packages/ag-grid-community/src/theming/core/css/range.css


alantreadway (approved):
LGTM

PR #2677, opened 2024-10-08
AG-11910 - Only trigger keyboard navigation when not in annotation se…

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-11910


packages/ag-charts-community/src/chart/interaction/interactionStateListener.ts
```diff
@@ -4,12 +4,14 @@ import { BaseManager } from '../baseManager';
...
-    ContextMenu = 2,
-    Animation = 1,
-    All = Default | ZoomDrag | Annotations | ContextMenu | Animation,
+    Default = 32,
+    ZoomDrag = 16,
+    Annotations = 8,
+    ContextMenu = 4,
+    Animation = 2,
+    AnnotationsSelected = 1,
+    AllButAnnotationsSelected = Default | ZoomDrag | Annotations | ContextMenu | Animation,
```
alantreadway: Having these composed enum values as part of `InteractionState` seems a bit odd to me, I wonder if we should have another type which is super-set of the individual `Interactionstate` enum and the composite types, so we can differentiate between cases where a composite value is allowed or not?

olegat: I don't think that it matters. Typescript already implicitly allows bitwise-OR'd values to be passed in places where the type is `InteractionState`.

For example:
```typescript
const mouseStates = InteractionState.Default | InteractionState.Annotations | InteractionState.AnnotationsSelected;
this.destroyFns.push(interactionManager.addListener('hover', () => this.onMouse(), mouseStates),
```

This does not raise an error despite the fact that the 3rd parameter is of type `InteractionState` and `mouseState` is not an exact value from that enum.

Therefore a super-set type of individual & composite types of `InteractionState` would be the same as `InteractionState` itself.

If we want stricter version of `InteractionState`, then we need to use a type union like this:

```diff
diff --git a/packages/ag-charts-community/src/chart/interaction/interactionStateListener.ts b/packages/ag-charts-community/src/chart/interaction/interactionStateListener.ts
index 86c665487..e44dba9f8 100644
--- a/packages/ag-charts-community/src/chart/interaction/interactionStateListener.ts
+++ b/packages/ag-charts-community/src/chart/interaction/interactionStateListener.ts
@@ -12,11 +12,13 @@ export enum InteractionState {
     All = Default | ZoomDrag | Annotations | ContextMenu | Animation,
 }
 
+export type InteractionStateDemasked = 16 | 8 | 4 | 2 | 1;
+
 export abstract class InteractionStateListener<
     TType extends string,
     TEvent extends { type: TType },
 > extends BaseManager<TType, TEvent> {
-    protected abstract getState(): InteractionState;
+    protected abstract getState(): InteractionStateDemasked;
 
     // Wrapper to only broadcast events when the InteractionManager is a given state.
     override addListener<T extends TType>(
```

alantreadway (approved):
LGTM

PR #2681, opened 2024-10-09
AG-11918 Allow preventing double click event on chart

jacobp100:
This adds a `stopImmediatePropagation` method for some events. This is slightly flaky as the registration order of listeners is important, but it does have precedent in web standards.

Alternatively, we could make `seriesAreaManager` the only module to call `regionManager.listenAll`, then other modules would call `seriesAreaManager.addListener('click', ...)`, which would have the defaultPrevented logic already checked


packages/ag-charts-community/src/util/listeners.ts
```diff
@@ -42,6 +66,8 @@ export class Listeners<EventType extends string, EventHandler extends Handler> {
             } catch (e) {
                 Logger.errorOnce(e);
             }
+
+            if (params.length === 1 && immediatePropagationStopped(params[0])) return;
```
olegat, line 70: Could you please clarify which specific handlers you are trying to stop calling? Intuitively, it might make more sense to just check the `defaultPrevented` property in the handlers that are causing problems.

olegat (changes requested):

PR #2684, opened 2024-10-09
Fix reset

jacobp100:
Our reset is probably too aggressive for user-rendered HTML. This is a fix for that

Best to hold off until next release


alantreadway (approved):
LGTM

PR #2685, opened 2024-10-09
AG-12695 Temporarily set `pointer-events: none` on annotation drags

olegat:
https://ag-grid.atlassian.net/browse/AG-12695


packages/ag-charts-website/e2e/toolbar.spec.ts-snapshots/settings-button-ignored-hover-event-chromium-linux.png
olegat, line 1: > [!TIP]
> The comparison of this snapshot before & after the bug fix can be viewed here: https://github.com/ag-grid/ag-charts/commit/7b7d2675d441d697ab69a5c4c91f609637696628

packages/ag-charts-community/src/chart/toolbar/toolbar.ts
```diff
@@ -222,6 +226,21 @@ export class Toolbar extends BaseModuleInstance implements ModuleInstance {
         this.translateFloatingElements(FloatingTop, false);
     }
 
+    // AG-12695 Temporarily set `pointer-events: none` on the annotationOptions when dragging, because the
+    // buttons block to mouse from hovering over the canvas.
+    private readonly oldPointerEvents: Map<HTMLButtonElement, string> = new Map();
+    private onStopPointerEvents() {
+        if (this.oldPointerEvents.size !== 0) throw new Error('AG Charts - unexpected drag-start event');
```
alantreadway: I'd avoid a hard fail here, it could leave the chart in an odd/unrecoverable state?

packages/ag-charts-community/src/chart/toolbar/toolbar.ts
```diff
@@ -222,6 +226,21 @@ export class Toolbar extends BaseModuleInstance implements ModuleInstance {
...
+        this.groupButtons['annotationOptions'].forEach((b) => {
+            this.oldPointerEvents.set(b, b.style.pointerEvents);
```
alantreadway: I think it would be better not to keep this state around, could we perhaps instead use a CSS class for this specific case, and add/remove it to apply/remove the `pointer-events: none`? This would be less likely to conflict with other processing (which is presumably why we're defensively copying the original `pointerEvents`)?

packages/ag-charts-community/src/chart/toolbar/toolbar.ts
```diff
@@ -222,6 +226,21 @@ export class Toolbar extends BaseModuleInstance implements ModuleInstance {
         this.translateFloatingElements(FloatingTop, false);
     }
 
+    // AG-12695 Temporarily set `pointer-events: none` on the annotationOptions when dragging, because the
+    // buttons block to mouse from hovering over the canvas.
+    private readonly oldPointerEvents: Map<HTMLButtonElement, string> = new Map();
+    private onStopPointerEvents() {
+        if (this.oldPointerEvents.size !== 0) throw new Error('AG Charts - unexpected drag-start event');
+        this.groupButtons['annotationOptions'].forEach((b) => {
```
alantreadway: I agree, I think it might be better to apply `pointer-events: none` at one spot in the DOM tree as per your other comments? I'm not sure what the scope of impact is for that change though

packages/ag-charts-community/src/chart/toolbar/toolbar.ts
olegat: What is the advantage of toggling a class over this approach? The advantage of setting style values on the element directly that these have priority over class values. This tells the browser ignore the class styling, which is what is intended here.

packages/ag-charts-community/src/chart/toolbar/toolbar.ts
olegat: I'll experiment with it. There's still time to restore this version if we see undesirable side-effects.

packages/ag-charts-community/src/chart/toolbar/toolbar.ts
olegat: I believe that the error logging is useful because this should never happen in theory.

I've replaced it with `Logger.errorOnce` to avoid unrecoverable states.

packages/ag-charts-community/src/chart/toolbar/toolbar.ts
olegat: Unfortunately setting `pointer-events: none` of the canvas overlay does not work, because some child elements override that value.

```diff
diff --git a/packages/ag-charts-community/src/chart/toolbar/toolbar.ts b/packages/ag-charts-community/src/chart/toolbar/toolbar.ts
index a2f2ff543..c3e608eea 100644
--- a/packages/ag-charts-community/src/chart/toolbar/toolbar.ts
+++ b/packages/ag-charts-community/src/chart/toolbar/toolbar.ts
@@ -7,7 +7,6 @@ import { BBox } from '../../scene/bbox';
 import { setAttribute, setAttributes } from '../../util/attributeUtil';
 import { createElement, getWindow } from '../../util/dom';
 import { initToolbarKeyNav, makeAccessibleClickListener } from '../../util/keynavUtil';
-import { Logger } from '../../util/logger';
 import { clamp } from '../../util/number';
 import { ObserveChanges } from '../../util/proxy';
 import { BOOLEAN, Validate } from '../../util/validation';
@@ -148,13 +147,9 @@ export class Toolbar extends BaseModuleInstance implements ModuleInstance {
         }
         this.toggleVisibilities();
 
-        const dragStates = InteractionState.Default | InteractionState.Annotations;
-        const seriesRegion = ctx.regionManager.getRegion('series');
         this.destroyFns.push(
             ctx.interactionManager.addListener('hover', this.onHover.bind(this), InteractionState.All),
             ctx.interactionManager.addListener('leave', this.onLeave.bind(this), InteractionState.All),
-            seriesRegion.addListener('drag-start', this.onStopPointerEvents.bind(this), dragStates),
-            seriesRegion.addListener('drag-end', this.onResumePointerEvents.bind(this), InteractionState.All),
             ctx.toolbarManager.addListener('button-toggled', this.onButtonToggled.bind(this)),
             ctx.toolbarManager.addListener('button-updated', this.onButtonUpdated.bind(this)),
             ctx.toolbarManager.addListener('group-toggled', this.onGroupToggled.bind(this)),
@@ -227,24 +222,6 @@ export class Toolbar extends BaseModuleInstance implements ModuleInstance {
         this.translateFloatingElements(FloatingTop, false);
     }
 
-    // AG-12695 Temporarily set `pointer-events: none` on the annotationOptions when dragging, because the
-    // buttons block to mouse from hovering over the canvas.
-    private readonly oldPointerEvents: Map<HTMLButtonElement, string> = new Map();
-    private onStopPointerEvents() {
-        if (this.oldPointerEvents.size !== 0) {
-            Logger.errorOnce('unexpected drag-start event');
-            return;
-        }
-        this.groupButtons['annotationOptions'].forEach((b) => {
-            this.oldPointerEvents.set(b, b.style.pointerEvents);
-            b.style.pointerEvents = 'none';
-        });
-    }
-    private onResumePointerEvents() {
-        this.oldPointerEvents.forEach((oldPointerEvent, b) => (b.style.pointerEvents = oldPointerEvent));
-        this.oldPointerEvents.clear();
-    }
-
     private onGroupChanged(group: ToolbarGroup) {
         if (this[group] == null || this.groupProxied.has(group)) return;
 
diff --git a/packages/ag-charts-community/src/dom/domManager.ts b/packages/ag-charts-community/src/dom/domManager.ts
index 5bc18434c..fe84e43c9 100644
--- a/packages/ag-charts-community/src/dom/domManager.ts
+++ b/packages/ag-charts-community/src/dom/domManager.ts
@@ -80,6 +80,18 @@ export class DOMManager extends BaseManager<Events['type'], Events> {
     private readonly observer?: IntersectionObserver;
     private readonly sizeMonitor = new SizeMonitor();
 
+    // AG-12695 Temporarily set `pointer-events: none !important` on the canvas-only when dragging, because some
+    // element like annotation option buttons may block to mouse from hovering over the canvas.
+    private readonly onCanvasProxyDrag = (event: MouseEvent) => {
+        if (event.button === 0) {
+            if (event.type === 'mouseup') {
+                this.rootElements['canvas-overlay'].element.style.removeProperty('pointer-events');
+            } else if (event.type === 'mousedown') {
+                this.rootElements['canvas-overlay'].element.style.setProperty('pointer-events', 'none', 'important');
+            }
+        }
+    };
+
     constructor(container?: HTMLElement) {
         super();
 
@@ -117,11 +129,17 @@ export class DOMManager extends BaseManager<Events['type'], Events> {
         if (container) {
             this.setContainer(container);
         }
+
+        this.rootElements['canvas-proxy'].element.addEventListener('mouseup', this.onCanvasProxyDrag);
+        this.rootElements['canvas-proxy'].element.addEventListener('mousedown', this.onCanvasProxyDrag);
     }
 
     override destroy() {
         super.destroy();
 
+        this.rootElements['canvas-proxy'].element.removeEventListener('mouseup', this.onCanvasProxyDrag);
+        this.rootElements['canvas-proxy'].element.removeEventListener('mousedown', this.onCanvasProxyDrag);
+
         this.observer?.unobserve(this.element);
         if (this.container) {
             this.sizeMonitor.unobserve(this.container);
```

PR #2686, opened 2024-10-09
AG-12094 Add volume to measurer statistics

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12094

Move a few functions around, see [this commit](https://github.com/ag-grid/ag-charts/pull/2686/commits/06e522858a200c7c19a0cedfbaca8e4b3099925c) for the main changes.

### DIFF
--- a/packages/ag-charts-enterprise/src/features/annotations/annotations.ts
+++ b/packages/ag-charts-enterprise/src/features/annotations/annotations.ts
@@ -181,7 +181,7 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
                     return;
                 }
 
-                return this.translateNode({ node, datum, translation });
+                return this.translateNode(node, datum, translation);
             },
 
             copy: (index: number) => {
@@ -191,7 +191,7 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
                     return;
                 }
 
-                return this.createAnnotationDatumCopy({ node, datum });
+                return this.createAnnotationDatumCopy(node, datum);
             },
 
             paste: (datum: AnnotationProperties) => {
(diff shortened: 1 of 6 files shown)


packages/ag-charts-enterprise/src/features/annotations/annotations.ts
```diff
@@ -514,15 +541,55 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
...
+        if (!isValidDate(from) || !isValidDate(to) || !this.processedData) return 0;
+
+        return this.processedData.data.reduce((sum, datum) => {
+            const key = datum.keys[0];
+            if (isValidDate(key) && key >= from && key <= to) {
+                return sum + datum.values[0];
+            }
+            return sum;
+        }, 0);
+    }
```
lsjroberts: This will reduce the data on every change to the measurement. It could be memoised, but I wasn't sure if there was a better approach in general to read the data?

alantreadway: This will work, but I wonder if it would be better to delegate to the relevant `Scale` to answer the question of what the distance is?

packages/ag-charts-enterprise/src/features/annotations/annotations.ts
```diff
@@ -514,15 +541,55 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
...
+        if ('getVolume' in datum) {
+            datum.getVolume = this.getDatumRangeVolume.bind(this);
+        }
+    }
+
+    private getDatumRangeVolume(from: Point['x'], to: Point['x']) {
+        if (!isValidDate(from) || !isValidDate(to) || !this.processedData) return 0;
+
+        return this.processedData.data.reduce((sum, datum) => {
+            const key = datum.keys[0];
```
alantreadway: Nit: Not sure if we should rely on an index of `0`? We should probably use the `DataModel.resolveProcessedDataIndexById()` method to resolve this index?

packages/ag-charts-enterprise/src/features/annotations/annotations.ts
alantreadway: Discussed offline, there isn't much we can do right now for the volume sum.

packages/ag-charts-enterprise/src/features/annotations/annotations.ts
lsjroberts: Thanks! Updated.

alantreadway (approved):
LGTM

PR #2691, opened 2024-10-10
Sparklines

jacobp100:
https://ag-grid.atlassian.net/browse/AG-13063


markdurrant (approved):
Website/CSS changes LGTM 👍

packages/ag-charts-website/e2e/examples.spec.ts
```diff
@@ -119,6 +119,11 @@ const exampleOptions: Record<string, Record<string, ExampleOverrides>> = {
         // Buttons have no visible rendering change
         'interaction-range': { skipCanvasUpdateCheck: true },
     },
+
+    sparklines: {
+        // FWs not ready yet
```
alantreadway, line 124: Nit: Could we instead assert that we expect a specific failure for FWs, I think it would be quite easy to forget to re-enable these.

jacobp100: https://ag-grid.atlassian.net/browse/AG-13119

PR #2697, opened 2024-10-10
Remove extraneous canvas instantiations from sprite renderer

jacobp100:


olegat (changes requested):
I suspect that this PR is necessary because we're allocating multiple `Legend` instances and want to reduce memory consumption. Can you confirm this?

A sprite-renderer instance without an offscreen canvas instance isn't really a renderer at all. Would it be simpler and more efficient to make `Legend.spriteRenderer` an optional property?

jacobp100:
https://github.com/ag-grid/ag-charts/pull/2718

PR #2700, opened 2024-10-10
AG-12074 - Update annotations to snap angle on keyDown in drag state

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-12074

### DIFF
--- a/packages/ag-charts-enterprise/src/features/annotations/annotations.ts
+++ b/packages/ag-charts-enterprise/src/features/annotations/annotations.ts
@@ -110,6 +110,7 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
     private readonly defaults = new AnnotationDefaults();
     private dataModel?: _ModuleSupport.DataModel<any, any>;
     private processedData?: _ModuleSupport.ProcessedData<any>;
+    private hoverCoords?: _ModuleSupport.Vec2;
 
     // Elements
     private seriesRect?: _Scene.BBox;
@@ -172,6 +173,10 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
                 );
 
                 return hovered;
+            },
+
+            getHoverCoords: () => {
+                return this.hoverCoords;
             },
 
             getNodeAtCoords: (coords: _ModuleSupport.Vec2, active: number) => {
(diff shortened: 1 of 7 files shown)


packages/ag-charts-enterprise/src/features/annotations/annotations.ts
```diff
@@ -1162,13 +1177,9 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
...
-                ? invertCoords(
-                      snapToAngle(offset, origin ? convertPoint(origin, context) : Vec2.origin(), angleStep),
-                      context
-                  )
-                : invertCoords(offset, context);
+        this.hoverCoords = offset;
+
+        const point = this.getPointFn(shiftKey, offset, context);
 
         state.transition('hover', { offset, point });
```
lsjroberts, line 1184: The state machine could store the `offset` passed through here and then it wouldn't need to use `getHoverCoords()` and we could avoid storing that state on the `Annotations` module class.

Then the state machine would also have all the data that is passed to `getPointFn()`. So it should be able to do that internally as well. 

It has `snapping` state, the hover `offset` and the `context`, so we should be able to push all that down into the state machine?

We should try to avoid adding any state to the `Annotations` module class beyond the user options and `annotationData`.

packages/ag-charts-enterprise/src/features/annotations/states/dragState.ts
```diff
@@ -42,13 +56,8 @@ export class DragStateMachine<
...
             dragging: {
-                keyDown: ({ shiftKey }: { shiftKey: boolean }) => {
-                    ctx.setSnapping(shiftKey);
-                },
-
-                keyUp: ({ shiftKey }: { shiftKey: boolean }) => {
-                    ctx.setSnapping(shiftKey);
-                },
+                keyDown: onKeyChange,
+                keyUp: onKeyChange,
```
lsjroberts: `actionKeyChange` to match the other action fns

lsjroberts (changes requested):

packages/ag-charts-enterprise/src/features/annotations/annotations.ts
manapeirov: Discussed offline and decided to leave as is

alantreadway (approved):
LGTM, @lsjroberts confirmed he's happy to proceed offline.

PR #2702, opened 2024-10-11
AG-13136 In-place layer rendering

jacobp100:
https://ag-grid.atlassian.net/browse/AG-13136


packages/ag-charts-community/src/scene/layerNew.ts
```diff
@@ -0,0 +1,163 @@
+import { BBox } from './bbox';
+import type { HdpiCanvas } from './canvas/hdpiCanvas';
+import { nodeCount } from './debug.util';
+import { Group } from './group';
+import type { LayersManager, ZIndexSubOrder } from './layersManager';
+import { type ChildNodeCounts, RedrawType, type RenderContext } from './node';
+import { Translatable } from './transformable';
+
+export class LayerNew extends Group {
```
alantreadway: Can we roll this back into `Layer` please, with the new behaviour controlled by a constructor option, it makes it much easier to reason about the difference between this and the existing `Layer` implementation?

packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -258,18 +261,19 @@ export abstract class Chart extends Observable {
         const container = resources?.container ?? options.processedOptions.container ?? undefined;
 
         const root = new Group({ name: 'root' });
-        const titleGroup = new Layer({ name: 'titles', zIndex: ZIndexMap.SERIES_LABEL });
         // Prevent the scene from rendering chart components in an invalid state
         // (before first layout is performed).
         root.visible = false;
-        root.append(titleGroup);
+        root.append(this.backgroundRoot);
```
alantreadway: So I don't think these changes are necessarily needed after #2705, we just need to add a z-index on the series root.

packages/ag-charts-community/src/scene/layerNew.ts
jacobp100: Updated

packages/ag-charts-community/src/chart/chart.ts
jacobp100: Updated

packages/ag-charts-community/src/scene/layer.ts
```diff
@@ -125,16 +105,95 @@ export class Layer extends Group {
...
+        ctx.save();
+        ctx.resetTransform();
+        ctx.globalAlpha = this.getComputedOpacity();
+        this.layer.drawImage(ctx);
+        ctx.restore();
+
+        return true;
+    }
+
+    override render(renderCtx: RenderContext) {
```
alantreadway, line 153: Nit: Maybe move this before the new methods above to better preserve line history? (Might need to rebase/squash to achieve this)

packages/ag-charts-community/src/scene/layer.ts
```diff
@@ -125,16 +105,95 @@ export class Layer extends Group {
...
+            return; // Nothing to do.
+        }
+
+        if (forceRender !== 'dirtyTransform') {
+            forceRender = isChildDirty || this.dirtyZIndex;
+        }
+
+        if (this.dirtyZIndex) {
+            this.sortChildren(Group.compareChildren);
+        }
```
alantreadway, line 179: Side-note: We should review the overlap with `Group.render()`, as much of the logic here is repeated there.

alantreadway (commented):
Thanks, this is looking much more concise :-)

packages/ag-charts-community/src/scene/layer.ts
jacobp100: This is from here - https://github.com/ag-grid/ag-charts/pull/2705

packages/ag-charts-community/src/scene/layer.ts
jacobp100: Will do this after you merge your z-index branch - otherwise I'll get conflicts

alantreadway: Merged

PR #2705, opened 2024-10-11
Ensure scene-graph child z-order is persistent across renders.

alantreadway:
Reverts removal of persisted z-order in #2582.

Before:
Screenshot 2024-10-11 at 12 46 42

After:
Screenshot 2024-10-11 at 12 50 59


packages/ag-charts-community/src/scene/node.ts
```diff
@@ -193,6 +193,21 @@ export abstract class Node {
...
 
+    protected sortChildren(compareFn?: (a: Node, b: Node) => number) {
+        this.dirtyZIndex = false;
+        if (!this.childNodes) return;
+
+        // Virtual children need to be sorted on the fly.
+        if (this.hasVirtualChildren()) return;
+
+        // Sort children, and re-add in new order (Set preserves insertion order).
+        const sortedChildren = [...this.childNodes].sort(compareFn);
```
jacobp100, line 204: Would it be easier if we just kept `children` as an array? In theory, you couldn't have duplicate children, because you remove a child from its parent before adding it

jacobp100 (approved):

packages/ag-charts-community/src/scene/node.ts
alantreadway: @iMoses any thoughts? This was originally an array, it's not clear to me if it's practically more useful to be an Array or a Set?

iMoses: If there's a good reason, and caching is a good reason, then I'm not against.
My main concern is that the implementation doesn't leak and for `children()` to remain a generator, preventing potentially copying very large arrays.

PR #2710, opened 2024-10-11
AG-12886 - Prevent text input focus causing window jump/scroll

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-12886


alantreadway (approved):
LGTM

PR #2711, opened 2024-10-11
AG-12968 - Only hide crosshairs if zoom has actually changed to be co…

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-12968


alantreadway (approved):
LGTM

PR #2718, opened 2024-10-14
AG-13132 Remove unnecessary sprite renderer instances

jacobp100:
https://ag-grid.atlassian.net/browse/AG-13132

### DIFF
--- a/packages/ag-charts-community/src/chart/legend.ts
+++ b/packages/ag-charts-community/src/chart/legend.ts
@@ -189,7 +189,7 @@ export class Legend extends BaseProperties {
         LegendMarkerLabel
     );
 
-    private readonly spriteRenderer = new SpriteRenderer();
+    private spriteRenderer: SpriteRenderer | undefined = undefined;
 
     private readonly oldSize: [number, number] = [0, 0];
     private pages: Page[] = [];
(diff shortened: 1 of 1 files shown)


packages/ag-charts-community/src/chart/legend.ts
```diff
@@ -467,7 +468,7 @@ export class Legend extends BaseProperties {
             markerLabel.fontSize = fontSize;
             markerLabel.fontFamily = fontFamily;
 
-            const paddedSymbolWidth = this.updateMarkerLabel(markerLabel, datum, spriteDims);
+            const paddedSymbolWidth = this.updateMarkerLabel(this.spriteRenderer!, markerLabel, datum, spriteDims);
```
olegat, line 471: optional nit: would using a for-loop instead of `Selection.each()` avoid the need to add `!`?

olegat (approved):
LGTM. Much better and cleaner than before I think. Thanks, nicely done! 😃

packages/ag-charts-community/src/chart/legend.ts
alantreadway: From a risk perspective let's follow up on this distinctly, I think this is probably the minimalist change to remove the redundant allocation when there is no legend for the release.

alantreadway (approved):
LGTM

packages/ag-charts-community/src/chart/legend.ts
olegat: SGTM.

PR #2719, opened 2024-10-14
Sonar - Refactor methods that use boolean selector param

lsjroberts:
See https://sonarcloud.io/organizations/ag-grid/rules?open=typescript%3AS2301&rule_key=typescript%3AS2301

While I agree that `doThing(false)` is bad, it does add increase the lines of code in some instances (though to be fair, with greater clarity).


alantreadway (approved):
LGTM

PR #2722, opened 2024-10-14
AG-12973 - Reset should de-select and pop annotations interaction states

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-12973


alantreadway (approved):
LGTM

PR #2725, opened 2024-10-14
AG-12876 - Fix rapid stacked option toggling leading to corrupt series grouping state.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-12876

### DIFF
--- a/packages/ag-charts-community/src/chart/series/series.ts
+++ b/packages/ag-charts-community/src/chart/series/series.ts
@@ -279,10 +279,10 @@ export abstract class Series<
         const { internalId, type, visible } = this;
 
         if (prev) {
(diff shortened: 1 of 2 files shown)


packages/ag-charts-community/src/chart/series/series.ts
```diff
@@ -464,7 +464,7 @@ export abstract class Series<
     }
 
     private visibleMaybeChanged() {
-        this.ctx.seriesStateManager.registerSeries(this);
+        this.ctx.seriesStateManager.updateSeries(this);
```
alantreadway, line 467: This is the actual fix - previously `destroy()`ed series were having visibility updated (due to the nature of queued pending updates), leading to re-registration; we now differentiate between create vs. update cases.

lsjroberts (approved):
LGTM

PR #2730, opened 2024-10-14
AG-13110 TC1 Update zoom pan when focus indicator in outside series rect

olegat:
https://ag-grid.atlassian.net/browse/AG-13110

### DIFF
--- a/packages/ag-charts-community/src/chart/interaction/zoomManager.ts
+++ b/packages/ag-charts-community/src/chart/interaction/zoomManager.ts
@@ -148,7 +148,7 @@ export class ZoomManager extends BaseManager<ZoomEvents['type'], ZoomEvents> imp
         this.applyChanges(callerId);
     }
 
-    panToBBox(callerId: string, seriesRect: BBox, target: BBoxValues) {
+    public panToBBox(callerId: string, seriesRect: BBox, target: BBoxValues) {
         const { x: zoomX, y: zoomY } = this.getZoom() ?? {};
         if (!zoomX || !zoomY) return;
         if (target.width > seriesRect.width || target.height > seriesRect.height) {
--- a/packages/ag-charts-community/src/chart/interaction/zoomManager.ts
+++ b/packages/ag-charts-community/src/chart/interaction/zoomManager.ts
@@ -1,6 +1,9 @@
 import type { AgZoomRange, AgZoomRatio } from 'ag-charts-types';
 
 import type { MementoOriginator } from '../../api/state/memento';
+import type { BBox } from '../../scene/bbox';
+import { Logger } from '../../sparklines-util';
+import type { BBoxValues } from '../../util/bboxinterface';
 import { deepClone } from '../../util/json';
 import { StateTracker } from '../../util/stateTracker';
 import { isFiniteNumber, isObject } from '../../util/type-guards';
@@ -52,6 +55,22 @@ export type ChartAxisLike = {
 };
 
 type ZoomEvents = ZoomChangeEvent | ZoomPanStartEvent | ZoomRestoreEvent;
+
+function calcAxesPanTranslate(viewportMin: number, viewportMax: number, targetMin: number, targetMax: number): number {
+    const mindiff = targetMin - viewportMin;
+    const maxdiff = targetMax - viewportMax;
+    return Math.abs(mindiff) < Math.abs(maxdiff) ? mindiff : maxdiff;
+}
+
+function calcBBoxPanTranslate(viewport: BBox, target: BBoxValues): { dx: number; dy: number } {
+    const dx = calcAxesPanTranslate(viewport.x, viewport.x + viewport.width, target.x, target.x + target.width);
+    const dy = calcAxesPanTranslate(viewport.y, viewport.y + viewport.height, target.y, target.y + target.height);
+    return { dx, dy };
+}
+
+function normalize(screenMin: number, min: number, screenMax: number, max: number, target: number): number {
+    return min + (max - min) * ((target - screenMin) / (screenMax - screenMin));
+}
 
 /**
  * Manages the current zoom state for a chart. Tracks the requested zoom from distinct dependents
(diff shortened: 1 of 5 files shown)


packages/ag-charts-community/src/chart/interaction/zoomManager.ts
```diff
@@ -144,6 +148,18 @@ export class ZoomManager extends BaseManager<ZoomEvents['type'], ZoomEvents> imp
         this.applyChanges(callerId);
     }
 
+    panToBBox(callerId: string, seriesRect: BBox, target: BBoxValues) {
```
lsjroberts: nit: to match the other function styles in this file, helps make the intention of the api clearer

```suggestion
    public panToBBox(callerId: string, seriesRect: BBox, target: BBoxValues) {
```

packages/ag-charts-community/src/chart/interaction/zoomManager.ts
```diff
@@ -144,6 +148,18 @@ export class ZoomManager extends BaseManager<ZoomEvents['type'], ZoomEvents> imp
...
+        const { x: zoomX, y: zoomY } = this.getZoom() ?? {};
+        if (!zoomX || !zoomY) return;
```
lsjroberts: This should default to using `1` for `x` and `y` if they don't have a value yet.

packages/ag-charts-community/src/chart/keyboardUtil.ts
```diff
@@ -14,13 +14,19 @@ function computeCenter(bboxOrPath: Path | BBox | undefined) {
...
+function drawPickedFocus(focusIndicator: FocusIndicator | undefined, pick: PickProperties) {
     const { bounds, showFocusBox } = pick;
     if (showFocusBox) {
         focusIndicator?.updateBounds(bounds);
     }
 }
 
+export function getPickedFocusBBox({ bounds }: PickProperties): BBox {
+    if (bounds instanceof BBox) return bounds;
+    else if (bounds != null) return bounds.getBBox();
```
lsjroberts: nit: don't need the `else` since the previous block returns.

```suggestion
    if (bounds != null) return bounds.getBBox();
```

packages/ag-charts-community/src/util/panToBBox.ts
```diff
@@ -0,0 +1,64 @@
...
+type XYRatios = { x: Ratios, y: Ratios };
+
+function normalize(screenMin: number, min: number, screenMax: number, max: number, target: number): number {
+    return min + (max - min) * ((target - screenMin) / (screenMax - screenMin));
+}
+
+function unnormalize(screenMin: number, min: number, screenMax: number, max: number, ratio: number): number {
+    return screenMin + (ratio - min) * ((screenMax - screenMin) / (max - min));
+}
+function calcWorld(viewportMin: number, viewportMax: number, ratio: Ratios): [number, number] {
```
lsjroberts: nit:

```suggestion
}

function calcWorld(viewportMin: number, viewportMax: number, ratio: Ratios): [number, number] {
```

packages/ag-charts-community/src/util/panToBBox.ts
```diff
@@ -0,0 +1,64 @@
...
+    return clamp(worldMin, viewportMin + diff, worldMax);
+}
+
+// The calculations of the new desired viewport (i.e. ZoomState) is done in pixel coords (unnormalised).
+// The desired (x, y) for the new viewport is found, the pixel coords are converted into normalized values
+export function calcPanToBBoxRatios(viewport: BBoxValues, ratioX: Ratios, ratioY: Ratios, target: BBoxValues): XYRatios {
+    const [targetXMin, targetXMax] = [target.x, target.x + target.width];
+    const [targetYMin, targetYMax] = [target.y, target.y + target.height];
+    const [viewportXMin, viewportXMax] = [viewport.x, viewport.x + viewport.width];
+    const [viewportYMin, viewportYMax] = [viewport.y, viewport.y + viewport.height];
```
lsjroberts: Entirely optional, but I added a `Vec4.from(bbox) === { x1: x, y1: y, x2: x + width, y2: y + height }` util in `vector4.ts` as I was doing lots of work with bboxes and coordinates.

It might help with some of the repetition (and similarly in the `containsBBox()` function perhaps).

```suggestion
    const target = Vec4.from(target);
    const viewport = Vec4.from(viewport);
```

packages/ag-charts-community/src/util/panToBBox.ts
```diff
@@ -0,0 +1,64 @@
...
+    worldMin: number,
+    worldMax: number,
+    viewportMin: number,
+    viewportMax: number,
+    targetMin: number,
+    targetMax: number
+): number {
+    if (viewportMin <= targetMin && targetMax <= viewportMax) return viewportMin;
+    const mindiff = targetMin - viewportMin;
+    const maxdiff = targetMax - viewportMax;
```
lsjroberts: nit:

```suggestion
    const minDiff = targetMin - viewportMin;
    const maxDiff = targetMax - viewportMax;
```

lsjroberts (approved):
LGTM other than some nits.

lsjroberts:
> Snaspshots automatically updated, please review before merge: [diff](https://github.com/ag-grid/ag-charts/compare/AG-13110/focus_indicator_outside_series_rect...gha/snapshots-AG-13110/focus_indicator_outside_series_rect)

The "pan to here" snapshot change is quite significant, is that correct?

olegat:
> The "pan to here" snapshot change is quite significant, is that correct?

Yes I noticed this too. It's probably incorrect. I need to look into it.

packages/ag-charts-community/src/util/panToBBox.ts
olegat: Good suggestion 👍  `Vector4` feels a bit like a misnomer. Perhaps `Line2D` is a better name. But that's a discussion for another time.

packages/ag-charts-community/src/chart/interaction/zoomManager.ts
olegat: Could you clarify what you mean? At this point, `x` and `y` are of type `{ min: number, max: number } | undefined`.

My thoughts are that, when zooming is disabled, `x` and `y` would be `undefined` which lets us short-circuit and skip panning.

lsjroberts: Ah yeh, I was just thinking of the enabled case, so you're probably right it can skip the panning.

The only case might be if `x` is defined but y is not, like with just the navigator enabled? I'm not sure if that'd need different handling? Might work fine.

lsjroberts (approved):
LGTM

packages/ag-charts-community/src/chart/interaction/zoomManager.ts
olegat: That's a good point. I'll add explicit typing to ensure that `panToBBox` works when only 1 axis has zoom enabled

olegat: I've changed `calcPanToBBoxRatios` to accept optional axes zoom ratios.

olegat:
> [!NOTE]
> The e2e test failures were genuine. There `panToBBox()` call was missing a `isFocusVisible()` check.

alantreadway (approved):
LGTM

PR #2731, opened 2024-10-14
Layers 3

jacobp100:


packages/ag-charts-enterprise/src/features/crosshair/__image_snapshots__/crosshair-test-ts-crosshair-should-follow-mouse-pointer-on-series-hover-8-snap.png
iMoses, line 1: This z-index issue needs to be addressed

PR #2732, opened 2024-10-15
AG-12985 - Fix radial gauge download.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-12985


packages/ag-charts-community/src/module/optionsModule.ts
```diff
@@ -266,7 +265,7 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
         };
 
         const processedSeries = (options.series as SeriesOptionsTypes[])?.map((series) => {
-            series.type ??= defaultSeriesType;
+            series.type ??= this.getDefaultSeriesType(options);
```
alantreadway, line 268: NOTE: This now lazily calls `this.getDefaultSeriesType()` if there is no `series.type`.

packages/ag-charts-community/src/chart/mapping/types.ts
```diff
@@ -42,19 +42,18 @@ export type SeriesOptionsTypes =
...
 export type SeriesType = SeriesOptionsTypes['type'];
 
-export function optionsType(input: { series?: { type?: SeriesType }[] }): NonNullable<SeriesType> {
-    return input.series?.[0]?.type ?? 'line';
+export function optionsType(input: { series?: { type?: SeriesType }[] }): SeriesType {
+    const { series } = input;
+    if (series) {
+        return input.series?.[0]?.type ?? 'line';
+    }
+    return undefined;
```
iMoses:
```suggestion
    if (series) {
        return series?.[0]?.type ?? 'line';
    }
```
nit: cleaner + I'm on the side that thinks `undefined` shouldn't be used as a value 😄

iMoses (approved):
LGTM, one nit

packages/ag-charts-community/src/chart/mapping/types.ts
lsjroberts: My version:

```suggestion
    if (!series) return;
    return series[0]?.type ?? 'line';
```

jacobp100 (approved):

packages/ag-charts-community/src/chart/mapping/types.ts
alantreadway: Thanks both, was meaning to clean this up.

PR #2736, opened 2024-10-15
AG-13051 Update tooltip&highlight on click events

olegat:


alantreadway (approved):
LGTM, one snapshot failure though

olegat:
At first glance, the snapshot failure look like a genuine bug regression on an edge case. I'll look into it.

packages/ag-charts-enterprise/src/features/zoom/__image_snapshots__/zoom-test-ts-zoom-listeners-should-not-zoom-out-when-double-clicking-a-node-1-snap.png
olegat, line 1: > [!IMPORTANT]
> This new snapshot is correct. The old snapshot does not include the crosshair/highlight because the [`doubleClickAction`](https://github.com/ag-grid/ag-charts/blob/2492a51fdaef4e867620d7d6d0356ac0755030d1/packages/ag-charts-community/src/chart/test/utils.ts#L352) test function does not dispatch a `'mousemove'` event before the first `click` event.

I'll update the `clickAction` and `doubleClickAction` to fire `'mousemove'` events in a separate PR because doing so triggers other unrelated snapshot updates.

alantreadway (approved):
LGTM

PR #2738, opened 2024-10-15
AG-13163 Prevent default on all unconsumed arrow keys in our DOM

olegat:
https://ag-grid.atlassian.net/browse/AG-13163

This is to stop page scrolling in some browsers like Firefox and Chrome.


alantreadway (approved):
LGTM

PR #2739, opened 2024-10-15
AG-13051 Dispatch mousemouse event in click/dblclick event tests

olegat:


olegat:
> [!NOTE]
> The current CI failures look like they are caused by the same race condition in `dataSource.test.ts` that's also affecting https://github.com/ag-grid/ag-charts/pull/2759. We'll need to merge `latest` into `next` and rebase this PR once that other PR is merged in.

olegat:
Closing. We may have genuine cases where we want to test what happens when a `'click'` event is fired without `'mousemove'` event.

PR #2743, opened 2024-10-15
AG-12522 - Fix firing of click event after drag.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-12522

### DIFF
--- a/packages/ag-charts-community/src/chart/interaction/interactionManager.ts
+++ b/packages/ag-charts-community/src/chart/interaction/interactionManager.ts
@@ -10,8 +10,6 @@ import { type PreventableEvent, type Unpreventable, buildPreventable, dispatchTy
 
 export { InteractionState };
 
-const DRAG_COOLDOWN_MS = 50;
-
 export const DRAG_INTERACTION_TYPES = ['drag-start', 'drag', 'drag-end'] as const;
 
 export const POINTER_INTERACTION_TYPES = [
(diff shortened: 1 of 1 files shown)


olegat (changes requested):
> [!CAUTION]
> This solution is not compliant with WCAG 2 Success Criteria 2.2.3:
> > 2.2.3 No Timing, Level AAA
> > Timing is not an essential part of the event or activity presented by the content, except for non-interactive synchronized media and real-time events.
> https://www.w3.org/WAI/WCAG22/quickref/?versions=2.1#no-timing

As we discussed on Slack, another solution could be to record `mousemove` events in `InteractionManager` as we do with `mousedown` and `mouseup` events. AFAICT, this approach would be compliant with WCAG 2. Are there any side-effects from this approach that I am unaware of?

olegat (changes requested):
[deleted]

alantreadway:
> Would it be better to merge this breaking change `next` rather than `latest`?

We don't document the expected type of `event`, so although it is a change in behaviour, it is still correct from an API/contract perspective.

packages/ag-charts-community/src/chart/interaction/interactionManager.ts
```diff
@@ -374,15 +373,15 @@ export class InteractionManager extends InteractionStateListener<InteractionType
                     return;
                 }
                 this.mouseDown = false;
-                return this.recordUp(event) ? 'drag-end' : undefined;
+                return this.recordUp(event) ? 'drag-end' : 'click';
```
olegat, line 376: test

olegat (approved):
LGTM

PR #2749, opened 2024-10-16
Layer compositing

jacobp100:
- Removes `Layer` and makes this a property on `Group` so it can be dynamically turned on and off
- Defaults most things to non-composited
- Series can now specify if they want to be composited, which we only do if there's likely to be a real benefit

We have heuristics that I've sort of guessed here. We should do some benchmarking to find good defaults

I've intentionally set the threshold very low so our examples will have a mix of composited/non-composited, so if there's any bugs, we'll notice them sooner


packages/ag-charts-community/src/scene/group.ts
```diff
@@ -142,21 +215,18 @@ export class Group extends Node {
...
 
-        ctx.save();
         ctx.beginPath();
         ctx.rect(x, y, width, height);
         ctx.clip();
 
         this._debug?.(() => ({
             name: this.opts?.name,
-            clipRect: this.clipRect,
+            clipRect: clipRect,
```
alantreadway, line 216:
```suggestion
            clipRect,
```

alantreadway (approved):
LGTM

packages/ag-charts-community/src/chart/series/seriesLayerManager.ts
```diff
@@ -150,12 +180,13 @@ export class SeriesLayerManager {
...
                 this.seriesRoot.removeChild(groupInfo.group);
-            }
-        }
-        (this as any).groups = {};
-        (this as any).series = {};
+            });
+        });
+
+        (this as any).groups = new Map();
+        (this as any).series = new Map();
```
iMoses: nit: micro-optimisation, I prefer using `for..of` loops, they create less garbage to be collected.

iMoses:
```suggestion
        this.groups.clear();
        this.series.clear();
```
nit: cleaner, and respects being `readonly`

iMoses (approved):
LGTM 👍 
minor nits

alantreadway (approved):
LGTM

PR #2750, opened 2024-10-16
AG-13161 Fix annotation line cap pixel alignments

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13161


packages/ag-charts-website/e2e/toolbar.spec.ts-snapshots/line-8-paste-chromium-linux.png
lsjroberts, line 1: This is a bit more change than expected, but is perhaps nicer anyway as it maintains the angle of the line?

alantreadway (approved):
LGTM

PR #2751, opened 2024-10-16
Reduce cognitive complexity

lsjroberts:
It's rather debatable if some of complexity issues raised are actually simpler when refactored. So just two instances here for now.

Diff is clearest with hiding whitespace.


alantreadway (approved):
LGTM

PR #2756, opened 2024-10-16
AG-13163 Do not call preventDefault on `<select>` elements

olegat:
https://ag-grid.atlassian.net/browse/AG-13163


alantreadway (approved):
LGTM

PR #2759, opened 2024-10-16
AG-13051 Clear tooltip/highlight on series area clicks

olegat:

Calling the onHoverLikeEvent inside the checkSeriesNodeClick branch only clears the tooltip/highlight when the hit-testing finds a node. Therefore, always call the onHoverLikeEvent method when the mouse is clicked.

Also add an e2e test for this case.


alantreadway (approved):
LGTM

olegat:
I think the snapshots need updates. The `onClick` handler now behaves much like a hover event and updates the tooltip/highlight.

These test fire clicks event so a crosshair/highlight should be shown.

https://github.com/ag-grid/ag-charts/blob/162d0607ffe97eac0a9ea0584b2513be5fb84ede/packages/ag-charts-enterprise/src/features/features.test.ts#L96

https://github.com/ag-grid/ag-charts/blob/162d0607ffe97eac0a9ea0584b2513be5fb84ede/packages/ag-charts-enterprise/src/features/data-source/dataSource.test.ts#L80

alantreadway (approved):
LGTM

PR #2760, opened 2024-10-17
Offscreen canvas

jacobp100:
Currently we use unmounted canvas elements for offscreen canvases. Normally any update to a canvas element would cause the browser to at least partially redraw the page, and I'm not so sure they have optimisations to not do this for unmounted canvases. Using OffscreenCanvas provides a better guarantee that updates to the canvas won't lead to redrawing the browser page

### DIFF
--- a/packages/ag-charts-community/src/scene/canvas/hdpiOffscreenCanvas.ts
+++ b/packages/ag-charts-community/src/scene/canvas/hdpiOffscreenCanvas.ts
@@ -1,0 +1,73 @@
+import { getWindow } from '../../util/dom';
+import { hasConstrainedCanvasMemory } from '../../util/userAgent';
+import { clearContext, debugContext } from './canvasUtil';
+
+// Work-around for typing issues with Angular 13+ (see AG-6969),
+type OffscreenCanvasRenderingContext2D = any;
+
+export interface CanvasOptions {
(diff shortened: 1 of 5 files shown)


packages/ag-charts-community/src/scene/canvas/hdpiOffscreenCanvas.ts
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    destroy() {
+        // Workaround memory allocation quirks in iOS Safari by resizing to 0x0 and clearing.
+        // See https://bugs.webkit.org/show_bug.cgi?id=195325.
+        this.canvas.width = 0;
+        this.canvas.height = 0;
+        this.context.clearRect(0, 0, 0, 0);
+
+        (this as any).canvas = null!;
```
iMoses, line 69: nit: Is this necessary?

iMoses (approved):

packages/ag-charts-community/src/scene/canvas/hdpiOffscreenCanvas.ts
jacobp100: It's copied from the normal hdpi canvas - I think just to reduce the impact of a memory leak

PR #2762, opened 2024-10-17
AG-13145 Add typing to state machine event data

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13145

This adds typings to the state machine event data passed in from the transitions to the actions, for example:

```
new StateMachine<'state', { myEvent: { foo: string } }>({
    myEvent: ({ foo }) => {} 
});

state.transition('myEvent', { foo: 'bar' });
```

It can be a bit verbose when events don't require data and end up as `{ foo: undefined }`, but on balance I think it's more useful than not. (It showed up a couple of minor issues in the annotations state machine).


alantreadway (approved):
LGTM

PR #2765, opened 2024-10-17
AG-13166 Do not always show focus indicator on -/+ key presses

olegat:
https://ag-grid.atlassian.net/browse/AG-13166


alantreadway (approved):
LGTM

PR #2776, opened 2024-10-18
AG-13187 Initialise `overrideFocusVisible(false)` when `mode: 'integrated'`

olegat:
https://ag-grid.atlassian.net/browse/AG-13197

### DIFF
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -263,8 +263,8 @@ export abstract class Chart extends Observable {
 
         this.tooltip = new Tooltip();
         this.seriesLayerManager = new SeriesLayerManager(this.seriesRoot, this.highlightRoot, this.annotationRoot);
-        const mode: ChartMode = (options.userOptions as { mode?: ChartMode }).mode ?? this.mode;
-        const ctx = (this.ctx = new ChartContext(this, mode, {
+        this.mode = (options.userOptions as { mode?: ChartMode }).mode ?? this.mode;
+        const ctx = (this.ctx = new ChartContext(this, {
             scene,
             root,
             container,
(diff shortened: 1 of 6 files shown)


packages/ag-charts-community/src/chart/interaction/keyNavManager.ts
```diff
@@ -43,6 +45,8 @@ export class KeyNavManager extends InteractionStateListener<KeyNavEventType, Key
             interactionManager.addListener('focus', (e) => this.onFocus(e), InteractionState.All),
             interactionManager.addListener('keydown', (e) => this.onKeyDown(e), InteractionState.All)
         );
+
+        this.focusIndicator.overrideFocusVisible(mode === 'integrated' ? false : undefined); // AG-13197
```
olegat, line 49: > [!IMPORTANT]
> This line (as well as the removal in `onFocus`) is the key change. The rest of changes is just plumbing/logistics.

packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -263,7 +263,8 @@ export abstract class Chart extends Observable {
 
         this.tooltip = new Tooltip();
         this.seriesLayerManager = new SeriesLayerManager(this.seriesRoot, this.highlightRoot, this.annotationRoot);
-        const ctx = (this.ctx = new ChartContext(this, {
+        const mode: ChartMode = (options.userOptions as { mode?: ChartMode }).mode ?? this.mode;
+        const ctx = (this.ctx = new ChartContext(this, mode, {
```
olegat: > [!WARNING]
> This change is not type-safe, but I cannot find a better way to retrieve the `mode` option because the `Chart.mode` property is still set to its default `'standalone'` value at this point. Open to better suggestions.

PR #2778, opened 2024-10-18
AG-13203 AG-13201 Mark <buttons> dirty only when the legend data changes

olegat:
https://ag-grid.atlassian.net/browse/AG-13201
https://ag-grid.atlassian.net/browse/AG-13203

### DIFF
--- a/packages/ag-charts-community/src/chart/legend.ts
+++ b/packages/ag-charts-community/src/chart/legend.ts
@@ -202,9 +202,9 @@ export class Legend extends BaseProperties {
     private _data: CategoryLegendDatum[] = [];
     private _symbolsDirty = true;
     set data(value: CategoryLegendDatum[]) {
+        this.onDataUpdate(this._data, value);
         this._data = value;
         this._symbolsDirty = true;
-        this.proxyLegendDirty = true;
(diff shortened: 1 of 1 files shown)


packages/ag-charts-community/src/chart/legend.ts
```diff
@@ -369,6 +369,15 @@ export class Legend extends BaseProperties {
         this.proxyLegendDirty = false;
     }
 
+    private onDataUpdate(oldData: CategoryLegendDatum[], newData: CategoryLegendDatum[]) {
+        this.proxyLegendDirty =
+            oldData.length !== newData.length ||
+            oldData.some((_v, index, _a) => {
+                const [newValue, oldValue] = [newData[index], oldData[index]];
+                return newValue.id !== oldValue.id;
```
olegat, line 377: > [!CAUTION]
> I do not know if just checking the legend datum ID is sufficient. This condition may need to be tweaked.

alantreadway: I think it's probably a reasonable approximation - we need to exercise it in Integrated Charts cases to be sure though.

alantreadway (approved):
LGTM

PR #2785, opened 2024-10-21
CRT-559 Change text to use dragStart to ensure new mouse action

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-559

`drag` events from a single mouse move bled into the next state and caused it to create on one drag event and save on the next drag event, which led to it being deleted as the text was empty.


alantreadway (approved):
LGTM

PR #2791, opened 2024-10-21
CRT-535 Add missing docs for measurer extends options

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-535


alantreadway (approved):
LGTM

PR #2793, opened 2024-10-21
CRT-548 Only consider cartesian axes for region manager

jacobp100:
https://ag-grid.atlassian.net/browse/CRT-548


alantreadway (approved):
LGTM

PR #2794, opened 2024-10-21
CRT-558 Ensure display: none is set when legend is hidden

olegat:
https://ag-grid.atlassian.net/browse/CRT-558


alantreadway (approved):
LGTM

PR #2797, opened 2024-10-21
CRT-561 Fix axis button to hide when tab out like crosshairs

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-561


alantreadway (approved):
LGTM

PR #2799, opened 2024-10-21
CRT-535 Remove extendStart & extendEnd from measurer options

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-535


alantreadway (approved):
LGTM

PR #2800, opened 2024-10-21
CRT-529 - Fix zoom pan not working when an annotation is selected

manapeirov:
PR for https://ag-grid.atlassian.net/browse/CRT-529


alantreadway (approved):
LGTM

PR #2803, opened 2024-10-21
CRT-544 - Fix polar keyboard context menu.

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-544


olegat:
LGTM, just make sure that the addition of the `Transformable.toCanvasPoint` call does not break other series that currently work as intended like bars.

olegat (approved):

alantreadway:
> LGTM, just make sure that the addition of the `Transformable.toCanvasPoint` call does not break other series that currently work as intended like bars.

Agreed, confirmed that `BarSeries` is still working after this change - I'll do a quick sanity check of a few others now.

UPDATE: I ran through a bunch of different series types in the gallery (as we need Enterprise examples), and this seems to be working as expected.

PR #2815, opened 2024-10-22
AG-13145 Add state machine hierarchy properties and parallelism

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13145

- Adds properties that are propagated down the hierarchy of a state machine
- Adds parallel state machines

### Property propagation

Previously a child state machine (e.g. `LineStateMachine` when creating a line annotation) would be required to use `ctx.node()` to get the active node.

Now it can just use `this.node`.

This works by copying any property from a parent state machine that is decorated by `StateMachineProperty()` into the child state machine that has the same name and is also decorated by `StateMachineProperty()`.

e.g.

```
// AnnotationsStateMachine
@StateMachineProperty()
private node?: AnnotationScene;
```

```
// LineStateMachine
@StateMachineProperty()
private node?: LineScene;
```

### Parallel State Machines

These pass events to all their children simultaneously, they do not hold a specific state themselves.

The annotations machine uses this to have a `SnappingStateMachine` that is always listening to all relevant events to set the `snapping` property, that is then propagated, even if the main `AnnotationsMainStateMachine` is in a child state.

This means the child state machines do not have to each separately handle setting the snapping value themselves.


lsjroberts:
> Snaspshots automatically updated, please review before merge: [diff](https://github.com/ag-grid/ag-charts/compare/AG-13145-annotation-hierarchy-data...gha/snapshots-AG-13145-annotation-hierarchy-data)

Merged. The change makes sense as when you press enter to save a callout it deselects it. The handle used to remain as it detected the mouse there but it shouldn't since the mouse is not shown while typing until after pressing enter and moving the mouse.

lsjroberts:
> Snaspshots automatically updated, please review before merge: [diff](https://github.com/ag-grid/ag-charts/compare/AG-13145-annotation-hierarchy-data...gha/snapshots-AG-13145-annotation-hierarchy-data)

Merged, expected change.

alantreadway (approved):
LGTM

PR #2816, opened 2024-10-22
Incremental axis cleanup (1)

iMoses:
The reason for removing methods from the axis context is to gradually make other module depend on the scale directly instead of the axis instance/context.


alantreadway (approved):
LGTM

PR #2817, opened 2024-10-22
AG-13233 - Add styleContainer to avoid detached DOM causing styling issues.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-13233

Published as `10.3.0-beta.20241022.1441`

### DIFF
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -70,6 +70,7 @@ const debug = Debug.create(true, 'opts');
 
 export type TransferableResources = {
     container?: HTMLElement;
+    styleContainer?: HTMLElement;
     scene: Scene;
 };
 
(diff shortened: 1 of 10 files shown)


packages/ag-charts-community/src/chart/chartProperties.ts
```diff
@@ -25,6 +25,9 @@ export class ChartProperties extends BaseProperties {
     @Validate(OBJECT.restrict(HTMLElement), { optional: true })
     container?: HTMLElement;
 
+    @Validate(OBJECT.restrict(HTMLElement), { optional: true })
```
iMoses: this file is actually dead code 😶

packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -246,6 +247,7 @@ export abstract class Chart extends Observable {
 
         const scene: Scene | undefined = resources?.scene;
         const container = resources?.container ?? options.processedOptions.container ?? undefined;
+        const styleContainer = resources?.styleContainer ?? options.processedOptions.styleContainer ?? undefined;
```
iMoses:
```suggestion
        const styleContainer = resources?.styleContainer ?? options.processedOptions.styleContainer;
```
nit: `?? undefined` is redundant.

packages/ag-charts-community/src/chart/chartProperties.ts
alantreadway: Removed :+1:

packages/ag-charts-community/src/chart/chart.ts
alantreadway: Fixed

PR #2819, opened 2024-10-22
AG-13167 Defer `focus()` call until after CSS bounds are set

olegat:
https://ag-grid.atlassian.net/browse/AG-13167


alantreadway (approved):
LGTM

PR #2828, opened 2024-10-23
Incremental axis cleanup (3)

iMoses:


packages/ag-charts-community/src/chart/__image_snapshots__/cartesian-chart-test-ts-cartesian-chart-small-chart-height-should-render-chart-correctly-at-height-80-1-snap.png
iMoses, line 1: The change is due to BBox.shrink now allowing width & height to be set to negative values.
Original subtitle spacing does look slightly better in previous version, but that's due to negative bbox sizes.

alantreadway (approved):
LGTM

PR #2835, opened 2024-10-24
AG-13258 - Remove deprecated code for 11.0.0 release.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-13258

Breaking changes:
- `AgPriceVolumePreset.rangeToolbar` is removed, use `rangeButtons` instead.
- `AgPriceVolumePreset.xKey` is removed, use `dateKey` instead.
- `AgPriceVolumePreset.annotations` is removed, use `toolbar` instead.
- Financial Charts `chartType` of `range-area` is removed. Use `hlc` instead.
- L10n string `toolbarAnnotationsColor` is removed, use `toolbarAnnotationsLineColor` instead.
- All toolbar icons with `-legacy` suffix have been removed.
- `AgChartSerializableState` is removed, use `AgChartState` instead.


packages/ag-charts-types/src/chartBuilderOptions.ts
```diff
@@ -202,17 +203,6 @@ export interface ImageDataUrlOptions {
 
 export interface AgChartState {
     version: string;
-    annotations?: AgChartSerializableState;
```
alantreadway, line 205: @lsjroberts does this change look right?

lsjroberts (approved):
LGTM, very satisfying 🤩

PR #2837, opened 2024-10-24
AG-3390 Improve docs menu contrast

kylerphillips:


markdurrant (approved):
Nitpick: Add a space above nested items in Sass.

Otherwise looking good.

PR #2846, opened 2024-10-25
AG-13213 Change crossline label position options to kebab-case

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13213


alantreadway (approved):
LGTM

PR #2847, opened 2024-10-25
Incremental axis cleanup (4)

iMoses:


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1417,20 +1418,18 @@ export abstract class Chart extends Observable {
...
+                [ChartAxisDirection.Y]: [],
+            };
 
-        chart.axes.forEach((axis) => {
-            const { id } = axis;
-            const node = axis.getRegionNode();
-            if (node != null) {
-                axisGroups[axis.direction].push({ id, node });
+            for (const axis of chart.axes) {
+                if (axis instanceof CartesianAxis) {
```
alantreadway, line 1428: Nit: I'm not sure about this change, it's probably better if `Chart` doesn't vary it's behaviour here based upon the type of `axis`, but the `axis` controls this.

iMoses: My reasoning is that this is a patch and a patch shouldn't change public APIs because that's how a temporarily problem becomes permanent :)

alantreadway: Discussed offline.

PR #2854, opened 2024-10-28
AG-13065 Add class support for series tooltips

jacobp100:
https://ag-grid.atlassian.net/browse/AG-13065


iMoses:
The reason most known APIs use `className` instead of `class` is that the later is a reserved keyword in ecmascript, which can cause some annoyance when using common features when it'll force you to rename local variable. (i.e. destructuring).
```
// Can't do that
const { name, class, size } = this;
// ...
return { name, class, size };

// Must change everywhere to
const { name, class: className, size } = this;
// ...
return { name, class: className, size };
```
For this reason I do think we should opted for a better/consistent UX and not use reserved keywords when naming options.

jacobp100:
@iMoses we need to fix this because we have different names in different places. However, this has to match the chart tooltip options

PR #2858, opened 2024-10-28
AG-13142 Fix bug with Chrome sometimes focusing on aria-hidden elems

olegat:
https://ag-grid.atlassian.net/browse/AG-13142

When a screen-reader is enabled, Chrome allows the user to browse to elems with `tabindex="-1"` even if `aria-hidden="true"`. We do not want that, therefore remove the tabindex attribute on the hidden swapchain element to make it non-focusable.

### DIFF
--- a/packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
+++ b/packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
@@ -6,7 +6,6 @@ import type { BBox } from '../../scene/bbox';
 import type { TranslatableGroup } from '../../scene/group';
 import { Transformable } from '../../scene/transformable';
 import { createId } from '../../util/id';
-import { Logger } from '../../util/logger';
 import { clamp } from '../../util/number';
 import type { TypedEvent } from '../../util/observable';
 import { debouncedAnimationFrame } from '../../util/render';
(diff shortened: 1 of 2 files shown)


packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
```diff
@@ -169,7 +170,11 @@ export class SeriesAreaManager extends BaseManager {
     }
 
     public setTabIndex(tabIndex: number) {
-        this.swapChain.setTabIndex(tabIndex);
+        // FIXME: Remove keyboard.tabIndex option in next major release
+        if (tabIndex !== 0 && tabIndex !== -1) {
+            Logger.warnOnce(`Tab indices other than 0 and -1 are deprecated`);
+        }
```
alantreadway: No need for this, we're in a major release cycle so let's just put the new constraint in place and add `Breaking Changes` notes to the JIRA?

olegat: I agree, but we need consensus with David before we can go ahead with that.

alantreadway: Discussed offline.

olegat: Removed the warning. The `tabIndex > 0` issue will be addressed in a separate PR.

alantreadway (approved):
LGTM

PR #2863, opened 2024-10-29
AG-11993 Remove navigator min & max options

lsjroberts:


packages/ag-charts-community/src/chart/interaction/zoomManager.ts
```diff
@@ -103,13 +102,20 @@ export class ZoomManager extends BaseManager<ZoomEvents['type'], ZoomEvents> imp
...
-    public guardMemento(blob: unknown): blob is ZoomMemento {
-        return (
-            isObject(blob) && (blob.ratioX != null || blob.ratioY != null || blob.rangeX != null || blob.rangeY != null)
-        );
+    public guardMemento(blob: unknown): blob is ZoomMemento | undefined {
+        if (blob == null) return true;
+        if (!isObject(blob)) return false;
+
+        for (const key of Object.keys(blob)) {
+            if (!['ratioX', 'ratioY', 'rangeX', 'rangeY'].includes(key)) {
```
alantreadway: Nit: How type-safe is this array of keys - would it be better to define a `const` array of element type `keyof ZoomMemento` so we'll catch contract changes here?

alantreadway (commented):
LGTM, but CI is failing

packages/ag-charts-community/src/chart/interaction/zoomManager.ts
lsjroberts: Good point, fixed.

alantreadway (approved):
LGTM

PR #2871, opened 2024-10-30
AG-10564 - Only add datum to legend item formatter params if not unde…

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-10564


alantreadway (approved):
LGTM

PR #2874, opened 2024-10-30
AG-3390 - Update algolia script so unix chrome path is the default

taktran:
To use the MacOSX chrome, pass in `--local`


alantreadway (approved):
LGTM

PR #2875, opened 2024-10-30
AG-13282 Upgrade target to es2022

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13282


lsjroberts:
Pending further discussion.

PR #2885, opened 2024-10-31
Column data model (4 - Heatmap, Map, Chord, Sankey Series)

jacobp100:


packages/ag-charts-website/src/content/docs/bullet-series-test/index.mdoc
alantreadway, line 1: Good spot, we need to remove `bullet` series in this release too!

alantreadway (approved):
LGTM

PR #2886, opened 2024-10-31
AG-13304 - Improve sparkline performance.

alantreadway:


packages/ag-charts-community/src/chart/mapping/themes.ts
alantreadway, line 1: Changes here are core to the improvement; caching of `ChartTheme` instances.

packages/ag-charts-community/src/chart/factory/processEnterpriseOptions.ts
```diff
@@ -86,3 +88,14 @@ export function removeUsedEnterpriseOptions<T extends Partial<AgChartOptions>>(o
         );
     }
 }
+
+export function removeUnusedEnterpriseOptions<T extends Partial<AgChartOptions>>(options: T) {
```
alantreadway, line 92: This function prevents unused/disabled enterprise modules from being initialised downstream by completely removing their configuration.

packages/ag-charts-community/src/api/preset/sparkline.ts
```diff
@@ -147,7 +149,9 @@ const SPARKLINE_THEME: AgChartTheme = {
     },
 };
 
-function setInitialBaseTheme(
+const setInitialBaseTheme = simpleMemorize(createInitialBaseTheme);
```
alantreadway, line 152: Required to ensure correct caching of sparkline themes.

packages/ag-charts-community/src/api/preset/sparkline.ts
```diff
@@ -74,6 +75,7 @@ const SPARKLINE_THEME: AgChartTheme = {
             animation: {
                 enabled: false,
             },
+            toolbar: { enabled: false },
```
alantreadway, line 78: Prevents toolbar module initialisation, and DOM additions that are relatively expensive.

packages/ag-charts-community/src/module/optionsModule.ts
```diff
@@ -123,12 +111,13 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
...
 
-            const presetParams = options as any as AgPresetOptions;
+            const presetParams = userOptions as any as AgPresetOptions;
 
             // Note financial charts defines the theme in its returned options
             // so we need to get the theme before and after applying the preset
             const presetSubType = (options as any).type as keyof AgPresetOverrides | undefined;
-            const presetTheme = presetSubType != null ? getChartTheme(options.theme).presets[presetSubType] : undefined;
+            const presetTheme =
+                presetSubType != null ? getChartTheme(userOptions.theme).presets[presetSubType] : undefined;
```
alantreadway, line 120: Tweaks here are to ensure caching of themes works.

packages/ag-charts-community/src/api/preset/sparkline.ts
jacobp100: Should this be a general default?

packages/ag-charts-community/src/integrated-charts-theme.ts
```diff
@@ -10,7 +9,13 @@ export * from './chart/themes/constants';
 export * from './chart/themes/util';
 export * from './module/theme';
 
-export const themes: Record<AgChartThemeName, AgChartTheme> = mapValues(
-    themeFactories,
-    (themeFactory) => themeFactory?.()!
-);
+export const themes: Record<AgChartThemeName, AgChartTheme> = {} as any;
```
jacobp100, line 12: It wouldn't work to provide this as a function?

alantreadway: Integrated probably needs to be able to iterate over the options - I just wanted to avoid changing the contract right now; we could probably easily change the values to be a function to simplify things here.

packages/ag-charts-community/src/util/memo.ts
```diff
@@ -12,3 +12,45 @@ export function memo<T, R>(params: T, fnGenerator: (params: T) => () => R): () =
 
     return memorizedFns.get(fnGenerator)?.get(serialisedParams) as () => R;
 }
+
+export function simpleMemorize<F extends (...args: any[]) => any>(fn: F) {
+    const primitveCache = new Map<string | number, {}>();
```
jacobp100, line 17: I think this will lead to memory leaks - there's no way to free strings that get added

jacobp100 (approved):

iMoses (approved):

packages/ag-charts-community/src/util/memo.ts
alantreadway: Agreed in theory - practically though the number of different primitive values we're going to see is limited to theme names, and I think it adds a lot of complexity to callers to use object-only parameters, so I think this is ok for now.

packages/ag-charts-community/src/api/preset/sparkline.ts
alantreadway: It's not really possible without a bigger module system rework, as we don't automatically enable dependencies yet.

PR #2895, opened 2024-11-01
Cleanup axis format function

iMoses:
Removed redundant creation of wrappers everytime format methods were invoked
+ adjusted access modifiers of axis methods
+ removed redundant `updatePosition` calls and removed it from the public API


alantreadway (approved):
LGTM

PR #2898, opened 2024-11-01
AG-13313 Remove unused code

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13313

This is a fairly aggressive removal of unused code within the `ag-charts-community` and `ag-charts-enterprise` packages via `npx knip --fix`.

It should not remove anything that is exported to the public api.

We should check if we want to keep some of this as it may be used in pending changes.


packages/ag-charts-community/src/chart/series/cartesian/lineInterpolation.ts
alantreadway, line 1: @jacobp100 ?

packages/ag-charts-enterprise/src/series/map-util/lineStringUtil.ts
alantreadway, line 1: @jacobp100 ?

alantreadway (approved):
LGTM

packages/ag-charts-community/src/chart/themes/constants.ts
```diff
@@ -6,9 +6,6 @@ export enum FONT_SIZE {
 
 export enum FONT_WEIGHT {
```
jacobp100, line 7: Is this enum used at all? We could just write 'normal' as a string

jacobp100 (approved):

alantreadway:
Going for merge, @lsjroberts can you follow-up on https://github.com/ag-grid/ag-charts/pull/2898/files/d57c04d817cf3dd21a7d7f0f1dbc3f7f27fb4096#r1827669151 distinctly?

PR #2900, opened 2024-11-01
AG-13304 - Sparkline performance improvements pt3.

alantreadway:

### DIFF
--- a/packages/ag-charts-community/src/api/agCharts.ts
+++ b/packages/ag-charts-community/src/api/agCharts.ts
@@ -251,7 +251,7 @@ class AgChartsInternal {
     }
 
     static updateUserDelta(proxy: AgChartInstanceProxy, deltaOptions: DeepPartial<AgChartOptions>) {
-        deltaOptions = deepClone(deltaOptions, { shallow: ['data'] });
+        deltaOptions = deepClone(deltaOptions, new Set(['data']));
 
         jsonWalk(
             deltaOptions,
@@ -263,7 +263,7 @@ class AgChartsInternal {
                     }
                 }
             },
-            { skip: ['data'] }
+            new Set('data')
         );
 
         const { chart } = proxy;
--- a/packages/ag-charts-community/src/api/agCharts.ts
+++ b/packages/ag-charts-community/src/api/agCharts.ts
@@ -250,21 +250,19 @@ class AgChartsInternal {
         return proxy;
     }
 
+    private static markRemovedProperties(node: any) {
+        if (typeof node !== 'object') return;
+        for (const [key, value] of Object.entries(node)) {
+            if (typeof value === 'undefined') {
+                Object.assign(node, { [key]: Symbol('UNSET') });
+            }
+        }
+    }
+
     static updateUserDelta(proxy: AgChartInstanceProxy, deltaOptions: DeepPartial<AgChartOptions>) {
         deltaOptions = deepClone(deltaOptions, new Set(['data']));
 
-        jsonWalk(
-            deltaOptions,
-            (node) => {
-                if (typeof node !== 'object') return;
-                for (const [key, value] of Object.entries(node)) {
-                    if (typeof value === 'undefined') {
-                        Object.assign(node, { [key]: Symbol('UNSET') });
-                    }
-                }
-            },
-            new Set(['data'])
-        );
+        jsonWalk(deltaOptions, AgChartsInternal.markRemovedProperties, new Set(['data']));
 
         const { chart } = proxy;
         const lastUpdateOptions = chart.getOptions();
--- a/packages/ag-charts-community/src/api/agCharts.ts
+++ b/packages/ag-charts-community/src/api/agCharts.ts
@@ -263,7 +263,7 @@ class AgChartsInternal {
                     }
                 }
             },
-            new Set('data')
+            new Set(['data'])
         );
 
         const { chart } = proxy;
--- a/packages/ag-charts-community/src/api/preset/sparkline.ts
+++ b/packages/ag-charts-community/src/api/preset/sparkline.ts
@@ -75,6 +75,7 @@ const SPARKLINE_THEME: AgChartTheme = {
             animation: {
                 enabled: false,
             },
+            contextMenu: { enabled: false },
             toolbar: { enabled: false },
             padding: {
                 top: 0,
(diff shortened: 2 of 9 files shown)


packages/ag-charts-community/src/api/preset/sparkline.ts
```diff
@@ -350,7 +348,7 @@ export function sparkline(opts: AgSparklineOptions): AgCartesianChartOptions {
     };
     const yAxisBase: AgCartesianAxisOptions = {
         ...axis(yAxis, 'number'),
-        gridLine: gridLine(xAxis, true),
+        gridLine: gridLine(xAxis, false),
```
alantreadway, line 351: Disabling grid-lines means we can skip axis tick generation, so I think this makes sense from a defaulting perspective.

packages/ag-charts-community/src/module/optionsModule.ts
alantreadway, line 1: The core rework is here - I've tried to minimise the number of times we copy/clone data-structures (as this seems to be EXPENSIVE), and re-ordered the processing to facilitate this.

I've also applied micro-optimisations of removing lambdas in favor of real functions/methods, which also seems to have yielded performance improvements.

packages/ag-charts-community/src/util/json.ts
```diff
@@ -111,26 +117,30 @@ export function shallowClone<T>(source: T): T {
...
  * @param json to traverse
  * @param visit callback for each non-primitive and non-array object found
- * @param opts
- * @param opts.skip property names to skip when walking
- * @param jsons to traverse in parallel
+ * @param skip property names to skip when walking
+ * @param parallelJson to traverse in parallel
  */
-export function jsonWalk<T>(json: T, visit: (...nodes: T[]) => void, opts?: { skip?: string[] }, ...jsons: T[]) {
+export function jsonWalk<T, C>(
```
alantreadway, line 123: More micro-optimisations - mostly removing use of `...` and `Array.includes()` in preference of a single param and `Set.has()` respectively (also removal of lambda use)

packages/ag-charts-community/src/api/preset/sparkline.ts
jacobp100: That's fine - it's getting turned off anyway as a design requirement

packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts
```diff
@@ -143,7 +143,7 @@ export interface CartesianSeriesNodeDataContext<
     visible: boolean;
 }
 
-export const RENDER_TO_OFFSCREEN_CANVAS_THRESHOLD = 10;
+export const RENDER_TO_OFFSCREEN_CANVAS_THRESHOLD = 20;
```
jacobp100, line 146: I was thinking this should be like 1000

alantreadway: I went there originally, but backed-off to make a Sparklines-specific change for now.

jacobp100 (approved):

packages/ag-charts-community/src/api/agCharts.ts
```diff
@@ -263,7 +263,7 @@ class AgChartsInternal {
                     }
                 }
             },
-            { skip: ['data'] }
+            new Set('data')
```
iMoses:
```suggestion
            new Set(['data'])
```
Missing array wrap, meaning the set is equal to `{ 'd', 'a', 't' }`

packages/ag-charts-community/src/chart/themes/chartTheme.ts
```diff
@@ -396,28 +396,30 @@ export class ChartTheme {
...
             }
-        });
+        }
+    }
+
+    templateTheme<T>(themeTemplate: T, clone = true): T {
+        const themeInstance = clone ? deepClone(themeTemplate) : themeTemplate;
+        const params = this.getTemplateParameters();
+
+        jsonWalk(themeInstance, this.templateThemeFn, undefined, undefined, params);
```
iMoses:
```suggestion
        jsonWalk(themeInstance, this.templateThemeFn, undefined, params);
```
Redundant `undefined` arguments?

alantreadway: Good spot - I'm intrigued that this didn't flag up on any tests 🤔

iMoses (changes requested):
One bug found, few nits.

Another nit: We now pass some private function directly to new methods, and the only reason it works without issues is that these functions don't access their context, which is broken in such cases as the methods lose their binding.
How about we make these methods static to prevent future bugs/confusion or binding of these methods, which will defeat the purpose of optimisation?

packages/ag-charts-community/src/util/json.ts
```diff
@@ -111,26 +117,31 @@ export function shallowClone<T>(source: T): T {
...
-        });
+        visit(json, parallelJson, ctx);
+        let index = 0;
+        for (const node of json) {
+            jsonWalk(node, visit, skip, (parallelJson as any[])?.[index], ctx);
+            index++;
+        }
     } else if (isPlainObject(json)) {
-        visit(json, ...jsons);
+        visit(json, parallelJson, ctx);
```
iMoses, line 138: Why do we need `ctx` here?

packages/ag-charts-community/src/api/agCharts.ts
alantreadway: Thanks, good spot

packages/ag-charts-community/src/chart/themes/chartTheme.ts
alantreadway: Actually this is correct, we want to pass the same `ctx` object into the handling function

alantreadway: Reverted

packages/ag-charts-community/src/util/json.ts
alantreadway: This is for the `ChartTheme.templateTheme()` case where we need to pass around the map of values to apply - so we need a common 'context' that gets passed in to every `visit()` callback.

packages/ag-charts-community/src/chart/themes/chartTheme.ts
iMoses: Oh, so `params` is the `ctx`...

iMoses (approved):
LGTM

PR #2901, opened 2024-11-04
AG-3390 - Move features to components folder

taktran:


markdurrant (approved):
All code LGTM 👍

May be missing changes to `vitest.config.mjs` in website & shared dirs ???

PR #2902, opened 2024-11-04
AG-11993 Change zoom state ratios to use Ratio type

lsjroberts:


alantreadway (approved):
LGTM

PR #2910, opened 2024-11-04
AG-13304 - Add Sparkline Benchmarks.

alantreadway:


packages/ag-charts-community/src/scene/layersManager.ts
```diff
@@ -52,6 +52,9 @@ export class LayersManager {
     }
 
     clear() {
+        for (const layer of this.layersMap.values()) {
```
alantreadway, line 55: `Group.setLayerManager()` calls `LayerManager.removeLayer()` to cleanup, however `LayerManager.removeLayer()` will do nothing if `this.layersMap` doesn't contain the Canvas.

jacobp100 (approved):

PR #2911, opened 2024-11-04
AG-12538 Update zoom to not block scrolling browser when at extents

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12538


packages/ag-charts-enterprise/src/features/zoom/zoom.ts
```diff
@@ -454,7 +457,20 @@ export class Zoom extends _ModuleSupport.BaseModuleInstance implements _ModuleSu
...
+            event.preventDefault();
+        }
+
+        // Prevent browser scrolling when smooth wheel events continue being fired after the chart
+        // reaches a min or max extent
+        this.isFirstWheelEvent = false;
+        if (this.wheelTimer != null) clearTimeout(this.wheelTimer);
+        this.wheelTimer = setTimeout(() => {
+            this.isFirstWheelEvent = true;
+        }, 100);
```
alantreadway: Nit: I don't like these inline `setTimeout()`s, as it can make testing tricky -  maybe we could use a utility function like `debouncedCallback()` to create a debounce queue/processor to clean this up a bit?

alantreadway (approved):
LGTM - thanks!

PR #2914, opened 2024-11-05
Column data model 9

jacobp100:


packages/ag-charts-community/src/chart/data/dataModel.test.ts
```diff
@@ -643,35 +599,29 @@ describe('DataModel', () => {
...
-                expect(result?.data.length).toEqual(2);
-                expect(result?.data[0].keys).toEqual(['Q1']);
-                expect(result?.data[1].keys).toEqual(['Q2']);
+                expect(result.type).toEqual('grouped');
+                expect(result.groups).toHaveLength(2);
+                expect(result.groups[0].keys).toEqual(['Q1']);
+                expect(result.groups[1].keys).toEqual(['Q2']);
             });
 
             it('should extract the configured values', () => {
```
alantreadway, line 610: This test case seems to have dropped some assertions?

packages/ag-charts-community/src/chart/data/dataModel.test.ts
```diff
@@ -742,35 +688,29 @@ describe('DataModel', () => {
...
-                expect(result?.data.length).toEqual(2);
-                expect(result?.data[0].keys).toEqual(['Q1']);
-                expect(result?.data[1].keys).toEqual(['Q2']);
+                expect(result.type).toEqual('grouped');
+                expect(result.groups).toHaveLength(2);
+                expect(result.groups[0].keys).toEqual(['Q1']);
+                expect(result.groups[1].keys).toEqual(['Q2']);
             });
 
             it('should extract the configured accumulated values', () => {
```
alantreadway, line 699: Ditto

packages/ag-charts-community/src/chart/data/dataModel.test.ts
```diff
@@ -1171,7 +1089,7 @@ describe('DataModel', () => {
 `);
         });
 
-        describe('property tests', () => {
+        describe.skip('property tests', () => {
```
alantreadway, line 1092: Nit: `.skip()?`

packages/ag-charts-community/src/chart/data/dataModel.test.ts
jacobp100: I've not been super consistent with the testing here. Some of the tests take these indices and then read from the columns to compare the data. This just assumes the columns are correct (they're tested elsewhere), so here it's just validating you'll read the right location in the column. It's sort of the same thing, but I can update it if you feel strongly?

packages/ag-charts-community/src/chart/data/dataModel.test.ts
jacobp100: Same - https://github.com/ag-grid/ag-charts/pull/2914/files/c3031a99beb3e71d9c0c28a1f7b73ea47d1932c6#r1829530633

packages/ag-charts-community/src/chart/data/dataModel.test.ts
jacobp100: Fixed in 10

packages/ag-charts-community/src/chart/data/dataModel.test.ts
alantreadway: Ok, if the values are covered elsewhere then SGTM

alantreadway (approved):

PR #2923, opened 2024-11-06
AG-13035: Add support for tree-shakable entry-points

iMoses:
cleanup _Util & _Scale (Part 1)


packages/ag-charts-community/src/sparklines-util.ts
```diff
@@ -1,26 +1,4 @@
...
-
-export { isValidDate as isDate, isFiniteNumber as isNumber, isString } from './util/type-guards';
-export { extent, normalisedExtentWithMetadata } from './util/array';
-export { isEqual as isNumberEqual } from './util/number';
-export { tickFormat } from './util/ticks';
-
-export * from './util/time';
-export { createTicks as ticks, tickStep, range, isDenseInterval } from './util/ticks';
+// Only these methods are used by ag-grid.
+// DO NOT ADD EXPORTS UNLESS REQUIRED BY INTEGRATED CHARTS.
```
lsjroberts, line 2: Would it be worth renaming this file to `integrated-utils.ts` or something?

alantreadway: Agreed - I believe from offline discussion that is the intent, and this is the first drop of several, so this should happen in a follow-on PR.

alantreadway (approved):
LGTM

PR #2929, opened 2024-11-06
AG-13305 Fix display for cone funnel with one datum

jacobp100:
https://ag-grid.atlassian.net/browse/AG-13305


alantreadway (approved):
LGTM

PR #2933, opened 2024-11-07
Add types for grid integration

iMoses:


alantreadway (approved):
LGTM

packages/ag-charts-community/src/main-modules.ts
```diff
@@ -1,6 +1,34 @@
...
+};
+
+export const SparklineModule: IntegratedSparklineModule = {
+    AgCharts,
+    VERSION,
+    // @ts-ignore
+    _Scene,
+    // @ts-ignore
+    _Theme,
+    _Util,
```
StephenCooper, line 30: Sparklines from the grid side of things does not actually require these sections but maybe leave for now.

PR #2934, opened 2024-11-07
AG-13304 - Add sparkline pooling.

alantreadway:


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -399,6 +400,11 @@ export abstract class Chart extends Observable {
         this._performUpdateSkipAnimations = true;
     }
 
+    detachAndClear() {
+        this.container = undefined;
+        this.ctx.scene.clear();
+    }
```
jacobp100, line 406: We might be able to speed stuff up by clearing on remount - rather than on detatch

jacobp100 (approved):

packages/ag-charts-community/src/chart/chart.ts
alantreadway: Fair - let's see where this lands us, you're right that it's redundant work if we never re-use the instnace.

PR #2938, opened 2024-11-07
Fix disabling the keyboard

jacobp100:


alantreadway (approved):
LGTM

olegat (changes requested):
This won't completely disable keyboard input. The user would still be about to focus on other `tabindex="0"` elements like the legend, navigator or FC toolbars.

The `keyboard.enabled` option exists only because it defaulted to `false` back when keyboard navigation was experimental. I would be in favour of simply removing it, because setting `keyboard.enabled = false` is not compliant with accessibility guidelines and

jacobp100:
Also breaks hovering

PR #2947, opened 2024-11-08
AG-12655 Skip unordered category animations

jacobp100:
https://ag-grid.atlassian.net/browse/AG-12655

### DIFF
--- a/packages/ag-charts-community/src/chart/axis/categoryAxis.ts
+++ b/packages/ag-charts-community/src/chart/axis/categoryAxis.ts
@@ -38,7 +38,7 @@ export class CategoryAxis<
 
             if (normalizedNextIndex === -1) {
                 // All subsequent values must be extending (i.e. appending to) the normalized domain
-                normalizedIndex = Number.MAX_SAFE_INTEGER;
+                normalizedIndex = Infinity;
             } else if (normalizedNextIndex <= normalizedIndex) {
                 return false;
             } else {
--- a/packages/ag-charts-community/src/chart/axis/categoryAxis.ts
+++ b/packages/ag-charts-community/src/chart/axis/categoryAxis.ts
@@ -30,6 +30,52 @@ export class CategoryAxis<
     @Validate(RATIO, { optional: true })
     paddingOuter?: number;
 
+    private domainOrderedToNormalizedDomain(seriesDomain: any[], normalizedDomain: any[]) {
+        let normalizedIndex = -1;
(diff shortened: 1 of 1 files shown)


packages/ag-charts-community/src/chart/axis/categoryAxis.ts
```diff
@@ -30,6 +30,52 @@ export class CategoryAxis<
...
+        return true;
+    }
+
+    private categoryAnimatable = true;
+    protected override calculateDomain() {
+        let normalizedDomain: any[] = [];
+
+        let categoryAnimatable = true;
+        for (const series of this.boundSeries) {
+            if (!this.includeInvisibleDomains && !series.isEnabled()) continue;
```
iMoses, line 58: nit: `includeInvisibleDomains` is always true in category axes, so this always returns false

packages/ag-charts-community/src/chart/axis/categoryAxis.ts
```diff
@@ -30,6 +30,52 @@ export class CategoryAxis<
...
 
+    private domainOrderedToNormalizedDomain(seriesDomain: any[], normalizedDomain: any[]) {
+        let normalizedIndex = -1;
+        for (let seriesIndex = 0; seriesIndex < seriesDomain.length; seriesIndex += 1) {
+            const value = seriesDomain[seriesIndex];
+            const normalizedNextIndex = normalizedDomain.indexOf(value);
+
+            if (normalizedNextIndex === -1) {
+                // All subsequent values must be extending (i.e. appending to) the normalized domain
+                normalizedIndex = Number.MAX_SAFE_INTEGER;
```
iMoses: nit: use `Infinity` to keep consistent with similar logic in our codebase

iMoses:
Can we extract this code into `cartesianChart.ts`?
It'll help us by reducing coupling between the axis and series classes, and will keep logic closer to where it's supposed to be.

packages/ag-charts-community/src/chart/axis/categoryAxis.ts
jacobp100:
```suggestion
                normalizedIndex = Infinity;
```

packages/ag-charts-community/src/chart/axis/categoryAxis.ts
jacobp100: It's to match the original logic - in case this value ever changes

alantreadway (approved):
LGTM

PR #2952, opened 2024-11-11
AG-13288 Don't call panToBBox if seriesRect contains focus center

olegat:
https://ag-grid.atlassian.net/browse/AG-13288

This fixes a bug where the SeriesAreaManager would try to pan to datums that are touching the origin of the axes. This change permits the bounds of the focus indicator to extend beyond the seriesRect, which is fine because that's how the datum shapes are drawn in some cases.


alantreadway (approved):
LGTM

PR #2953, opened 2024-11-11
AG-13316 Fix regression where `no-animation` class was not set

olegat:
https://ag-grid.atlassian.net/browse/AG-13316

The `Tooltip.toggle` method uses `Toggle.isVisible()` to calculate the `wasVisible` and `visible` variabled.

The `Tooltip.isVisible()` method checks if the `hidden` class exists, so we need to make sure that this class isn't removed because the `Tooltip.toggle` method is called.


alantreadway (approved):
LGTM

PR #2957, opened 2024-11-11
AG-4416: Support nested categories (2)

iMoses:
- align TickDatum with axis.ts to keep gridLine animation logic consistent
- Use default axis title implementation
https://ag-grid.atlassian.net/browse/AG-4416


alantreadway (approved):
LGTM

PR #2961, opened 2024-11-12
AG-13382 Move toolbar floating zoom buttons into zoom module

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13382

There's a bit of duplication and cross-over between this and the ranges buttons, which will be cleaned away later.

There's also some upcoming consideration required around toggling enabled, changing the button options and updating the locale, and how that affects the dom. Things the toolbar module currently handles.

### DIFF
--- a/packages/ag-charts-community/src/util/dom.ts
+++ b/packages/ag-charts-community/src/util/dom.ts
@@ -83,11 +83,30 @@ export function setWindow(window: Window) {
     verifiedGlobals.window = window;
 }
 
-export function setElementBBox(element: HTMLElement | undefined, bbox: BBoxValues) {
-    if (element) {
+export function setElementBBox(element: HTMLElement | undefined, bbox: Partial<BBoxValues>) {
+    if (!element) return;
+
+    if (bbox.width == null) {
+        element.style.removeProperty('width');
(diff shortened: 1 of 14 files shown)


packages/ag-charts-community/src/util/dom.ts
```diff
@@ -83,13 +83,12 @@ export function setWindow(window: Window) {
...
-export function setElementBBox(element: HTMLElement | undefined, bbox: BBoxValues) {
-    if (element) {
-        element.style.width = `${bbox.width}px`;
-        element.style.height = `${bbox.height}px`;
-        element.style.left = `${bbox.x}px`;
-        element.style.top = `${bbox.y}px`;
-    }
+export function setElementBBox(element: HTMLElement | undefined, bbox: Partial<BBoxValues>) {
+    if (!element) return;
+    if (bbox.width != null) element.style.width = `${bbox.width}px`;
```
alantreadway: Nit: Should `null`/`undefined` values remove the related `style` attribute too?

alantreadway (commented):
Generally LGTM, one nit

alantreadway (approved):
LGTM

PR #2965, opened 2024-11-12
AG-13392: Allow axis label padding to be applied consistently regardl…

iMoses:
…ess of tick state
https://ag-grid.atlassian.net/browse/AG-13392


alantreadway (approved):
LGTM

PR #2970, opened 2024-11-13
AG-11986 Add toolbar component with shared styles and button management

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13361


alantreadway (approved):
Nice, LGTM

PR #2971, opened 2024-11-13
Ordinal time axis perf

jacobp100:


packages/ag-charts-community/src/scale/ordinalTimeScale.ts
```diff
@@ -113,12 +137,22 @@ export class OrdinalTimeScale extends BandScale<Date, TimeInterval | number> {
...
+            low = precomputedSteps[i];
+            high = i < precomputedSteps.length - 1 ? precomputedSteps[i + 1] : sortedTimestamps.length - 1;
+        } else {
+            low = 0;
+            high = sortedTimestamps.length - 1;
+        }
 
         while (low <= high) {
-            const mid = Math.floor((low + high) / 2);
+            const mid = ((low + high) / 2) | 0;
```
alantreadway, line 158: Is `| 0` correct?

iMoses: It's a trick to floor which is more efficient.
I'd leave a comment here, because not all people know these sort of hacks.

packages/ag-charts-community/src/chart/axis/axis.ts
```diff
@@ -1085,8 +1085,12 @@ export abstract class Axis<S extends Scale<D, number, TickInterval<S>> = Scale<a
     }
 
     protected calculateDomain() {
-        const visibleSeries = this.boundSeries.filter((s) => this.includeInvisibleDomains || s.isEnabled());
-        const domains = visibleSeries.flatMap((series) => series.getDomain(this.direction));
+        const { includeInvisibleDomains, boundSeries, direction } = this;
+        const visibleSeries = boundSeries.filter((s) => includeInvisibleDomains || s.isEnabled());
```
iMoses:
```suggestion
        const visibleSeries = this.includeInvisibleDomains ? boundSeries : boundSeries.filter((s) => s.isEnabled());
```
nit: I assume that you destructed `includeInvisibleDomains` to reduce lookup time, but if that's the case wouldn't it be simpler to skip filtering altogether?

packages/ag-charts-community/src/chart/series/series.ts
```diff
@@ -420,9 +420,9 @@ export abstract class Series<
     // The union of the series domain ('community') and series-option domains ('enterprise').
     getDomain(direction: ChartAxisDirection): any[] {
         const seriesDomain: any[] = this.getSeriesDomain(direction);
-        const moduleDomains: any[][] = this.moduleMap.mapModules((module) => module.getDomain(direction));
+        const moduleDomains: any[][] = this.moduleMap.mapModules((module) => module.getDomain(direction)).flat();
         // Flatten the 2D moduleDomains into a 1D array and concatenate it with seriesDomain
-        return seriesDomain.concat(moduleDomains.flat());
+        return moduleDomains.length !== 0 ? seriesDomain.concat(moduleDomains.flat()) : seriesDomain;
```
iMoses:
```suggestion
        return moduleDomains.length !== 0 ? seriesDomain.concat(moduleDomains) : seriesDomain;
```
nit: If you already flattened it you don't need to do it again.

iMoses:
Did you notice that in the current version the last datum of a series is lost? (not rendered)

packages/ag-charts-community/src/scale/ordinalTimeScale.ts
jacobp100: It's a performance hack for JS engines - it means they know this is an int32. It's equivalent to Math.trunc - rounds towards zero

packages/ag-charts-community/src/chart/axis/axis.ts
jacobp100: Yes - good point

packages/ag-charts-community/src/chart/series/series.ts
jacobp100: Thanks!

alantreadway (approved):
LGTM

PR #2987, opened 2024-11-14
Candlestick perf 2

jacobp100:
- Ensures ordinal time axis' domain is always uniqued and sorted (forwards or backwards)
- sortedTimestamps is always sorted forwards
- timestamps is no longer needed


alantreadway (approved):
LGTM

PR #2989, opened 2024-11-14
ESLint v9 Upgrade

alantreadway:
Upgrades to eslint v9; follow-up changes to enable new rules will be needed.


packages/ag-charts-community/src/chart/data/processors.ts
```diff
@@ -307,16 +310,16 @@ const animationValidationProcessKey = (
...
 
         return validation;
     }
 
     let lastValue = column[0];
     for (let d = 1; validation !== 0 && d < column.length; d++) {
         const keyValue = column[d];
-        if (!(lastValue <= keyValue)) validation &= ~ANIMATION_VALIDATION_ORDERED_KEYS;
-        if (!(lastValue !== keyValue)) validation &= ~ANIMATION_VALIDATION_UNIQUE_KEYS;
+        if (lastValue > keyValue) validation &= ~ANIMATION_VALIDATION_ORDERED_KEYS;
```
jacobp100: Not quite - the old version handled NaNs (via undefined values)

packages/ag-charts-community/src/scene/changeDetectable.ts
```diff
@@ -89,7 +89,7 @@ function buildCheckDirtyChain(setterFn: Function, opts: SceneChangeDetectionOpti
         return function (this: any, value: any) {
             const change = setterFn.call(this, value);
 
-            if (change !== NO_CHANGE && value != null && value._dirty) {
+            if (change !== NO_CHANGE && value?._dirty) {
```
jacobp100: Just because this is a hot path -

```suggestion
            if (change !== NO_CHANGE && value?._dirty === true) {
```

Means the engine can do fewer checks

packages/ag-charts-community/src/util/async.ts
```diff
@@ -8,7 +8,7 @@ export class AsyncAwaitQueue {
                 resolve(true);
             };
             const timeoutFn = () => {
-                const queueIndex = this.queue.findIndex(successFn);
+                const queueIndex = this.queue.indexOf(successFn);
```
jacobp100, line 11: ?

packages/ag-charts-community/src/util/attributeUtil.ts
```diff
@@ -108,7 +108,7 @@ export function setAttribute<A extends keyof BaseAttributeTypeMap>(
     qualifiedName: A,
     value: BaseAttributeTypeMap[A] | undefined
 ) {
-    if (value === undefined || value === '') {
+    if (value == null) {
```
jacobp100:
```suggestion
    if (value == null || value === '') {
```

packages/ag-charts-community/src/util/mutex.ts
```diff
@@ -26,7 +26,7 @@ export class Mutex {
     }
 
     public async waitForClearAcquireQueue() {
-        return this.acquire(async () => undefined);
+        return this.acquire(() => Promise.resolve(undefined));
```
jacobp100, line 29: Why did it care here?

packages/ag-charts-community-examples/src/charts-overview/examples/log-axis/main.ts
```diff
@@ -78,18 +78,20 @@ const options: AgCartesianChartOptions = {
...
 const chart = AgCharts.create(options)
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 function setNumberAxis() {
   options.subtitle = {
     text: "linear scale",
   }
   options.axes = linearAxes
-  chart.update(options)
+  void chart.update(options)
```
jacobp100, line 87: 👏

packages/ag-charts-community/src/chart/data/dataService.ts
```diff
@@ -86,6 +86,7 @@ export class DataService<D extends object> extends Listeners<EventType, EventHan
     }
 
     private createThrottledFetch(requestThrottle: number) {
+        // eslint-disable-next-line sonarjs/no-misused-promises
```
jacobp100: Why did it think this was misused?

packages/ag-charts-community/src/chart/test/utils.ts
```diff
@@ -457,7 +458,7 @@ export function spyOnAnimationManager() {
 
         const forceTimeJumpMock = jest.spyOn(AnimationManager.prototype, 'forceTimeJump');
         forceTimeJumpMock.mockImplementation((controller: IAnimation, defaultDuration: number) => {
-            if (controller.isComplete) return true;
+            if (controller.isComplete) return true as boolean;
```
jacobp100: ???

packages/ag-charts-community/src/util/async.ts
alantreadway: successFn is the item we're trying to find, so `Array.findIndex()` is the wrong thing to be doing here.

packages/ag-charts-community/src/chart/data/processors.ts
alantreadway: We should make that a bit more explicit here perhaps, can we handle the `NaN` case explicitly?

packages/ag-charts-community/src/util/attributeUtil.ts
alantreadway:
```
        111:32  error  Remove this "===" check; it will always be false. Did you mean to use "=="?  sonarjs/different-types-comparison
```

Perhaps the rule is faulty in that case 🤔

packages/ag-charts-community/src/util/mutex.ts
alantreadway: There's a new rule for use of `async` keyword, you must use `await` for it to be valid - which seems to make sense generally?

packages/ag-charts-community/src/chart/data/dataService.ts
alantreadway: this.fetch() returns a promise, but we just throw it away - this is probably a bug, but I don't want to fix this right now.

alantreadway: Actually let me tune the rule config for this down, so we remember to address it.

packages/ag-charts-community/src/chart/data/processors.ts
jacobp100: Only thing to remember is it might have date objects in it. Maybe just call .valueOf to get a numeric value, then use Number.isFinite and the normal comparison operators

packages/ag-charts-community/src/chart/test/utils.ts
jacobp100: Maybe it wants you to annotate the return type as a boolean?

alantreadway: Without this:
```
      /Users/bls/git/ag-charts/packages/ag-charts-community/src/chart/test/utils.ts
        460:96  error  Refactor this function to not always return the same value  sonarjs/no-invariant-returns
```

I can probably rework this slightly differently to make it look less odd.

packages/ag-charts-community/src/chart/data/processors.ts
alantreadway: Reverted for now, we'll follow-up tomorrow.

.env
```diff
@@ -3,5 +3,4 @@
 # This is disabled for existing workspaces to maintain compatibility
 # For more info, see: https://nx.dev/concepts/inferred-tasks
 NX_ADD_PLUGINS=false
-ESLINT_USE_FLAT_CONFIG=false
```
taktran, line 6: 🎉

PR #2991, opened 2024-11-15
AG-12204 - Enable no-empty-function eslint rule.

alantreadway:

### DIFF
--- a/packages/ag-charts-community/src/api/preset/presetUtils.ts
+++ b/packages/ag-charts-community/src/api/preset/presetUtils.ts
@@ -1,5 +1,12 @@
+import { Logger } from '../../util/logger';
+
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-export function assertEmpty(_t: Record<string, never>) {}
+export function assertEmpty(t: Record<string, never>) {
+    const keys = Object.keys(t);
+    if (keys.length > 0) {
+        Logger.warn(`unexpected options, ignoring these: [${keys.join(',')}]`);
+    }
+}
 
 export const IGNORED_PROP = Symbol('IGNORED_PROP');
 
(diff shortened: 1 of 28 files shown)


packages/ag-charts-community/src/api/preset/presetUtils.ts
```diff
@@ -1,5 +1,12 @@
+import { Logger } from '../../util/logger';
+
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
iMoses: can't you remove the `eslint-disable` annotation as well?

packages/ag-charts-enterprise/src/features/crosshair/crosshair.ts
```diff
@@ -211,7 +211,7 @@ export class Crosshair extends _ModuleSupport.BaseModuleInstance implements _Mod
             const fractionDigits = (axisLayout?.label.fractionDigits ?? 0) + (isInteger(value) ? 0 : 1);
             return formatNumber(value, fractionDigits);
         }
-        return String(value ?? '');
+        return String(value?.toString() ?? '');
```
iMoses: This doesn't make too much sense 😶 
What was the lint error here?
If a change is needed then I think you can get rid of the outer `String` (i.e. `return value?.toString() ?? '';`

packages/ag-charts-community/src/api/preset/presetUtils.ts
alantreadway: I thought I'd seen eslint complain about inactive disables, but maybe it was a figment of my imagination :D

alantreadway: Enable detection of unused eslint inline disabled, and I've removed everything flagged.

packages/ag-charts-enterprise/src/features/crosshair/crosshair.ts
alantreadway: https://typescript-eslint.io/rules/no-base-to-string/

I'm not sure if we should take it tbh, I can undo this change

alantreadway: Reverted the fix

PR #2993, opened 2024-11-15
AG-13414 Use image elements in gallery

JamesSwinton:
Replace `background-image` with `` elements so that we can include an alt-tag that is optimised for SEO.

Does not produce a flicker or increase CLS on initial load, which I believe was the original reason for using `background-image`.


markdurrant (approved):
LGTM 👍

PR #2999, opened 2024-11-15
AG-12204 - Enable eslint @typescript-eslint/require-await rule.

alantreadway:


packages/ag-charts-community/src/chart/series/__image_snapshots__/series-labels-test-ts-series-labels-create-for-grouped-bubble-series-labels-it-should-render-to-canvas-as-expected-1-snap.png
jacobp100, line 1: Any idea why this changed?

jacobp100 (approved):
I think the screenshots broke, however

packages/ag-charts-community/src/chart/series/__image_snapshots__/series-labels-test-ts-series-labels-create-for-grouped-bubble-series-labels-it-should-render-to-canvas-as-expected-1-snap.png
alantreadway: I believe it's because previously the async-ness of the series processing lead to differences in when all of the label data becomes available, meaning we are seeing a different behaviour from the (scatter/bubble-specific) collision detection due to differences in when state becomes available.

PR #3002, opened 2024-11-18
AG-12204 - ESLint enable inverted condition and misused Promises rules.

alantreadway:

### DIFF
--- a/packages/ag-charts-community/src/util/function.ts
+++ b/packages/ag-charts-community/src/util/function.ts
@@ -1,3 +1,5 @@
+import { Logger } from './logger';
+
 interface DebounceOptions {
     leading?: boolean;
     trailing?: boolean;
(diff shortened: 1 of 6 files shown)


jacobp100 (approved):

packages/ag-charts-community/src/util/function.ts
```diff
@@ -72,7 +74,7 @@ export function debounce<T extends (...args: Parameters<T>) => void>(
     });
 }
 
-export function throttle<T extends (...args: Parameters<T>) => void>(
+export function throttle<T extends (...args: Parameters<T>) => void | Promise<void>>(
```
iMoses: `throttle`callback shouldn't return a promise, it's a bit of an anti-pattern.
I think it's better if we take care of it on the `dataService` side instead of the generic method.

alantreadway: Fair, I've pushed the error handling back out to the callback.

PR #2956, opened 2024-11-11
AG-13306 Migrate `SeriesAreaManager` to `widgets/` framework

olegat:
https://ag-grid.atlassian.net/browse/AG-13306

Depends on:
- https://github.com/ag-grid/ag-charts/pull/2955
- https://github.com/ag-grid/ag-charts/pull/2949
- https://github.com/ag-grid/ag-charts/pull/2945
- https://github.com/ag-grid/ag-charts/pull/2942

PR #3006, opened 2024-11-19
AG-13387 Fix dev regression with contextmenu events on navigator

olegat:
https://ag-grid.atlassian.net/browse/AG-13387

Depends on:
- https://github.com/ag-grid/ag-charts/pull/2956


alantreadway (approved):
LGTM

PR #3008, opened 2024-11-19
AG-12204 - Fix series label placement.

alantreadway:

Fixes label placement inconsistency introduced by #2999


packages/ag-charts-community/src/chart/series/cartesian/__image_snapshots__/histogram-series-test-ts-histogram-series-series-highlighting-should-highlight-scatter-datum-when-overlapping-histogram-1-snap.png
iMoses, line 1: Is this change correct?
It seems like we're rendering the labels twice.

packages/ag-charts-enterprise/src/series/map-marker/mapMarkerSeries.ts
```diff
@@ -530,8 +530,17 @@ export class MapMarkerSeries
...
+
+        this.ctx.seriesLabelLayoutManager
+            .placeLabels(this.id, this.getLabelData())
+            .then((layoutResult) => {
+                const placedLabels = layoutResult.get(this.id) ?? [];
+                opts.labelSelection.update(placedLabels);
+                this.updateLabelNodes(opts);
+            })
+            .catch((e) => Logger.error(e));
+        return opts.labelSelection;
```
iMoses: Shouldn't we wait for the placement to finish before returning `labelSelection`? (which isn't yet updated)
Same question apply to all other locations...

alantreadway: Technically yes, but practically no.

Without a lot of additional change to the rest of the `chart.update()` flow to support async execution, we can't 'wait' here. I've checked the control flow, and calling `this.updateLabelNodes()` is the scope of effects after the placed data is calculated.

The new `Chart.updateSeries()` methods coordinates with `SeriesLabelLayoutManager` to ensure these pending updates happen before the main process flow continues.

Practically we need to rework the entire label handling process, that is too large to address in this tactical PR.

packages/ag-charts-community/src/chart/series/cartesian/__image_snapshots__/histogram-series-test-ts-histogram-series-series-highlighting-should-highlight-scatter-datum-when-overlapping-histogram-1-snap.png
alantreadway: Will look into this, couldn't see what had changed locally

alantreadway: Reverted after latest updated.

packages/ag-charts-enterprise/src/series/map-marker/mapMarkerSeries.ts
alantreadway: Discussed offline, re-arranged the processing so that we aren't doing async processing, and the flow is a bit clearer.

jacobp100 (approved):

PR #3009, opened 2024-11-19
AG-13397 Move annotation options toolbar into annotations module

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13397

`AnnotationOptionsToolbar` takes the user options and instantiates a toolbar that uses the `AnnotationOptionsButtonWidget` variant of the button widget, which handles the fill color and stroke width indicators on the buttons (so they don't pollute unrelated toolbars).

The `FloatingToolbar` is a `Toolbar` that renders itself inside a `DraggablePopover` (which is now also used for the `Dialog`).

Also moves popover & menu to community, color picker to enterprise/components (from features).


packages/ag-charts-community/src/components/toolbar/toolbar.ts
```diff
@@ -0,0 +1,177 @@
...
+    public toggleActiveButtonByIndex(index: number) {
+        for (const [buttonIndex, button] of this.buttonWidgets.entries()) {
+            button.getElement().classList.toggle(BUTTON_ACTIVE_CLASS, index != null && index === buttonIndex);
+        }
+    }
+
+    public toggleButtonEnabledByIndex(index: number, enabled: boolean) {
+        const buttonWidget = this.buttonWidgets.at(index);
+        if (!buttonWidget) return;
+        buttonWidget.getElement().ariaDisabled = (!enabled).toString();
```
olegat:
```suggestion
        buttonWidget.getElement().setEnabled(enabled);
```

packages/ag-charts-community/src/components/toolbar/toolbar.ts
```diff
@@ -0,0 +1,177 @@
...
+    private readonly buttonWidgets: Array<ButtonWidget> = [];
+
+    constructor(
+        protected readonly ctx: ModuleContext,
+        private readonly onButtonPress: (button: ButtonOptions & { index: number; sourceEvent: MouseEvent }) => void,
+        private readonly onDragStart?: (sourceEvent: MouseEvent, element: HTMLElement) => void
+    ) {
+        super();
+
+        const element = this.getElement();
```
olegat: The `Widget.getElement()` method is primarily intended for internal use and compatibility glue with legacy code. It's preferable to interface directly with the `Widget` instance:

```diff
diff --git a/packages/ag-charts-community/src/widget/widget.ts b/packages/ag-charts-community/src/widget/widget.ts
index e7fe51e6e..17bf7b6c4 100644
--- a/packages/ag-charts-community/src/widget/widget.ts
+++ b/packages/ag-charts-community/src/widget/widget.ts
@@ -126,6 +126,10 @@ export abstract class Widget<
         this.onChildAdded(child);
     }
 
+    toggle(token: string, force?: boolean){
+        this.elem.classList.toggle(token, force);
+    }
+
     protected appendChildToDOM(child: TChildWidget) {
         this.elem.appendChild(child.getElement());
     }
```

packages/ag-charts-community/src/components/toolbar/toolbar.ts
```diff
@@ -0,0 +1,177 @@
...
+        this.refreshButtonClasses();
+    }
+
+    public updateButtonByIndex(index: number, button: ButtonOptions) {
+        this.buttonWidgets.at(index)?.update(button);
+    }
+
+    public clearActiveButton() {
+        for (const button of this.buttonWidgets) {
+            button.getElement().classList.toggle(BUTTON_ACTIVE_CLASS, false);
```
olegat:
```suggestion
            button.toggle(BUTTON_ACTIVE_CLASS, false);
```

packages/ag-charts-community/src/components/toolbar/toolbar.ts
```diff
@@ -0,0 +1,177 @@
...
+        }
+    }
+
+    public toggleActiveButtonByIndex(index: number) {
+        for (const [buttonIndex, button] of this.buttonWidgets.entries()) {
+            button.getElement().classList.toggle(BUTTON_ACTIVE_CLASS, index != null && index === buttonIndex);
```
olegat:
```suggestion
            button.toggle(BUTTON_ACTIVE_CLASS, index != null && index === buttonIndex);
```

packages/ag-charts-community/src/components/toolbar/toolbar.ts
```diff
@@ -0,0 +1,177 @@
...
+    public toggleButtonEnabledByIndex(index: number, enabled: boolean) {
+        const buttonWidget = this.buttonWidgets.at(index);
+        if (!buttonWidget) return;
+        buttonWidget.getElement().ariaDisabled = (!enabled).toString();
+    }
+
+    public toggleSwitchCheckedByIndex(index: number, checked: boolean) {
+        const buttonWidget = this.buttonWidgets.at(index);
+        if (!buttonWidget) return;
+        buttonWidget.getElement().ariaChecked = checked.toString();
```
olegat:
```suggestion
        buttonWidget.setChecked(checked);
```

With:
```diff
diff --git a/packages/ag-charts-community/src/widget/buttonWidget.ts b/packages/ag-charts-community/src/widget/buttonWidget.ts
index 3bc63501f..5f0e001ac 100644
--- a/packages/ag-charts-community/src/widget/buttonWidget.ts
+++ b/packages/ag-charts-community/src/widget/buttonWidget.ts
@@ -16,4 +16,8 @@ export class ButtonWidget extends Widget<HTMLButtonElement> {
         setAttribute(this.elem, 'aria-disabled', !enabled);
         setElementStyle(this.elem, 'pointer-events', enabled ? undefined : 'none');
     }
+
+    setChecked(checked: boolean) {
+        setAttribute(this.elem, 'aria-checked', checked);
+    }
 }
```

packages/ag-charts-community/src/components/toolbar/toolbar.ts
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    public toggleButtonVisibilities(visibleIndices: Array<number>) {
+        for (const [index, buttonWidget] of this.buttonWidgets.entries()) {
+            buttonWidget
+                .getElement()
+                .classList.toggle('ag-charts-toolbar__button--hidden-toggled', !visibleIndices.includes(index));
```
olegat:
```suggestion
            buttonWidget.toggle('ag-charts-toolbar__button--hidden-toggled', !visibleIndices.includes(index));
```

packages/ag-charts-community/src/components/toolbar/toolbar.ts
```diff
@@ -0,0 +1,177 @@
...
+        }
+        this.refreshButtonClasses();
+    }
+
+    public getButtonBounds() {
+        return this.buttonWidgets.map((buttonWidget) => {
+            const element = buttonWidget.getElement();
```
olegat, line 89: There is a `Widget.getBounds` method, however that would need to be modified in order add the `offsetParent` values.

packages/ag-charts-community/src/components/toolbar/toolbar.ts
```diff
@@ -0,0 +1,177 @@
...
+                index === buttonWidgets.length - 1 ||
+                element.getAttribute('data-section') !=
+                    buttonWidgets
+                        .at(index + 1)
+                        ?.getElement()
+                        .getAttribute('data-section');
+
+            element.classList.toggle('ag-charts-toolbar__button--first', first);
+            element.classList.toggle('ag-charts-toolbar__button--last', last);
+            element.classList.toggle('ag-charts-toolbar__button--gap', index > 0 && first);
```
olegat:
```suggestion
            buttonWidget.toggle('ag-charts-toolbar__button--first', first);
            buttonWidget.toggle('ag-charts-toolbar__button--last', last);
            buttonWidget.toggle('ag-charts-toolbar__button--gap', index > 0 && first);
```

packages/ag-charts-community/src/components/toolbar/toolbar.ts
```diff
@@ -0,0 +1,177 @@
...
+            const element = buttonWidget.getElement();
+
+            first = (onDragStart == null && index === 0) || section != element.getAttribute('data-section');
+            last =
+                index === buttonWidgets.length - 1 ||
+                element.getAttribute('data-section') !=
+                    buttonWidgets
+                        .at(index + 1)
+                        ?.getElement()
+                        .getAttribute('data-section');
```
olegat: Preferably, if we should follow the same pattern for all `'data-*'` attributes.

Here's how it's done with `'data-preventdefault'` (I'm open to hear alternative suggestions):
https://github.com/ag-grid/ag-charts/blob/a3ce49463e91367e67c17921bb4966637c48da23/packages/ag-charts-community/src/widget/widget.ts#L114

packages/ag-charts-community/src/components/toolbar/toolbar.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    private createDragHandle() {
+        const { onDragStart } = this;
+        if (!onDragStart) return;
+
+        const dragHandle = new NativeWidget<HTMLElement>(createElement('div', 'ag-charts-toolbar__drag-handle'));
+        dragHandle.getElement().innerHTML = `<span class="${getIconClassNames('drag-handle')} ag-charts-toolbar__icon"></span>`;
+        dragHandle.getElement().addEventListener('mousedown', (event) => {
+            onDragStart(event, dragHandle.getElement());
+        });
```
olegat, line 127: This is acceptable for now. But we might want to consider writing a `class DragHandle extends Widget` at some point for better abstraction.

packages/ag-charts-community/src/components/toolbar/toolbar.ts
```diff
@@ -0,0 +1,177 @@
...
+        this.appendChild(dragHandle);
+    }
+
+    private createButton(index: number, button: ButtonOptions) {
+        const buttonWidget = this.createButtonWidget();
+        const element = buttonWidget.getElement();
+        element.classList.add('ag-charts-toolbar__button');
+
+        element.addEventListener('click', (sourceEvent) => {
```
olegat:
```suggestion
        buttonWidget.toggle('ag-charts-toolbar__button', true);

        buttonWidget.addListener('click', (_widget, sourceEvent) => {
```

packages/ag-charts-enterprise/src/features/annotations/annotationOptionsToolbar.ts
```diff
@@ -41,12 +56,94 @@ interface EventMap {
...
+    private updateLineStrokeWidth(options: AnnotationOptionsButtonOptions) {
+        const element = this.getElement();
+        element.classList.add('ag-charts-annotations-stroke-width-button');
+        element.style.setProperty('--stroke-width', `${options.strokeWidth}px`);
+    }
+}
+
+class FloatingAnnotationOptionsToolbar extends FloatingToolbar<
+    AnnotationOptionsButtonOptions,
+    any // TODO: AnnotationOptionsButtonWidget
```
alantreadway, line 123: Nit: Remove the `any`?

packages/ag-charts-community/src/components/toolbar/floatingToolbar.ts
```diff
@@ -0,0 +1,146 @@
...
+        private readonly onToolbarMoved: (event: { buttonBounds: Array<BBoxValues>; popoverBounds: BBoxValues }) => void
+    ) {
+        super(ctx, onButtonPress, (event: MouseEvent, element: HTMLElement) =>
+            this.popover.startDragging(event, element)
+        );
+        this.popover = new FloatingToolbarPopover(ctx, id, this.onPopoverMoved.bind(this));
+    }
+
+    public show(options: PopoverOptions) {
+        this.popover.show([this.getElement()], options);
```
olegat, line 117: This is acceptable for now. But eventually we'll want to change the signature of `FloatingToolbarPopover.show` to take a `Widget[]` rather a `HTMLElement[]`.

olegat (approved):
LGTM. Okay to merge. I've added some code suggestions but we can put those in a follow-up PR.

packages/ag-charts-community/src/components/toolbar/toolbar.ts
lsjroberts: It seemed like the `ButtonWidget` probably shouldn't have this since there's the `SwitchWidget` which extends it and adds `setChecked()`?

So I added `setChecked()` to `ToolbarButtonWidget` instead. Did think about making that extend `SwitchWidget` but then it has to remove the role in the constructor, which seemed a bit wrong.

Ideally it would be a mixture of buttons and switches but that's more complex than I want to go for now.

packages/ag-charts-community/src/components/toolbar/toolbar.ts
lsjroberts: Yeah that seems reasonable, I don't imagine we'll have loads of custom `data-*` attributes, but we can always address that later. I added `data-toolbarsection`.

packages/ag-charts-community/src/components/toolbar/toolbar.ts
lsjroberts: I may add an `IconWidget` at some point, and then it can just be `dragHandle.appendChild(new IconWidget('drag-handle'))`.

I also looked at adding `MouseWidgetEvent<'mousedown'>` but that percolated through to the region manager and it needed to be converted to a `'click'` event here - https://github.com/ag-grid/ag-charts/blob/latest/packages/ag-charts-community/src/chart/interaction/regionManager.ts#L179 - which seemed a little risky? So I've left this for now.

packages/ag-charts-community/src/components/toolbar/floatingToolbar.ts
lsjroberts: Agreed, I'll look into converting popovers to widgets shortly.

packages/ag-charts-community/src/components/toolbar/toolbar.ts
lsjroberts: Yeah I didn't want to cause issues with existing uses of it. Not sure if we would need two separate styles bounds and offset bounds functions, or if offset would cover all cases?

packages/ag-charts-community/src/components/toolbar/toolbar.ts
lsjroberts: Realised that the data attribute is not needed, and changed it to just a property on `ToolbarButtonWidget`.

packages/ag-charts-enterprise/src/features/annotations/annotationOptionsToolbar.ts
lsjroberts: Ideally yeh, there's some typing issues that I hope will be resolved when I change the popovers to extend widgets.

packages/ag-charts-community/src/components/toolbar/toolbar.ts
olegat: That's fine. Leave it as is for now.

packages/ag-charts-community/src/components/toolbar/toolbar.ts
olegat: 🤔 Hmm. Yes, you are right. I was thinking of toggle-buttons, but that uses the `aria-pressed` attribute (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#associated_aria_roles_states_and_properties).

We want `SwitchWidget.setChecked` instead of `ButtonWidget.setChecked` to avoid misuse.

alantreadway (approved):

PR #3011, opened 2024-11-19
AG-13306 Move MockEvents creation into the test files

olegat:
https://ag-grid.atlassian.net/browse/AG-13306

Follow-up of:
- https://github.com/ag-grid/ag-charts/pull/2956


packages/ag-charts-community/src/chart/test/caster.ts
olegat, line 1: > [!IMPORTANT]
> The bulk of this PR's complexity is in this file. The `Caster` class accesses object properties and uses the jest `check()` method verify assumptions at runtime. This better than doing something like `((chart as any).seriesAreaManager) as SeriesAreaManager`, because it allows us to access private properties in the tests whilst still getting meaningful test failures if an assumption fails.

packages/ag-charts-community/src/chart/test/findTarget.ts
olegat, line 1: > [!NOTE]
> All the code `testFindTarget` methods from the ag-charts-community code has been moved into this file, with some modifications to codify assumptions.

packages/ag-charts-community/src/chart/test/caster.ts
alantreadway: LGTM, nice!

Could you add this into the `ag-charts-test` package instead of `ag-charts-community` as it's a completely standalone test utility?

packages/ag-charts-community/src/scene/bbox.ts
```diff
@@ -56,6 +56,8 @@ export class BBox implements BBoxValues, BBoxContainsTester, DistantObject, Inte
         return nearestSquared(x, y, boxes);
     }
 
+    constructor(x: unknown, y: unknown, width: unknown, height: unknown);
```
alantreadway: I'd rather we didn't do this, it weakens type safety in the library code for test convenience, if I'm understanding the intent and implication here correctly?

Could you roll-back the constructor overloads in the rest of this PR too if my assumptions are correct.

packages/ag-charts-community/src/chart/test/caster.ts
olegat: Of course. Will do 👍

packages/ag-charts-community/src/scene/bbox.ts
olegat: The reason for these constructor overloads is to avoid using the `any` type, which contagiously hides compilation errors. For example, in the current `HEAD` of this branch, the [`domProxy.axes`](https://github.com/ag-grid/ag-charts/blob/2cdff79200906ba89cc505b3d7a180e27cee9ff0/packages/ag-charts-community/src/chart/test/findTarget.ts#L84) variable is of type `NativeWidget<HTMLElement>[]`.

Without the `NativeWidget(elem: unknown)` constructor overloads, the compiler will raise errors like these:

```
packages/ag-charts-community/src/chart/test/findTarget.ts:82:56 - error TS2345: Argument of type 'typeof NativeWidget' is not assignable to parameter of type 'UnknownConstructor<NativeWidget<HTMLElement>>'.
  Types of construct signatures are incompatible.
    Type 'new <TElem extends HTMLElement = HTMLElement>(elem: TElem) => NativeWidget<TElem>' is not assignable to type 'new (...args: unknown[]) => NativeWidget<HTMLElement>'.
      Types of parameters 'elem' and 'args' are incompatible.
        Type 'unknown' is not assignable to type 'HTMLElement'.

  80 |             .castProperty('axes', Array)
  81 |             .findArrayElementProperties('axes', 'div')
> 82 |             .castArrayElementProperties('axes', 'div', NativeWidget).value;
     |                                                        ^
  83 | 
  84 |         for (const axis of domProxy.axes) {
  85 |             const bbox = axis.div.getBounds();
```

An easy way around this error is to change the constructor arguments type:
```diff
diff --git a/packages/ag-charts-community/src/chart/test/caster.ts b/packages/ag-charts-community/src/chart/test/caster.ts
index ad31c9a6f..14bb8b016 100644
--- a/packages/ag-charts-community/src/chart/test/caster.ts
+++ b/packages/ag-charts-community/src/chart/test/caster.ts
@@ -58,7 +58,7 @@ type ElementsOfArrayPropertiesOf<T> =
  */
 type KeysOfUnionOfArrayPropertiesOf<T> = keyof ElementsOfArrayPropertiesOf<T>;
 
-type UnknownConstructor<T> = new (...args: unknown[]) => T;
+type UnknownConstructor<T> = new (...args: any[]) => T;
 
 function convert<NewT>(caster: Caster<unknown>): Caster<NewT> {
     return caster as Caster<NewT>;
```

However, with this `unknown -> any` change, the [`domProxy.axes`](https://github.com/ag-grid/ag-charts/blob/2cdff79200906ba89cc505b3d7a180e27cee9ff0/packages/ag-charts-community/src/chart/test/findTarget.ts#L84) variable is would be of type `NativeWidget<any>[]`, which is definitely undesirable.

As a compromise though, we could do something like this (more verbose, but we can wrap the `as` statement in a function):
```diff
diff --git a/packages/ag-charts-community/src/chart/test/findTarget.ts b/packages/ag-charts-community/src/chart/test/findTarget.ts
index 0951eb86d..c67681b66 100644
--- a/packages/ag-charts-community/src/chart/test/findTarget.ts
+++ b/packages/ag-charts-community/src/chart/test/findTarget.ts
@@ -79,7 +79,7 @@ function findZoomTarget(zoomModule: unknown, canvasX: number, canvasY: number):
             .findProperty('axes')
             .castProperty('axes', Array)
             .findArrayElementProperties('axes', 'div')
-            .castArrayElementProperties('axes', 'div', NativeWidget).value;
+            .castArrayElementProperties('axes', 'div', NativeWidget as new (elem: unknown) => NativeWidget).value;
 
         for (const axis of domProxy.axes) {
             const bbox = axis.div.getBounds();
```

Please let me know your thoughts

olegat: The best solution that I could find is providing both the `NativeWidget` and `typeof NativeWidget` (i.e. constructor) to the generic methods. The downside is that the caller code would be very verbose (e.g. `.cast<NativeWidget, typeof NativeWidget>(NativeWidget)` 🫨 , so I've added a `CAST_INFO` to reduce some of that verbosity.

It's still not ideal because writing up the `CAST_INFO` declaration is tedious, but I still think it's the best that we can do because we could in theory auto-generate the Typescript code that declares this constant.

alantreadway (approved):
Nice, thanks for persevering on this one!

PR #3012, opened 2024-11-19
Use popover API for tooltips

jacobp100:
https://ag-grid.atlassian.net/browse/AG-13454


packages/ag-charts-community/src/chart/tooltip/springAnimation.ts
```diff
@@ -0,0 +1,114 @@
...
+    private y = NaN;
+    private vx = 0;
+    private vy = 0;
+    private t0 = NaN;
+    private animationFrameHandle: number | undefined = undefined;
+
+    private onUpdate: ((e: SpringAnimationUpdateEvent) => void) | null = null;
+
+    addEventListener(_type: 'update', handler: (e: SpringAnimationUpdateEvent) => void) {
+        this.onUpdate = handler;
```
alantreadway: This doesn't really follow the pattern for other event emitters, could we either support multiple listeners, or extend `BaseListener` to inherit all of the default behaviours and handling? Or fail by throwing an error if this would overwrite an already registered listener?

olegat (commented):
✅ I've had a quick look. LGTM because you're only using canvasX / canvasY and only updating the tooltip (which uses canvas coords). You're not reading anything from the scene-graph or any other nest DOM elements so you don't need relative-coords

alantreadway (approved):
LGTM

PR #3018, opened 2024-11-20
AG-13455: Axis `nice: false` does not work correctly with series cont…

iMoses:
…aining only values of 0
https://ag-grid.atlassian.net/browse/AG-13455


alantreadway (approved):
LGTM

PR #3019, opened 2024-11-20
AG-13390 Do not call updateZoom if the select rect is empty

olegat:
https://ag-grid.atlassian.net/browse/AG-13390


alantreadway (approved):
LGTM

PR #3020, opened 2024-11-21
Improve benchmark compatibility with v9

alantreadway:
- **Fix benchmark v9.x.y backwards compatibility.**
- **Improve historic version benchmark flow.**

### DIFF
--- a/packages/ag-charts-community/benchmarks/benchmark.ts
+++ b/packages/ag-charts-community/benchmarks/benchmark.ts
@@ -6,7 +6,17 @@ import { AgChartInstance, AgChartOptions } from 'ag-charts-types';
 import { AgCharts } from '../src/main';
 import { Point } from '../src/scene/point';
 import { extractImageData, setupMockCanvas } from '../src/util/test/mockCanvas';
-import { isAtOrAfterVersion, isHistoricBenchmarkTest, prepareTestOptions, waitForUpdate } from './compatibility.ts';
+import {
+    getVersion,
+    isAtOrAfterVersion,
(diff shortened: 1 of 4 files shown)


packages/ag-charts-community/benchmarks/benchmark.ts
```diff
@@ -36,7 +46,8 @@ export class BenchmarkContext<T extends AgChartOptions = AgChartOptions> {
             await this.chart?.update(this.options);
             return;
         }
-        await (AgCharts as any).update(this.chart, this.options);
+        (AgCharts as any).update(this.chart, this.options);
+        await this.waitForUpdate();
```
jacobp100, line 50: Has our API changed?

jacobp100: Or is this for backwards compatibility?

jacobp100 (approved):

packages/ag-charts-community/benchmarks/benchmark.ts
alantreadway: Yes, so in 10.0.0 we added the promise-based + chart-ref centric APIs, but prior to that we didn't have either of these aspects, it appears (or the promise only waited for option application, not rendering) - for v9 updates to be timed correctly, we need to explicitly call `waitForUpdate()` to correctly wait for the chart to re-render.

alantreadway: (this change is in the <10.0.0 path)

PR #3022, opened 2024-11-21
AG-7083 rescale y axis when zooming

jacobp100:


packages/ag-charts-community/src/chart/cartesianChart.ts
```diff
@@ -47,13 +47,19 @@ export class CartesianChart extends Chart {
...
 
     private firstSeriesTranslation = true;
 
     override onAxisChange(newValue: ChartAxis[], oldValue?: ChartAxis[]) {
         super.onAxisChange(newValue, oldValue);
-        this.ctx?.zoomManager.updateAxes(newValue);
+
+        if (this.ctx != null) {
+            this.ctx.zoomManager.updateAxes(newValue);
+            this.ctx.seriesBoundsManager.axes = newValue;
```
alantreadway: Nit: Can we follow the same pattern as `ZoomManager` and use an `updateAxes()` method?

packages/ag-charts-community/src/chart/cartesianChart.ts
```diff
@@ -47,13 +47,19 @@ export class CartesianChart extends Chart {
 
     constructor(options: ChartOptions, resources?: TransferableResources) {
         super(options, resources);
+
+        this.ctx.seriesBoundsManager.axes = this.axes;
```
alantreadway: This is probably redundant, since `onAxisChange()` is likely called after this.

packages/ag-charts-community/src/chart/series/cartesian/lineSeries.ts
```diff
@@ -149,7 +149,8 @@ export class LineSeries extends CartesianSeries<
                 id: `yValueRaw`,
                 ...common,
                 invalidValue: undefined,
-            })
+            }),
+            yRangeLookup('yValueRaw')
```
alantreadway: Just for the record/reference, I think a key part of this approach will be to ensure we don't do this calculation if autoScaling is not enabled.

packages/ag-charts-community/src/chart/series/cartesian/__image_snapshots__/line-series-test-ts-line-series-create-for-line-number-x-axis-missing-x-data-example-it-should-render-to-canvas-as-expected-1-snap.png
iMoses, line 1: Regression?

packages/ag-charts-community/src/chart/series/cartesian/lineSeries.ts
jacobp100: Made the computation lazy. I don't think it's the _correct_ way to do this - but it certainly is the easiest. Has the added benefit if a chart can zoom and never is, the work is never done

packages/ag-charts-community/src/chart/cartesianChart.ts
jacobp100: `ctx` isn't available because the constructor hasn't finished

packages/ag-charts-community/src/chart/cartesianChart.ts
jacobp100: Updated

alantreadway (approved):
Performance + memory use benchmarks look great, LGTM

PR #3023, opened 2024-11-21
Fix tooltip position in scroll

jacobp100:
https://ag-grid.atlassian.net/browse/AG-13454


alantreadway (approved):
LGTM, but test failures?

PR #3028, opened 2024-11-22
AG-13280 - update sparkline interfaces

StephM-W:


packages/ag-charts-community/src/api/preset/sparkline.ts
```diff
@@ -235,15 +242,27 @@ function axisPreset(
...
 }
 
-function gridLinePreset(opts: AgSparklineAxisOptions | undefined, defaultEnabled: boolean): AgAxisGridLineOptions {
+function gridLinePreset(
+    opts: AgSparklineAxisOptions | undefined,
+    defaultEnabled: boolean,
+    sparkOpts: AgSparklineOptions
+): AgAxisGridLineOptions {
     const gridLineOpts: AgAxisGridLineOptions = {};
+    const isBarVertical = sparkOpts.type === 'bar' && sparkOpts.direction !== 'horizontal';
```
jacobp100: We have a check for `isVertical` (called `swapAxes`) below. Maybe we could pass that boolean in instead

packages/ag-charts-community/src/api/preset/sparkline.ts
```diff
@@ -293,6 +312,10 @@ export function sparkline(opts: AgSparklineOptions): AgCartesianChartOptions {
     const { data, series: [seriesOverrides] = [] } = sparklineDataPreset(baseData);
 
     const seriesOptions = optsRest as any as AgCartesianSeriesOptions;
+
+    if (seriesOptions.highlightStyle?.item && seriesOptions.type !== 'bar') {
+    }
+
```
jacobp100:
```suggestion
```

packages/ag-charts-types/src/chart/labelOptions.ts
```diff
@@ -15,7 +15,7 @@ export interface AgChartLabelOptions<TDatum, TParams> extends Toggleable, FontOp
     formatter?: Formatter<AgChartLabelFormatterParams<TDatum> & TParams>;
 }
 
-export interface AgChartLabelFormatterParams<TDatum> extends AgChartCallbackParams<TDatum> {
+export interface AgChartLabelFormatterParams<TDatum = any> extends AgChartCallbackParams<TDatum> {
```
jacobp100: Preferably we don't do this so we remember to type them. Why was this needed?

packages/ag-charts-community/src/api/preset/sparkline.ts
```diff
@@ -182,18 +186,21 @@ export function sparklineDataPreset(data: any[] | undefined): {
             const mappedData = data.map((y, x) => ({ x, y }));
             return { data: mappedData, series: [{ xKey: 'x', yKey: 'y' }] };
         } else if (Array.isArray(firstItem)) {
-            const mappedData = data.map(([x, y]) => ({ x, y }));
+            const mappedData = data.filter((item) => !!item).map(([x, y]) => ({ x, y }));
```
alantreadway: We shouldn't be doing this filtering here - what is the motivation for this?

packages/ag-charts-types/src/presets/sparkline/sparklineAxisOptions.ts
```diff
@@ -28,17 +28,17 @@ export interface AgSparklineContinuousAxisOptions<TDatum extends Date | number>
 }
 
 export interface AgSparklineCategoryAxisOptions extends AgSparklineBaseCategoryAxisOptions {
-    type: 'category';
+    type?: 'category';
```
jacobp100:
```suggestion
    type: 'category';
```

We should enforce it at the type level, but if they don't pass it in, we can just let it slide

packages/ag-charts-types/src/chart/labelOptions.ts
StephM-W: The grid callbacks don't force users to type them, this is to align with the current convention

packages/ag-charts-community/src/api/preset/sparkline.ts
```diff
@@ -235,15 +242,27 @@ function axisPreset(
...
         gridLineOpts.enabled ??= true;
+    } else if (isBarVertical) {
+        gridLineOpts.style = [{ stroke: DEFAULT_AXIS_GRID_COLOUR }];
+        gridLineOpts.enabled ??= true;
     }
     if (opts?.strokeWidth != null) {
         gridLineOpts.width = opts?.strokeWidth;
         gridLineOpts.enabled ??= true;
+    } else if (isBarVertical) {
+        gridLineOpts.width = 2;
```
alantreadway: We should avoid hardcoding where possible, why do we need a vertical-case specific hardcoding too? I'd normally suggest where feasible to put this fallback default in the theme.

packages/ag-charts-types/src/chart/labelOptions.ts
alantreadway: Is this needed?

packages/ag-charts-types/src/chart/tooltipOptions.ts
```diff
@@ -61,7 +61,7 @@ export interface AgTooltipRendererResult {
     class?: CssColor;
 }
 
-export interface AgSeriesTooltipRendererParams<TDatum> extends AgChartCallbackParams<TDatum> {
+export interface AgSeriesTooltipRendererParams<TDatum = any> extends AgChartCallbackParams<TDatum> {
```
alantreadway: Ditto?

packages/ag-charts-community/src/api/preset/sparkline.ts
StephM-W: that other check isn't quite the same, it checks for !bar || vertical. In this case, this must only apply to vertical bar charts (we don't currently default to showing the gridline for line or area charts)

packages/ag-charts-types/src/presets/sparkline/sparklineAxisOptions.ts
alantreadway: I'd prefer we kept these required, what was the motivation here?

packages/ag-charts-types/src/presets/sparkline/sparklineAxisOptions.ts
```diff
@@ -28,17 +28,17 @@ export interface AgSparklineContinuousAxisOptions<TDatum extends Date | number>
...
 }
 
 export interface AgSparklineNumberAxisOptions
     extends AgSparklineBaseAxisOptions,
         AgSparklineContinuousAxisOptions<number> {
-    type: 'number';
+    type?: 'number';
```
alantreadway: Ditto

packages/ag-charts-types/src/presets/sparkline/sparklineAxisOptions.ts
```diff
@@ -28,17 +28,17 @@ export interface AgSparklineContinuousAxisOptions<TDatum extends Date | number>
...
 }
 
 export interface AgSparklineTimeAxisOptions extends AgSparklineBaseAxisOptions, AgSparklineContinuousAxisOptions<Date> {
-    type: 'time';
+    type?: 'time';
```
alantreadway: Ditto?

packages/ag-charts-types/src/series/cartesian/barOptions.ts
```diff
@@ -22,7 +22,7 @@ export interface AgBarSeriesLabelOptions<TDatum, TParams> extends AgChartLabelOp
     padding?: PixelSize;
 }
 
-export interface AgBarSeriesItemStylerParams<TDatum>
+export interface AgBarSeriesItemStylerParams<TDatum = any>
```
alantreadway: Ditto

packages/ag-charts-community/src/api/preset/sparkline.ts
```diff
@@ -182,18 +186,21 @@ export function sparklineDataPreset(data: any[] | undefined): {
...
 }
 
 function axisPreset(
     opts: AgSparklineAxisOptions | undefined,
     defaultType: AgCartesianAxisOptions['type']
 ): AgCartesianAxisOptions {
+    opts = {
+        ...opts,
+        type: opts?.type ?? defaultType,
+    } as any as AgSparklineAxisOptions;
```
alantreadway: Presuming was can avoid this if `axis.type` is required?

packages/ag-charts-types/src/presets/sparkline/sparklineAxisOptions.ts
StephM-W: Updated with a different approach, let me know what you think

packages/ag-charts-community/src/api/preset/sparkline.ts
StephM-W: Our current specs state that we ignore undefined entries in the array - happy to move the filtering elsewhere. Discovered it when one of our examples threw an NPE here

packages/ag-charts-community/src/api/preset/sparkline.ts
StephM-W: It's only applicable for vertical bar charts, and our themes don't allow that level of specificity. But, as the default for horizontal bar charts is hidden anyway, we can move this default to the theme

packages/ag-charts-community/src/api/preset/sparkline.ts
StephM-W: It currently defaults to 'category' in prod, so it's optional from the user perspective. There's a comment higher up where we're discussing al alternative approach

packages/ag-charts-types/src/presets/sparkline/sparklineAxisOptions.ts
StephM-W: Ok, my proposed doesn't fit the bill for other reasons - could you suggest a way to achieve the strong typing optionality you mention?

packages/ag-charts-types/src/chart/labelOptions.ts
StephM-W: As above, to align with the grid convention for callbacks that can optionally be untyped

packages/ag-charts-types/src/chart/tooltipOptions.ts
StephM-W: As above, to align with the grid convention for callbacks that can optionally be untyped

packages/ag-charts-types/src/series/cartesian/barOptions.ts
StephM-W: As above, to align with the grid convention for callbacks that can optionally be untyped

packages/ag-charts-community/src/api/preset/sparkline.ts
StephM-W: Simplified

packages/ag-charts-community/src/api/preset/sparkline.ts
StephM-W: Moved to theme

packages/ag-charts-community/src/api/preset/sparkline.ts
StephM-W: Removed, type is now required

packages/ag-charts-types/src/series/cartesian/barOptions.ts
StephM-W: Reverted

packages/ag-charts-types/src/presets/sparkline/sparklineAxisOptions.ts
StephM-W: Reverted

packages/ag-charts-types/src/presets/sparkline/sparklineAxisOptions.ts
StephM-W: Reverted

packages/ag-charts-types/src/presets/sparkline/sparklineAxisOptions.ts
StephM-W: Reverted

packages/ag-charts-types/src/chart/tooltipOptions.ts
StephM-W: Reverted

packages/ag-charts-types/src/chart/labelOptions.ts
StephM-W: Reverted

packages/ag-charts-community/src/api/preset/sparkline.ts
StephM-W: Moving to the grid's sparklineRenderer

packages/ag-charts-types/src/presets/sparkline/sparklineAxisOptions.ts
StephM-W: Reverted

packages/ag-charts-types/src/chart/labelOptions.ts
StephM-W: Reverted

packages/ag-charts-community/src/api/preset/sparkline.ts
```diff
@@ -245,6 +275,9 @@ function gridLinePreset(opts: AgSparklineAxisOptions | undefined, defaultEnabled
         gridLineOpts.width = opts?.strokeWidth;
         gridLineOpts.enabled ??= true;
     }
+    if (sparkOpts.type === 'bar' && sparkOpts.direction !== 'horizontal') {
+        gridLineOpts.enabled ??= true;
+    }
```
jacobp100, line 277: Could we just pass in `direction` here since we're working it out below?

jacobp100 (approved):

packages/ag-charts-community/src/api/preset/sparkline.ts
```diff
@@ -67,6 +68,26 @@ const crossHairTooltip = {
...
+    },
+    log: {
+        gridLine: barGridLineDefaults,
+    },
+    time: {
+        gridLine: barGridLineDefaults,
+    },
+    category: {
+        gridLine: barGridLineDefaults,
+    },
```
jacobp100, line 85: Do we need `ordinal-time` too?

StephM-W: Actually, current spec is only category, number and time. I'll remove log.

packages/ag-charts-community/src/api/preset/sparkline.ts
StephM-W: I don't think the change is trivial, especially when direction really only applies to bar sparklines here.

alantreadway (approved):
LGTM

PR #3040, opened 2024-11-25
AG-13314: Support configuring secondary time & log axes

iMoses:
https://ag-grid.atlassian.net/browse/AG-13314
note: support can be improved, this fix is to make sure we don't crash in such cases


alantreadway (approved):
LGTM

PR #3043, opened 2024-11-25
AG-12190 - Rework setLicenseKey() API to match Grid.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-12190

### DIFF
--- a/packages/ag-charts-enterprise/src/main-modules.ts
+++ b/packages/ag-charts-enterprise/src/main-modules.ts
@@ -1,9 +1,15 @@
 import { AgCharts, VERSION, _Scene, _Theme, _Util, setupCommunityModules } from 'ag-charts-community';
 import type { IntegratedModule } from 'ag-charts-types';
 
+import { LicenseManager as RealLicenseManager } from './license/licenseManager';
 import { setupEnterpriseModules as internalSetup } from './setup';
 
-export { AgCharts, VERSION };
+const LicenseManager = {
+    setLicenseKey(key: string) {
+        RealLicenseManager.setLicenseKey(key);
+    },
+};
+export { AgCharts, VERSION, LicenseManager };
 
 export function setupEnterpriseModules() {
     internalSetup();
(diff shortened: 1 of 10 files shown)


taktran (approved):
LGTM 👍🏼

packages/ag-charts-enterprise/src/main-modules.ts
```diff
@@ -1,9 +1,15 @@
...
+import { LicenseManager as RealLicenseManager } from './license/licenseManager';
 import { setupEnterpriseModules as internalSetup } from './setup';
 
-export { AgCharts, VERSION };
+const LicenseManager = {
+    setLicenseKey(key: string) {
+        RealLicenseManager.setLicenseKey(key);
+    },
+};
+export { AgCharts, VERSION, LicenseManager };
```
iMoses, line 12: why aren't we exporting the `RealLicenseManager`?

packages/ag-charts-types/src/integratedCharts.ts
```diff
@@ -97,7 +97,7 @@ export interface IntegratedModule {
     create(options: AgChartInstanceOptions): AgChartInstance<AgChartInstanceOptions>;
     createSparkline(options: AgSparklineOptions): AgChartInstance<AgSparklineOptions>;
     setup(): void;
-    setGridContext(gridContext: boolean): void;
-    setLicenseKey(licenseKey: string): void;
+    setGridContext?(gridContext: boolean): void;
+    setLicenseKey?(licenseKey: string): void;
```
iMoses, line 101: It was Stephen who asked for a consistent API between community and enterprise.
I personally think it's fine, but maybe worth checking that it's consistent with the grid's approach.

alantreadway: https://github.com/ag-grid/ag-grid/pull/9292/files

packages/ag-charts-enterprise/src/main-modules.ts
alantreadway: Just from the perspective of only exporting the absolutely necessary API, `LicenseManager` has a larger API surface than users need to consider, so this 1) simplifies the exported API (from a user perspective) and 2) avoids users being able to do unexpected things with the `LicenseManager`.

packages/ag-charts-types/src/integratedCharts.ts
alantreadway: I've requested review from @StephenCooper

PR #3045, opened 2024-11-26
AG-13359 Fix context menu focus with multiple charts

olegat:
https://ag-grid.atlassian.net/browse/AG-13359


packages/ag-charts-community/src/util/keynavUtil.ts
olegat, line 1: > [!IMPORTANT]
> The fix is entirely contained in this file. The rest of the PR just adds an e2e test for this edge case.

packages/ag-charts-website/src/content/docs/accessibility-test/_examples/opening-context-menu-second-chart/main.ts
```diff
@@ -0,0 +1,72 @@
...
+    <script>
+      options1.container = document.getElementById('myChart1'),
+      options2.container = document.getElementById('myChart2'),
+      AgCharts.create(options1);
+      AgCharts.create(options2);
+    </script>
+  </body>
+</html>
+`);
+document?.close();
```
alantreadway: Do we need to use this approach to replicate the failure? (We do have support for multiple charts in an example without needing to do this, see https://github.com/ag-grid/ag-charts/tree/latest/packages/ag-charts-website/src/content/docs/benchmarks/_examples/summary )

olegat: Perhaps not. I was copying https://github.com/ag-grid/ag-charts/tree/e333372f134b145ea5b542998f5c0b382723ad6b/packages/ag-charts-website/src/content/docs/layout-test/_examples/layout-inline

I'll try replicating the same pattern as the benchmark example.

olegat: Done.

alantreadway (approved):
LGTM

PR #3057, opened 2024-11-27
AG-13505 Fix TC1 using `.focus( {preventScroll: true} )`

olegat:
https://ag-grid.atlassian.net/browse/AG-13505


alantreadway (approved):
LGTM

PR #3060, opened 2024-11-27
AG-13396 AG-13398 Add shared widget for annotations and chart toolbars

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13396
https://ag-grid.atlassian.net/browse/AG-13398

This fixes the issue with not being able to use the arrow keys to go between the annotations toolbar buttons and the chart toolbar series type selector button.

They both now use the same shared toolbar widget which they retrieve and add their buttons to via `ToolbarManager.getSharedToolbar()`.

The shared toolbar then translates the indices and limits the event listeners to the right section.

In the long run we will either want this to be more dynamic, or handled by the layout magic.


packages/ag-charts-community/src/chart/toolbar/toolbarManager.ts
alantreadway, line 1: What's the reasoning for having this in Community, it seems relatively annotations-specific (and that there is a singleton `SharedToolbar` also seems to reinforce the thought)?

olegat: In production, the `'chartToolbar'` and `'annotations'` toolbars are in reality part of the same `<div role="toolbar" aria-label="Financial Charts">`. So there needs to be a getter somewhere in the `ChartContext` for these components to add buttons to.

However, it's debatable whether a new `ToolbarManager` class with just 1 getter unnecessary. We could also add a getter to existing class like `DOMManager`.

packages/ag-charts-community/src/chart/toolbar/sharedToolbar.ts
olegat, line 1: What is the responsibility of `SharedToolbar`? It looks like this class primary responsibility is to manage to the ordering of the buttons in the DOM. I've added the `Widget.domIndex` for that purpose.

https://github.com/ag-grid/ag-charts/blob/2ab3fa48fd4111b761d4303dbba13e01ba9de9cc/packages/ag-charts-community/src/widget/widget.ts#L62

Simply setting `domIndex = 0` on the chart-type button should take care of everything, in theory. Although upon closer inspection it looks like there's a bug in `Widget.getBefore`, the `.filter()` shouldn't ignore `child.domIndex === undefined`:
https://github.com/ag-grid/ag-charts/blob/2ab3fa48fd4111b761d4303dbba13e01ba9de9cc/packages/ag-charts-community/src/widget/widget.ts#L163

I think that changing the `Widget.domIndex` type and default should fix that:

```diff
diff --git a/packages/ag-charts-community/src/widget/widget.ts b/packages/ag-charts-community/src/widget/widget.ts
index a32f51a1a..1c17a4d90 100644
--- a/packages/ag-charts-community/src/widget/widget.ts
+++ b/packages/ag-charts-community/src/widget/widget.ts
@@ -16,7 +16,7 @@ type EventType = keyof WidgetEventMap;
 
 interface IWidget<TElement extends HTMLElement> {
     index: number;
-    domIndex?: number;
+    domIndex: number;
     destroy(): void;
     getElement(): TElement;
 }
@@ -59,7 +59,7 @@ export abstract class Widget<
 
     // WARNING (not implemented): setting domIndex will not move it in the DOM. This property is currently only used
     // when adding child widgets.
-    public domIndex?: number;
+    public domIndex: number = Infinity;
 
     protected readonly children: TChildWidget[] = [];
     protected htmlListener?: WidgetListenerHTML;
@@ -165,7 +165,6 @@ export abstract class Widget<
 
         if (domIndex === undefined) return undefined;
         return this.children
-            .filter((child): child is R => child.domIndex !== undefined && child.domIndex > domIndex)
             .reduce<R | undefined>((prev, curr) => (!prev || curr.domIndex < prev.domIndex ? curr : prev), undefined);
     }
 
```

packages/ag-charts-community/src/chart/toolbar/toolbarManager.ts
lsjroberts: Alan's added a `contextModule` thing which I'll try out. It'll make this a small module in enterprise that sits in the module context.

https://github.com/ag-grid/ag-charts/pull/3065

packages/ag-charts-community/src/chart/toolbar/sharedToolbar.ts
lsjroberts: Will have a look.

I also wanted it to handle translating indices so the modules don't need to know where they sit within the shared toolbar. They use these indices to update the buttons (since value is not common across the toolbars, only index).

For example, the annotations module can assume the 0th index is its first button, and the shared toolbar will actually translate that to the 1st index.

packages/ag-charts-enterprise/src/features/annotations/annotationsToolbar.ts
```diff
@@ -32,29 +33,38 @@ class AnnotationsToolbarButtonProperties extends ToolbarButtonProperties {
...
     private readonly annotationMenu = new Menu(this.ctx, 'annotations');
 
     private readonly destroyFns: (() => void)[] = [];
 
     constructor(private readonly ctx: _ModuleSupport.ModuleContext) {
         super();
 
-        this.toolbar.addClass('ag-charts-annotations__toolbar');
-        ctx.domManager.addChild('canvas-overlay', 'annotations-toolbar', this.toolbar.getElement());
+        console.log(ctx);
```
alantreadway: Nit: ?

alantreadway (approved):
LGTM, thanks for the updates!

packages/ag-charts-community/src/chart/toolbar/sharedToolbar.ts
lsjroberts: I'm not sure this is going to help in this case, due to the translating index thing I mentioned above.

alantreadway (approved):
LGTM

PR #3063, opened 2024-11-27
AG-13349 Fix dragging events

olegat:
https://ag-grid.atlassian.net/browse/AG-13349

### DIFF
--- a/packages/ag-charts-community/src/chart/interaction/regionManager.ts
+++ b/packages/ag-charts-community/src/chart/interaction/regionManager.ts
@@ -186,7 +186,7 @@ export class RegionManager {
         return map[widgetEvent.type];
     }
 
-    private computeEventOffsets(currentWidget: Widget, widget: Widget, widgetEvent: TWidgetEvent) {
+    private computeEventOffsets(currentWidget: Widget, rootWidget: Widget, widgetEvent: TWidgetEvent) {
         const { deltaX, deltaY } = InteractionManager.getWheelDeltas(widgetEvent.sourceEvent);
 
         if ('mockRegion' in widgetEvent.sourceEvent) {
@@ -196,11 +196,11 @@ export class RegionManager {
             }
         }
 
-        const widgetRect = widget.getElement().getBoundingClientRect();
+        const rootRect = rootWidget.getElement().getBoundingClientRect();
         const currentWidgetRect = currentWidget.getElement().getBoundingClientRect();
         return {
-            canvasX: widgetEvent.clientX - widgetRect.x,
-            canvasY: widgetEvent.clientY - widgetRect.y,
+            canvasX: widgetEvent.clientX - rootRect.x,
+            canvasY: widgetEvent.clientY - rootRect.y,
             regionX: widgetEvent.clientX - currentWidgetRect.x,
             regionY: widgetEvent.clientY - currentWidgetRect.y,
             deltaX,
@@ -209,17 +209,13 @@ export class RegionManager {
     }
 
     // Create and dispatch a copy of the InteractionEvent.
-    private dispatch(
-        current: Region | undefined,
-        widget: Widget,
-        widgetEvent: TWidgetEvent,
-        regionEventType?: 'leave' | 'enter'
-    ) {
+    private dispatch(current: Region | undefined, widgetEvent: TWidgetEvent, regionEventType?: 'leave' | 'enter') {
         const { widget: currentWidget } = current?.properties ?? {};
-        if (current == null || currentWidget == null) return;
+        const { widget: rootWidget } = this.regions.root.properties;
+        if (current == null || currentWidget == null || rootWidget == null) return;
 
         const event: RegionEvent = buildPreventable({
-            ...this.computeEventOffsets(currentWidget, widget, widgetEvent),
+            ...this.computeEventOffsets(currentWidget, rootWidget, widgetEvent),
             sourceEvent: widgetEvent.sourceEvent,
             type: this.widgetEventTypeToRegionEventType(widgetEvent, regionEventType),
             region: current.properties.name,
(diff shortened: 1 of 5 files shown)


packages/ag-charts-community/src/widget/widgetListenerInternal.ts
```diff
@@ -91,30 +98,62 @@ export class WidgetListenerInternal {
...
 
-        window.addEventListener('mousemove', mousemove);
-        window.addEventListener('mouseup', mouseup);
+        window.addEventListener('mousedown', mousegeneral, { capture: true });
+        window.addEventListener('mouseenter', mousegeneral, { capture: true });
+        window.addEventListener('mouseleave', mousegeneral, { capture: true });
+        window.addEventListener('mouseout', mousegeneral, { capture: true });
+        window.addEventListener('mouseover', mousegeneral, { capture: true });
+        window.addEventListener('mousemove', mousemove, { capture: true });
+        window.addEventListener('mouseup', mouseup, { capture: true });
```
alantreadway: Bit of an edge case, but if the chart is destroyed in the middle of a drag, will these event handlers leak?

packages/ag-charts-community/src/chart/interaction/regionManager.ts
```diff
@@ -261,12 +251,16 @@ export class RegionManager {
                 this.blockNextClickEvent = false;
                 break;
             }
-            case 'hover': {
-                if (!this.isDragMoving) {
+            case 'leave':
+            case 'enter':
+                if (!this.isDragMoving || this.deferredDragStart != null) {
```
alantreadway, line 256: I'm not 100% following the logic here, but just to confirm that we won't issue drag-related events until a mouse move happens after a mouse down still, to avoid firing conflicting click and drag events on mouse-up?

packages/ag-charts-community/src/widget/widgetListenerInternal.ts
olegat: No. How could that happen? `WidgetListenerInternal.destroy()` only removes `'mousedown'` event handlers on the HTMLElement, but if a drag is currently being processed then those `'mousedown'` event handles do not matter. The `'mouseup'` listener would still be registered in the `window` object, so it'll remove the handlers when it's called.

alantreadway: >  so it'll remove the handlers when it's called.

The chart may have been destroyed before this point though, so there is a stale clousre still referencing the destroyed chart hanging around, and it would be very easy for that chart being in an unexpected state - to mean that these handlers fail in unexpected ways, before being removed.

Could you take a more defensive approach here, and ensure that on `destroy()` any of the temporary handlers are disposed of?

packages/ag-charts-community/src/chart/interaction/regionManager.ts
olegat: Correct. That's what the `blockNextClickEvent` state handles: We fire either 1 `'click'` event or a set of `'drag*'` events but not both.

The `case 'leave','enter':` logic stops us broadcasting which region changes when dragging. Example steps:

1. `'mousedown'` fired on the series-area.
2. `'mousemove'` fired such that the pointer leaves the bounds the series-area HTML element.
3. `'mouseleave'` fired as pointer hovers over a new HTML element (we don't want to broadcast this, because the user is still dragging).
4. `'mousemove'` fired such that the pointer returns to the bounds the series-area HTML element.
3. `'mouseenter'` fired as pointer hovers back to the series-area element (we don't want to broadcast this either; the user never really left because they are still dragging)

packages/ag-charts-community/src/widget/widgetListenerInternal.ts
olegat: I understand the importance of risk mitigation, but there are two important things to remember here:

1. We agreed on the design decision that the `widgets/` directory should be as isolated and independent as possible for separation of concerns (similar to the `scene/` directory). Therefore the state of the chart is of no relevance at this point. Only `this: WidgetListenerListener`, `window` and `current: T` (which is essentially an HTMLElement) are relevant.
2. The `this: WidgetListenerListener` is responsible for allocating these handlers but those not "own" them, they are conceptually owned by the `window`. Therefore removing the handlers in `WidgetListenerListener.destroy()` is more likely to cause problems than to solve anything. We would, for example, need to make sure that they are no bugs caused by double-destruction.

The only thing that the `mouseup` handler really needs from `this` is a callback. Therefore, we must go with an approach like this which is much much simpler (and safer):

```diff
diff --git a/packages/ag-charts-community/src/widget/widgetListenerInternal.ts b/packages/ag-charts-community/src/widget/widgetListenerInternal.ts
index 164a36904..588f9f780 100644
--- a/packages/ag-charts-community/src/widget/widgetListenerInternal.ts
+++ b/packages/ag-charts-community/src/widget/widgetListenerInternal.ts
@@ -128,9 +128,8 @@ export class WidgetListenerInternal {
                 window.removeEventListener('mouseover', mousegeneral, { capture: true });
                 window.removeEventListener('mousemove', mousemove, { capture: true });
                 window.removeEventListener('mouseup', mouseup, { capture: true });
-                const dragEndEvent = makeDragEvent('drag-end', origin, upEvent);
-                this.dispatch('drag-end', current, dragEndEvent);
-                this.endDrag(current, dragEndEvent);
+                const callback: undefined | ((upEvent) => void) = getCallback(upEvent);
+                callback?.(upEvent);
             }
         };
 
```

olegat: Fixed here: https://github.com/ag-grid/ag-charts/commit/a4709c65aa6f5dac828bf95b51b73c2c7c14a5f2

alantreadway: Ok, let's roll with this then.

alantreadway (approved):
LGTM

PR #3067, opened 2024-11-28
AG-13099: Fix grouped category labels on mini-chart

iMoses:
Fix rotation support for depth zero.
Update tests to include entire chart.
Cleanup examples.

https://ag-grid.atlassian.net/browse/AG-13099


alantreadway (approved):
LGTM

PR #3073, opened 2024-11-29
AG-13353 - Update homepage hero to have brand colour

taktran:


packages/ag-charts-website/src/components/homepage/components/hero/ImageCarousel.tsx
```diff
@@ -9,15 +8,6 @@ const ImageCarousel = () => {
     const rightColumnRef = useRef(null);
     const [scrollHeight, setScrollHeight] = useState(0);
 
-    const [darkMode] = useDarkmode();
```
taktran, line 12: Not sure if we want the carousel images to be dark or not. If we want them to be dark, we can revert the last commit of this PR

PR #3074, opened 2024-11-29
AG-13464 Fix zoom toolbar hide when zoom module disabled

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13464

### DIFF
--- a/packages/ag-charts-enterprise/src/features/zoom/zoom.ts
+++ b/packages/ag-charts-enterprise/src/features/zoom/zoom.ts
@@ -205,7 +205,8 @@ export class Zoom extends _ModuleSupport.BaseModuleInstance implements _ModuleSu
             ctx.updateService.addListener('update-complete', (event) => this.onUpdateComplete(event)),
             ctx.zoomManager.addListener('zoom-change', (event) => this.onZoomChange(event)),
             ctx.zoomManager.addListener('zoom-pan-start', (event) => this.onZoomPanStart(event)),
-            this.panner.addListener('update', (event) => this.onPanUpdate(event))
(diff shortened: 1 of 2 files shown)


packages/ag-charts-enterprise/src/features/zoom/zoomModule.ts
```diff
@@ -44,6 +44,7 @@ export const ZoomModule: _ModuleSupport.Module = {
     packageType: 'enterprise',
     chartTypes: ['cartesian', 'topology'],
     moduleFactory: (ctx) => new Zoom(ctx),
+    removable: false, // Toggling this module requires tear down of dom elements
```
lsjroberts: This is an issue on all modules that have associated dom elements. Changing it to `false` fixes this, but does that leave another issue for which the modules were removed in the first place?

alantreadway: This is probably a gap in the module implementations - they should handle removal correctly by destroying all resources they manage.

For the Sparklines implementation, we identified that all modules were being initialised, and so made a change to actually lazy initialise, and therefore destroy modules on an update to minimise additional resource allocations.

I'd prefer we fixed this by handling module destruction properly - and raise a distinct ticket for the other cases perhaps to be followed up seperately?

lsjroberts: I've updated it to destroy the elements when the module is removed then recreate in the constructor, and hide them when it's just toggling the `zoom.buttons.enabled` property.

I'll use this pattern for the ranges and annotation toolbars.

alantreadway (approved):
LGTM - thanks!

PR #3076, opened 2024-11-29
AG-13361 Fix ranges toolbar to teardown dom when disabled

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13361


alantreadway (approved):
LGTM

PR #3079, opened 2024-11-29
AG-13286 - Add Vue package tests.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-13286


tests/vue-package-tests/patches/App.vue
```diff
@@ -0,0 +1,14 @@
+<script setup lang="ts">
+import { AgCharts } from "ag-charts-vue3";
+
+// Insert options.partial here.
+</script>
+
+<template>
+  <main>
+    <AgCharts :options="options" />
+  </main>
```
jacobp100: I think if you get rid of the `main`s, it'll fix the legend in the screenshots

jacobp100 (approved):

tests/vue-package-tests/patches/App.vue
alantreadway: It didn't seem to help 🤷 Might be best to sit together on Monday and see if there is a simple fix for this, otherwise it's probably good enough right now to verify the package works.

PR #3082, opened 2024-12-02
AG-13396 Fix annotations and chart toolbars to hide when disabled

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13396

- Fixes create and destroy of shared toolbar, keeping it alive until both annotations and chartToolbar call `destroy()`.
- Changes shared toolbar to simplify the external api, removing the constant need to pass in the section e.g. `'annotations'`.
- Removes the `ToolbarSharedWidget` and just uses the base toolbar component.


packages/ag-charts-enterprise/src/features/shared-toolbar/sharedToolbar.ts
```diff
@@ -1,20 +1,168 @@
...
+            this.ctx.domManager.removeChild('canvas-overlay', 'shared-toolbar');
+        });
+    }
+
+    private toolbarWithSection<ButtonOptions extends _ModuleSupport.ToolbarButtonOptions>(
+        section: SharedToolbarSection
+    ): SharedToolbarWithSection<ButtonOptions> {
+        const sharedToolbar = this.sharedToolbar!;
+
+        const withSection = {
```
lsjroberts, line 63: This isn't entirely ideal, but the other option was duplicating tracking of active sections in both the `SharedToolbar` manager and the `SharedToolbarWidget`, along with the kinda meh duplication of functions with `section` prepended.

This essentially overrides the functions it needs to using the hidden section value from `.getSharedToolbar(section)`.

alantreadway (approved):
LGTM

PR #3087, opened 2024-12-02
AG-13553 Fix the focus indicator of box-plot series

olegat:
https://ag-grid.atlassian.net/browse/AG-13553

For reason, the box-plot datums BBoxes have negative heights. This is unusual, but there is nothing mathematically wrong about that (it just means that `y` is the bottom rather than the top). It is a problem when setting the CSS bounds however; we want a positive height for that. Therefore, normalize the input BBox before setting the CSS bounds.


alantreadway (approved):
LGTM

PR #3088, opened 2024-12-02
AG-13398 Change chartToolbar option to only be in theme

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13398


alantreadway (approved):
LGTM

PR #3093, opened 2024-12-02
AG-13503 Fix range bar animation

jacobp100:
https://ag-grid.atlassian.net/browse/AG-13503

### DIFF
--- a/packages/ag-charts-community/src/chart/data/dataController.ts
+++ b/packages/ag-charts-community/src/chart/data/dataController.ts
@@ -175,6 +175,11 @@ export class DataController {
     }
 
     private static groupMatch({ data, opts }: RequestedProcessing<any, any, any>) {
+        function isMergeable(group: RequestedProcessing<any, any, any>) {
+            // AG-13503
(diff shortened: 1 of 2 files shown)


packages/ag-charts-community/src/chart/data/dataController.ts
```diff
@@ -175,6 +175,11 @@ export class DataController {
     }
 
     private static groupMatch({ data, opts }: RequestedProcessing<any, any, any>) {
+        function isMergeable(group: RequestedProcessing<any, any, any>) {
+            // AG-13503
+            return group.opts.props.every((p) => p.type !== 'processor' && p.type !== 'reducer');
```
jacobp100, line 180: Crux of the issue is these will get overwritten if we merge these requests. With more time, we could fix this to scope the reduced data per series

alantreadway: Unfortunately we can't take this approach, as stacked cases rely on shared processing with processors in the mix - I had a recollection that the processors also used to get merged, would be worth checking if that has changed?

PR #3095, opened 2024-12-02
AG-13397 Fix popover drag handles and disabled buttons

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13397

Changes drag handles to be widgets, the existing `'mousemove'` events were being consumed somewhere else while the mouse was down.

Dialog and popover components will be properly converted to use widgets at a later date.

### DIFF
--- a/packages/ag-charts-community/src/components/popover/draggablePopover.ts
+++ b/packages/ag-charts-community/src/components/popover/draggablePopover.ts
@@ -1,4 +1,6 @@
 import { Vec2 } from '../../util/vector';
+import type { NativeWidget } from '../../widget/nativeWidget';
+import type { DragWidgetEvent } from '../../widget/widgetEvents';
 import { Popover, type PopoverOptions } from './popover';
 
 export abstract class DraggablePopover<Options extends PopoverOptions = PopoverOptions> extends Popover<Options> {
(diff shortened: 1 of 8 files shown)


olegat:
The regression looks like a bug in `RegionManager`, which is not surprising because its implementation has recently changed to use `widgets/` rather than `InteractionManager` (https://github.com/ag-grid/ag-charts/commit/24b8db92baa394cd1f884249ea3a4d38efd4dfa5). The `RegionManager.dispatchEvent` is broadcasting an `'enter'` event to the `'series'` region when the dragging starts. That shouldn't happen though.

It would be preferable to fix this regression and correcting the dispatch bug of `RegionManager`.

**EDIT** Actually, it seems like the `RegionManager.dispatchEvent` only sends `'enter'` when the pointer leaves the drag handle bound, which is in fact correct. The `'mousemove'` events are almost certainly being highjacked by the `startDrag` handler on the .ag-charts-canvas-proxy element:
https://github.com/ag-grid/ag-charts/blob/8b09067ce80900ccfdcb75785d9a67ca090722c0/packages/ag-charts-community/src/widget/widgetListenerInternal.ts#L38

Therefore using a `Widget` for the drag handle is a correct approach 👍

olegat (approved):
LGTM subject to green CI

alantreadway (approved):
LGTM

PR #3110, opened 2024-12-03
AG-10574 Tooltip changes

jacobp100:
https://ag-grid.atlassian.net/browse/AG-10574


packages/ag-charts-community/src/chart/interaction/tooltipManager.css
```diff
@@ -17,115 +17,97 @@
     box-shadow: 0 4px 8px rgba(0, 0, 0, 0.08);
 }
 
-.ag-chart-tooltip--compact {
+.ag-charts-tooltip--compact {
     --ag-charts-content-padding: var(--ag-charts-content-compact-padding);
 }
 
-.ag-chart-tooltip-wrap-always {
+.ag-charts-tooltip--arrow::before {
```
jacobp100, line 25: This replaces the previous arrow mechanism, it means we can inherit the background and border, so you can just style the tooltip, and it cascades through

alantreadway (approved):
LGTM

PR #3111, opened 2024-12-03
AG-13564 - Remove tooltip + crossline label CSS class options.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-13564


alantreadway:
Merged to allow Jacob to progress on #3110

PR #3112, opened 2024-12-03
AG-13570 - Fix naming of crosshair label CSS classes.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-13570

### DIFF
--- a/packages/ag-charts-enterprise/src/features/crosshair/crosshairLabel.css
+++ b/packages/ag-charts-enterprise/src/features/crosshair/crosshairLabel.css
@@ -21,6 +21,6 @@
     color: var(--ag-charts-axis-label-color);
 }
 
-.ag-charts-crosshair-label-hidden {
+.ag-charts-crosshair-label--hidden {
     visibility: hidden !important;
 }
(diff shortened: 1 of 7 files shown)


packages/ag-charts-enterprise/src/features/crosshair/crosshairLabel.css
```diff
@@ -13,14 +13,14 @@
...
+.ag-charts-crosshair-label-content {
     padding: 0 8px;
     border-radius: 2px;
     line-height: 20px;
     background-color: var(--ag-charts-axis-label-background-color);
     color: var(--ag-charts-axis-label-color);
 }
 
-.ag-crosshair-label-hidden {
+.ag-charts-crosshair-label-hidden {
```
jacobp100: If it's not too much hassle, I think we should still keep the modifiers clear (this is what I did in tooltips)

```suggestion
.ag-charts-crosshair-label--hidden {
```

jacobp100 (approved):

packages/ag-charts-enterprise/src/features/crosshair/crosshairLabel.css
alantreadway: Applied

PR #3113, opened 2024-12-03
AG-13357 Add `seriesVisibilityChange` listener API

olegat:
https://ag-grid.atlassian.net/browse/AG-13357


eslint.config.mjs
```diff
@@ -45,6 +45,24 @@ export const sonarjsConfig = [
...
+            // Example:
+            //   function f (a?: undefined);
+            //   function g (a: undefined);
+            //
+            //   f(undefined); // OK: compiles.
+            //   f();          // OK: compiles. (equivalent to `f(undefined)`)
+            //   g(undefined); // OK: compiles.
+            //   g();          // FAIL: will not compile because parameter #0 is missing.
+            //
+            'sonarjs/no-redundant-optional': 0,
```
alantreadway: Where was this causing an issue? I managed to fix up all violations of this during the v9 eslint upgrade?

packages/ag-charts-community/src/util/object.ts
```diff
@@ -131,3 +131,22 @@ export function deepFreeze<T>(obj: T): T {
...
+
+export function defineSetterGetter<T, P extends string, V>(
+    o: T,
+    p: P,
+    attributes: { set: (v: V) => void; get: () => V }
+): T & { [K in P]: V };
+export function defineSetterGetter<T, P extends string, V>(
+    o: T,
+    p: P,
+    attributes: { set?: undefined; get: () => V }
```
alantreadway: This looks wrong - probably we should either:
- Omit `set` here.
- Make it `unknown` or `never` rather than `undefined`.

eslint.config.mjs
alantreadway: Also - these SonarJS rules are still enforced in our SonarCloud reports irrespective of the ESLint config, so this specific issue will still be reported there. Ergo, we should actually address the issue raised.

See my comments on `object.ts`, but I think we can tackle this by being more accurate with the typing in `object.ts`.

packages/ag-charts-community/src/chart/legend/legendEvent.ts
```diff
@@ -0,0 +1,42 @@
...
+): AgChartLegendClickEvent;
+
+export function makeLegendItemEvent<T extends 'dblclick'>(
+    type: T,
+    itemId: string,
+    seriesId: string,
+    event: Event
+): AgChartLegendDoubleClickEvent;
+
+export function makeLegendItemEvent(
```
jacobp100, line 17: Potentially we could change the legend click event to include `type: 'click' | 'dblclick'`, and we could just instantiate a single class rather than use this helper. I don't feel massively strongly here though

packages/ag-charts-community/src/chart/legend/legendEvent.ts
```diff
@@ -0,0 +1,42 @@
...
+        preventDefault: (): void => {
+            _defaultPrevented = true;
+        },
+    };
+
+    return defineSetterGetter(that, 'defaultPrevented', {
+        get: function () {
+            return _defaultPrevented;
+        },
+    });
```
jacobp100:
```suggestion
    return {
        type,
        itemId,
        seriesId,
        event,
        preventDefault: (): void => {
            _defaultPrevented = true;
        },
        get defaultPrevented() {
          return _defaultPrevented;
        }
    };
```

jacobp100 (approved):

packages/ag-charts-community/src/util/object.ts
```diff
@@ -131,3 +131,22 @@ export function deepFreeze<T>(obj: T): T {
...
+): T & { readonly [K in P]: V };
+export function defineSetterGetter<T, P extends string, V>(
+    o: T,
+    p: P,
+    attributes: { set?: (v: V) => void; get: () => V }
+): T {
+    Object.defineProperty(o, p, attributes);
+    return o;
+}
```
alantreadway: This keeps the same signature but without needing to disable linting rules:
```suggestion
export function defineSetterGetter<T, P extends string, V, S extends ((v: V) => void) | undefined>(
    o: T,
    p: P,
    attributes: { set?: S; get: () => V }
) {
    Object.defineProperty(o, p, attributes);
    return o as S extends undefined ? T & { readonly [K in P]: V } : T & { [K in P]: V };
}
```

packages/ag-charts-community/src/util/object.ts
olegat: It's not wrong. The overload simply states that the type of the `set` is undefined, and that it can be explicitly set to undefined or implicitly set to by omitting the property.

However, it seems that `never` accomplishes the same thing so I'll try with that.

packages/ag-charts-community/src/util/object.ts
olegat: Thanks for the feedback. I've tried this out, but the return type isn't quite what you'd expect.
```typescript
const a = defineSetterGetter( {}, 'p', { get: () => 'foobar'} );
// a is of type:
// * `{ readonly p: 'foobar'} | { p: 'foobar' }` when using `as`
// * `{ readonly p: 'foobar'}`  when using an overload
```

The two types are the same in practice, but the overload is tidier.

packages/ag-charts-community/src/chart/legend/legendEvent.ts
olegat: I don't think that this matters. The aim of this function to just use a raw object rather than a class to avoid exposing private members in API event.

eslint.config.mjs
olegat: obsolete.

packages/ag-charts-community/src/chart/legend/legendEvent.ts
olegat: Thanks for the suggestion. I've done this. I think it's much cleaner.

packages/ag-charts-community/src/util/object.ts
olegat: obsolete.

packages/ag-charts-community/src/util/object.ts
olegat: obsolete.

alantreadway (approved):
LGTM

PR #3124, opened 2024-12-04
AG-13542 Use a ButtonWidget in axisButton.ts

olegat:
https://ag-grid.atlassian.net/browse/AG-13542

Also add a `'drag-start'` event handler to said ButtonWidget. This fixes a regression with drag events on axis buttons.


alantreadway (approved):
LGTM - thanks for this!

PR #3125, opened 2024-12-04
AG-13559 Fix annotation dialog closing when options toolbar moves

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13559


packages/ag-charts-enterprise/src/features/annotations/annotationOptionsToolbar.ts
```diff
@@ -387,8 +387,6 @@ export class AnnotationOptionsToolbar extends _ModuleSupport.BaseProperties {
     }) {
         const { buttonBounds, popoverBounds } = event;
 
-        this.hideOverlays();
```
lsjroberts, line 390: This was a carry-over from before the refactor, it's not needed here as the menus and dialog are already closed by clicking elsewhere.

alantreadway (approved):
LGTM

PR #3136, opened 2024-12-04
AG-13581 Fix style ordering

jacobp100:
https://ag-grid.atlassian.net/browse/AG-13581

### DIFF
--- a/packages/ag-charts-community/src/dom/domManager.ts
+++ b/packages/ag-charts-community/src/dom/domManager.ts
@@ -367,12 +367,30 @@ export class DOMManager extends BaseManager<Events['type'], Events> {
         };
 
         const addStyleElement = (el: HTMLElement) => {
-            for (const child of el.children as any as Iterable<Element>) {
+            const metaElements = new Set(['TITLE', 'META']);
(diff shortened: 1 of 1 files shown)


alantreadway (approved):
LGTM

packages/ag-charts-community/src/dom/domManager.ts
```diff
@@ -367,12 +367,30 @@ export class DOMManager extends BaseManager<Events['type'], Events> {
         };
 
         const addStyleElement = (el: HTMLElement) => {
-            for (const child of el.children as any as Iterable<Element>) {
+            const metaElements = new Set(['TITLE', 'META']);
```
BernieSumption, line 370: In the grid we don't bother skipping title and meta elements and we just add the styles to the beginning of the head. Is there a reason to do this? I'm not sure, but if there is maybe the grid should do so too, and if there isn't maybe charts shouldn't bother.

alantreadway: From recollection, there was some evidence that the `meta` tag specifying document character encoding was expected before anything else, which is partly to do with this choice.

PR #3144, opened 2024-12-05
AG-13577 Fix changing legend item label

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13577

Also minor cleanup on state manager.

### DIFF
--- a/packages/ag-charts-community/src/chart/legend/legend.ts
+++ b/packages/ag-charts-community/src/chart/legend/legend.ts
@@ -26,6 +26,7 @@ import { getWindow } from '../../util/dom';
 import { createId } from '../../util/id';
 import { Logger } from '../../util/logger';
 import { clamp } from '../../util/number';
+import { objectsEqual } from '../../util/object';
 import { BaseProperties } from '../../util/properties';
 import { ObserveChanges } from '../../util/proxy';
 import { CachedTextMeasurerPool, TextUtils } from '../../util/textMeasurer';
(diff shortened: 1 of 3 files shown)


packages/ag-charts-community/src/chart/legend/legend.ts
```diff
@@ -302,13 +303,10 @@ export class Legend extends BaseProperties {
...
 
-        const propertiesToCheck: (keyof CategoryLegendDatum)[] = ['seriesId', 'itemId', 'enabled', 'legendItemName'];
+        const propertiesToCheck = ['seriesId', 'itemId', 'enabled', 'legendItemName', 'label.text'];
         const shouldUpdate =
             data.length !== legendData.length ||
-            data.some((_v, index, _a) => {
-                const [newValue, oldValue] = [legendData[index], data[index]];
-                return propertiesToCheck.some((p) => newValue[p] !== oldValue[p]);
-            });
+            data.some((_v, index, _a) => !objectsEqualKeys(propertiesToCheck, legendData[index], data[index]));
```
lsjroberts: It might be we can just do the complete `objectsEqual()` check here, not sure if there is reason we are comparing specific properties? @manapeirov 

But this just adds `'label.text'` to the list to compare.

alantreadway (approved):
LGTM, @manapeirov ?

packages/ag-charts-community/src/chart/legend/legend.ts
manapeirov: Yeah we can just do objectsEqual() technically everything should be the same for us not to update it

manapeirov (approved):
Lgtm thank you

packages/ag-charts-community/src/chart/legend/legend.ts
lsjroberts: Updated to use `objectsEqual()`.

alantreadway (approved):

PR #3164, opened 2024-12-06
AG-13526 Update Locale translations

JamesSwinton:


packages/ag-charts-locale/src/es-ES.ts
```diff
@@ -83,6 +89,11 @@ export const AG_CHARTS_LOCALE_ES_ES: Record<string, string> = {
     overlayLoadingData: 'Cargando datos...',
     overlayNoData: 'No hay datos para mostrar',
     overlayNoVisibleSeries: 'No hay series visibles',
+    overlayUnsupportedBrowser: 'Versión del navegador no compatible. Por favor, actualice su navegador.',
+    seriesHistogramTooltipCount: '${yName} (cuenta)',
+    seriesHistogramTooltipFrequency: 'Frecuencia',
+    seriesHistogramTooltipMean: '${yName} (media)',
+    seriesHistogramTooltipSum: '${yName} (suma)',
```
alantreadway, line 96: These won't work, the names inside curly braces are variable names, which shouldn't be translated - for these specific cases we should probably just clone the source values.

alantreadway (approved):
LGTM

PR #3167, opened 2024-12-06
AG-13603 Fix zoom anchor pointer to use canvas point

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13603


alantreadway (approved):
LGTM

PR #3169, opened 2024-12-06
Focus box fixes

jacobp100:


packages/ag-charts-community/src/chart/overlay/overlay.ts
```diff
@@ -19,7 +19,7 @@ export class Overlay extends BaseProperties {
     renderer?: () => string | HTMLElement;
 
     private content?: HTMLElement;
-    public focusBox?: { x: number; y: number; width: number; height: number };
+    public focusBox?: BBox;
```
olegat, line 22: Keep this type as-is, or better yet change it to `BBoxValue`

packages/ag-charts-community/src/dom/focusIndicator.ts
```diff
@@ -1,6 +1,6 @@
+import { BBox } from '../integrated-charts-scene';
```
olegat: Caution with these imports

packages/ag-charts-community/src/dom/focusIndicator.ts
```diff
@@ -23,10 +23,28 @@ export class FocusIndicator {
...
+        if (hoverRect == null || bounds == null) {
+            boundsVisible = true;
         } else if (bounds instanceof Path) {
+            boundsVisible = hoverRect.intersection(bounds.getBBox()) != null;
+        } else {
+            boundsVisible = hoverRect.intersection(bounds) != null;
+        }
+
+        if (!boundsVisible) {
+            this.element.style.display = 'none';
```
olegat: Do not touch the `display` property. The WCAG success criteria requires the focus indicator to remain visible at all times for keyboard users. If the input `bounds` does not intersect with the `hoverRect`, then it's much better to draw a focus indicator with a height or width of 0 than to draw nothing at all.

packages/ag-charts-community/src/dom/focusIndicator.ts
```diff
@@ -23,10 +23,28 @@ export class FocusIndicator {
         this.swapChain.addListener('swap', (parent) => this.onSwap(parent));
     }
 
-    updateBounds(bounds: Path | BBoxValues | undefined, hoverRect?: { x: number; y: number }) {
+    updateBounds(bounds: Path | BBox | undefined, hoverRect?: BBox) {
```
olegat, line 26: Keep the inputs as `BBoxValues`.

packages/ag-charts-community/src/scene/bbox.ts
```diff
@@ -149,6 +149,24 @@ export class BBox implements BBoxValues, BBoxContainsTester, DistantObject, Inte
         return dx * dx + dy * dy;
     }
 
+    /* Clip this BBox with the assumption that [this.x, this.y] are relative to [clipRect.x, clipRect.y] */
+    relativeClip(clipRect: BBox | undefined): this {
```
olegat: Move this `src/util/bboxinterface.ts`

packages/ag-charts-community/src/dom/focusIndicator.ts
```diff
@@ -36,7 +54,7 @@ export class FocusIndicator {
             this.path.setAttribute('d', bounds.svgPathData(transform));
             this.show(this.svg);
         } else {
-            setElementBBox(this.div, bounds);
+            setElementBBox(this.div, bounds.relativeClip(hoverRect));
```
olegat, line 52: Change this to `BBoxValues.relativeClip(bounds, hoverRect)` to avoid the unnecessary `BBox` type requirement.

olegat (changes requested):
Great work Jacob! I think that adding the clipping logic to `FocusIndicator.updateBounds` seems like the good approach.

Key takeaways:
- It's preferable to use `BBoxValues` and only use the `BBox` type when it is absolutely necessary
- Do not set the `display: none` CSS property if the focus indicator does not intersect with the hoverRect.
- The `ZoomManager.panToBBox` will still log an error if the focus indicator does not fit in the series area's bounds. Make sure to update that too.

packages/ag-charts-community/src/chart/overlay/overlay.ts
jacobp100: It's needed to use the BBox methods

packages/ag-charts-community/src/dom/focusIndicator.ts
```diff
@@ -23,9 +23,18 @@ export class FocusIndicator {
...
+    updateBounds(bounds: Path | BBox | undefined, hoverRect?: BBox) {
         if (bounds === undefined) {
-            // skip
+            return;
+        }
+
+        const boundsBBox = bounds instanceof Path ? bounds.getBBox() : bounds;
+        const boundsVisible = hoverRect != null && boundsBBox != null ? hoverRect.intersects(boundsBBox) : true;
+
+        if (!boundsVisible) {
```
olegat, line 38: This condition is unnecessary. The `relativeClip` in the `else` clause will set the height or width to 0 if necessary.

The only advantage of this `if(!boundsVisible)` condition is that it also handle SVG-based focus indicators which currently has no clipping in production. But I would prefer to leave the SVG rendering logic unchanged, even if our current implementation is a bit buggy.

alantreadway (approved):
Verified locally, pushed snapshot updates.

PR #3181, opened 2024-12-09
CRT-610 - Fix homepage automated example

taktran:
Archive is also production, so needs to be used in the conditionals first


alantreadway (approved):
LGTM

PR #3184, opened 2024-12-09
CRT-566 Fix zoom reset to last restored state

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-566


lsjroberts:
Targetting wrong branch

PR #3188, opened 2024-12-09
Focus overlay fixes

jacobp100:
https://ag-grid.atlassian.net/browse/CRT-580
https://ag-grid.atlassian.net/browse/CRT-569
https://ag-grid.atlassian.net/browse/CRT-588
https://ag-grid.atlassian.net/browse/CRT-612
https://ag-grid.atlassian.net/browse/CRT-592

### DIFF
--- a/packages/ag-charts-community/src/chart/series/dataModelSeries.ts
+++ b/packages/ag-charts-community/src/chart/series/dataModelSeries.ts
@@ -24,6 +24,7 @@ export abstract class DataModelSeries<
     protected dataModel?: DataModel<any, any, any>;
     protected processedData?: ProcessedData<any>;
     protected showFocusBox: boolean = true;
+    protected clipFocusBox: boolean = true;
 
     protected getScaleInformation({
         xScale,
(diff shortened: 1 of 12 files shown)


packages/ag-charts-community/src/chart/series/dataModelSeries.ts
```diff
@@ -24,6 +24,7 @@ export abstract class DataModelSeries<
     protected dataModel?: DataModel<any, any, any>;
     protected processedData?: ProcessedData<any>;
     protected showFocusBox: boolean = true;
+    protected clipFocusBox: boolean = true;
```
olegat, line 27: This is fine for this release.

However, ideally this should be part of a utility function like `computeBarFocusBounds` and `computeMarkerFocusBounds` to avoid needing to override this value for these marker-based and bar-based chart type. That would require a significant refactor however, so please file a ticket in the backlog to tidy this up post-release. (In the fact, the showFocusBox property is obsolete too, because it's always `true`).

packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
```diff
@@ -411,7 +412,15 @@ export class SeriesAreaManager extends BaseManager {
...
-        const keyboardEvent = makeKeyboardPointerEvent(hoverRect, this.focusIndicator, pick);
+        drawPickedFocus(
+            seriesRect != null ? new BBox(0, 0, seriesRect.width, seriesRect?.height) : undefined,
+            this.focusIndicator,
+            pick,
+            seriesRect?.x,
+            seriesRect?.y
+        );
+
+        const keyboardEvent = makeKeyboardPointerEvent(hoverRect, pick);
```
olegat, line 423: nit (optional): you could add the `seriesRect` parameter to `makeKeyboardPointerEvent` instead of exporting `drawPickedFocus`.

olegat (approved):

PR #3190, opened 2024-12-09
CRT-576 Ensure empty `role="list"` widgets stay hidden

olegat:
https://ag-grid.atlassian.net/browse/CRT-576


alantreadway (approved):
LGTM

PR #3192, opened 2024-12-09
CRT-572 - Fix label formatting issue

taktran:


taktran:
Should be a library fix.

PR #3194, opened 2024-12-09
CRT-595 Item styler fixes

jacobp100:
https://ag-grid.atlassian.net/browse/CRT-595

The fix flagged up way more issues than expected


alantreadway (approved):
LGTM

PR #3199, opened 2024-12-09
CRT-574 Fix dragging of annotations out of y-bounds, ignoring channels

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-574

Also does a bit of cleaning.


alantreadway (approved):
LGTM

PR #3200, opened 2024-12-09
CRT-570 Align copy and hero button style

kylerphillips:


markdurrant (changes requested):
Sass stuff looks good. 

I'm not a fan of changing `See the charts` => `See demos`. Just because we don't have a "Demos" nav item. 


I'd happily keep it the same, or we could go for `See Gallery`?

markdurrant (approved):
LGTM 👍

PR #3209, opened 2024-12-09
CRT-631 - Improve false drag detection in favour of detecting for small time/distance moves with the mouse down.

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-631


lsjroberts (approved):
LGTM

PR #3218, opened 2024-12-10
CRT-625

jacobp100:
https://ag-grid.atlassian.net/browse/CRT-625

Manually verified every series. There was an issue where providing series area padding could offset the positions too - so I fixed that also


alantreadway (approved):
LGTM

PR #3221, opened 2024-12-10
CRT-626 Use datum refPoint in makeKeyboardPointerEvent

olegat:
https://ag-grid.atlassian.net/browse/CRT-626

This fixes the tooltip positioning on radial columns, because the center of sector's bounds is not necessary inside the sector.


alantreadway (approved):
LGTM

PR #3222, opened 2024-12-10
CRT-629- Fix note annotation text position when rendered below icon

manapeirov:
PR for https://ag-grid.atlassian.net/browse/CRT-629


lsjroberts (approved):
LGTM

PR #3249, opened 2024-12-11
AG-13619 Noramlize coordinates for focus indicator

jacobp100:
https://ag-grid.atlassian.net/browse/AG-13619

### DIFF
--- a/packages/ag-charts-community/src/chart/series/cartesian/barUtil.ts
+++ b/packages/ag-charts-community/src/chart/series/cartesian/barUtil.ts
@@ -7,6 +7,7 @@ import type { Scale } from '../../../scale/scale';
 import { BBox } from '../../../scene/bbox';
 import type { DropShadow } from '../../../scene/dropShadow';
 import type { Rect } from '../../../scene/shape/rect';
+import { Transformable } from '../../../scene/transformable';
 import { isNegative } from '../../../util/number';
 import { mergeDefaults } from '../../../util/object';
 import type { ChartAxis } from '../../chartAxis';
@@ -14,6 +15,7 @@ import { ChartAxisDirection } from '../../chartAxisDirection';
 import { createDatumId } from '../../data/processors';
 import type { Series } from '../series';
 import type { SeriesItemHighlightStyle } from '../seriesProperties';
+import type { ISeries } from '../seriesTypes';
 import type { CartesianSeriesNodeDatum } from './cartesianSeries';
 
 export type RectConfig = {
@@ -302,10 +304,11 @@ export function resetBarSelectionsFn(_node: Rect, { x, y, width, height, clipBBo
 }
 
 export function computeBarFocusBounds(
+    series: ISeries<any, any>,
     datum: { x: number; y: number; width: number; height: number } | undefined
 ): BBox | undefined {
     if (datum === undefined) return undefined;
 
     const { x, y, width, height } = datum;
-    return new BBox(x, y, width, height);
+    return Transformable.toCanvas(series.contentGroup, new BBox(x, y, width, height));
 }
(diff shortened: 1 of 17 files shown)


packages/ag-charts-community/src/chart/series/cartesian/barUtil.ts
```diff
@@ -302,10 +304,11 @@ export function resetBarSelectionsFn(_node: Rect, { x, y, width, height, clipBBo
 }
 
 export function computeBarFocusBounds(
+    series: ISeries<any, any>,
```
olegat: nit: `ISeries<any, any>` is a bloated and contagious type. In util functions, it's preferable to explicitly redeclare the specific properties required:
```suggestion
    series: { contentGroup: Group },
```

olegat (changes requested):
The intent of this PR is to address a discrepancy with focuses of type `Path` (using canvas coords) and `BBox` (using relative coords). I support the idea of normalising the coord space, but this approach of changes `BBox` focus to canvas coords feels counter intuitive.

The `FocusIndicator` class is only intended to indicate focus within the series area.

Screenshot 2024-12-12 at 10 05 28

All focus indicator outside of the series area (e.g. legend and navigator) use standard `:focus-visible` CSS styling and do not rely on `FocusIndicator`. Therefore, making the `FocusIndicator` receive focus paths/boxes in canvas coords is misleading. It also seems counter-productive because the scene-nodes also use the same relative coords as the `.ag-charts-swapchain` elements, so we need to convert the BBox's relative coords to canvas coords and then back to relative coords again.

To me, the root cause seems to be that `Path` class does not honour the scene-graph's relative coordinate system that's expected. Would be possible to address that instead?

jacobp100:
> To me, the root cause seems to be that Path class does not honour the scene-graph's relative coordinate system that's expected. Would be possible to address that instead?

It's not quite that - it _does_ honour the scene graph's relative coordinate system, but there's two issues: we have no way to apply a transform to a path without modifying the path, and we no longer have a way to compute the transform relative to a node - only the canvas. So using canvas coordinates for both was the least bad option

packages/ag-charts-community/src/chart/series/cartesian/barUtil.ts
jacobp100: Duplicating the interfaces like this makes more work if you ever change stuff like `contentGroup` - which we have done in the past

olegat: Yes that's true. However, the use of `any` is definitely an anti-pattern and we mustn't encourage its use. As compromise, you could try to see whether one of the alternatives is sufficient:
* `ISeries<never, never>`
* `ISeries<unknown, unknown>`
* `ISeries<{}, {}>`

olegat:
> It's not quite that - it _does_ honour the scene graph's relative coordinate system, but there's two issues: we have no way to apply a transform to a path without modifying the path, and we no longer have a way to compute the transform relative to a node - only the canvas. So using canvas coordinates for both was the least bad option

Okay. Thanks for the explanation. That's unfortunate, because if we ever need to write something else that uses either boxes or paths like the `FocusIndicator` then we'll run into this exact same problem again. But let's go with this option because the alternative doesn't sounds feasible right now.

olegat (approved):
LGTM (just 1 nit regarding the use of `any`).

packages/ag-charts-community/src/chart/series/cartesian/barUtil.ts
jacobp100: Updated

alantreadway (approved):
LGTM

PR #3256, opened 2024-12-12
AG-13669 Add a sanity-check for panToBBox when refresh=true

olegat:
https://ag-grid.atlassian.net/browse/AG-13669


alantreadway (approved):
LGTM, thanks for this!

PR #3262, opened 2024-12-13
AG-12786 - Add trend based fibonacci retracement

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-12786


packages/ag-charts-enterprise/src/features/annotations/fibonacci-retracement-trend-based/fibonacciRetracementTrendBasedScene.ts
```diff
@@ -0,0 +1,331 @@
...
+import { createFibonacciRangesData, getFibonacciCoords } from '../utils/fibonacci';
+import { validateDatumPoint } from '../utils/validation';
+import { convertLine, convertPoint, invertCoords } from '../utils/values';
+import type { FibonacciRetracementTrendBasedProperties } from './fibonacciRetracementTrendBasedProperties';
+
+const { Vec2, Vec4 } = _ModuleSupport;
+
+type ActiveHandle = StartEndHandle | 'endRetracement';
+
+export class FibonacciRetracementTrendBasedScene extends FibonacciScene<FibonacciRetracementTrendBasedProperties> {
```
lsjroberts, line 20: Can some of these functions be moved into `FibonnaciScene`? Looks like `dragHandle()` is identical.

For others its just the `endRetracement` handle has been added in a few statements and calls. I'm not sure if there's any real benefit to making these ones more generic? At some point it becomes too convoluted when instead a bit of duplication is easier to work with.

But I'm interested in your thoughts.

lsjroberts (approved):
LGTM. One comment but I'm happy for that to be considered after merging so it can head to testing.

packages/ag-charts-enterprise/src/features/annotations/fibonacci-retracement-trend-based/fibonacciRetracementTrendBasedScene.ts
manapeirov: Discussed offline and decided to leave as is for now

alantreadway (approved):

PR #3269, opened 2024-12-13
AG-13651 Fix toolbars to not change tab order when toggled

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13651

### DIFF
--- a/packages/ag-charts-enterprise/src/features/ranges/ranges.ts
+++ b/packages/ag-charts-enterprise/src/features/ranges/ranges.ts
@@ -11,15 +11,19 @@ export class Ranges extends _ModuleSupport.BaseModuleInstance implements _Module
     @Validate(OBJECT)
     public buttons = new PropertiesArray(RangesButtonProperties);
 
+    private readonly container: HTMLElement;
     private readonly toolbar: _ModuleSupport.BaseToolbar;
     private readonly verticalSpacing = 10;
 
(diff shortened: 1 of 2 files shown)


packages/ag-charts-enterprise/src/features/ranges/ranges.ts
```diff
@@ -11,15 +11,19 @@ export class Ranges extends _ModuleSupport.BaseModuleInstance implements _Module
...
     public buttons = new PropertiesArray(RangesButtonProperties);
 
+    private readonly container: HTMLElement;
     private readonly toolbar: _ModuleSupport.BaseToolbar;
     private readonly verticalSpacing = 10;
 
     constructor(private readonly ctx: _ModuleSupport.ModuleContext) {
         super();
 
+        this.container = ctx.domManager.addChild('canvas-overlay', 'range-buttons');
```
alantreadway, line 21: When does this element get destroyed/removed? Would be good to check what happens in Integrated Charts scenarios where chart-type is changed multiple times?

lsjroberts: This element is destroyed by the dom manager here - https://github.com/ag-grid/ag-charts/blob/latest/packages/ag-charts-community/src/dom/domManager.ts#L149

If it already exists it will be reused.

alantreadway (approved):

PR #3271, opened 2024-12-13
AG-13086 - Fix Numpad enter support.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-13086

### DIFF
--- a/packages/ag-charts-community/src/chart/interaction/keyNavManager.ts
+++ b/packages/ag-charts-community/src/chart/interaction/keyNavManager.ts
@@ -94,6 +94,10 @@ export class KeyNavManager extends InteractionStateListener<KeyNavEventType, Key
         if (altKey || shiftKey || metaKey || ctrlKey) return;
 
         this.focusIndicator?.overrideFocusVisible(true);
+        if (key === 'Enter') {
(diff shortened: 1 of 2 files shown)


packages/ag-charts-community/src/chart/interaction/keyNavManager.ts
```diff
@@ -110,7 +114,6 @@ export class KeyNavManager extends InteractionStateListener<KeyNavEventType, Key
             case 'Substract':
                 return this.dispatch('nav-zoom', -1, event);
             case 'Space':
-            case 'Enter':
```
olegat: Keep this line as a safe-guard.

Although it's probably redundant, it might not be. The `KeyboardEvent.key` property is localised and is therefore not guaranteed to be `'Enter'` on all keyboard layouts.

olegat (approved):
LGTM (just 1 minor nit), subject to e2e snapshot rendering as expected.

packages/ag-charts-community/src/chart/interaction/keyNavManager.ts
alantreadway: Fair point - reverted this line.

PR #3274, opened 2024-12-16
AG-13708 Remove `id` attributes from legend `<button>` elements

olegat:
https://ag-grid.atlassian.net/browse/AG-13708


alantreadway (approved):
LGTM, thanks!

PR #3275, opened 2024-12-16
AG-13715 Set data-preventdefault: false on text-edit elements

olegat:
This fixes a regression where the DOMManager calls `preventDefault()` on Arrow keydown events using the `stopPageScrolling()` method.


alantreadway (approved):
LGTM

PR #3278, opened 2024-12-16
AG-13489 Migrate interactive component away from Canvas based events

olegat:
https://ag-grid.atlassian.net/browse/AG-13489

### DIFF
--- a/packages/ag-charts-community/src/dom/domManager.ts
+++ b/packages/ag-charts-community/src/dom/domManager.ts
@@ -6,6 +6,7 @@ import { createElement, getDocument, getWindow } from '../util/dom';
 import { GuardedElement } from '../util/guardedElement';
 import { stopPageScrolling } from '../util/keynavUtil';
 import { type Size, SizeMonitor } from '../util/sizeMonitor';
+import { NativeWidget } from '../widget/nativeWidget';
 // TODO move to utils
 import BASE_DOM from './domLayout.html';
 
@@ -89,6 +90,10 @@ export class DOMManager extends BaseManager<Events['type'], Events> {
     private readonly observer?: IntersectionObserver;
     private readonly sizeMonitor = new SizeMonitor();
 
+    public readonly seriesWidget: NativeWidget<HTMLElement>;
+    public readonly chartWidget: NativeWidget<HTMLElement>;
+    public readonly containerWidget: NativeWidget<HTMLElement>;
+
     constructor(container?: HTMLElement, styleContainer?: HTMLElement) {
         super();
 
@@ -135,10 +140,17 @@ export class DOMManager extends BaseManager<Events['type'], Events> {
         const topGuard = createTabGuardElement(guardedElement, 'beforebegin');
         const botGuard = createTabGuardElement(guardedElement, 'afterend');
         this.tabGuards = new GuardedElement(guardedElement, topGuard, botGuard);
+
+        this.seriesWidget = new NativeWidget(this.getParent('series-area'));
+        this.chartWidget = new NativeWidget(this.getParent('canvas-proxy'));
+        this.containerWidget = new NativeWidget(this.getParent('canvas-container'));
     }
 
     override destroy() {
         super.destroy();
+        this.seriesWidget.destroy();
+        this.chartWidget.destroy();
+        this.containerWidget.destroy();
 
         this.observer?.unobserve(this.element);
         if (this.container) {
--- a/packages/ag-charts-enterprise/src/features/annotations/annotations.ts
+++ b/packages/ag-charts-enterprise/src/features/annotations/annotations.ts
@@ -1,4 +1,10 @@
-import { type AgAnnotation, type AgAnnotationLineStyleType, type Direction, _ModuleSupport } from 'ag-charts-community';
+import {
+    type AgAnnotation,
+    type AgAnnotationLineStyleType,
+    type Direction,
+    _ModuleSupport,
+    _Widget,
+} from 'ag-charts-community';
 
 import { TextInput } from '../text-input/textInput';
 import { AxesButtons } from './annotationAxesButtons';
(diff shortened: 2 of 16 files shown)


packages/ag-charts-community/src/dom/domManager.ts
```diff
@@ -89,6 +90,10 @@ export class DOMManager extends BaseManager<Events['type'], Events> {
     private readonly observer?: IntersectionObserver;
     private readonly sizeMonitor = new SizeMonitor();
 
+    public readonly seriesWidget: NativeWidget<HTMLElement>;
+    public readonly chartWidget: NativeWidget<HTMLElement>;
+    public readonly containerWidget: NativeWidget<HTMLElement>;
+
```
alantreadway, line 96: I don't think this is the right spot for these:
- `DOMManager` is mostly concerned with lower-level DOM primitives (e.g. raw `HTMLElement`s)
- use and configuration of these seems to happen in other parts of the codebase:
  -  bounds being set in `SeriesAreaManager`
   - event listeners being setup in `RegionManager`
   - references are pushed in `root` and `series` regions for use there

packages/ag-charts-enterprise/src/features/annotations/annotations.ts
```diff
@@ -447,8 +453,8 @@ export class Annotations extends _ModuleSupport.BaseModuleInstance implements _M
             seriesRegion.addListener('drag-start', this.onDragStart.bind(this), annotationsState),
             seriesRegion.addListener('drag', this.onDrag.bind(this), annotationsState),
             seriesRegion.addListener('drag-end', this.onDragEnd.bind(this), All),
-            ctx.interactionManager.addListener('keydown', this.onKeyDown.bind(this), All),
-            ctx.interactionManager.addListener('keyup', this.onKeyUp.bind(this), All),
+            ctx.domManager.seriesWidget.addListener('keydown', this.onKeyDown.bind(this)),
+            ctx.domManager.seriesWidget.addListener('keyup', this.onKeyUp.bind(this)),
```
alantreadway, line 457: Could these be listeners added to the `seriesRegion`?

packages/ag-charts-enterprise/src/features/zoom/zoomToolbar.ts
```diff
@@ -103,8 +102,8 @@ export class ZoomToolbar extends BaseProperties {
         this.destroyFns.push(
             this.toolbar.addToolbarListener('button-pressed', this.onButtonPress.bind(this)),
             this.toolbar.addToolbarListener('button-focused', this.onButtonFocus.bind(this)),
-            ctx.interactionManager.addListener('hover', this.onHover.bind(this), InteractionState.All),
-            ctx.interactionManager.addListener('leave', this.onLeave.bind(this), InteractionState.All),
+            ctx.domManager.containerWidget.addListener('mousemove', this.onHover.bind(this)),
+            ctx.domManager.containerWidget.addListener('mouseleave', this.onLeave.bind(this)),
```
alantreadway, line 106: Could these be listeners added to the `root` region?

packages/ag-charts-enterprise/src/features/annotations/annotations.ts
olegat: The `RegionManager` does not listen to `'keydown'` and `'keyup'` events.

In fact, the `seriesRegion` does not even listen to events on the series-area, it listens to events on the container. In the past, we have seen bugs from add listeners that are too broadly scoped. For example, ArrowLeft/ArrowRight key presses on the legend buttons could bubble up to the container and cause the focus indicator in the series-area to unintentionally move. Therefore using the `seriesWidget` is the preferred approach in the new tree-based interaction mechanism.

packages/ag-charts-community/src/dom/domManager.ts
olegat: That's true. `DOMManager` may not be the best place for this, but `SeriesAreaManager` wasn't much better either. The only reason I've put these here is because the `containerWidget` and `seriesWidget` need to be publicly available so that modules can add listeners at the desired depth in the DOM tree.

Can you propose a better place for these? It seems like making these part of the `RegionManager` would make more sense.

packages/ag-charts-enterprise/src/features/zoom/zoomToolbar.ts
olegat: The `'root'` region only fires `'hover'` events when hovering directly above the container (i.e. root) element, it does not bubble and therefore does not receive `'hover'` events when hovering over the series-area. As a consequence, the conditions in the `onHover` handler that determine the visibility state will always be `false`, because the zoom toolbar can only become visible when hovering in the series-area.

In theory, it could be added to the `'series'` and the `event.canvasY` value could be used to determine whether the zoom toolbar should be visible. However, this will fire a `'leave'` event when hovering over a zoom button, which toggles the visibility back to `false`, causing the zoom toolbar to wiggle up & down:

https://github.com/user-attachments/assets/f6020655-8d95-4a64-8d50-a8386bcb8692

In the past, we've addressed this sort of problem by adding a condition to `shouldIgnore`, but we want to move away from the approach because it's difficult to maintain.
https://github.com/ag-grid/ag-charts/blob/aa3e0f1bfe492cb0c1e2b86142895f5633748e94/packages/ag-charts-community/src/chart/interaction/regionManager.ts#L85

I initially tried adding event listeners to the `seriesWidget` instead. The issue with this however is that the zoom toolbar HTML elements are not a descendant of the `seriesWidget` element, so this still fires `'mouseleave'` events when hovering over a zoom button and produces the same affect as in this video above.

Adding a listener to the `containerWidget` ensures that `'mouseleave'` is only fired when the mouse completely exits the bounds of the entire chart, and always fire `'mousemove'` event within the chart to ensure that the correct visibility state is computed by `onHover`.

alantreadway:
Discussed offline, we'll merge this change and continue the dicussion about where the widgets should live distinctly.

PR #3282, opened 2024-12-17
Axis refactor

jacobp100:


packages/ag-charts-community/src/chart/axis/axis.ts
```diff
@@ -1282,14 +648,14 @@ export abstract class Axis<S extends Scale<D, number, TickInterval<S>> = Scale<a
...
-    private getScaleValueFormatter(format?: string): (value: any) => string {
-        if (format && this.scale.tickFormat) {
-            try {
-                return this.scale.tickFormat({ specifier: format });
-            } catch {
-                Logger.warnOnce(`the format string ${format} is invalid, ignoring.`);
-            }
-        }
+    private getScaleValueFormatter(_format?: string): (value: any) => string {
+        // if (format && this.scale.tickFormat) {
```
jacobp100: Need to check what to do here

libraries/ag-charts-test/src/canvas/mock-canvas-text.ts
```diff
@@ -10,6 +10,20 @@ export function mockCanvasText() {
     };
 }
 
+const replacements: Partial<Record<string, string>> = {
+    // Math minus
+    '\u2212': '-',
```
alantreadway, line 15: I'm confused why this is needed now, was it a long standing issue?

alantreadway (approved):
Awesome, one query but otherwise LGTM - let's get @iMoses to review too.

libraries/ag-charts-test/src/canvas/mock-canvas-text.ts
jacobp100: We didn't technically need it in the end - but if you add a label formatter using `f`, it uses this unicode character for the minus sign (which is correct). I left it in because it might be needed later

iMoses (approved):
LGTM

PR #3286, opened 2024-12-18
AG-11705 Add initial font theme globals

lsjroberts:

A few todos and will likely be iterated a bunch, but ready for an 'in principle' review.


packages/ag-charts-community/src/module/optionsModule.ts
```diff
@@ -256,6 +257,8 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
         // so we aim to only do this once in the processing flow.
         processedOptions = deepClone(processedOptions, ChartOptions.OPTIONS_CLONE_OPTS);
 
+        this.processOptionsParams(processedOptions);
```
lsjroberts, line 260: Since this visits every node within the object, it may be possible to combine it with the cloning and only do a single iteration.

But that may depend on how we handle ref's that point to ref's (if we don't just throw an error on that).

packages/ag-charts-types/src/chart/themeOptions.ts
```diff
@@ -78,9 +78,31 @@ export interface AgChartThemePalette {
...
+        ? ExtendLiteralLeaves<U, V>[]
+        : T[P] extends string | symbol | number | undefined
+          ? T[P] | V
+          : ExtendLiteralLeaves<T[P], V>;
+};
+
+type ThemeParam = keyof AgChartThemeParams;
+type ThemeParamsRef = { ref: ThemeParam };
+
+export type WithThemeParams<T> = ExtendLiteralLeaves<T, ThemeParamsRef>;
```
lsjroberts, line 99: For now, this type is only used internally. It wraps the series modules and annotations theme definitions, so we can use `as const` instead of `as unknown as string`.

It extends every literal leaf with `| { ref: ThemeParam }` (and other future ones like calc etc).

We could be more specific and instead only allow refs on specific options, e.g. `fontFamily: ThemeParamsRef<'fontFamily'>` etc, depending on how restrictive we'd want to be. But I figured being looser is a good starting point for now.

packages/ag-charts-community/src/module/optionsModule.ts
alantreadway: I think this is fine, although it would be preferable to do this resolution of refs in the `ChartTheme` creation step, since we cache `ChartTheme` instances based on the same object reference in the `theme` option, which would mean in update cases we don't need to redo ref resolution?

packages/ag-charts-types/src/chart/themeOptions.ts
alantreadway: LGTM, caveat we might need to see how that plays with the API doc rendering (it might be we need to adjust the rendering to ensure the docs are still readable).

packages/ag-charts-community/src/module/optionsModule.ts
lsjroberts: Yeh agreed, that was my initial attempt. However, it would fail when the theme option was a string. In that case it would only have a theme config for each series, and not the top level values like `params`. The defaults would then need to be injected in different places depending on the different times the theme was accessed.

I'll look into it more as we go, as it's possible the theme processing can be simplified down.

packages/ag-charts-types/src/chart/themeOptions.ts
lsjroberts: Note: I was able to remove the `as const` bit as well in the next PR https://github.com/ag-grid/ag-charts/pull/3299

PR #3294, opened 2024-12-19
AG-13694 Fix toggling legend item strokes

jacobp100:
https://ag-grid.atlassian.net/browse/AG-13694


packages/ag-charts-community/src/chart/legend/legend.ts
```diff
@@ -516,26 +513,24 @@ export class Legend extends BaseProperties {
             paddedSymbolWidth += spacing + markerWidth;
         }
 
-        if (this._symbolsDirty) {
-            const { marker, line } = markerLabel;
+        const { marker, line } = markerLabel;
```
jacobp100, line 516: Verified this change has no impact on performance - it doesn't re-render the marker/line if they didn't change

alantreadway (approved):
LGTM

PR #3295, opened 2024-12-19
Scale cleanup

jacobp100:


alantreadway (approved):
LGTM, maybe get @iMoses to take a look too?

packages/ag-charts-community/src/chart/axis/categoryAxis.ts
```diff
@@ -84,17 +46,7 @@ export class CategoryAxis<
...
-        const uniqueValues = new Set();
-        for (const v of d) {
-            const key = v instanceof Date ? v.getTime() : v;
-            if (!uniqueValues.has(key)) {
-                uniqueValues.add(key);
-                // Only add unique values
-                domain.push(v);
-            }
-        }
+    override normaliseDataDomain(domain: Array<string | object>) {
```
iMoses, line 49: That's a breaking change, isn't it?
Currently we do not support duplicate categories.
Not sure how important it is, but worth asking David, since it was used by integrated to implement certain behaviours, but I'm not sure how relevant it is with all of our latest changes.

iMoses: Talked offline - I understand this logic has moved elsewhere and did change.

iMoses (approved):
LGTM

PR #3298, opened 2024-12-20
AG-13753 Fix regression with interactions on single item legends

olegat:
https://ag-grid.atlassian.net/browse/AG-13753


packages/ag-charts-community/src/chart/legend/legendDOMProxy.ts
olegat, line 1: > [!IMPORTANT]
> This hunk is the root cause of the regression.

alantreadway (approved):
LGTM

PR #3299, opened 2024-12-20
AG-11705 Add backgroundColor and foregroundColor theme params

lsjroberts:

### DIFF
--- a/packages/ag-charts-community/src/module/optionsModule.ts
+++ b/packages/ag-charts-community/src/module/optionsModule.ts
@@ -239,6 +239,7 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
         } = this.getSeriesThemeConfig(chartType, activeTheme);
 
         const [annotationsOptions, annotationsThemes] = this.splitAnnotationsOptions(annotations);
+        this.annotationThemes = deepClone(annotationsThemes);
 
         let processedOptions = mergeDefaults(
             processedOverrides,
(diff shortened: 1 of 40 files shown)


packages/ag-charts-community/src/module/optionsModule.ts
```diff
@@ -251,13 +252,17 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
...
         processedOptions = deepClone(processedOptions, ChartOptions.OPTIONS_CLONE_OPTS);
 
-        this.processOptionsParams(processedOptions);
+        let publicParameters = activeTheme.getPublicParameters();
+        if (isPlainObject(processedOptions.theme) && processedOptions.theme.params) {
+            publicParameters = mergeDefaults(processedOptions.theme.params, publicParameters);
+        }
+
+        this.resolveOptionsParams(publicParameters, processedOptions);
+        this.resolveOptionsParams(publicParameters, this.annotationThemes);
```
lsjroberts, line 265: The primary change in this PR is here, moving the default theme params to the `getPublicParameters()` fn and applying them also the the annotation theme (which is split out of the processed options above).

jacobp100 (approved):
Looks good! The only thing I might add is it might be better to call them `background` and `foreground` - just in case we ever add support for gradients and patterns

lsjroberts:
> Looks good! The only thing I might add is it might be better to call them `background` and `foreground` - just in case we ever add support for gradients and patterns

Good point, I'll check in with the design team, names come from them.

jacobp100:
> > Looks good! The only thing I might add is it might be better to call them `background` and `foreground` - just in case we ever add support for gradients and patterns
> 
> Good point, I'll check in with the design team, names come from them.

Sounds good. Just to add some extra context, SwiftUI named everything `-Color`, then deprecated them all when they added support for patterns etc. https://developer.apple.com/documentation/swiftui/view/foregroundcolor(_:)

PR #3304, opened 2024-12-23
Datum index refactor

jacobp100:
https://ag-grid.atlassian.net/browse/AG-13772


packages/ag-charts-website/e2e/keyboard-nav.spec.ts-snapshots/sankey-link-highlight-chromium-linux.png
olegat, line 1: > [!CAUTION]
> This snapshot appears to change the ArrowLeft / ArrowRight order from declaration order (i.e. first link focused is the first datum in the charts `data` option) to render order (i.e. first link focus is the one in the top-left). Can you confirm this?

If we're changing the focus order to follow the rendering order then this new behaviour is drastically different to how we deal with keyboard navigation on all other series types. Some examples:

- In line series the ArrowDown always navigates to the next series, even if the line of the first series is drawn below the second series.
- In all cartesian series, the ArrowLeft always navigates to the next datum, even if the X axis is configured to be drawn in reverse (meaning that ArrowLeft navigates right and ArrowRight navigates left).
- The treemap series, the ArrowLeft / ArrowRight keys follow the declaration order of the children and ignores the order in which the nodes are drawn.

Therefore, I believe that the `ArrowDown` should still focus on the `Wind - Renewables` link because that's the first datum declared in the chart options.

I would strongly advise discussing this new keyboard navigation behaviour with David and/or Kiril before merging them.

packages/ag-charts-community/src/chart/cartesianChart.ts
olegat, line 1: Does `Series<unknown, any, any>` successfully compile? If so, then please replace all `Series<any, any, any>` occurrences with that instead.

packages/ag-charts-community/src/chart/series/seriesTypes.ts
```diff
@@ -54,10 +54,11 @@ export interface ISeries<TDatum, TProps, TLabel = TDatum> {
  * Processed series datum used in node selections,
  * contains information used to render pie sectors, bars, markers, etc.
  */
-export interface SeriesNodeDatum {
+export interface SeriesNodeDatum<I> {
```
olegat, line 57: nit: Consider using `SeriesNodeDatum<I = unknown>` to avoid needed to write `SeriesNodeDatum<unknown>` everywhere.

olegat (changes requested):
LGTM (2 nits), but subject to approval from David and/or Kiril regarding the behavioural change to the sankey keyboard navigation.

packages/ag-charts-community/src/chart/series/seriesTypes.ts
jacobp100: If it's ok, I'll revisit this after everything is migrated over. I suspect this will be a required property, but if I'm wrong, I can always update the code later

packages/ag-charts-community/src/chart/cartesianChart.ts
jacobp100: Updated

packages/ag-charts-website/e2e/keyboard-nav.spec.ts-snapshots/sankey-link-highlight-chromium-linux.png
jacobp100: Will check with David when we're all back

packages/ag-charts-community/src/chart/series/seriesTypes.ts
olegat: Okay. SGTM

packages/ag-charts-community/src/chart/cartesianChart.ts
olegat: Thanks.

packages/ag-charts-website/e2e/keyboard-nav.spec.ts-snapshots/sankey-link-highlight-chromium-linux.png
AG-DavidG: I agree with @olegat that focus order should follow the data order without regard for the rendering.

packages/ag-charts-community/src/chart/series/hierarchy/hierarchySeries.ts
```diff
@@ -50,74 +53,62 @@ export class HierarchyNode<TDatum = Record<string, any>>
...
-        public readonly rootIndex: number,
+        public readonly datumIndex: number[],
         public readonly datum: TDatum | undefined,
         public readonly sizeValue: number,
         public readonly colorValue: number | undefined,
         public readonly sumSize: number,
         public readonly depth: number | undefined,
-        public readonly parent: HierarchyNode<TDatum> | undefined,
-        public readonly children: HierarchyNode<TDatum>[]
+        public readonly parent: This | undefined,
```
alantreadway, line 62: Could you use the actual `this` here and elsewhere?

jacobp100: I was not able to get it to work

PR #3310, opened 2024-12-24
AG-13757 Line series tooltip formatter params

jacobp100:
https://ag-grid.atlassian.net/browse/AG-13757


olegat (approved):
LGTM.

We might need to factor in the strokeWidth in the calculations of the bounding boxes (such as the focus indicator for example), but we can wait for QA feedback first.

PR #3318, opened 2025-01-03
AG-13759 Ensure nice domain is generated even when the axis is not vi…

jacobp100:
…sible

https://ag-grid.atlassian.net/browse/AG-13759


iMoses (approved):
LGTM 👍

PR #3322, opened 2025-01-03
AG-13780 Fix regression with Navigator handle z-index

olegat:
https://ag-grid.atlassian.net/browse/AG-13780

This change ensures that the currently drag handle (min or max) is always at the top of the z-ordering.

### DIFF
--- a/packages/ag-charts-enterprise/src/features/navigator/navigatorDOMProxy.ts
+++ b/packages/ag-charts-enterprise/src/features/navigator/navigatorDOMProxy.ts
@@ -125,7 +125,7 @@ export class NavigatorDOMProxy {
         if (index === 1) return; // ignore pan-handle
 
         const frontSlider = this.sliders[index];
-        const otherSlider = this.sliders[[2, NaN, 0][index]];
+        const otherSlider = this.sliders[2 - index];
         this.toolbar.moveChild(otherSlider, frontSlider.domIndex! - 1);
     }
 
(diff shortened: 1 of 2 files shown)


packages/ag-charts-enterprise/src/features/navigator/navigatorDOMProxy.ts
```diff
@@ -120,14 +121,20 @@ export class NavigatorDOMProxy {
...
-    private onDragStart(
-        slider: _ModuleSupport.SliderWidget,
-        event: _ModuleSupport.DragWidgetEvent<'drag-start'>,
-        key: NavigatorButtonType
-    ) {
+    private moveToFront(index: number) {
+        if (index === 1) return; // ignore pan-handle
+
+        const frontSlider = this.sliders[index];
+        const otherSlider = this.sliders[[2, NaN, 0][index]];
```
jacobp100: This might be easier to follow than the double index lookup

```suggestion
    private moveToFront(index: 0 | 1 | 2) {
        if (index === 1) return; // ignore pan-handle

        const frontSlider = this.sliders[index];
        const otherSlider = this.sliders[2 - index];
```

jacobp100 (approved):

packages/ag-charts-enterprise/src/features/navigator/navigatorDOMProxy.ts
olegat: Thanks for the feedback. I've updated the PR.

PR #3323, opened 2025-01-06
AG-11705 Update theme css to use theme params

lsjroberts:

Applies the theme params as css variables on the main dom manager element.

Maintains current compatibility with grid themes.

### DIFF
--- a/packages/ag-charts-community/src/dom/theme.css
+++ b/packages/ag-charts-community/src/dom/theme.css
@@ -1,17 +1,25 @@
 .ag-charts-theme-default,
 .ag-charts-theme-default-dark {
+    /* Theme Parameters */
+    --ag-charts-background-color: #fff;
+    --ag-charts-foreground-color: #181d1f;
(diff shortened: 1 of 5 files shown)


packages/ag-charts-community/src/dom/theme.css
```diff
@@ -121,7 +137,7 @@
     --ag-charts-input-focus-outline: var(--ag-charts-focus-border);
     --ag-charts-input-focus-box-shadow: var(--ag-charts-focus-border-shadow);
     --ag-charts-input-font-family: var(--ag-charts-font-family);
-    --ag-charts-input-font-size: var(--ag-charts-font-size);
+    --ag-charts-input-font-size: 14px;
```
lsjroberts, line 140: This should probably be (in the future) a derived value of default font size plus/times some value.

alantreadway (approved):
LGTM, we should have @markdurrant review the CSS updates.

markdurrant (approved):
All the CSS LGTM 👍

PR #3327, opened 2025-01-06
Add bundle size validations + cleanup main entry points

iMoses:
https://ag-grid.atlassian.net/browse/AG-13804


packages/ag-charts-community/project.json
```diff
@@ -125,6 +125,14 @@
       "options": {
         "commands": ["node tools/package/sanity-check-package.js dist/packages/contents/ag-charts-community/package"]
       }
+    },
+    "validate-size": {
```
alantreadway, line 129: We should either tie this into the `package:verify` or `test` tasks as a dependency so that it is run routinely.

iMoses: I think it should be tied to the tests, question is do we already want to do it or should we wait until there's actually something to test?
There's also a question whether we should already add this to the `core` and `enterprise` packages.

alantreadway (approved):
LGTM

PR #3328, opened 2025-01-06
Introduce an `ag-charts-core` package

iMoses:
A new internal package to share utilities and core functionality between the community and enterprise packages, to improve our tree-shaking abilities.


packages/ag-charts-core/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "not ie_mob >= 0",
+    "not blackberry > 0",
+    "not op_mini all",
+    "not operamobile >= 0"
+  ],
+  "homepage": "https://ag-grid.com/charts/",
+  "dependencies": {},
+  "devDependencies": {},
+  "publishConfig": {
+    "access": "public"
```
alantreadway, line 55: Can we mark this as private, to avoid accidental publishing?

iMoses: My concern was that it may cause issues with the way we publish packages, but I can set it private and we can find out if it does 😅

alantreadway: Yes, I'd prefer that way around. We publish to the private Verdaccio routinely, so I would hope any issues in that respect should show up there; if `ag-charts-core` is a `devDependency` I don't foresee any issue with publishing of dependents either?

On another note, as long as we don't generate a `.tgz` when `nx pack --prod` is run, then nothing in the publishing pipeline should care AFAIK

alantreadway (approved):
LGTM

packages/ag-charts-core/package.json
alantreadway: As a heads-up, I followed up by removing the `package` and `package:verify` targets to ensure we don't generate publishable artifacts: https://github.com/ag-grid/ag-charts/commit/45b0457721568c58602c230239193d7f53556807

PR #3332, opened 2025-01-06
AG-13798 Fix 'wheel' zoom events in Safari

olegat:
https://ag-grid.atlassian.net/browse/AG-13798

This might be because the `passive` defaults in Safari is different to other browsers. Regardless, we're adding `'wheel'` event listeners to the .ag-charts-series-area element which has the CSS value `pointer-events: none`. This is sort of asking for trouble.

Making the .ag-charts-series-area have `pointer-event: auto` works on all browsers (Safari, Firefox, Chrome), although we need to set the .ag-charts-series-area's tabindex to -1 and focus on the .ag-charts-swapchain child when the series area is clicked on.

### DIFF
--- a/packages/ag-charts-community/src/dom/container.css
+++ b/packages/ag-charts-community/src/dom/container.css
@@ -37,14 +37,14 @@
 
 .ag-charts-series-area {
     outline: none;
-    pointer-events: none;
+    pointer-events: auto;
     position: absolute;
 }
 
 .ag-charts-swapchain {
     outline: none;
     opacity: 0;
-    pointer-events: auto;
+    pointer-events: none;
     position: absolute;
     width: 100%;
     height: 100%;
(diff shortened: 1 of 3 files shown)


packages/ag-charts-community/src/dom/container.css
```diff
@@ -37,14 +37,14 @@
 
 .ag-charts-series-area {
     outline: none;
-    pointer-events: none;
+    pointer-events: auto;
```
jacobp100, line 40: I don’t feel too strongly here, but currently we only do this if we’re overriding ‘none’ from a parent element

jacobp100 (approved):
Looks good!

packages/ag-charts-community/src/dom/container.css
olegat: AFAIK it's necessary because the immediate parent has `pointer-events: none`

Screenshot 2025-01-07 at 08 48 32

alantreadway (approved):
LGTM

PR #3333, opened 2025-01-06
AG-11705 Add more theme params

lsjroberts:

- Adds `fontWeight`, `axisColor`, `borderColor`, `gridLineColor` and `padding` theme params.
- Refactors theme param json walk to not use inline fns.
- Adds params docs to the themes page


packages/ag-charts-website/src/content/docs/themes/index.mdoc
```diff
@@ -114,18 +144,18 @@ var myTheme: AgChartTheme = {
             axes: {
                 category: {
                     line: {
-                        stroke: 'gray',
+                        stroke: '#881008',
```
alantreadway: Shouldn't these be removed now?

packages/ag-charts-website/src/content/docs/themes/_examples/advanced-theme/main.ts
```diff
@@ -19,18 +26,18 @@ var myTheme: AgChartTheme = {
             axes: {
                 category: {
                     line: {
-                        stroke: 'gray',
+                        stroke: '#881008',
```
alantreadway: Ditto?

packages/ag-charts-website/src/content/docs/themes/index.mdoc
lsjroberts: Good spot, thanks! Replaced with `axisColor` and then added `width` instead so we still have an axis override.

alantreadway (approved):
LGTM

PR #3338, opened 2025-01-07
Ag 13658 homepage fw animation

markdurrant:
- Add FW animation to Charts homepage
- Fix Sass nesting order issues 
- Sync website shared


packages/ag-charts-website/src/pages/index.astro
```diff
@@ -113,10 +114,18 @@ const mapExampleHeights = 360;
         <div class={styles.homepageHero}>
             <div class:list={[styles.heroInner, 'layout-max-width-small']}>
                 <div class={styles.heroHeadings}>
-                    <p>The Best</p>
-                    <h1>JavaScript Charts</h1>
-                    <p>in the World</p>
+                    <noscript>
+                        <h1>The Best JavaScript Charts in the World</h1>
```
taktran, line 118: Wouldn't this make it show twice when JS is disabled?

markdurrant: Now fixed with `.only-script` class.

PR #3339, opened 2025-01-07
AG-13799 Fix updates when series data is empty

jacobp100:
https://ag-grid.atlassian.net/browse/AG-13799

Previously we used to check `rawData.length` because calls to `dataModel.resolveProcessedDataDefById` would throw errors for empty data. I'm not sure why this is no longer the case - but it seems to just work now.


alantreadway (approved):
LGTM

PR #3351, opened 2025-01-08
AG-12393 - Remove axes buttons from non-financial chart type options

manapeirov:

### DIFF
--- a/packages/ag-charts-community/src/module/optionsModule.ts
+++ b/packages/ag-charts-community/src/module/optionsModule.ts
@@ -361,7 +361,19 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
             return [{}, annotationsThemes];
         }
 
-        return [{ annotations: { axesButtons, enabled, optionsToolbar, toolbar } }, annotationsThemes];
+        const { presetType } = this.optionMetadata;
+
+        if (presetType == 'price-volume') {
+            return [{ annotations: { axesButtons, enabled, optionsToolbar, toolbar } }, annotationsThemes];
+        }
+
+        if (axesButtons != null) {
+            Logger.warn(`unable to set ['axesButtons'] - property is only supported for financial charts`);
+
+            delete annotations.axesButtons;
+        }
+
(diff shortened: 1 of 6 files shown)


packages/ag-charts-community/src/module/optionsModule.ts
```diff
@@ -332,7 +332,19 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
             return [{}, annotationsThemes];
         }
 
-        return [{ annotations: { axesButtons, enabled, optionsToolbar, toolbar } }, annotationsThemes];
+        const { presetType } = this.optionMetadata;
+
+        if (presetType == 'price-volume') {
```
alantreadway: Can we make this less FC specific, and assume that for preset cases they are passing the right options?

Practically:
```suggestion
        if (presetType != null) {
```

packages/ag-charts-types/src/chart/annotationsOptions.ts
alantreadway, line 1: Should the removed `axesButtons` appear elsewhere for FC cases and backwards compatibility?

manapeirov: It's broken at the moment so removing it won't impact any users

packages/ag-charts-enterprise/src/features/annotations/annotationsTheme.ts
```diff
@@ -94,16 +89,6 @@ const toolbar: AgAnnotationsToolbar = {
...
-        {
-            icon: 'measurer-drawing',
-            tooltip: 'toolbarAnnotationsMeasurerAnnotations',
-            value: 'measurer-menu',
-        },
-        {
-            icon: 'delete',
-            tooltip: 'toolbarAnnotationsClearAll',
-            value: 'clear',
-        },
```
lsjroberts: Do we not want this button in non-financial cases?

packages/ag-charts-community/src/module/optionsModule.ts
alantreadway: Discussed and updated - we don't need to be this specific here.

alantreadway (approved):

PR #3352, opened 2025-01-08
AG-12393 - Disable tooltip when in annotation active state

manapeirov:

### DIFF
--- a/packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
+++ b/packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
@@ -599,10 +599,7 @@ export class SeriesAreaManager extends BaseManager {
     }
 
     private handleHoverTooltip(event: TooltipWidgetEvent, redisplay: boolean) {
-        if (!this.isState(InteractionState.Default)) {
-            this.clearTooltip();
-            return;
-        }
+        if (!this.isState(InteractionState.Clickable)) return;
         const { type, currentX, currentY } = event;
         const canvasX = currentX + (this.hoverRect?.x ?? 0);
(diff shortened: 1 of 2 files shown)


jacobp100 (approved):

packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
```diff
@@ -599,8 +599,10 @@ export class SeriesAreaManager extends BaseManager {
     }
 
     private handleHoverTooltip(event: TooltipWidgetEvent, redisplay: boolean) {
-        if (!this.isState(InteractionState.Clickable)) return;
-
+        if (!this.isState(InteractionState.Default)) {
```
alantreadway: Does the `Clickable` case still need to be work?

packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
```diff
@@ -599,8 +599,10 @@ export class SeriesAreaManager extends BaseManager {
...
+            this.clearTooltip();
+            return;
+        }
```
olegat: The change to the `if` condition is almost certainly harmless. The added `this.clearTooltip()` may have some unintended consequences however.

The `annotations.ts` uses `TooltipManager.suppressTooltip` and `TooltipManager.unsuppressTooltip` in some areas, but it doesn't call these after `InteractionManager.pushState`. So perhaps there's a bug with missing `suppressTooltip` call in `annotations.ts`.

alantreadway (approved):
LGTM, reviewed offline by @olegat

PR #3353, opened 2025-01-08
AG-5805 Reduce y domain if x domain is clipped

jacobp100:
https://ag-grid.atlassian.net/browse/AG-5805

### DIFF
--- a/packages/ag-charts-community/src/chart/series/cartesian/abstractBarSeries.ts
+++ b/packages/ag-charts-community/src/chart/series/cartesian/abstractBarSeries.ts
@@ -14,7 +14,6 @@ import { ChartAxisDirection } from '../../chartAxisDirection';
 import { fixNumericExtent } from '../../data/dataModel';
 import type { SeriesNodePickMatch } from '../series';
 import type { SeriesNodeDatum } from '../seriesTypes';
-import { visibleRangeIndices } from '../util';
 import type { CartesianSeriesNodeDataContext, CartesianSeriesNodeDatum } from './cartesianSeries';
 import { CartesianSeries, CartesianSeriesProperties } from './cartesianSeries';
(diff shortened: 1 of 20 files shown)


packages/ag-charts-community/src/chart/series/cartesian/abstractBarSeries.ts
```diff
@@ -88,18 +87,11 @@ export abstract class AbstractBarSeries<
...
         const xScale = this.axes[ChartAxisDirection.X]!.scale;
-        const xValues = dataModel.resolveKeysById(this, `xValue`, processedData);
         const barWidth = xScale.bandwidth ?? 0;
-
-        return visibleRangeIndices(xValues.length, visibleRange, (index) => {
-            const x = xScale.convert(xValues[index]);
-            return [x, x + barWidth];
-        });
+        const x = xScale.convert(xValue);
+        return [x, x + barWidth];
```
alantreadway, line 94: Q: Shouldn't this be `[x - barWidth/2, x + barWdith/2]`?

alantreadway (commented):
One query - we should verify performance impact via the benchmarks, I can assist with that.

packages/ag-charts-community/src/chart/series/cartesian/abstractBarSeries.ts
jacobp100: I believe not - but I do have some off-by-one errors on these that I want to fix in an immediate follow up. Because it also affects the auto-scaling, I'd rather do it in its own piece

alantreadway: SGTM

alantreadway (approved):

PR #3359, opened 2025-01-09
AG-12204 - Next step in ESLlint v9 upgrade

alantreadway:

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -52,7 +52,6 @@ export const sonarjsConfig = [
 
             // For review - new for eslint 9.
             'sonarjs/no-redeclare': 1,
-            'sonarjs/updated-loop-counter': 1,
             'sonarjs/function-return-type': 1,
 
             // Duplicates @typescript-eslint
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -52,7 +52,6 @@ export const sonarjsConfig = [
 
             // For review - new for eslint 9.
             'sonarjs/no-redeclare': 1,
-            'sonarjs/public-static-readonly': 1,
             'sonarjs/updated-loop-counter': 1,
             'sonarjs/function-return-type': 1,
 
--- a/packages/ag-charts-community/src/chart/interaction/tooltipManager.ts
+++ b/packages/ag-charts-community/src/chart/interaction/tooltipManager.ts
@@ -5,7 +5,7 @@ import { type ErrorBoundSeriesNodeDatum, type SeriesNodeDatum, getDatumRefPoint 
 import type { Tooltip, TooltipContent, TooltipMeta, TooltipPointerEvent } from '../tooltip/tooltip';
 
 interface TooltipState {
-    content?: TooltipContent | string;
+    content?: TooltipContent;
     meta?: TooltipMeta;
 }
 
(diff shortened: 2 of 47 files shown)


eslint.config.mjs
```diff
@@ -52,8 +52,6 @@ export const sonarjsConfig = [
 
             // For review - new for eslint 9.
             'sonarjs/no-redeclare': 1,
-            'sonarjs/public-static-readonly': 1,
```
alantreadway, line 55: Note: Removing these entries activates the rules at error level.

packages/ag-charts-community/src/chart/interaction/tooltipManager.ts
```diff
@@ -5,7 +5,7 @@ import { type ErrorBoundSeriesNodeDatum, type SeriesNodeDatum, getDatumRefPoint
 import type { Tooltip, TooltipContent, TooltipMeta, TooltipPointerEvent } from '../tooltip/tooltip';
 
 interface TooltipState {
-    content?: TooltipContent | string;
+    content?: TooltipContent;
```
jacobp100, line 8: Is string definitely not used? I thought it could be used for raw html

packages/ag-charts-community/src/chart/tooltip/tooltip.ts
```diff
@@ -64,13 +64,13 @@ export interface TooltipContent {
     title?: string;
     symbol?: LegendSymbolOptions;
     data?: TooltipContentDataRow[];
+    rawHtmlString?: string;
 }
```
jacobp100: This is one or the other. Might be better to represent like this:-

```suggestion
    title?: string;
    symbol?: LegendSymbolOptions;
    data?: TooltipContentDataRow[];
} | {
    rawHtmlString: string;
}
```

jacobp100 (approved):

packages/ag-charts-community/src/util/textWrapper.ts
```diff
@@ -68,7 +68,10 @@ export class TextWrapper {
                 continue;
             }
 
-            for (let i = 0, estimatedWidth = 0, lastSpaceIndex = 0; i < line.length; i++) {
+            let i = 0;
+            let estimatedWidth = 0;
+            let lastSpaceIndex = 0;
+            while (i < line.length) {
```
iMoses, line 74: Wondering, what was the warning here about?

jacobp100: Mutating the iteratee in a for-loop

packages/ag-charts-community/src/chart/tooltip/tooltip.ts
alantreadway: I've tried applying this - but the simplest thing to do is add a `type` discriminator field, as `'rawHtmlString' in content` isn't enough of a hint to the TS compiler that on the negative control flow that the inferred typing of `content` is a single union member type.

https://github.com/ag-grid/ag-charts/pull/3359/commits/9e3aeac5c35bf6dcd1983b47ba308eaefbbc281f is the smallest additional change encoding use of a type union for `TooltipContent`.

alantreadway (commented):
PTAL

PR #3360, opened 2025-01-09
AG-13488 QA feedback & regressions

olegat:
https://ag-grid.atlassian.net/browse/AG-13488


packages/ag-charts-website/e2e/interactive-tooltip.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+
+        await expect(page).toHaveScreenshot('interactive-tooltip-hidden.png');
+
+        await page.mouse.move(400, 150);
+        await expect(page).toHaveScreenshot('interactive-tooltip-visible.png');
+        await expect(page.locator(SELECTORS.wrapper)).toHaveAttribute('data-scene-renders');
+        const expectedRenders: string = await page.locator(SELECTORS.wrapper).getAttribute('data-scene-renders');
+
+        const bbox = await page.getByText('Click here').boundingBox();
+        await page.mouse.move(bbox.x, bbox.y);
```
alantreadway, line 24: Nit: It looks like by default this will emit a single `mousemove` event at the point specific - but we can add a third parameter of `{ steps: 5 }` to better simulate a user moving their mouse in incremental steps from the previous position to this new position. Could we use that option to more accurately simulate this test-case?
https://playwright.dev/docs/api/class-mouse#mouse-move

olegat: A single `'mousemove'` event is what's desired.

We want to assert that hovering over the interactive tooltip does not cause it to disappear, flicker or move. If we fire 5 steps, then the interactive tooltip could potentially genuinely disappear or move if one or more of the `'mousemove'` steps does not hover over the tooltip.

alantreadway: Discussed offline, we agreed that we need to exercise both scenarios (immediate jump and gradual move).

olegat: `{ steps: 5 }` was causing the tooltip to move, so I've used `{ step: 4 }` instead.

alantreadway (approved):
LGTM, thanks!

PR #3361, opened 2025-01-10
Move global utils to `ag-charts-core` (1)

iMoses:
- moved global types & type guards
- moved new validation implementation + string utils
- added test support for the core package
- added myself to the core package codeowners


alantreadway (approved):
LGTM

PR #3363, opened 2025-01-10
AG-12936 Change zoom minVisibleItems to actually count visible items

lsjroberts:

- Adds `zoom.minVisibleItems`, defaulting to `2`.
- Deprecates `zoom.minVisibleItemsX` and `zoom.minVisibleItemsY`.
- Changes zoom selector behaviour – there is no minimum size for the selection box.
- Changes context menu "zoom in" button behaviour – changes zoom by 1 step of `zoom.scrollingStep` to match the zoom toolbar buttons.
  - It enables/disables the button depending on the location the user right-clicked, and if zooming a step there would be valid or not.


packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts
```diff
@@ -841,99 +831,77 @@ export abstract class CartesianSeries<
...
+        for (let index = 0; index < xValues.length; index++) {
+            const x = xAxis.scale.convert(xValues[index]);
+            const y = yAxis.scale.convert(yValues[index]);
+            if (x >= xMin && x <= xMax && y >= yMin && y <= yMax) {
+                visibleItems += 1;
+                if (visibleItems >= minVisibleItems) {
+                    return visibleItems;
+                }
             }
         }
```
lsjroberts: @jacobp100 Is there an existing method that works like this? From what I can tell, the related fns are looking for indicies, whereas I need an actual count of items.

lsjroberts: Jacob suggested modifying `domainForVisibleRange()` and adding a `yCoordinateRange()` method, similar to the existing `xCoordinateRange()`.

I did the latter, but I decided to keep this counting a separate method as it has a different optimisation technique (see `countExpandingSearch()`).

packages/ag-charts-website/e2e/context-menu.spec.ts-snapshots/pan-to-here-chromium-linux.png
alantreadway, line 1: These seem like quite big changes, is the updated image correct?

packages/ag-charts-website/e2e/context-menu.spec.ts-snapshots/zoom-to-here-chromium-linux.png
alantreadway, line 1: Ditto

packages/ag-charts-website/e2e/context-menu.spec.ts-snapshots/pan-to-here-chromium-linux.png
lsjroberts: These relate to the behaviour change on the context menu "zoom to here" button. It now zooms in 1 step rather than to the maximum zoom (since we no longer have a known maximum zoom).

It probably should zoom in more than 1 step as it is a little tedious, but will chat with design/product on that.

PR #3365, opened 2025-01-13
AG-7083 Fix visible range for unsorted data series

jacobp100:


alantreadway (approved):
LGTM

PR #3366, opened 2025-01-13
AG-7670 Fix zoom event example

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-7670

The website example compiler can't handle `chart.update(options)` within a callback on the chart. Just changed it to do a `console.log()` per other examples.


alantreadway (approved):
LGTM

PR #3368, opened 2025-01-13
Remove unused code (cleanup)

iMoses:


alantreadway (approved):
LGTM, I think @jacobp100 is best placed to review these changes too

packages/ag-charts-community/src/chart/axis/axis.ts
```diff
@@ -309,10 +309,6 @@ export abstract class Axis<
         return value >= min - tolerance && value <= max + tolerance;
     }
 
-    protected defaultDatumFormatter(datum: any, fractionDigits: number): string {
```
jacobp100, line 312: ThI think this should be used an a mistake was made in axis.ts

```
        this.datumFormatter =
            scale.datumFormatter({ domain: tickDomain, specifier, ticks, fractionDigits }) ??
            ((x: any) => this.defaultLabelFormatter(x, fractionDigits));
```

packages/ag-charts-community/src/chart/data/processors.ts
```diff
@@ -350,7 +350,7 @@ export function animationValidation(valueKeyIds?: string[]): ProcessorOutputProp
                 for (let i = 0; validation !== 0 && i < valuesDef.length; i++) {
                     const value = valuesDef[i];
 
-                    if (!valueKeyIds?.includes(value.id as string) === true) continue;
+                    if (!valueKeyIds?.includes(value.id as string)) continue;
```
jacobp100, line 353: Minor - but `=== true` is better for engines because it's a bitwise operation vs. dynamic dispatch

jacobp100 (approved):
Other than the datum formatter - I'm pretty sure that _should_ be used

packages/ag-charts-community/src/chart/data/processors.ts
iMoses: I think that in this very specific case it doesn't matter since we are converting the value to a boolean (`!`) meaning no additional casting will happen anyhow

packages/ag-charts-community/src/chart/axis/axis.ts
iMoses: Reverted the removal of `defaultDatumFormatter` and applied it in the correct place.
Also it seems that `defaultDatumFormatter` and `defaultLabelFormatter` were reversed, so fixed that as well.

packages/ag-charts-community/src/chart/data/processors.ts
jacobp100: Ah yeah, it should be `valueKeyIds?.includes(value.id as string) !== true` 😅 But yes, semantically equivalent

PR #3369, opened 2025-01-13
AG-7083 Fix restoring initial state for autozoom

jacobp100:

### DIFF
--- a/packages/ag-charts-types/src/chart/zoomOptions.ts
+++ b/packages/ag-charts-types/src/chart/zoomOptions.ts
@@ -6,7 +6,7 @@ export type AgZoomAnchorPoint = 'pointer' | 'start' | 'middle' | 'end';
 export type AgZoomAxes = 'x' | 'y' | 'xy';
 export type AgZoomPanKey = 'alt' | 'ctrl' | 'meta' | 'shift';
 export type AgZoomDeceleration = 'off' | 'short' | 'long' | Ratio;
-export type AgAutoScaledAxes = 'y' | '';
+export type AgAutoScaledAxes = Array<'y'>;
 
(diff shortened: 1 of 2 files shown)


packages/ag-charts-types/src/chart/zoomOptions.ts
```diff
@@ -6,6 +6,7 @@ export type AgZoomAnchorPoint = 'pointer' | 'start' | 'middle' | 'end';
 export type AgZoomAxes = 'x' | 'y' | 'xy';
 export type AgZoomPanKey = 'alt' | 'ctrl' | 'meta' | 'shift';
 export type AgZoomDeceleration = 'off' | 'short' | 'long' | Ratio;
+export type AgAutoScaledAxes = 'y' | '';
```
lsjroberts: This type is a bit odd with the empty string, presumably here because it has to allow the user to override the a default `'y'` with a null? But I guess the alternative is `'y' | false` which is not really better.

lsjroberts (approved):
LGTM

packages/ag-charts-types/src/chart/zoomOptions.ts
alantreadway: I'd prefer we dropped `''` - can't we just using `undefined`/`null` for unset cases?

jacobp100: I did do null but I thought semantic differences between null and undefined was worse. I copied the zoom axes which can be x, y, or xy

jacobp100: Updated to an array

alantreadway (approved):
LGTM

PR #3371, opened 2025-01-13
Align FW package dependency type with Grid.

alantreadway:


alantreadway:
Discussed offline with @seanlandsman and @jacobp100 - this aligns with the new Grid approach for 33.0.0.

PR #3389, opened 2025-01-15
AG-13242 - Fix batch nx executors.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-13242


plugins/plugin-utils/src/executors-utils.ts
```diff
@@ -138,10 +137,15 @@ export function batchWorkerExecutor<ExecutorOptions>(workerModule: string) {
 
         // Run yield loop after dispatch to avoid serializing execution.
         for (let taskIndex = 0; taskIndex < tasks.length; taskIndex++) {
-            const taskName = tasks[taskIndex++];
+            const taskName = tasks[taskIndex];
             yield results.get(taskName);
         }
 
+        await Promise.allSettled([...results.values()]);
```
jacobp100: Minor but I think you can just do this

```suggestion
        await Promise.allSettled(results.values());
```

jacobp100 (approved):

iMoses (approved):
LGTM

packages/all/project.json
```diff
@@ -64,8 +64,8 @@
       "options": {
         "parallel": false,
         "commands": [
-          "if (test \"${BUILD_BATCH}\" != \"0\") ; then nx run-many --output-style compact -t generate-examples,generate-thumbnails -p ag-charts-website --batch ; fi",
-          "if (test \"${BUILD_BATCH}\" == \"0\") ; then nx run-many --output-style compact -t generate-examples,generate-thumbnails -p ag-charts-website ; fi"
+          "if (test \"${BUILD_BATCH}\" != \"0\") ; then npx nx run-many --output-style compact -t generate-examples,generate-thumbnails -p ag-charts-website --batch ; fi",
```
taktran, line 67: This is kind of weird - why do you need to do this?

taktran (approved):
Looks good 👍🏼 Just a minor q

packages/all/project.json
alantreadway: I was getting a different `nx` version than the repo's version, so this was to rule that out.

```
$ nx --version
Nx Version:
- Local: v18.3.4
- Global: v20.3.1
$ npx nx --version
Nx Version:
- Local: v18.3.4
- Global: Not found
```

PR #3393, opened 2025-01-15
AG-13216 Fix TC1: Join segmented gauge nodes into 1 node

olegat:
https://ag-grid.atlassian.net/browse/AG-13216


olegat:
> [!NOTE]
> The snapshot updates for the needle e2e tests are acceptable:

combined-diffs

alantreadway (approved):
LGTM, @jacobp100 could you review too?

jacobp100 (approved):
The animations seem to have broken compared to prod (https://www.ag-grid.com/charts/javascript/linear-gauge/). The bar should animate from 0 to the value on mount, and animate value changes too. I'm not 100% sure if this is caused by this PR as it's also broken on staging

Not 100% following the datumUnion bit, but everything looks reasonable

olegat:
> Not 100% following the datumUnion bit, but everything looks reasonable

`datumUnion` joins the coloured segments (which are individual scene nodes) into a single scene node. It ignores empty segments on the right that aren't partially/fully filled by the gauge value.

PR #3399, opened 2025-01-16
AG-13880 - Fix infinite loop when computing focusDatumIndex.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-13880


olegat (approved):
LGTM

PR #3402, opened 2025-01-17
AG-13781 Add warning when zoom range state does not match domain type

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13781


lsjroberts:
It looks like this will cause issues with valid values that are of different types, e.g. a timestamp vs a date object.

PR #3404, opened 2025-01-17
AG-9958 Disable rect crisp-ness during animations

jacobp100:
https://ag-grid.atlassian.net/browse/AG-9958

A fair bit was needed to achieve this:-

- Nodes now type their datums, but defaults to `any` as before
- `crisp` is now required on node datums that use the bar animations

This was required as `boolean` is not animatable, and there was no way to pass this down to `applyFn` other than by reading from the node's datum property - which was typed as `any`


alantreadway (approved):
LGTM

PR #3405, opened 2025-01-17
AG-13318 Fix TC9: Default enabled to true with `tooltip: {}`

olegat:
https://ag-grid.atlassian.net/browse/AG-13318


olegat:
> [!NOTE]
> Linear/Radial gauge snapshots updates are valid. (Tooltips removed from the examples)

packages/ag-charts-website/e2e/keyboard-nav.spec.ts-snapshots/linear-gauge-bar-highlight-chromium-linux.png
alantreadway, line 1: Should we enable tooltips explicitly for these test cases, as it feels like we're losing some coverage otherwise?

olegat: We could, but there's minimal gain in doing so because the tooltips that we've got on our examples are very rudimentary. It would be nice to update the docs to include prettier tooltips.

alantreadway (approved):
LGTM

PR #3416, opened 2025-01-21
Register charts as modules

iMoses:
Known regression - if an enterprise chart type is used on the community package error message would be for unknown type instead of enterprise should've been imported.


packages/ag-charts-community/src/api/agCharts.ts
```diff
@@ -296,7 +258,7 @@ class AgChartsInternal {
             false
         );
 
-        debugOptions('>>> AgCharts.updateUserDelta() user delta', deltaOptions);
+        debug(() => ['>>> AgCharts.updateUserDelta() user delta', deltaOptions]);
```
alantreadway: Nit: Re-add `deepClone()` here and other cases where we were calling `debugOptions()`?

alantreadway (approved):
LGTM

PR #3419, opened 2025-01-21
AG-13892 Add json operations to the theme resolution

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13892

- Adds operations with `$` prefix
- Implements the bar series vertical/horizontal label colour automatic switching.


packages/ag-charts-community/src/chart/themes/chartTheme.ts
```diff
@@ -135,31 +135,31 @@ export class ChartTheme {
                 enabled: false,
                 text: 'Axis Title',
                 spacing: 25,
-                fontWeight: { ref: 'fontWeight' },
-                fontSize: FONT_SIZE.MEDIUM,
-                fontFamily: { ref: 'fontFamily' },
-                color: { ref: 'foregroundColor' },
+                fontWeight: { $ref: 'fontWeight' },
+                fontSize: { $round: [{ $mul: [{ $ref: 'fontSize' }, FONT_SIZE.MEDIUM / FONT_SIZE.SMALL] }] },
```
lsjroberts, line 139: These font sizes are unchanged for the default size, but scale at the same ratio. There's a question of whether we should just round in `$mul` itself to be more concise, something to iterate on.

packages/ag-charts-community/src/util/json.ts
```diff
@@ -283,3 +293,157 @@ function classify(value: any): Classification | null {
...
+ * Resolve logical operations within a json object.
+ *
+ * @param source JSON object to walk and onto which to apply the resolved values.
+ * @param params An object of parameters to use with the `ref` operation.
+ *
+ * @returns An object of modified paths.
+ */
+export function jsonResolveOperations<T, P>(source: T, params: P, skip?: Set<string>) {
+    return jsonResolveInner(source, params, source, skip);
+}
```
lsjroberts, line 307: Main change in the PR is this fn.

packages/ag-charts-enterprise/src/series/radar/radarThemes.ts
```diff
@@ -30,18 +30,17 @@ const BASE_RADAR_SERIES_THEME: _ModuleSupport.SeriesModule<'radar-line' | 'radar
...
+                fillOpacity: 0.8,
+                strokeWidth: 2,
+                marker: { enabled: false },
+            },
         },
-    },
-    BASE_RADAR_SERIES_THEME
-);
+        BASE_RADAR_SERIES_THEME
+    );
```
lsjroberts, line 46: Typescript hits an issue with depth of type evaluation and required explicit typing here. I'm not sure if it is due to the recursive nature of the `ThemeParamsOperation` & `ThemeParamsLeaf` types, of it they were just one step too far.

Something to keep an eye on when we introduce this further.

alantreadway (approved):
LGTM

iMoses (approved):
Looks good. Just adding an idea regarding the multiplying operator; Perhaps we can keep it cleaner by introducing support for em units, based on the theme's global fontSize, which would result in a simpler interface and would push the responsibility of converting and rounding down the chain and away from the theme.

alantreadway:
> Looks good. Just adding an idea regarding the multiplying operator; Perhaps we can keep it cleaner by introducing support for em units, based on the theme's global fontSize, which would result in a simpler interface and would push the responsibility of converting and rounding down the chain and away from the theme.

@AG-DavidG ^

SGTM, something for future consideration, something we could add into the mix prior to making the API public perhaps.

PR #3424, opened 2025-01-22
Enable nx batching by default.

alantreadway:
- **Guard against destroyed charts throwing errors from Chart.update().**
- **Enable Nx batching by default.**


taktran (approved):
nice 👍🏼

PR #3426, opened 2025-01-23
AG-13917 Hide axes divs when `zoom.enabledAxisDragging=false`

olegat:
https://ag-grid.atlassian.net/browse/AG-13917


alantreadway (approved):
LGTM

PR #3431, opened 2025-01-27
AG-13914 Fix slow zoom when discarding large numbers of ticks

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13914

This was caused by the `range()` fn iterating millions of times, just for the calculated ticks to be discarded down to the few that are visible.

https://github.com/ag-grid/ag-charts/blob/2196e75307f794fabb45e1dbbd92c79d2a2ee2a2/packages/ag-charts-community/src/util/ticks.ts#L175-L192


packages/ag-charts-enterprise/src/features/zoom/__image_snapshots__/zoom-test-ts-zoom-axis-dragging-should-zoom-the-y-axis-from-the-middle-1-snap.png
lsjroberts, line 1: This is an incidental change. The behaviour is essentially the same as before.

alantreadway (approved):
LGTM, maybe @iMoses can sanity check too?

iMoses (approved):
Push a minor change, now filtering visible-range ticks is done as part of the `createTicks` method to prevent unnecessary creation of invisible ticks.

alantreadway (approved):
LGTM

PR #3433, opened 2025-01-27
AG-13860 - Area series gradient fill

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-13860


packages/ag-charts-community/src/scene/shape/shape.ts
```diff
@@ -67,13 +81,14 @@ export abstract class Shape<D = any> extends Node<D> {
     strokeOpacity: number = 1;
 
     @SceneChangeDetection({ changeCb: (s: Shape) => s.onFillChange() })
-    fill?: string | Gradient = Shape.defaultStyles.fill;
+    fill?: string | AgGradientFill | Gradient = Shape.defaultStyles.fill;
```
alantreadway: I'm seeing the type `string | AgGradientFill | Gradient` repeated a lot, could we add a type-alias for this union to centrally define this type?

packages/ag-charts-community/src/scene/shape/shape.ts
```diff
@@ -180,12 +212,26 @@ export abstract class Shape<D = any> extends Node<D> {
...
 
+    protected getGradientFillBBox() {
+        const { fill } = this;
+        if (!isGradientFill(fill)) {
+            return this.getBBox();
+        }
+
+        if (fill.bounds === 'axes' || fill.bounds === 'series') {
+            return this.gradientFillOptions.bbox ?? this.getBBox();
+        }
```
alantreadway: The bounds being `axes` or `series` doesn't seem like a concern of the `Shape` type - we should extract this out to be a concern of the series implementation(s).

alantreadway (approved):
LGTM

PR #3435, opened 2025-01-28
AG-11705 Theme params defaults fixes

lsjroberts:

- Updates various font options to use the theme params
- Separates the chrome & non-chrome css variables so the params do not unintentionally affect the UI elements. There's a few elements that are html purely as implementation detail that use the non-chrome variables, e.g. the no data overlay.


alantreadway (approved):
LGTM

PR #3440, opened 2025-01-29
Ag 12393/non financial charts annotations

manapeirov:


packages/ag-charts-enterprise/src/features/annotations/states/dragState.ts
```diff
@@ -67,7 +67,6 @@ export class DragStateMachine<
                     target: StateMachine.parent,
                     action: () => {
                         this.node?.stopDragging();
-                        ctx.stopInteracting();
```
lsjroberts, line 70: Just double checking this change, does it not need to exit the annotations interation state at this point?

manapeirov: When we start drag the annotation is selected and the floating toolbar options shows, when we stop we don't deselect but for some reason we pop the annotation interaction state even though it's still selected. That's why I removed this line because the annotation is still active

alantreadway (approved):
LGTM

PR #3441, opened 2025-01-29
AG-13967 Touch docs placeholder

olegat:
https://ag-grid.atlassian.net/browse/AG-13967

Depends on
- https://github.com/ag-grid/ag-charts/pull/3439

PR #3442, opened 2025-01-30
AG-13363 Long-Tap iOS Safari

olegat:

Depends on:
- https://github.com/ag-grid/ag-charts/pull/3439
- https://github.com/ag-grid/ag-charts/pull/3441

### DIFF
--- a/packages/ag-charts-community/src/widget/widgetListenerInternal.ts
+++ b/packages/ag-charts-community/src/widget/widgetListenerInternal.ts
@@ -144,7 +144,6 @@ function startOneFingerTouch(
     target: HTMLElement
 ) {
     if (that.globalTouchDragCallbacks != null || gIsInLongTap) return;
-    console.log(`touchstart`, `(callback)`, { initialTouch });
 
     let longTapInterrupted = false;
     setTimeout(() => {
@@ -159,6 +158,7 @@ function startOneFingerTouch(
             const longTapMove = (e: Event) => {
                 e.preventDefault();
             };
+            // Unblock new 'drag-start' events:
             const longTapEnd = (e: Event) => {
                 gIsInLongTap = false;
                 e.preventDefault();
@@ -180,9 +180,6 @@ function startOneFingerTouch(
                 clientY,
             });
             target.dispatchEvent(contextMenuEvent);
-            console.log(`longtap`, `(callback)`, { initialTouch });
-        } else {
-            console.log(`longtap`, `(ignored)`, { initialTouch });
         }
     }, LONG_TAP_DURATION);
 
@@ -197,10 +194,7 @@ function startOneFingerTouch(
             longTapInterrupted =
                 longTapInterrupted || deltaClientSquared(initialTouch, touch) < LONG_TAP_INTERRUPT_MIN_TOUCHMOVE;
             if (dragTouchEnabled && touch != null) {
-                console.log(`touchmove`, `(callback)`, { touch });
                 that.globalTouchDragCallbacks?.touchmove(moveEvent, touch);
-            } else {
-                console.log(`touchmove`, `(ignored)`, { touch });
             }
         }
     };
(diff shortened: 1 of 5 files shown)


packages/ag-charts-community/src/widget/widgetListenerInternal.ts
```diff
@@ -142,16 +191,21 @@ function startTouchDrag(
...
-            that.globalTouchDragCallbacks?.touchmove(moveEvent, touch);
+            longTapInterrupted =
+                longTapInterrupted || deltaClientSquared(initialTouch, touch) < LONG_TAP_INTERRUPT_MIN_TOUCHMOVE;
+            if (dragTouchEnabled && touch != null) {
+                that.globalTouchDragCallbacks?.touchmove(moveEvent, touch);
+            }
         }
     };
     const touchend = (endEvent: TouchEvent) => {
+        longTapInterrupted = true;
```
alantreadway, line 202: In the cases where `longTapInterrupted` is set to `true`, could you cancel the `setTimeout()` via `clearTimeout()` so that we don't have dangling JS variable contexts hanging around?

It would also be good if we could clear these up on `chart.destroy()` too, otherwise the callback could happen after chart destruction, with inevitable errors - also references in the callback context can prevent garbage collection?

packages/ag-charts-community/src/widget/widgetListenerInternal.ts
```diff
@@ -126,13 +132,56 @@ function makeTouchDrag<K extends DragEvents>(
     };
 }
 
-function startTouchDrag(
+const LONG_TAP_DURATION = 500; /* milliseconds */
```
alantreadway: Nit: Add `_MS` suffix?

packages/ag-charts-community/src/widget/widgetListenerInternal.ts
```diff
@@ -126,13 +132,56 @@ function makeTouchDrag<K extends DragEvents>(
...
+const LONG_TAP_INTERRUPT_MIN_TOUCHMOVE = 100; /* px²*/
```
alantreadway: Nit: Add `_PX` suffix?

packages/ag-charts-community/src/widget/widgetListenerInternal.ts
olegat: > In the cases where `longTapInterrupted` is set to `true`, could you cancel the `setTimeout()` via `clearTimeout()` so that we don't have dangling JS variable contexts hanging around?

Can you clarify how we would end up with a dangling reference? I don't mind using `clearTimeout()`, but I cannot see how that improves memory safety. Without a `clearTimeout()`, the `setTimeout()` will eventually call then remove the function, which would in turn make its variable context eligible for garbage collection.


> It would also be good if we could clear these up on `chart.destroy()` too, otherwise the callback could happen after chart destruction, with inevitable errors

The callbacks will not cause errors because we always remove the listeners on `chart.destroy()` (see `WidgetListenerInternal.destroy()`). So these callbacks will, at worst, loop through an undefined or empty array of listeners.

Calling `chart.destroy()` whilst a finger is being dragged on that chart can indeed delay garbage collection until after the finger is lifted. Removing the listeners on `chart.destroy()` isn't necessary, but it would stop this delay. However, the challenge is that these are global callbacks shared amongst all chart instances, we must be cautious to ensure that destroying one chart doesn't interrupt another chart that's currently being dragged.

We could perhaps refactor the `startMouseDrag` and `startOneFingerTouch` to be temporary member objects of `WidgetListenerInternal` to manage the memory better. But the way I see it there's very little gain in doing that: there's only one mouse & touch device in the `window` and therefore there's only ever one drag motion in progress. So we can safely manage the memory regardless of whether or not the chart gets destroyed.


> also references in the callback context can prevent garbage collection?

Are you referring to `that.globalTouchDragCallbacks`? If yes, that variable is used correctly AFAIK. It sets to `undefined` either in the `touchend` handler (called on 'touchstart', 'touchend' or 'touchcancel') or in the `WidgetListenerInternal.destroy()` method (called by `chart.destroy()`).

olegat: Discussed on slack: I've tested the spark-line use case and didn't notice anything alarming.

olegat: See follow-up https://github.com/ag-grid/ag-charts/pull/3446 (currently untested)

alantreadway (approved):
LGTM

PR #3446, opened 2025-01-31
AG-13363 Move drag logic into objects with a destroy() method

olegat:

Depends on
- https://github.com/ag-grid/ag-charts/pull/3442
- https://github.com/ag-grid/ag-charts/pull/3448


alantreadway (approved):
LGTM

PR #3451, opened 2025-01-31
AG-11705 Update css variables hierarchy with theme params

lsjroberts:

- Adds `accentColor`, `spacing` and chrome theme params.
- Updates css theme variables to inherit from the theme params and removes unnecessary indirection/inheritance.
- Pushes some variables into their local context where they did not need to be global.
- Removes references to grid variables that are no longer used.
- Fixes inconsistent colours across context menu, dialogs and other popovers.


alantreadway (approved):
LGTM

PR #3458, opened 2025-02-03
AG-13217 Make gauge-bars unclickable when `bar.enabled=false`

olegat:
https://ag-grid.atlassian.net/browse/AG-13217


alantreadway (approved):
LGTM

PR #3465, opened 2025-02-03
AG-11720 - Fix series stacking cases

alantreadway:
https://ag-grid.atlassian.net/browse/AG-11720


packages/ag-charts-community/src/chart/data/dataModel.ts
```diff
@@ -14,7 +14,7 @@ export interface DataGroup {
     keys: any[];
     datumIndices: number[][];
     aggregation: any[][];
-    validScopes: Set<ScopeId> | undefined;
+    validScopes: Set<ScopeId>;
```
jacobp100, line 17: Is it worth trying to keep the `| undefined` (i.e. all scopes are valid) to reduce allocations on large datasets?

jacobp100 (approved):

packages/ag-charts-community/src/chart/data/dataModel.ts
alantreadway: This is handled by using a single shared `Set` instance, as it's simpler for naive cases just to always have a `Set` present - we can handle performance sensitive cases as exceptions (there are already a couple).

packages/ag-charts-community/src/chart/series/cartesian/barSeries.ts
```diff
@@ -539,24 +540,26 @@ export class BarSeries extends AbstractBarSeries<
         if (processedData.type === 'grouped') {
             const width = barWidth;
 
-            const yStartValues = dataModel.resolveColumnById(this, `yValue-start`, processedData);
-            const yEndValues = dataModel.resolveColumnById(this, `yValue-end`, processedData);
-            const yRangeIndex = dataModel.resolveProcessedDataIndexById(this, `yValue-range`);
+            const stacked = dataModel.hasColumnById(this, `yValue-start`);
+            const yStartValues = stacked ? dataModel.resolveColumnById(this, `yValue-start`, processedData) : [];
+            const yEndValues = stacked ? dataModel.resolveColumnById(this, `yValue-end`, processedData) : [];
+            const yRangeIndex = stacked ? dataModel.resolveProcessedDataIndexById(this, `yValue-range`) : -1;
```
jacobp100, line 546: This helps ensuring you don't forget to do the checks later:-

```suggestion
            const yStartValues = stacked ? dataModel.resolveColumnById(this, `yValue-start`, processedData) : undefined;
            const yEndValues = stacked ? dataModel.resolveColumnById(this, `yValue-end`, processedData) : undefined;
            const yRangeIndex = stacked ? dataModel.resolveProcessedDataIndexById(this, `yValue-range`) : undefined;
```

jacobp100 (approved):

PR #3470, opened 2025-02-04
Data model perf

jacobp100:
We had a case in the high performance case where an array couldn't directly store the floats inside it, and instead has to allocate an object to store the number, then reference the object by pointer in array. This happens because by a JS quirk `new Array(nonZeroLength)` creates the most inefficient array storage method possible. The effect is the array takes up much more memory, and is more expensive for the GC.

image

I've updated these performance-sensitive cases to use `.map` The heap snapshot for the 1m points dropped from 223mb to 214mb with no change in time to process the data. We might need to debug the performance, because it's taking just under 3s to process 1m points

### DIFF
--- a/packages/ag-charts-community/src/chart/data/dataModel.ts
+++ b/packages/ag-charts-community/src/chart/data/dataModel.ts
@@ -802,7 +802,7 @@ export class DataModel<
                     continue;
                 }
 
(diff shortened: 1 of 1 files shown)


alantreadway (approved):
LGTM

PR #3471, opened 2025-02-04
AG-12936 Fix zoom buttons direction check to before layout

lsjroberts:

Caused by `shouldFlipXY` not getting set on the zoom module until after a layout. So the toolbar buttons were a step behind and did their check with the wrong x/y zoom values.

This moves it into the process data step on the update service (and a small change on that service to allow mixed event types).


alantreadway (approved):
LGTM

PR #3473, opened 2025-02-04
AG-13663 Force disable `:focus-visible` with touch (iOS Safari bug)

olegat:


alantreadway (approved):
LGTM

PR #3476, opened 2025-02-04
AG-13945 Fix measure date range annotations with grouping values

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13945


alantreadway (approved):
LGTM

PR #3482, opened 2025-02-05
AG-13216 Update focus indicator to not clip

jacobp100:
https://ag-grid.atlassian.net/browse/AG-13216


packages/ag-charts-community/src/dom/focusStyles.css
```diff
@@ -18,6 +18,7 @@
     width: 100%;
     height: 100%;
     fill: none;
+    overflow: visible;
```
olegat, line 21: I think that we can probably enable `overflow: visible;` on boxed focus indicators too, either by copying this to `.ag-charts-focus-indicator > div` or moving it under `.ag-charts-focus-indicator`. This will fix clipping of line strokes:
https://ag-grid.atlassian.net/browse/AG-13216?focusedCommentId=77248

olegat (changes requested):

packages/ag-charts-community/src/dom/focusStyles.css
jacobp100: All those elements already overflow implicitly. Svg is the only element with a hidden overflow by default

olegat: Just checked. SGTM 👍

olegat (approved):

PR #3488, opened 2025-02-05
AG-14001 - Hide Plunker and CodeSandbox buttons on iOS

markdurrant:
Hide Plunker and CodeSandbox buttons on iOS as it's not possible to make them function satisfactorily.


taktran (approved):
👍🏼

PR #3500, opened 2025-02-06
AG-12936 Fix zoom on bar series with only one series

lsjroberts:

### DIFF
--- a/packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts
+++ b/packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts
@@ -848,7 +848,7 @@ export abstract class CartesianSeries<
         if (!dataModel || !processedData) return Infinity;
 
         const crossValues = this.keysOrValues(crossAxisKey);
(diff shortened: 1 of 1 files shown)


packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts
```diff
@@ -848,7 +848,9 @@ export abstract class CartesianSeries<
         if (!dataModel || !processedData) return Infinity;
 
         const crossValues = this.keysOrValues(crossAxisKey);
-        const allAxisValues = axisKeys.map((axisKey) => this.keysOrValues(axisKey));
+        const allAxisValues = axisKeys.map(
+            (axisKey) => processedData.columns[dataModel.resolveProcessedDataIndexById(this, axisKey)] ?? []
```
alantreadway: Nit: Maybe use `dataModel.resolveColumnById()`?

lsjroberts: That's the one, forgot which fn it was that I used previously.

alantreadway (approved):
LGTM

PR #3501, opened 2025-02-06
AG-13736 Async updateZoom

olegat:
https://ag-grid.atlassian.net/browse/AG-13736


olegat:
Deferred for next release.

PR #3502, opened 2025-02-06
Reduce pixel color diff threshold from 0.2 on playwright config

lsjroberts:
Default color diff threshold is `0.2` (20%).

Setting it to 0% makes the tests a bit flaky with varying handfuls of pixels on aliased rounded corners being detected as changing (though with no actual human recognisable change).

So I've set it to 1% which should catch any actual errors.

(`maxDiffPixels` is only relevant if a pixel breaches this threshold, else it is considered to not be different).


alantreadway (approved):
LGTM

PR #3503, opened 2025-02-06
AG-13951 Fix reading of pixel ratio for axes

jacobp100:
https://ag-grid.atlassian.net/browse/AG-13951

Now takes into account the pending pixel ratio


alantreadway (approved):
LGTM

PR #3505, opened 2025-02-07
AG-13951 Fix legend marker rendering when browser zooming

jacobp100:
https://ag-grid.atlassian.net/browse/AG-13951

Caused by the series area having an animation applied, meaning the pixel alignment failed


alantreadway (approved):
LGTM

PR #3510, opened 2025-02-07
AG-14051 update installation page

JamesSwinton:
Updates installation page content to include CDN & Download for JS, and bring more inline with Grid


AG-DavidG (approved):
LGTM

PR #3514, opened 2025-02-07
key features tweaks

AG-DavidG:


JamesSwinton:
LGTM

packages/ag-charts-website/src/content/docs/key-features/_examples/enterprise-features-example/main.ts
```diff
@@ -42,8 +43,16 @@ const options: AgCartesianChartOptions = {
...
     ],
 };
 
-AgCharts.create(options);
+let chart = AgCharts.create(options);
+
+function reloadChart() {
+    chart.destroy();
+    chart = AgCharts.create(options);
+}
```
alantreadway: The example generator can't handle this correctly for FWs (see e2e test failures).

AG-DavidG: Have removed it now.

PR #3516, opened 2025-02-07
AG-13363 Fix TC8: Do not dispatch synthetic clicks after noop touch drags

olegat:


alantreadway (approved):
LGTM

PR #3517, opened 2025-02-07
AG-14059 - Stop using window.alert() for examples.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-14059

### DIFF
--- a/packages/ag-charts-website/src/content/docs/accessibility-test/_examples/opening-context-menu-second-chart/alert.ts
+++ b/packages/ag-charts-website/src/content/docs/accessibility-test/_examples/opening-context-menu-second-chart/alert.ts
@@ -1,0 +1,18 @@
+// Less obtrusive alert using HTML Popover API.
+let currentPopover;
+window.alert = (msg) => {
+    if (currentPopover) {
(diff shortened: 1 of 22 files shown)


packages/ag-charts-website/src/content/docs/accessibility-test/_examples/opening-context-menu-second-chart/alert.ts
```diff
@@ -0,0 +1,20 @@
+// Less obtrusive alert using HTML Popover API.
+(() => {
```
jacobp100, line 2: Is the IIFE needed?

jacobp100 (approved):

packages/ag-charts-website/src/content/docs/accessibility-test/_examples/opening-context-menu-second-chart/alert.ts
alantreadway: Yes unfortunately, but mostly because TS validation fails without it (`nx validate-examples`). The variable is declared at the global scope, and then complains that all the examples clash with one another :-(

PR #3520, opened 2025-02-07
AG-14066 Prevent default touchmove events when dragAction: 'hover'

olegat:
https://ag-grid.atlassian.net/browse/AG-14066


alantreadway (approved):
LGTM

PR #3528, opened 2025-02-10
CRT-663 Fix zoom on maps with no series bound to axes

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-663


alantreadway (approved):
LGTM

PR #3529, opened 2025-02-10
CRT-669 Change zoom custom buttons example to always show buttons

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-669


alantreadway (approved):
LGTM

PR #3531, opened 2025-02-10
CRT-664 - Fix Histogram with Date-value xKey.

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-664


packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
```diff
@@ -192,8 +192,11 @@ export class HistogramSeries extends CartesianSeries<
             this.calculatedBins = [...bins];
 
             return (keys) => {
-                const xValue = keys[0];
-                if (typeof xValue !== 'number') return [];
+                let xValue = keys[0];
+                if (isDate(xValue)) {
+                    xValue = xValue.getTime();
+                }
```
jacobp100, line 198: Minor - but we have `dateToNumber` which is what's used for the category scale

jacobp100 (approved):

PR #3535, opened 2025-02-10
CRT-661: Funnel - Item styler params do not all include `datum` on hover

iMoses:
https://ag-grid.atlassian.net/browse/CRT-661


jacobp100 (approved):

alantreadway (approved):
LGTM

PR #3552, opened 2025-02-10
CRT-686 Focus on chart when receiving our synthetic clicks from touch

olegat:
https://ag-grid.atlassian.net/browse/CRT-686


alantreadway (approved):
LGTM

PR #3563, opened 2025-02-11
CRT-685 Make panToBBox fail if zoom is disabled

olegat:
https://ag-grid.atlassian.net/browse/CRT-685

This fixes keyboard navigation on non-zoom-pannable charts.


alantreadway (approved):
LGTM

PR #3564, opened 2025-02-11
CRT-684 Fix measurer stats visible when parent off canvas

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-652


alantreadway (approved):
LGTM

PR #3568, opened 2025-02-11
CRT-665 Fix maps zoom anchor point when pressing zoom buttons

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-665


alantreadway (approved):
LGTM

PR #3572, opened 2025-02-11
CRT-654 Fix parallel channel resizing, does not need validation

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-654


alantreadway (approved):
LGTM

PR #3574, opened 2025-02-11
CRT-697: Website - API scroll-to doesn't take header size into account

iMoses:
https://ag-grid.atlassian.net/browse/CRT-697


alantreadway (approved):
LGTM

PR #3576, opened 2025-02-11
AG-13991: Interactivity not working correctly in iframes

iMoses:
https://ag-grid.atlassian.net/browse/AG-13991


olegat (approved):
LGTM.

Although technically the user and manually overwrite the `constructor.name` property, I believe that this solution is sufficient.

alantreadway (approved):

PR #3582, opened 2025-02-12
AG-14095 Fix toggling axis drag zoom cursor when zoom invalid

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14095


alantreadway (approved):
LGTM

PR #3588, opened 2025-02-13
AG-14117 - Remove use of object entries.

alantreadway:


packages/ag-charts-enterprise/src/features/annotations/scenes/startEndScene.ts
```diff
@@ -41,6 +41,7 @@ export abstract class StartEndScene<Datum extends StartEndProperties> extends Li
             this.start.visible = show;
             this.end.visible = show;
         } else {
+            // eslint-disable-next-line no-restricted-properties
             for (const [handle, visible] of Object.entries(show) as [StartEndHandle, boolean][]) {
```
iMoses: Why wasn't the new method used here?

packages/ag-charts-enterprise/src/features/annotations/fibonacci-retracement-trend-based/fibonacciRetracementTrendBasedScene.ts
```diff
@@ -283,6 +283,7 @@ export class FibonacciRetracementTrendBasedScene extends FibonacciScene<Fibonacc
             this.end.visible = show;
             this.endRetracement.visible = show;
         } else {
+            // eslint-disable-next-line no-restricted-properties
             for (const [handle, visible] of Object.entries(show) as [StartEndHandle, boolean][]) {
```
iMoses: Why wasn't the new method used here?

iMoses (approved):
LGTM

packages/ag-charts-core/src/utils/iterators.ts
```diff
@@ -31,3 +31,22 @@ export function first<T>(iterable: Iterable<T>): T | never {
     }
     throw new Error('AG Charts - no first() value found');
 }
+
+/**
+ * Efficient key/value iterator. Note that the returned tuple is always the same array, only
+ * elements change from one iteration to the next, so should only be used in destructing for
+ * statements.
+ *
+ * NOTE: For performance sensitive code, prefer using `Object.keys()` directly.
```
olegat: Is `Objects.keys()` preferable over using a `Map`? MDN mentions that the `Map` type is more efficient with additions & removals (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) but I see so mention of performance with respect to iteration.

StephenCooper: Would it not be better to make all the changes now and not have this extra code?

alantreadway: @olegat I'm not intending to increase the scope of this change to reworking data structures at this point, that is a much riskier change - we should tackle this type of tech-debt as we go. For a bunch of cases (e.g. options processing) we wouldn't want to require use of a `Map`, but agree internally it might make sense in some cases.

@StephenCooper Aside from the annotation-related cases which I'll do another pass over now, we can't use `entries()` for some cases without a bigger performance drop - see https://ag-grid.atlassian.net/browse/AG-14117?focusedCommentId=78358

packages/ag-charts-enterprise/src/features/annotations/scenes/startEndScene.ts
alantreadway: Fixed

packages/ag-charts-enterprise/src/features/annotations/fibonacci-retracement-trend-based/fibonacciRetracementTrendBasedScene.ts
alantreadway: Fixed

packages/ag-charts-core/src/utils/iterators.ts
alantreadway: I've amended the `NOTE` to suggest considering `Map` too, FWIW.

olegat: @alantreadway Just to clarify, what I meant was that comment might mislead someone into thinking that `Object.keys()` and `object[key]` is always the optimal, but this may not always be the case. Perhaps the comment could be changed to
```
For performance sensitive code, prefer using `Object.keys()` directly or a `Map`.
```

PR #3592, opened 2025-02-13
AG-14137 - Make ag-charts-core a non-dev dependency.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-14137


iMoses (approved):
LGTM

PR #3593, opened 2025-02-13
AG-14133: Fix `grouped-category` axis line position when initialised …

iMoses:
…without data

https://ag-grid.atlassian.net/browse/AG-14133


alantreadway (approved):
LGTM

PR #3594, opened 2025-02-13
AG-13959 Fix annotation drag to be contained by non-visible pixel range

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13959


alantreadway (approved):
LGTM

PR #3597, opened 2025-02-13
AG-14117 - Fix Sparkline datumKey warning.

alantreadway:


packages/ag-charts-community/src/module/optionsModule.ts
```diff
@@ -180,7 +180,8 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
         const processor = presetType ? PRESET_DATA_PROCESSORS[presetType] : undefined;
         if (presetType != null && deltaOptions.data != null && processor != null) {
             // Handle preset data transforms gracefully.
-            deltaOptions = mergeDefaults(processor(deltaOptions.data), deltaOptions) as DeepPartial<T>;
+            const { series, data } = processor(deltaOptions.data);
```
alantreadway, line 183: This is the fix, we only take `series` and `data` properties from the processor output.

jacobp100 (approved):

PR #3584, opened 2025-02-12
AG-14140 Touch Node.js tests

olegat:
https://ag-grid.atlassian.net/browse/AG-14140

PR #3601, opened 2025-02-14
AG-14140 Add smoke tests for `touch.dragAction`

olegat:
https://ag-grid.atlassian.net/browse/AG-14140

Depends on:
- https://github.com/ag-grid/ag-charts/pull/3584

Unfortunately, Playwright only provides a Touchscreen.tap for `'touchstart'` & `'touchend'` event simulation, which is insufficient for a test cases.
    
Therefore, add a Node.js test that closely mimics the `single-finger-touch-dragging` docs example.


alantreadway (approved):
LGTM

PR #3602, opened 2025-02-14
AG-13707 Allow fibonacci and channels to drag some handles out of continuous domains

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13707


alantreadway (approved):
LGTM

PR #3605, opened 2025-02-14
AG-14150 Fix options toolbar to respect enabled property

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14150


alantreadway (approved):
LGTM

PR #3607, opened 2025-02-14
AG-14117 - Allow tooltip.renderer() in options fast-path.

alantreadway:

### DIFF
--- a/packages/ag-charts-community/src/module/optionsModule.ts
+++ b/packages/ag-charts-community/src/module/optionsModule.ts
@@ -105,7 +105,7 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
         options = ChartOptions.FAST_PATH_OPTIONS,
         path = ''
     ) {
-        const slowKeys = new Set<string>();
+        let slowKeys: Set<string> | undefined;
         for (const key of Object.keys(deltaOptions)) {
             const config = options.get(key);
             if (config === undefined) {
@@ -127,7 +127,7 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
             }
         }
 
-        return slowKeys.size === 0;
+        return slowKeys === undefined;
     }
 
     activeTheme: ChartTheme;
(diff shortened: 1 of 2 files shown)


packages/ag-charts-community/src/module/optionsModule.ts
```diff
@@ -93,14 +93,41 @@ const unthemedSeries = new Set<SeriesType>(['map-shape-background', 'map-line-ba
...
+        ['height', null],
+        ['container', null],
+        ['tooltip', new Map([['renderer', null]])],
+    ]);
+    private static isFastPathDelta(
+        deltaOptions: DeepPartial<AgChartOptions>,
+        options = ChartOptions.FAST_PATH_OPTIONS,
+        path = ''
+    ) {
+        const slowKeys = new Set<string>();
```
jacobp100: This is only used for debugging, so we could initiate it lazily

```suggestion
        let slowKeys: Set<string> | undefined;
```

packages/ag-charts-community/src/module/optionsModule.ts
```diff
@@ -93,14 +93,41 @@ const unthemedSeries = new Set<SeriesType>(['map-shape-background', 'map-line-ba
...
+                    `ChartOptions.isFastPathDelta() - slow path required due to presence of:`,
+                    slowKeys
+                );
+            } else {
+                ChartOptions.perfDebug(`ChartOptions.isFastPathDelta() - fast path possible.`);
             }
         }
-        return true;
+
+        return slowKeys.size === 0;
```
jacobp100:
```suggestion
        return slowKeys === undefined;
```

jacobp100 (approved):

packages/ag-charts-community/src/module/optionsModule.ts
```diff
@@ -93,14 +93,41 @@ const unthemedSeries = new Set<SeriesType>(['map-shape-background', 'map-line-ba
...
-            if (!this.FAST_PATH_OPTIONS.has(key as keyof AgChartOptions)) {
-                return false;
+            const config = options.get(key);
+            if (config === undefined) {
+                slowKeys.add(path + key);
+            } else if (config === null) {
+                // Supported key, no need to recurse.
+            } else if (!ChartOptions.isFastPathDelta((deltaOptions as any)[key], config, `${key}.`)) {
+                slowKeys.add(path + key);
+            }
```
jacobp100, line 117:
```suggestion
            if (config === undefined) {
                slowKeys ??= new Set();
                slowKeys.add(path + key);
            } else if (config === null) {
                // Supported key, no need to recurse.
            } else if (!ChartOptions.isFastPathDelta((deltaOptions as any)[key], config, `${key}.`)) {
                slowKeys ??= new Set();
                slowKeys.add(path + key);
            }
```

alantreadway:
Closing out as there are some gaps with this change which don't address the fast-path optimisation for the intended case.

PR #3612, opened 2025-02-17
AG-14100 Fix regression with range area ArrowRight keyboard navigation

olegat:
https://ag-grid.atlassian.net/browse/AG-14100


alantreadway (approved):
LGTM

PR #3614, opened 2025-02-17
AG-14097 Fix pointer tooltips dismissal when focus indicator is shown

olegat:
https://ag-grid.atlassian.net/browse/AG-14097


alantreadway (approved):
LGTM

PR #3622, opened 2025-02-18
AG-13707 Allow channels to drag by handles out of continuous domains

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13707

- When channels dragged by a single handle, it allows 2 of the handles to be dragged outside a continuous domain
- Remove validation from annotations except when restoring. This was preventing annotations from being dragged outside the domains, but this is now managed more precisely by the `translate()` fn.
  - I will split out a new ticket related to this but a tangent from this bug to refactor the `initialState` annotation validation to use the new validation methods and perhaps become part of the annotation manager instead.


alantreadway (approved):
LGTM

PR #3628, opened 2025-02-19
AG-3390 - Update validate dist folder script to check for `ag-charts-core`

taktran:


seanlandsman (commented):
~I think this should probably be renamed to be clearer as to what it does~

tools/release/checkFilesExist.js
```diff
@@ -0,0 +1,70 @@
+const fs = require('fs');
```
seanlandsman: I think this should probably be renamed to be clearer as to what it does

seanlandsman (approved):

PR #3630, opened 2025-02-19
AG-14106 Allow channels to create with inactive handle out of domain

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14106


alantreadway (approved):
LGTM

PR #3634, opened 2025-02-19
AG-8254 Add conditional pie callout line colors to use strokes/fills

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-8254


packages/ag-charts-types/src/chart/themeOptions.ts
```diff
@@ -191,12 +191,19 @@ export interface AgChartThemeParams {
  * @param E leaf types to exclude and keep their original type
  */
 type ExtendLiteralLeaves<T, V, E> = {
-    [P in keyof T]: T[P] extends Array<infer U>
+    [P in keyof T]: NonNullable<T[P]> extends Array<infer U>
+        ? U extends E
+            ? Array<U> | V
```
lsjroberts, line 196: This lets you replace `foo: string[]` with a single operation, rather than having to do each potential item in the array. It only allows it for arrays of `E`, i.e. primitives.

There's no check to see if the operation resolves to an equivalent array, but that will be caught by the normal validation after the operation resolves.

packages/ag-charts-community/src/chart/series/polar/pieTheme.ts
```diff
@@ -34,6 +34,9 @@ export const pieTheme: ExtensibleTheme<'pie'> = {
         calloutLine: {
             length: 10,
             strokeWidth: 2,
+            colors: {
+                $if: [{ $eq: [{ $path: '../strokeWidth' }, 0] }, { $path: '../fills' }, { $path: '../strokes' }],
+            },
```
lsjroberts, line 39: The actual change

alantreadway (approved):
Nice, LGTM

PR #3635, opened 2025-02-19
AG-14102 Clipped focus indicator of markers outside the series viewport

olegat:
https://ag-grid.atlassian.net/browse/AG-14102


alantreadway (approved):
LGTM

PR #3640, opened 2025-02-20
AG-7642 Add support for operations for grid cross filters

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-7642


packages/ag-charts-core/src/options/commonOptionsDefs.ts
```diff
@@ -1,24 +1,62 @@
...
+    fill: or(string, gradient, operation),
     fillOpacity: ratio,
 };
 
 export const strokeOptionsDef: OptionsDefs<StrokeOptions> = {
-    stroke: string,
+    stroke: or(string, operation),
     strokeWidth: positiveNumber,
     strokeOpacity: ratio,
 };
```
lsjroberts, line 63: In the long run, we probably don't want to have to manually attach this validator to every single option.

packages/ag-charts-core/src/options/commonOptionsDefs.ts
```diff
@@ -1,24 +1,62 @@
...
+    $or: array,
+    $and: array,
+    $eq: array,
+    $mul: array,
+    $round: array,
+    $rem: array,
+    $mix: array,
+    $mixEach: array,
+    $foregroundBackgroundMix: array,
+    $foregroundBackgroundAccentMix: array,
```
lsjroberts, line 34: Maybe worth making these more specific later e.g. `tuple(string, string, number)`

packages/ag-charts-core/src/options/commonOptionsDefs.ts
```diff
@@ -1,24 +1,62 @@
...
+
 export const fillOptionsDef: OptionsDefs<FillOptions> = {
-    fill: or(string, object),
+    fill: or(string, gradient, operation),
     fillOpacity: ratio,
 };
 
 export const strokeOptionsDef: OptionsDefs<StrokeOptions> = {
-    stroke: string,
+    stroke: or(string, operation),
```
StephM-W, line 60: do we need to support gradient for stroke too? (example implies this)

lsjroberts: It's not part of the API, if it's in a current example it would be an error too since this previous validator would have blocked it.

alantreadway (approved):
LGTM

PR #3653, opened 2025-02-21
AG-14187 - Fix ProcessedData invalidData metadata generation.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-14187


packages/ag-charts-community/src/chart/data/dataModel.ts
```diff
@@ -888,15 +895,16 @@ export class DataModel<
             const column = columnSource.map((valueDatum, datumIndex) => {
                 const invalidKey = invalidKeys.get(columnScope)?.[datumIndex];
 
-                let value = processValue(def, valueDatum, datumIndex, def.scopes);
+                const result = processValue(def, valueDatum, datumIndex, def.scopes);
```
jacobp100, line 898: Does this returning of the object hurt the high performance benchmarks?

packages/ag-charts-community/src/chart/data/dataModel.ts
```diff
@@ -1185,12 +1186,17 @@ export class DataModel<
 
         const accessors = this.buildAccessors(iterate(keyDefs, valueDefs));
 
+        const reusableResult: ProcessedValue = {
+            value: INVALID_VALUE,
```
jacobp100: Is this constant still needed?

packages/ag-charts-community/src/chart/data/dataModel.ts
```diff
@@ -1240,19 +1249,23 @@ export class DataModel<
...
+            } else {
+                reusableResult.valid = true;
             }
 
             if (def.processor) {
                 if (!processorFns.has(def)) {
                     processorFns.set(def, def.processor());
                 }
-                value = processorFns.get(def)?.(value);
+                value = processorFns.get(def)?.(value, idx);
```
jacobp100: Small point - but it's faster to do `let value = processorFns.get(...)` then `if (value == null) { value = ...; processorFns.set(...) }` rather than `.has` then `.get`

jacobp100 (approved):

packages/ag-charts-community/src/chart/data/dataModel.ts
alantreadway: Fair point, done

packages/ag-charts-community/src/chart/data/dataModel.ts
alantreadway: Removed

packages/ag-charts-community/src/chart/data/processors.ts
```diff
@@ -83,17 +84,37 @@ export function rowCountProperty<K>(propName: K, opts: Partial<DatumPropertyDefi
...
+    return v;
+};
+
+function processorChain(...chain: ((() => ProcessorFn) | undefined)[]): () => ProcessorFn {
+    const filteredChain = chain.filter((fn): fn is () => ProcessorFn => fn != null);
+    if (filteredChain.length === 0) {
+        return () => noopProcessor;
+    }
+    if (filteredChain.length === 1) {
+        return filteredChain[0];
```
jacobp100, line 97: Does this need to be called?

alantreadway:
Comments addressed, merging

PR #3654, opened 2025-02-21
AG-14228 Only fire synthetic dblclick events if touches are nearby

olegat:
https://ag-grid.atlassian.net/browse/AG-14228


alantreadway (approved):
LGTM

PR #3655, opened 2025-02-21
AG-14117 - More sparkline optimisations.

alantreadway:

### DIFF
--- a/packages/ag-charts-community/src/chart/axis/axis.ts
+++ b/packages/ag-charts-community/src/chart/axis/axis.ts
@@ -499,7 +499,11 @@ export abstract class Axis<
 
     _scaleNiceDomainInputDomain: D[] | undefined = undefined;
     _scaleNiceDomainRangeExtent: number = NaN;
-    calculateLayout(initialPrimaryTickCount?: number) {
+    calculateLayout(initialPrimaryTickCount?: number): {
+        primaryTickCount?: number;
+        bbox: BBox;
+        niceDomain?: unknown[];
+    } {
         const { scale, label, visibleRange, nice } = this;
 
         this.updateScale();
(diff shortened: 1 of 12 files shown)


packages/ag-charts-community/src/chart/axis/axis.ts
```diff
@@ -526,12 +530,14 @@ export abstract class Axis<
...
-            visibleRange,
-            initialPrimaryTickCount
-        );
+        const {
+            niceDomain,
+            primaryTickCount,
+            ticks,
+            tickDomain,
+            fractionDigits,
+            bbox = BBox.zero,
```
jacobp100: Why was defaulting bbox to zero needed?

packages/ag-charts-community/src/dom/domManager.ts
```diff
@@ -93,6 +93,7 @@ export class DOMManager extends BaseManager<Events['type'], Events> {
     private readonly styles = new Map<string, string>();
     private readonly element: HTMLElement;
     private readonly styleRootElement?: HTMLElement;
+    private newContainer?: HTMLElement = undefined;
```
jacobp100: Should we call this `pendingContainer` similar to `pendingSize`?

jacobp100 (approved):
Only concern is the bbox zero - because a zero bbox has defined coordinates where as `undefined` doesn't, and some code might make assumptions there

packages/ag-charts-community/src/dom/domManager.ts
alantreadway: Fixed.

packages/ag-charts-community/src/chart/axis/axis.ts
alantreadway: Fair point - removed.

PR #3656, opened 2025-02-24
AG-10006 Multi datum tooltips

jacobp100:
https://ag-grid.atlassian.net/browse/AG-10006


packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
alantreadway, line 1: @olegat could you review the changes here?

packages/ag-charts-community/src/chart/tooltip/tooltip.ts
```diff
@@ -329,6 +214,9 @@ export class Tooltip extends BaseProperties {
     @Validate(OBJECT)
     readonly position = new TooltipPosition();
 
+    @Validate(OBJECT)
```
alantreadway, line 217: @iMoses should we be adjusting validations elsewhere too?

packages/ag-charts-community/benchmarks/__image_snapshots__/large-dataset-test-ts-large-dataset-benchmark-after-load-1-x-datum-highlight-1-snap.png
olegat, line 1: > [!WARNING]
> These snapshot updates are suspicious.

In both the snapshot updates the highlighted markers shifts to the left by a few pixel. Consider checking the that the hit-testing on other series types (e.g. bar) hasn't been affected.

packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
olegat: ✅ LGTM. Nothing alarming.

packages/ag-charts-community/benchmarks/__image_snapshots__/large-dataset-test-ts-large-dataset-benchmark-after-load-1-x-datum-highlight-1-snap.png
jacobp100: The logic slightly changed so that if the nearest node had zero distance, we do an exact shape match. The exact shape match can return multiple shapes - normally ordered by data order - then we pick the first. So this is sort of expected

packages/ag-charts-community/src/chart/tooltip/tooltip.ts
alantreadway: Discussed offline, this will be caught if/when the options contract changes.

PR #3659, opened 2025-02-24
AG-14243 Move base theme params shared with grid to separate type

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14243


alantreadway (approved):
LGTM - thanks!

PR #3674, opened 2025-02-26
AG-14117 - Improve low latency rendering speed.

alantreadway:

### DIFF
--- a/packages/ag-charts-community/src/chart/tooltip/tooltip.ts
+++ b/packages/ag-charts-community/src/chart/tooltip/tooltip.ts
@@ -436,13 +436,19 @@ export class Tooltip extends BaseProperties {
     private toggle(visible: boolean) {
         if (!this.element?.isConnected) return;
 
+        // Avoid touching the DOM if invisible and visibility status hasn't changed.
+        if (!this._visible && !visible) return;
+
+        const changed = this._visible !== visible;
         this._visible = visible;
 
         if (!visible) {
             clearTimeout(this._showTimeout);
         }
 
-        this.element.togglePopover(visible);
+        if (changed) {
+            this.element.togglePopover(visible);
+        }
 
         if (visible) {
             // We can only measure the element when it's actually visible
--- a/packages/ag-charts-community/src/dom/domManager.ts
+++ b/packages/ag-charts-community/src/dom/domManager.ts
@@ -87,6 +87,8 @@ function createTabGuardElement(guardedElem: HTMLElement, where: 'beforebegin' | 
 }
 
 export class DOMManager extends BaseManager<Events['type'], Events> {
+    private static readonly batchedUpdateContainer: DOMManager[] = [];
+
     readonly anchorName = `--${createId(this)}`;
 
     private readonly rootElements: Record<DOMElementClass, LiveDOMElement>;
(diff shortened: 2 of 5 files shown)


olegat (approved):
LGTM

packages/ag-charts-community/src/chart/tooltip/tooltip.ts
```diff
@@ -436,13 +436,19 @@ export class Tooltip extends BaseProperties {
     private toggle(visible: boolean) {
         if (!this.element?.isConnected) return;
 
+        // Avoid touching the DOM if invisible and visibility status hasn't changed.
+        if (!this._visible && !visible) return;
```
jacobp100, line 440: Did this check improve perf? It feels like it'll be tricky to get this one right because it also affects timeouts

packages/ag-charts-community/src/dom/domManager.ts
```diff
@@ -240,14 +254,15 @@ export class DOMManager extends BaseManager<Events['type'], Events> {
...
-        if (!isShadowDom) {
+        if (this.documentRoot != null) {
             for (const id of this.rootElements['styles'].children.keys()) {
                 this.removeChild('styles', id);
             }
         }
 
         this.container = pendingContainer;
         this.pendingContainer = undefined;
+        this.documentRoot = this.getShadowDocumentRoot(pendingContainer);
```
jacobp100, line 265: We did previously support a pathological case where you weren't in a shadow DOM, then you moved into a shadow DOM (and vice-versa). Is this something that still works, or do we not care?

packages/ag-charts-community/src/module/optionsModule.ts
```diff
@@ -263,12 +263,14 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
         }
 
         const validatedSeriesOptions: any[] = [];
-        options.series?.forEach((seriesOptions, index) => {
+        let index = -1;
+        for (const seriesOptions of options.series ?? []) {
```
jacobp100, line 267: Was this faster?

jacobp100 (approved):

packages/ag-charts-community/src/chart/tooltip/tooltip.ts
alantreadway: Yes, this code path was clearly showing up as a hot-spot due to (usually redundant) DOM updates before this change.

packages/ag-charts-community/src/dom/domManager.ts
alantreadway: I'll verify, if it's an issue we can probably use the `Node.isConnected()` check to know when to be more pessimistic.

packages/ag-charts-community/src/module/optionsModule.ts
alantreadway: Yes - lambda functions are much harder for the JIT compiler to handle, so they don't get converted to compiled code anywhere near as quickly as normal functions.

This change was a little more tenuous as this did show up as a hotspot in at least one case, and after this change I didn't see it again - but in these situations its generally preferable to err on the side of not using lambdas.

PR #3676, opened 2025-02-26
AG-14131 Fix gauge fonts, remove default label font family & size

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14131


alantreadway (approved):
LGTM

PR #3679, opened 2025-02-26
AG-13024 Add `context` object to API

olegat:

### DIFF
--- a/packages/ag-charts-community/src/chart/apiContext.test.ts
+++ b/packages/ag-charts-community/src/chart/apiContext.test.ts
@@ -3,6 +3,7 @@ import { AgCartesianChartOptions } from 'ag-charts-types';
 import type { MockItemStyler, MockLabelFormatter, MockTooltipRenderer } from './test/freezableMock';
 import { newFreezableMock } from './test/freezableMock';
 import {
+    AgCartesianChartOptionsWithContext,
     Chart,
     createChart,
     hoverAction,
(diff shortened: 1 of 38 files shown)


packages/ag-charts-community/src/chart/apiContext.test.ts
```diff
@@ -0,0 +1,97 @@
...
+            data: [
+                { quarter: 'q1', Toyota: 120000, Ford: 95000, BMW: 80000 },
+                { quarter: 'q2', Toyota: 150000, Ford: 110000, BMW: 90000 },
+                { quarter: 'q3', Toyota: 170000, Ford: 120000, BMW: 95000 },
+                { quarter: 'q4', Toyota: 160000, Ford: 115000, BMW: 92000 },
+            ],
+            series: [
+                { type: 'bar', xKey: 'quarter', yKey: 'Toyota', context: seriesContext },
+                { type: 'bar', xKey: 'quarter', yKey: 'Ford', context: seriesContext },
+                { type: 'bar', xKey: 'quarter', yKey: 'BMW', context: seriesContext },
```
alantreadway: Nit: It would be more realistic (from an Integrated perspective) if each series had a distinct `context`, as I believe we would need series/column specific properties such as Grid `columnId` per series.

packages/ag-charts-community/src/util/properties.ts
```diff
@@ -3,6 +3,9 @@ import { Logger, isArray } from 'ag-charts-core';
 import { extractDecoratedPropertyMetadata, listDecoratedProperties } from './decorator';
 
 export class BaseProperties<T extends object = object> {
+    // user pass-through option: no validation-decorator required.
+    context?: unknown;
```
alantreadway, line 7: I don't think this is the right class for this, as it's used as the base class for a lot of use-cases where this isn't relevant - perhaps we could have a `ContextProperties extends BaseProperties` base class with the `context`-related behaviours, which is extended in the specific cases where this behaviour is needed?

alantreadway (changes requested):

packages/ag-charts-community/src/util/properties.ts
olegat: > it's used as the base class for a lot of use-cases where this isn't relevant

Why does this matter? 

Removing this property will raise this compiler error:

```
packages/ag-charts-community/src/util/properties.ts:41:18 - error TS2339: Property 'context' does not exist on type 'BaseProperties<T>'.
  39 |         }
  40 |         if ('context' in properties) {
> 41 |             this.context = properties.context;
     |                  ^
  42 |             keys.delete('context');
  43 |         }
  44 |         for (const unknownKey of keys) {
```

This `if` is necessary to avoid warnings

packages/ag-charts-community/src/chart/apiContext.test.ts
olegat: Done. Thanks for the feedback.

packages/ag-charts-community/src/util/properties.ts
olegat: This is coming in a follow-up PR https://github.com/ag-grid/ag-charts/commit/a15cdbcdf3c5752f815dd392fef87004d2378b68

alantreadway (approved):
LGTM - thanks for addressing the feedback taking follow-up PRs into account.

PR #3685, opened 2025-02-27
AG-14117 - Optimise DOM for Sparklines.

alantreadway:

### DIFF
--- a/packages/ag-charts-community/src/dom/container.css
+++ b/packages/ag-charts-community/src/dom/container.css
@@ -61,6 +61,8 @@
 }
 
 .ag-charts-swapchain {
+    top: 0;
+    left: 0;
     outline: none;
     opacity: 0;
     pointer-events: none;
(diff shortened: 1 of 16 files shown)


packages/ag-charts-community/src/dom/container.css
```diff
@@ -45,8 +45,8 @@
     -webkit-user-select: none;
 }
 
-.ag-charts-canvas-container > *,
-.ag-charts-canvas > * {
+.ag-charts-canvas-container > *:not([popover]),
```
jacobp100: I didn't know we had this rule, and it's definitely not right. We need to remove it.

We need to be careful not to change the specificity, however.

```suggestion
.ag-charts-canvas-container > :where(:not([popover])),
```

alantreadway: Discussed offline and fixed.

PR #3691, opened 2025-02-28
AG-14185 Fix $path operation not warning when default value provided

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14185


alantreadway (approved):
LGTM

PR #3696, opened 2025-03-03
AG-14274 Fix pie and donut default gradient colours

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14274


alantreadway (approved):
LGTM

PR #3701, opened 2025-03-03
AG-14158 - Upgrade to Astro 5.x

taktran:
## Changes

* Upgrade to [Astro 5.4](https://docs.astro.build/en/guides/upgrade-to/v5)
* Updated redundant references in `env.d.ts` file
* Update `tsconfig.ts` for Astro 5
* Migrate content for content collections
* Move `migration.mdoc` outside of `content/docs`
* Add verbose configuration for building docs
* Remove `.astro` folder during `nx clean`
* Update use of boolean html values for `data-is-homepage`
* Add plunkr/codesandbox to local dev CORS

## Notes

* Based off Grid PR: https://github.com/ag-grid/ag-grid/pull/9977
* Content file names need to be lowercased because of https://github.com/withastro/astro/issues/13265#issuecomment-2666270862 (collection names can be capitalised though)
* Need to pin `vitest` to `v2.1.6` because it is not compatible with vite 6: https://github.com/withastro/astro/issues/12384


packages/ag-charts-website/astro.config.mjs
```diff
@@ -118,6 +131,5 @@ export default defineConfig({
             skip: CHECK_REDIRECTS !== 'true',
         }),
     ],
-    experimental: { contentCollectionCache: EXPERIMENTAL_ASTRO_BUILD_CACHING === 'true' },
```
alantreadway, line 121: Is this now enabled by default, and/or are there other caching implications here that we need to be aware of? (E.g. how does this play with the Nx cache?)

packages/ag-charts-website/vitest.config.mjs
```diff
@@ -8,11 +7,8 @@ function resolvePath(srcPath) {
 
 export default getViteConfig({
     root: __dirname,
-    cacheDir: '../../node_modules/.vite/test',
```
alantreadway, line 11: Is the caching disabled now, or is the path defaulted somewhere else now?

alantreadway (approved):
LGTM

packages/ag-charts-website/vitest.config.mjs
taktran: This defaults to `node_modules/.vite`: https://vite.dev/config/shared-options.html#cachedir

My thinking was that it was in the `.vite` directory already. Does it need to specifically be in `.vite/test` for nx caching or anything?

packages/ag-charts-website/astro.config.mjs
taktran: This has been removed in v5: https://docs.astro.build/en/guides/upgrade-to/v5/#experimental-flags
I presumed it was being handled by the vite 6 upgrade, but I found your commit (https://github.com/ag-grid/ag-charts/commit/24195dbdf205196e31be03fa6510ac10d1b78927) related to it, and need to understand what it was about.

PR #3707, opened 2025-03-04
AG-10112 Invalidate callback caches

olegat:
https://ag-grid.atlassian.net/browse/AG-10112


packages/ag-charts-community/src/module/optionsModule.ts
```diff
@@ -364,7 +366,12 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
         if (this === other) return {};
         if (other == null) return this.processedOptions;
 
-        return (this.fastDelta as Partial<T>) ?? jsonDiff(other.processedOptions, this.processedOptions);
+        // Callbacks must always be considered "outdated" because it is impossible for us to know if the return value
+        // (which is what we actually care about) has changed.
+        return (
+            (this.fastDelta as Partial<T>) ??
+            jsonDiff(other.processedOptions, this.processedOptions, ChartOptions.JSON_DIFF_OPTS)
+        );
```
alantreadway: We can't take this approach as for Sparkline cases this will completely kill update optimisations. From an options processing perspective nothing has changed, we should just dump any cached callback values.

alantreadway (changes requested):

packages/ag-charts-community/src/module/optionsModule.ts
olegat: > We can't take this approach as for Sparkline cases this will completely kill update optimisations. 

Are you sure about this?

I've got two observations:
1. `this.fastDelta` should theoretically be defined in the Sparkline cases, so it shouldn't be calling `jsonDiff` and therefore shouldn't be affected by this change.
2. I would expect that returning a non-null `jsonDiff` only updates the node-datums and only redraw the canvas if it detects a change. Is that not the case?

This PR already clears the callback cache on updates, but that's insufficient. Without this change, jsonDiff would be `null` and the callbacks will not be called because `update(opts)` assumes that nothing needs to be redrawn.

alantreadway: Discussed offline (see JIRA commentary).

alantreadway (approved):
LGTM, pending performance verifications.

alantreadway (approved):
LGTM

PR #3738, opened 2025-03-10
AG-14200 Fix nightingale and radial column default axis shapes

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14200


alantreadway (approved):
LGTM

PR #3740, opened 2025-03-10
Gradient wip

jacobp100:


jacobp100:
Merged into AG-13860/gradient-fill-fixes

PR #3741, opened 2025-03-11
AG-14319 Update palette fills to take gradients

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14319

- Callout lines check if it is a gradient and then pick the stroke instead (or foreground color if no stroke for that index)
- Box plot mixes the palette fill with the background, so `$mix` can now handle gradients


alantreadway (approved):
LGTM

PR #3744, opened 2025-03-11
AG-14117 - Fix tooltip forced relayout + misc optimisations.

alantreadway:

### DIFF
--- a/packages/ag-charts-community/src/chart/data/dataModel.ts
+++ b/packages/ag-charts-community/src/chart/data/dataModel.ts
@@ -890,7 +890,7 @@ export class DataModel<
             const columnScopes = new Set(def.scopes);
             const columnScope = first(def.scopes);
             const columnSource = sources.get(columnScope) as unknown[];
-            const column = new Array<unknown>(columnSource.length);
+            const column = new Array<unknown>();
             for (let datumIndex = 0; datumIndex < columnSource.length; datumIndex++) {
                 const valueDatum = columnSource[datumIndex];
                 const invalidKey = invalidKeys.get(columnScope)?.[datumIndex];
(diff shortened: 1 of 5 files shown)


packages/ag-charts-community/src/chart/data/dataModel.ts
```diff
@@ -890,7 +890,9 @@ export class DataModel<
             const columnScopes = new Set(def.scopes);
             const columnScope = first(def.scopes);
             const columnSource = sources.get(columnScope) as unknown[];
-            const column = columnSource.map((valueDatum, datumIndex) => {
+            const column = new Array<unknown>(columnSource.length);
```
jacobp100: Use `.map` here - if you use `new Array(length)`, it basically forces the array into a `Map<number, element>` behind the scenes. That was one of the big wins for the high performance work

packages/ag-charts-community/src/scene/shape/text.ts
```diff
@@ -94,7 +94,16 @@ export class Text<D = any> extends Shape<D> {
...
 
-        ctx.font = TextUtils.toFontString(this);
+        if (!fill && !(stroke != null && strokeWidth > 0)) {
+            // Short circuit early if nothing will be rendered.
+            return super.render(renderCtx);
+        }
+
+        const font = TextUtils.toFontString(this);
+        // Try to avoid this assignment, which typically always incurs a font switch cost.
+        if (ctx.font !== font) {
```
jacobp100, line 104: If this is showing as a hot path, it might be worth storing the resolved font on the instance too, and reading that to avoid the overhead of crossing the language barrier since the `.font` getter is C++

packages/ag-charts-community/src/chart/data/dataModel.ts
alantreadway: Based upon https://jsbench.me/t6m85tuhgv/1, our best option seems to be to just use `new Array()` without `.map()` - the main issue is likely that the closure can't be easily optimised at runtime for this particular case.

packages/ag-charts-community/src/scene/shape/text.ts
alantreadway: This is probably ok, as with a prior iteration of this change the hotspot was significantly reduced with just this guard added. We'd need to track the current font in the renderContext for this to be accurate, which seems more complexity than I'd like to add right now.

jacobp100 (approved):

PR #3746, opened 2025-03-11
Ag 13860/gradient fill fixes

manapeirov:


packages/ag-charts-types/src/series/markerOptions.ts
```diff
@@ -19,5 +19,6 @@ export interface AgSeriesMarkerOptions<TDatum, TParams> extends AgSeriesMarkerSt
 }
 
 export interface ISeriesMarker<TParams> extends AgSeriesMarkerOptions<unknown, TParams> {
+    defaultColorRange: string[];
```
alantreadway: Does this need to be in a public interface?

alantreadway (approved):
LGTM

PR #3750, opened 2025-03-12
AG-14201 Update default axes to use operations

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14201

### DIFF
--- a/packages/ag-charts-community/src/module/optionsModule.ts
+++ b/packages/ag-charts-community/src/module/optionsModule.ts
@@ -424,11 +424,10 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
 
     private getDefaultAxes(options: T, seriesTheme: object) {
         const optionsType = this.optionsType(options);
-        let firstSeriesOptions = options.series?.find((series) => (series.type ?? 'line') === optionsType) ?? {};
-        if (seriesRegistry.isDerivedDefaultAxes(optionsType)) {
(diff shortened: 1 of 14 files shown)


packages/ag-charts-community/src/module/optionsModule.ts
```diff
@@ -424,11 +424,10 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
...
         const optionsType = this.optionsType(options);
-        let firstSeriesOptions = options.series?.find((series) => (series.type ?? 'line') === optionsType) ?? {};
-        if (seriesRegistry.isDerivedDefaultAxes(optionsType)) {
-            firstSeriesOptions = mergeDefaults(firstSeriesOptions, seriesTheme);
-        }
-        return seriesRegistry.cloneDefaultAxes(optionsType, firstSeriesOptions) as T;
+        const firstSeriesOptions = options.series?.find((series) => (series.type ?? 'line') === optionsType) ?? {};
+        const axes = seriesRegistry.cloneDefaultAxes(optionsType) as T;
+        jsonResolveOperations(axes, {}, undefined, mergeDefaults(firstSeriesOptions, seriesTheme));
+        return axes;
```
lsjroberts, line 430: I considered adding a `hasOperation()` check here but it would need to be recursive, at which point it may as well just do the recursion to merge the objects.

alantreadway (approved):
LGTM

PR #3751, opened 2025-03-12
AG-14258 Fix integrated resolveOperations() fn return type

lsjroberts:
This now matches https://github.com/ag-grid/ag-charts/commit/03c2179f3d0be7f52434ab245a5516abfff55fb7


alantreadway (approved):
LGTM

PR #3759, opened 2025-03-13
AG-14059 - Add console log component

taktran:
## Changes

* Refactor darkmode snippet insertion, so it's a generic function that gets passed into `frameworkFilesGenerator`
* Add console log override for examples
  * Add snippet in example generator, and remove it from view in `CodeViewer.tsx`
  * Send console message to parent window
  * Listen to iframe console messages and display them underneath the example runner
* Replace all `window.alert`s in examples to use `console.log`


packages/ag-charts-website/src/content/docs/events/_examples/chart-click-event/main.ts
```diff
@@ -34,10 +32,10 @@ const options: AgChartOptions = {
     ],
     listeners: {
         click: (_event: AgChartClickEvent) => {
-            window.alert('[click]');
+            console.log('[click] 216');
```
alantreadway:
```suggestion
            console.log('[click]');
```

taktran: Ah, caching hack 🤦🏼‍♂️ - good catch 😄

alantreadway (approved):
LGTM aside from one typo

PR #3764, opened 2025-03-13
Fix memory leak and benchmark assertions.

alantreadway:
- **Fix DOMManager memory leak.**
- **Fix benchmark memory use assertions.**


iMoses (approved):
LGTM

PR #3768, opened 2025-03-13
AG-14342 Fix leaking operations on merged objects

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14342

### DIFF
--- a/packages/ag-charts-community/src/util/json.ts
+++ b/packages/ag-charts-community/src/util/json.ts
@@ -378,10 +378,22 @@ function jsonResolveVisitor<T extends object, P extends object>(
             node[i] = jsonResolveVisitorValue(node[i], params, source, meta, [...path, `${i}`], modifiedPaths);
         }
     } else {
+        let hasOperation = false;
         for (const name of Object.keys(node)) {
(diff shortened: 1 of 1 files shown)


packages/ag-charts-community/src/util/json.ts
```diff
@@ -378,12 +378,24 @@ function jsonResolveVisitor<T extends object, P extends object>(
...
             if (node[name] === operationResolvedUndefined) {
                 delete node[name];
             }
         }
+
+        // Purge excess unused operations to prevent leaks. These occur when the default operation object is merged an
+        // object at the option key.
+        if (hasOperation && Object.keys(node).length > 1) {
+            for (const key of Object.keys(node)) {
+                if (isKey(key, operations)) {
```
alantreadway, line 394: Nit: Should this be restricted to recognised keys, or should we just ignore everything that didn't go through the resolve process?

lsjroberts: No. The valid operation key is resolved, which leaves us with the normal item styler keys and the unresolved operation key(s). It is only the latter that should be removed.

PR #3774, opened 2025-03-14
AG-14417 Add operations and path values for integrated cross filters

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14417


alantreadway (approved):
LGTM - thanks!

PR #3779, opened 2025-03-17
Add debug tooling for multiple scene node updates.

alantreadway:


packages/ag-charts-community/src/scene/node.ts
```diff
@@ -399,29 +410,71 @@ export abstract class Node<D = any> {
         return;
     }
 
-    markDirty() {
+    markDirty(property: string) {
         const { _dirty } = this;
 
+        if (this._debugDirtyProperties && !DEBUG_IGNORE_DIRTY_SOURCES.has(property)) {
```
jacobp100: We can avoid some overhead by 

```suggestion
    markDirty(property: string | undefined) {
        const { _dirty } = this;

        if (property != null && this._debugDirtyProperties) {
```

Then explicitly pass in `undefined` for these cases

alantreadway: Good spot, fixed.

PR #3782, opened 2025-03-17
AG-14319 Fix default color range when up/down/neutral are gradients

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14319


lsjroberts:
The sonarqube error is a duplication issue on the small refactor I did that reduces duplication

alantreadway (approved):
LGTM

PR #3783, opened 2025-03-17
Fix docs example logger to stringify values

lsjroberts:


taktran (commented):
I had this initially, and we decided we'd change the examples for logs with objects, but I'll discuss with the others, and update this later

taktran (approved):
Had a chat with the web team, and we think it's good to go for this pass of the console component. I've added a `safeStringify` to cater for cases where there are circular dependencies (more an issue on grid).

alantreadway (approved):
LGTM

PR #3788, opened 2025-03-17
AG-14318 `@SceneChangeDetection` decorator refactoring

olegat:
https://ag-grid.atlassian.net/browse/AG-14318


packages/ag-charts-community/src/scene/changeDetectable.ts
olegat, line 1: > [!IMPORTANT]
> This file includes most of the important changes

packages/ag-charts-community/src/scene/shape/path.ts
olegat, line 1: > [!IMPORTANT]
> This file includes important changes (removable of `ScenePathChangeDetection` and a new `onChangeDetection` override).

packages/ag-charts-community/src/scene/transformable.ts
olegat, line 1: > [!IMPORTANT]
> I initially tried to implement `markDirtyTransform()` using an `onChangeDetected` override, but unfortunately this clashes with `Path._dirtyPath`).
>
> See https://github.com/ag-grid/ag-charts/commit/261c1e38e59cb78430dbf24f6d4fd6a80d29f78a

packages/ag-charts-community/src/scene/changeDetectable.ts
alantreadway: As a heads-up, we'll need to validate the performance impact of these changes before merge - historically even single line changes made here have made a decent dent in performance, e.g. event something as innocuous as `changeCb?.()` only featured in one of the code-paths was a performance improvement to the others.

olegat: Agreed.

I did refactor it such as that we keep the function-chaining to avoid branching-code on each set operation (https://github.com/ag-grid/ag-charts/commit/8aa8d2e0686d45c051674455a0d1af7402a37472).

I'll check what's performance impact of removing the `changeCb?.()`, but I would be surprised to see much of a difference because from the debugging that I've done it always gets call here in `type="normal"` code-paths: https://github.com/ag-grid/ag-charts/blob/77a58fe6089591bc66d9ee5bddf7d8316f2b5cd9/packages/ag-charts-community/src/scene/changeDetectable.ts#L76

alantreadway (approved):
LGTM

PR #3791, opened 2025-03-17
AG-14222 Fix legend orientation default when invalid position

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14222


alantreadway (approved):
LGTM

PR #3796, opened 2025-03-18
AG-13024 Add root `context` object

olegat:


packages/ag-charts-community/src/util/callbackCache.ts
olegat, line 1: > [!IMPORTANT]
> This file  contains the most relevant change

packages/ag-charts-community/src/util/callbackCache.ts
```diff
@@ -1,25 +1,38 @@
...
+    params: Parameters<F>
+): ReturnType<F> {
+    if (!maybeSetContext(caller1, params)) maybeSetContext(caller2, params);
     return fn(...params);
 }
 
 export class CallbackCache {
     private cache: WeakMap<Function, Map<string, any>> = new WeakMap();
 
-    call<F extends AnyFn>(caller: Caller, fn: F, ...params: Parameters<F>): ReturnType<F> | undefined {
```
olegat, line 22: > [!NOTE]
> Callbacks are called once per datum in some cases. Therefore, I decided to use two separate arguments `caller1: Caller` and `caller2: Caller` instead of single & more general-use `callers: Caller[]` due to performance concerns of allocating a temporary 2-element array for each callback.

alantreadway (approved):
LGTM

PR #3800, opened 2025-03-18
AG-14436 - Fix floating toolbar top/left bound restriction.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-14436


lsjroberts (approved):
LGTM

PR #3804, opened 2025-03-18
AG-14447 Fix annotations translation to record action on key up

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14447


alantreadway (approved):
LGTM

PR #3806, opened 2025-03-18
AG-14442 Do not call stopPropagation on `'contextmenu'` events

olegat:
https://ag-grid.atlassian.net/browse/AG-14442


alantreadway (approved):
LGTM

PR #3809, opened 2025-03-19
AG-13860 - Remove options for release.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-13860

### DIFF
--- a/packages/ag-charts-community/src/scene/util/fill.ts
+++ b/packages/ag-charts-community/src/scene/util/fill.ts
@@ -7,6 +7,8 @@ export interface InternalAgGradientColor extends AgGradientColor {
     gradient?: AgGradientType;
     /** The domain of the color gradient, defaults to item. */
     bounds?: AgGradientColorBounds;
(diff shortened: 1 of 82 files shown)


packages/ag-charts-community/src/scene/util/fill.ts
```diff
@@ -1,12 +1,23 @@
 import { isArray, isObject } from 'ag-charts-core';
-import type { AgGradientColor } from 'ag-charts-types';
+import type { AgGradientColor, AgGradientColorBounds, AgGradientType, CssColor } from 'ag-charts-types';
 import type { AgPatternColor } from 'ag-charts-types';
 
-export function isGradientFill(fill: any): fill is AgGradientColor {
+export interface InternalAgGradientColor extends AgGradientColor {
```
jacobp100, line 5: Could we extend this to allow `colorStops` to also be an array of strings, as that's what's generated by the themes?

jacobp100 (approved):

PR #3812, opened 2025-03-19
AG-14405: Page crashes when time axis interval is too small

iMoses:
https://ag-grid.atlassian.net/browse/AG-14405


alantreadway (approved):
LGTM

PR #3829, opened 2025-03-20
AG-14469 Fix tick generation for ordinal time scale with dense interval

jacobp100:
https://ag-grid.atlassian.net/browse/AG-14469


alantreadway (approved):
LGTM

PR #3836, opened 2025-03-20
AG-14467 Fix pattern defaults when palette is pattern or gradient

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14467


alantreadway (approved):
LGTM

PR #3852, opened 2025-03-21
AG-3390 - Update grid version

taktran:
From running `node ./tools/update-grid-version 33.2.1`


taktran:
Replaced by https://github.com/ag-grid/ag-charts/pull/3854

PR #3857, opened 2025-03-21
Crt 740 743 api docs UI fixes

markdurrant:


packages/ag-charts-website/src/components/api-documentation/components/ApiReference.tsx
```diff
@@ -117,7 +117,7 @@ export function ChildPropertiesButton({
             aria-label={`See child properties of ${name}`}
         >
             <Icon svgClasses={styles.childChevron} name="chevronRight" />
-            <span>See child properties</span>
+            <span>{!isExpanded ? `See` : `Hide`} child properties</span>
```
taktran: nit: It's easier to read when the conditional is not negated

```suggestion
            <span>{isExpanded ? 'Hide' : 'See'} child properties</span>
```

taktran (approved):
Looks good, just a nitpick

PR #3868, opened 2025-03-21
CRT-731 Workaround for path hit test and bbox calculation being incor…

jacobp100:
…rect

https://ag-grid.atlassian.net/browse/CRT-731


alantreadway (approved):
LGTM

PR #3869, opened 2025-03-21
CRT-717: Chart blows up when first series is an invalid type

iMoses:
https://ag-grid.atlassian.net/browse/CRT-717


alantreadway (approved):
LGTM

PR #3875, opened 2025-03-21
CRT-762 - Fix console logging objects in an array of arguments

taktran:
Same as https://github.com/ag-grid/ag-grid/pull/10130


markdurrant (approved):
👍

PR #3886, opened 2025-03-24
CRT-749 Add tests for zoom dragging secondary axis

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-749


lsjroberts:
These were some draft tests I added to find the bug. They cover both the independent and non-independent axes cases so thought I'd PR them and cleanup the zoom tests a bit too.

alantreadway (approved):
LGTM

PR #3901, opened 2025-03-25
AG-14516 Set `tabindex="0"` on interactive tooltip `<a>` tag

olegat:
https://ag-grid.atlassian.net/browse/AG-14516


alantreadway (approved):
LGTM

PR #3912, opened 2025-03-28
Set pattern origin to shape origin

jacobp100:
Node canvas has bugs here so the snapshots won't change - but it does work in the browser

https://github.com/user-attachments/assets/0a0d59b5-3230-4269-8f57-c495fd4a28b9


alantreadway (approved):
LGTM

PR #3929, opened 2025-04-02
AG-8876 Add placeholder `contextMenu.items[]` contract option

olegat:

### DIFF
--- a/packages/ag-charts-community/src/chart/commonOptionsDefs.ts
+++ b/packages/ag-charts-community/src/chart/commonOptionsDefs.ts
@@ -332,7 +332,7 @@ export const commonChartOptionsDefs: OptionsDefs<Omit<AgBaseThemeableChartOption
     },
     contextMenu: {
         enabled: boolean,
-        items: unknown,
+        items: defined,
         extraActions: contextMenuActionsArray,
         extraSeriesAreaActions: contextMenuActionsArray,
         extraNodeActions: contextMenuActionsArray,
--- a/packages/ag-charts-community/src/chart/commonOptionsDefs.ts
+++ b/packages/ag-charts-community/src/chart/commonOptionsDefs.ts
@@ -120,19 +120,6 @@ const chartOverlayOptionsDefs: OptionsDefs<AgChartOverlayOptions> = {
     renderer: callbackOf(or(string, htmlElement)),
 };
 
-const contextMenuItemLiteral = union(
-    'defaults',
-    'download',
-    'zoom-to-cursor',
-    'pan-to-cursor',
-    'toggle-series-visibility',
-    'toggle-other-series',
-    'reset-zoom'
-);
-const contextMenuItemShowOn = union('series-area', 'series-node', 'legend-item');
-const contextMenuItemType = union('action', 'submenu', 'separator');
-contextMenuItemLiteral || contextMenuItemShowOn || contextMenuItemType;
-
 // eslint-disable-next-line sonarjs/deprecation
 const contextMenuActionsArray = arrayOfDefs<AgContextMenuAction>(
     {
--- a/packages/ag-charts-community/src/chart/commonOptionsDefs.ts
+++ b/packages/ag-charts-community/src/chart/commonOptionsDefs.ts
@@ -36,7 +36,7 @@ import type {
     AgChartAutoSizedBaseLabelOptions,
     AgChartCaptionOptions,
     AgChartLabelOptions,
-    AgChartOverlayOptions,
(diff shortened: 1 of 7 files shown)


packages/ag-charts-community/src/chart/commonOptionsDefs.ts
```diff
@@ -324,6 +324,7 @@ export const commonChartOptionsDefs: OptionsDefs<Omit<AgBaseThemeableChartOption
     },
     contextMenu: {
         enabled: boolean,
+        items: unknown,
```
iMoses: please use the `defined` validator

olegat: Done.

packages/ag-charts-types/src/chart/contextMenuOptions.ts
```diff
@@ -5,16 +5,64 @@ import type {
...
+    label: string;
+    /**  TODO: writeme. */
+    iconUrl?: string;
+    /**  TODO: writeme. */
+    enable?: boolean;
+    /**  TODO: writeme. */
+    action?: (event: ShowOnActionEventMap[TShowOn]) => void;
+    /**  TODO: writeme. */
+    items?: AgContextMenuItem;
+}
```
alantreadway: Unless there's an immediate use for it, I would avoid the generic parameter unless needed as it will likely confuse the doc generation?

If we are going with a `type` argument, could we use a union type with that as the discriminator so that TypeScript users will also get some help with understanding which `type`-specific fields are available (e.g. I'm assuming `items` is only specific to the `submenu` case? and for `separator` I'd guess most of the options aren't available?)

packages/ag-charts-types/src/chart/contextMenuOptions.ts
```diff
@@ -5,16 +5,64 @@ import type {
...
+
 export interface AgContextMenuOptions {
     /**  Whether to show the context menu. */
     enabled?: boolean;
-    /**  Custom actions displayed in the context menu when right-clicking anywhere on the chart. */
+    /**  TODO: writeme. */
+    items?: readonly AgContextMenuItem[];
+    /**
+     * Custom actions displayed in the context menu when right-clicking anywhere on the chart.
```
alantreadway, line 78: Nit: Maybe mark these as @deprecated now since we're adding the replacement?

olegat: I agree. I initially added `@deprecated` but reverted it (https://github.com/ag-grid/ag-charts/commit/210cf455697220af83e188976139a4ce3f16310c) to mute lint errors in the CI (https://github.com/ag-grid/ag-charts/actions/runs/14223277340/job/39856573087)

Do you know how to mute these lint errors?

packages/ag-charts-types/src/chart/contextMenuOptions.ts
olegat: The `type` and `items` properties are not relevant for the `action` callback.

The immediate use is to preserve the `TEvent` typing info from the deprecated properties `extraActions, extraSeriesAreaActions, extraNodeActions, extraLegendItemActions`. For example `extraNodeActions` callback parameter include extra properties like `datum, xKey, yKey, ...` which other callback parameters don't have.

The reason that the `TShowOn` parameter is required is because the `showOn` property is an array, but it's not entirely obvious to whether it needs to be. I'll investigate other options.

packages/ag-charts-types/src/chart/contextMenuOptions.ts
olegat: Discussed on Slack. I've added comments to mute the lint errors
```c
// eslint-disable-next-line sonarjs/deprecation
```

packages/ag-charts-types/src/chart/contextMenuOptions.ts
olegat: Two proposed alternatives:
- **Alternative 1:** https://github.com/ag-grid/ag-charts/blob/645c59cef0cee93d32680ab966215dde1d64e23b/packages/ag-charts-types/src/chart/contextMenuOptions.ts
- **Alternative 2:** https://github.com/ag-grid/ag-charts/blob/888de7ed069ba1cb3fed56a48417f0038f15c8ca/packages/ag-charts-types/src/chart/contextMenuOptions.ts

Let's discuss on JIRA & video-call

olegat: We've decided on going with alternative 1

alantreadway (approved):
LGTM

PR #3932, opened 2025-04-03
AG-14622: Fix grouped-category labels when only one datum is provided

iMoses:
https://ag-grid.atlassian.net/browse/AG-14622


alantreadway (approved):
LGTM

PR #3940, opened 2025-04-03
Ag 13031 example controls prototype

markdurrant:
- Improve examples controls styling 
- Move examples controls styling to single CSS
- Remove `layout` & `extra-styles` options from examples 
- Create example controls tester/demo page


packages/ag-charts-website/src/components/example-runner/components/ExampleRunner.tsx
```diff
@@ -9,8 +9,6 @@ import { type FunctionComponent, type ReactElement, useState } from 'react';
 import { CodeViewer } from './CodeViewer';
 import { ExampleIFrame } from './ExampleIFrame';
 import styles from './ExampleRunner.module.scss';
-// Charts specific example runner styles
-import chartsStyles from './LegacyExampleRunner.module.scss';
```
taktran, line 13: ✨

packages/ag-charts-website/src/content/docs/examples-controls-test/index.mdoc
```diff
@@ -0,0 +1,58 @@
...
+
+```html
+<button onclick="myFunction()">Apply <code>'{ myValue: 30 }'</code></button>
+```
+
+### Additional styles
+
+You should not need to apply any additional styles in an example’s `styles.css` file in order to style your example controls. CSS in `styles.css` should be reserved for code that is relevant to the function of the example.
+
+If you need to use an input or element that’s not covered below, please speak to the website team and we implement style new controls.
```
taktran:
```suggestion
If you need to use an input or element that’s not covered below, please speak to the website team and we can style new controls.
```

taktran (approved):
Looks really good 😍 

Just one issue with the react `examples-controls-test` page, where the checked checkbox and radio, aren't checked. Might not be trivial to fix for react though, so might be worth another ticket.

Screenshot 2025-04-07 at 10 47 18 am

packages/ag-charts-website/src/content/docs/touch/_examples/long-tap/index.html
```diff
@@ -1,5 +1,3 @@
-<div class="toolbar"></div>
```
taktran, line 1: Deleting this causes the `plugin.js:385 TypeScript [Error] JSX expressions must have one parent element. (TS2657)` error because previously it was relying on this to create the `<Fragment>` wrapper.

We can either

* Add `.gesture-demo` to the example generator to check for
* Add an empty `example-controls` div here
* Fix the example generator to be smarter

PR #3943, opened 2025-04-04
AG-14623 - Fix handling of plain-text overlay renderer result.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-14623

### DIFF
--- a/packages/ag-charts-community/src/chart/overlay/overlay.ts
+++ b/packages/ag-charts-community/src/chart/overlay/overlay.ts
@@ -44,7 +44,12 @@ export class Overlay extends BaseProperties {
             } else {
                 const tempDiv = createElement('div');
                 tempDiv.innerHTML = htmlContent;
-                this.content = tempDiv.firstElementChild! as HTMLElement;
+                const { firstElementChild } = tempDiv;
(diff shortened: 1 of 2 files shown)


packages/ag-charts-community/src/chart/overlay/overlay.ts
```diff
@@ -44,7 +44,12 @@ export class Overlay extends BaseProperties {
             } else {
                 const tempDiv = createElement('div');
                 tempDiv.innerHTML = htmlContent;
-                this.content = tempDiv.firstElementChild! as HTMLElement;
+                const { firstElementChild } = tempDiv;
+                if (firstElementChild instanceof HTMLElement && tempDiv.childElementCount === 1) {
+                    this.content = firstElementChild;
+                } else {
+                    this.content = tempDiv;
```
jacobp100, line 51: This might make it harder to predict what would happen - we could just log a warning instead that only the first element will get used?

jacobp100 (approved):

packages/ag-charts-community/src/chart/overlay/overlay.ts
alantreadway: The renderer() functions are for more advanced use-cases, so I'm happy that there is some risk that users don't have guard-rails here.

PR #3950, opened 2025-04-04
AG-13024 Add `context` to callback for gauges

olegat:


packages/ag-charts-core/src/utils/validation.ts
alantreadway, line 1: @iMoses could you review these changes?

packages/ag-charts-core/src/utils/validation.ts
```diff
@@ -254,6 +259,10 @@ export function undocumented<T extends Validator | OptionsDefs<any>>(validatorOr
     ) as T;
 }
 
+export const contextPropertyValidator: Validator = () => true;
+contextPropertyValidator[undocumentedSymbol] = true;
+contextPropertyValidator[alwaysvalidSymbol] = true;
+
```
iMoses: Why do we need a special case just for the context?
What is currently missing from the existing implementation?
I'd rather we keep this module self-contained and decoupled, so if we can we should provide the missing feature in a generic manner IMHO

olegat: > Why do we need a special case just for the context?
> What is currently missing from the existing implementation?

The objective is ensure that the AG Charts library doesn't care about the property's value. However, the reoccurring challenge seems to that AG Charts tries to "process" the `context` property, which is an obstacle. I'm opening to hearing about better suggestions

> I'd rather we keep this module self-contained and decoupled, so if we can we should provide the missing feature in a generic manner IMHO

Agreed. The `contextPropertyValidator` variable should be moved to another module. I didn't pay attention to that, well spotted!

olegat: This change was removed. See JIRA ticket for more info.

alantreadway (approved):
LGTM

PR #3952, opened 2025-04-07
AG-14649: Prevent hidden options from being suggested in validations

iMoses:
https://ag-grid.atlassian.net/browse/AG-14649


alantreadway (approved):
LGTM

PR #3958, opened 2025-04-08
AG-14647 Run `dataSource.getData()` on `charts.update()` call

olegat:
https://ag-grid.atlassian.net/browse/AG-14647

### DIFF
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -520,6 +520,7 @@ export abstract class Chart extends Observable implements ModuleInstance {
 
     requestFactoryUpdate(cb: (chart: Chart) => Promise<void> | void) {
         if (this.destroyed) return;
+        this.apiUpdate = true;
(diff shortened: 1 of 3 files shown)


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -520,6 +520,7 @@ export abstract class Chart extends Observable implements ModuleInstance {
 
     requestFactoryUpdate(cb: (chart: Chart) => Promise<void> | void) {
         if (this.destroyed) return;
+        this.apiUpdate = true;
```
alantreadway: We should probably add this to `UpdateOpts`, which track the requirements/parameters of update cycles, rather than trying to track distinctly - there maybe a delay between this update being made and `this.updateMutex.acquire()` being called back.

Practically, this would mean the `Chart.applyOptions()` call to `this.update()` would then set this flag at a point much closer to the actual execution stage, which will likely have less side-effects.

alantreadway (approved):
LGTM

PR #3967, opened 2025-04-08
AG-8876 Implement deprecated context menu option adapter

olegat:


packages/ag-charts-community/src/chart/legend/legend.ts
```diff
@@ -291,21 +288,10 @@ export class Legend extends BaseProperties {
...
-                id: ID_LEGEND_OTHER_SERIES,
-                type: 'legend-item',
-                label: 'contextMenuToggleOtherSeries',
-                action: (params) => this.contextToggleOtherSeries(params),
-            }),
-            ctx.legendManager.addListener('legend-change', this.onLegendDataChange.bind(this))
-        );
+        const { items } = ctx.contextMenuRegistry.builtins;
+        items['toggle-series-visibility'].action = (params) => this.contextToggleVisibility(params);
+        items['toggle-other-series'].action = (params) => this.contextToggleOtherSeries(params);
```
alantreadway, line 293: We probably need to handle removal of these in the the `destroy()` case?

Practically this would happen when users toggle the legend on off (see Integrated charts).

alantreadway (commented):
Generally LGTM - @lsjroberts could you also take a look?

packages/ag-charts-community/src/chart/legend/legend.ts
olegat: b41257f8bfd6d98a3dfc62269e7ef314411cbcd5

lsjroberts (approved):
Looks great, nice one.

Only minor comment is that while the builtins are definitely simpler now, they are possibly less "correct", e.g. the zoom items being defined in context menu module rather than in the zoom module (though with the item actions being added in the zoom module is good).

olegat:
> Only minor comment is that while the builtins are definitely simpler now, they are possibly less "correct", e.g. the zoom items being defined in context menu module rather than in the zoom module (though with the item actions being added in the zoom module is good).

Yes, that's true. It's unfortunate consequence that the built-ins are no longer defined by the corresponding module. 

However, trade-off is that simplicity allows us to perform better compile-time checks. For example, if I add a new string to the `AgContextMenuItemLiteral` union, then the Typescript compiler will immediately raise an error that this property is missing in `ContextMenuBuiltinItems`.

I feel like this is an acceptable trade-off.

alantreadway (approved):
LGTM

PR #3982, opened 2025-04-14
AG-NONE Add `isNever` to enforce exhaustive switch statements

olegat:


packages/ag-charts-community/src/scene/bbox.ts
```diff
@@ -210,6 +210,9 @@ export class BBox implements BBoxValues, BBoxContainsTester, DistantObject, Inte
                 this.width -= value * 2;
                 this.height -= value * 2;
                 break;
+
+            default:
+                position satisfies never;
```
olegat, line 215: > [!IMPORTANT]
> All uses of `isNever(x)` could essentially be replaced with `x satisfies never` to achieve the same result.
>
> The advantage of using the `isNever(x)` function is that it throws an error if called at runtime, which is a useful debugging tool. However, the downside is that the `isNever(x)` function call leaves behind a small footprint in the final library.
>
> We may want to consider adding some post-processing to remove the debugging footprint of `isNever(x)` when bundling the library for production.

olegat:
Diverged too far from `latest`. There are many merge conflicts now. Closing.

PR #3985, opened 2025-04-14
AG-8876 Add `MenuWidget` class and use it in `contextMenu.ts`

olegat:

This refactoring is necessary for good interaction management (a11y, keynav, focus, etc..) of submenus (which is currently unimplemented).

### DIFF
--- a/packages/ag-charts-community/src/widget/widgetListenerInternal.ts
+++ b/packages/ag-charts-community/src/widget/widgetListenerInternal.ts
@@ -119,6 +119,8 @@ export class WidgetListenerInternal {
                 this.registerDragTrigger(target);
                 break;
             }
+            default:
+                (type) satisfies 'close-widget';
         }
     }
 
--- a/packages/ag-charts-community/src/widget/widgetListenerInternal.ts
+++ b/packages/ag-charts-community/src/widget/widgetListenerInternal.ts
@@ -119,8 +119,6 @@ export class WidgetListenerInternal {
                 this.registerDragTrigger(target);
                 break;
             }
-            default:
-                (type) satisfies 'close-widget';
         }
     }
 
--- a/packages/ag-charts-community/src/widget/widgetListenerInternal.ts
+++ b/packages/ag-charts-community/src/widget/widgetListenerInternal.ts
@@ -103,7 +103,8 @@ export class WidgetListenerInternal {
         this.listeners ??= new Map();
         let result: Set<EventHandler<T, K>> | undefined = this.listeners.get(type);
         if (result === undefined) {
-            this.listeners.set(type, (result = new Set()));
+            result = new Set();
+            this.listeners.set(type, result);
         }
         return result;
     }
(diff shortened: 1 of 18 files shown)


packages/ag-charts-community/src/widget/widgetListenerInternal.ts
```diff
@@ -94,52 +94,39 @@ export class WidgetListenerInternal {
...
+            case 'drag-start':
+            case 'drag-move':
             case 'drag-end': {
-                this.dragEndListeners ??= [];
-                this.dragEndListeners.push(handler);
                 this.registerDragTrigger(target);
                 break;
             }
+            default:
+                (type) satisfies 'close-widget';
```
jacobp100: Might be worth looking into https://typescript-eslint.io/rules/switch-exhaustiveness-check/

jacobp100: This might be more readable anyway

```suggestion
            case 'close-widget':
                break;
            default:
                (type) satisfies never;
```

packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts
```diff
@@ -315,23 +267,18 @@ export class ContextMenu extends _ModuleSupport.BaseModuleInstance implements _M
...
-        el.classList.add(`${DEFAULT_CONTEXT_MENU_CLASS}__item`);
-        el.classList.toggle(DEFAULT_CONTEXT_MENU_DARK_CLASS, this.darkTheme);
-        el.ariaDisabled = disabled.toString();
-        el.textContent = this.ctx.localeManager.t(label);
-        el.role = 'menuitem';
-        if (callback != null) {
-            el.onclick = makeAccessibleClickListener(el, this.createButtonOnClick(showOn, callback));
+    private createButtonElement(item: ContextMenuItem): _Widget.ButtonWidget {
+        const button = new _Widget.ButtonWidget();
+        button.addClass(`${DEFAULT_CONTEXT_MENU_CLASS}__item`);
```
jacobp100, line 272: Could we move this to its own variable with the whole classname written out. At one point, I want to add some automation to check whether CSS classes exist in JS and vice versa - but that's not possible with interpolatiom

jacobp100 (approved):

packages/ag-charts-enterprise/src/features/context-menu/contextMenu.ts
olegat: Good idea. I think that's outside of the scope of this PR though, because the removed code used the same `${DEFAULT_CONTEXT_MENU_CLASS}__item` interpolation.

packages/ag-charts-community/src/widget/widgetListenerInternal.ts
olegat: Well spotted! I've got a separate PR for this https://github.com/ag-grid/ag-charts/pull/3982, so I'll just remove this check for now.

PR #3986, opened 2025-04-14
Ag 14557/custom svg patterns

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-14557


packages/ag-charts-community/src/scene/pattern/pattern.ts
```diff
@@ -34,14 +36,23 @@ export class Pattern implements Omit<RequiredInternalAgPatternColor, 'type'> {
...
         this.path = patternOptions.path;
     }
 
     private getPath(pixelRatio: number) {
         const { pattern, width, height, padding, strokeWidth, path: svgPath } = this;
 
         const path = new ExtendedPath2D();
-        PATTERNS[pattern](path, { width, height, pixelRatio, strokeWidth, padding, svgPath });
+
+        const svgParts = parseSvg(svgPath);
```
alantreadway, line 48: Just wondering if we should be doing this as an actual validation, which can fail much earlier than here, and give users more specific feedback (e.g. if they are using multiple SVGs)?

alantreadway (approved):
LGTM

PR #3987, opened 2025-04-15
Improve unit time scale for high performance cases

jacobp100:


jacobp100:
Moved to https://github.com/ag-grid/ag-charts/pull/3990

PR #3989, opened 2025-04-15
AG-14722 - Update AG Grid to v33

taktran:
* Updated in changelog and pipeline
* Remove grid darkmode handling, as it's handled by AG Grid now
* Add darkmode handling for grids


markdurrant (approved):
All LGTM 👍

PR #3991, opened 2025-04-15
AG-14724: When a heatmap colorKey value is null or undefined, no colo…

iMoses:
…ur should be shown
https://ag-grid.atlassian.net/browse/AG-14724


alantreadway (approved):
LGTM

PR #3993, opened 2025-04-16
AG-14724: When a heatmap colorKey value is invalid, don't show warning

iMoses:
https://ag-grid.atlassian.net/browse/AG-14724


alantreadway (approved):
LGTM

PR #4009, opened 2025-04-22
AG-14581: Validate preset options

iMoses:


packages/ag-charts-community/src/api/preset/gauge.ts
```diff
@@ -271,48 +261,23 @@ export function gauge(
...
     presetTheme: AgRadialGaugeThemeOverrides | AgLinearGaugeThemeOverrides | undefined
 ): AgGaugeChartOptions {
-    if (isRadialGauge(opts)) {
-        const radialGaugeOpts = applyThemeDefaults(opts, presetTheme as any);
-        return radialGaugeOptions(radialGaugeOpts);
-    } else if (isLinearGauge(opts)) {
-        const linearGaugeOpts = applyThemeDefaults(opts, presetTheme as any);
-        return linearGaugeOptions(linearGaugeOpts);
+    // PATCH for backwards compatibility - remove in v12.x.x
+    if (opts.listeners) {
```
jacobp100, line 265: Will this mutate the user options?

packages/ag-charts-enterprise/src/series/linear-gauge/linearGaugeSeriesProperties.ts
```diff
@@ -254,7 +247,7 @@ class LinearGaugeScaleProperties extends BaseProperties {
     readonly label = new LinearGaugeScaleLabelProperties();
 }
 
-export class LinearGaugeLabelProperties extends AutoSizedLabel<AgLinearGaugeLabelFormatterParams> {
+export class LinearGaugeLabelProperties extends AutoSizedLabel<unknown> {
```
jacobp100, line 250: Why unknown?

jacobp100 (approved):

packages/ag-charts-enterprise/src/series/linear-gauge/linearGaugeSeriesProperties.ts
iMoses: `AgLinearGaugeLabelFormatterParams` was an empty interface, so it's equivalent. (it eventually intersects with `TParams`, so it shouldn't be `any` or `never`)

packages/ag-charts-community/src/api/preset/gauge.ts
iMoses: Could be... I'll fix it

PR #4011, opened 2025-04-22
AG-8876 Implement submenus in context-menu

olegat:

Depends on
- https://github.com/ag-grid/ag-charts/pull/3985

### DIFF
--- a/packages/ag-charts-community/src/widget/menuWidget.ts
+++ b/packages/ag-charts-community/src/widget/menuWidget.ts
@@ -45,6 +45,7 @@ export class MenuWidget extends RovingTabContainerWidget {
         const subMenu = new MenuWidget();
         subMenuButton.setAriaHasPopup('menu');
         subMenuButton.addListener('click', (ev) => subMenu.open(ev));
+        subMenuButton.addListener('mouseenter', (ev) => subMenu.open(ev));
         this.addChild(subMenuButton);
         return { subMenuButton, subMenu };
     }
@@ -72,6 +73,7 @@ export class MenuWidget extends RovingTabContainerWidget {
         }
 
         this.children[0]?.focus({ preventScroll: true });
+        this.internalListener?.dispatch('open-widget', this, { type: 'open-widget' });
     }
 
     private selfClose() {
--- a/packages/ag-charts-community/src/widget/menuWidget.ts
+++ b/packages/ag-charts-community/src/widget/menuWidget.ts
@@ -16,12 +16,19 @@ import type { WidgetEvent } from './widgetEvents';
 
 type OpenScope = {
     lastFocus: HTMLElement | undefined;
+    openSubMenu: MenuWidget | undefined;
     removers: DestroyFns;
     abort: () => void;
     close: () => void;
 };
 
-enum CloseEnum { CLOSE = '0', ABORT = '1', DESTROY = '2' };
+enum CloseEnum {
+    CLOSE = '0',
+    ABORT = '1',
+    DESTROY = '2',
+    PARENT_CLOSED = '3',
+    SIDLING_OPENED = '4',
+}
 
 export class MenuWidget extends RovingTabContainerWidget {
     private openScope?: OpenScope;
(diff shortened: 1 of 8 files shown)


packages/ag-charts-community/src/widget/menuWidget.ts
```diff
@@ -4,24 +4,32 @@ import type { Direction } from 'ag-charts-types';
...
 type OpenScope = {
     lastFocus: HTMLElement | undefined;
-    lastFocusAborted: boolean;
+    openSubMenu: MenuWidget | undefined;
     removers: DestroyFns;
     abort: () => void;
     close: () => void;
 };
 
+enum CloseEnum {
```
jacobp100: Minor - `CloseMode` or something might be a better name

packages/ag-charts-community/src/widget/menuWidget.ts
```diff
@@ -4,24 +4,32 @@ import type { Direction } from 'ag-charts-types';
...
+    CLOSE = '0',
```
jacobp100, line 26: Are these constants needed? It'll default to integers of the same values

packages/ag-charts-enterprise/src/features/context-menu/contextMenuItem.ts
```diff
@@ -54,6 +61,28 @@ export function expandBuiltin(
...
+        } else {
+            const menuItem = appendItem(showing, item, result);
+            if (item.items && menuItem && menuItem.type === 'submenu') {
+                expandItems(showing, registry, item.items, menuItem.items);
+            }
+        }
+    }
+    // remove trailing 'separator' menu item
+    if (result[result.length - 1]?.type === 'separator') {
+        result.length = result.length - 1;
```
jacobp100:
```suggestion
        result.pop()
```

A bit more idiomatic

packages/ag-charts-enterprise/src/features/context-menu/contextMenuStyles.css
```diff
@@ -19,9 +12,16 @@
 }
 
 .ag-charts-context-menu__menu {
+    background: var(--ag-charts-chrome-background-color);
```
jacobp100, line 15: I'm not following why the background etc. is able to move?

packages/ag-charts-community/src/widget/menuWidget.ts
olegat: Integer enums are not type safe.

```typescript
enum CloseEnum { Close = 1, Abort = 2 };
const myVar1: CloseEnum = CloseEnum.Close + 1000;
const myVar2: CloseEnum = CloseEnum.Close | CloseEnum.Abort;
```

The snippet above is valid Typescript code despite the fact that `myVar1` & `myVar2` are neither 1 nor 2.

packages/ag-charts-enterprise/src/features/context-menu/contextMenuStyles.css
olegat: I do not understand the question. But I can say that, without this change, the styling in is not properly applied to submenus, which are also of class `.ag-charts-context-menu__menu` like the root menu.


https://github.com/user-attachments/assets/72651b14-0bd6-43a9-bf7b-fddd2e59845f

Perhaps some CSS properties like `white-space` and `transition` can stay as part of `.ag-charts-context-menu`; in all honesty I do not completely understand what all the properties in the file are intended for.

packages/ag-charts-community/src/widget/menuWidget.ts
olegat: No objections. Done.

packages/ag-charts-enterprise/src/features/context-menu/contextMenuStyles.css
alantreadway: Let's roll with these changes and see if we can spot any regressions in testing?

alantreadway (approved):
LGTM

PR #4017, opened 2025-04-23
AG-8876 `contextMenu.items` cycle detection

olegat:

Depends:
- https://github.com/ag-grid/ag-charts/pull/4011

### DIFF
--- a/packages/ag-charts-community/src/util/json.ts
+++ b/packages/ag-charts-community/src/util/json.ts
@@ -19,7 +19,7 @@ import { SKIP_JS_BUILTINS, getPath, mergeDefaults, without } from './object';
 import { isProperties } from './properties';
 
 type StringSet = { has(value: string): boolean };
-export type CloneOptions = { shallow?: StringSet; assign?: StringSet; seen?: Set<unknown> };
+export type CloneOptions = { shallow?: StringSet; assign?: StringSet; seen?: unknown[] };
 
 const CLASS_INSTANCE_TYPE = 'class-instance';
 
@@ -119,13 +119,14 @@ function cloneArray<T>(source: T[], opts?: CloneOptions): T[] {
     const result: T[] = [];
     result.length = source.length;
     let i = 0;
+    const parentSeen = opts?.seen;
     for (const item of source) {
-        if (typeof item === 'object' && opts?.seen?.has(item)) {
+        if (typeof item === 'object' && parentSeen?.includes(item)) {
             Logger.warn('cycle detected in array', item);
             continue;
         }
-        opts?.seen?.add(item);
-        result[i++] = deepClone(item, opts);
+        const seen = parentSeen ? [...parentSeen, item] : undefined;
+        result[i++] = deepClone(item, { ...opts, seen });
     }
     result.length = i;
     return result;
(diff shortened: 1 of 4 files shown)


packages/ag-charts-community/src/util/json.ts
olegat, line 1: I'm not sure if this is a best place to implement the cycle detection. I feel like writing this cycle detection as part of the validation would help us to ensure that error message reporting is consistent.

Thoughts?

packages/ag-charts-community/src/util/json.ts
```diff
@@ -115,6 +115,24 @@ export function deepClone<T>(source: T, opts?: CloneOptions): T {
     return shallowClone(source);
 }
 
+function cloneArray<T>(source: T[], opts?: CloneOptions): T[] {
+    const result: T[] = [];
+    result.length = source.length;
```
alantreadway: Nit: Skip this, it can actually harmful to performance to attempt to pre-allocate since the array gets initialised with `undefined` values, which de-optimises the array (JS will use optimised array types if all elements are of the same type, and `undefined` values with `object` values will give you an unoptimised array implementation).

packages/ag-charts-community/src/util/json.ts
alantreadway: I'm not sure we need this additional sanity check, but if it improves defensiveness without much additional cost then it's probably a good idea.

With that in mind, this is a performance sensitive part of the code base, and there are some learnings that we should apply here to avoid unexpected pitfalls - I've added nits for these.

packages/ag-charts-community/src/util/json.ts
```diff
@@ -115,6 +115,24 @@ export function deepClone<T>(source: T, opts?: CloneOptions): T {
...
+    let i = 0;
+    const seen = opts?.seen;
+    for (const item of source) {
+        if (typeof item === 'object' && seen?.includes(item)) {
+            Logger.warn('cycle detected in array', item);
+            continue;
+        }
+        seen?.push(item);
+        result[i++] = deepClone(item, { ...opts, seen });
```
alantreadway: Nit: It would be preferable to avoid the instantiation of an extra `object` for the `opts` here - can we instead do something like this at the top of this function:
```
opts ??= {};
opts.seen ??= new Set();
```

...and then just pass `opts` down to recursive calls without needing further allocations?

olegat: Well spotted! Actually I think that the `{ ...opt, seen }` is not longer required, it was initially required because the first implementation cloned `seen` rather than pushing & popping. See https://github.com/ag-grid/ag-charts/commit/466d2f51ad76fd62dbf00946d63622d2c8ed269a

packages/ag-charts-community/src/util/json.ts
olegat: Agreed.

One advantage of moving cycle-detection to be part of validation is that we could restrict it to just check `contextMenu.items[]`.

packages/ag-charts-community/src/util/json.ts
olegat: Done. `{ ...opts, seen }` replaced with `opts`

packages/ag-charts-community/src/util/json.ts
olegat: Done.

alantreadway (approved):
LGTM

PR #4029, opened 2025-04-28
Improve performance for continuous ticks

jacobp100:
For some zoomed-in cases, maxTickCount would be 6,000. The continuous mode will reduce these ticks one-by-one until it finds a suitable tick step. Instead, we can use binary search to find this value faster


alantreadway (approved):
LGTM

PR #4031, opened 2025-04-28
AG-8876 submenu a11y improvements

olegat:

Depends on:
- https://github.com/ag-grid/ag-charts/pull/4037

### DIFF
--- a/packages/ag-charts-core/src/utils/id.ts
+++ b/packages/ag-charts-core/src/utils/id.ts
@@ -1,7 +1,12 @@
 const ID_MAP = new Map<string, number>();
+const elementIDBrand = Symbol('ElementID');
+let nextElementID: number = 1;
+
+export type ElementID = string & { readonly [elementIDBrand]: true };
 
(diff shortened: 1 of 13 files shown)


packages/ag-charts-core/src/utils/id.ts
```diff
@@ -1,7 +1,12 @@
 const ID_MAP = new Map<string, number>();
+const elementIDBrand = Symbol('ElementID');
+let nextElementID: number = 1;
+
+export type ElementID = string & { readonly [elementIDBrand]: true };
```
alantreadway, line 7: Nit: Maybe worth a comment to explain this construction is to allow enforcement of `ElementID` types distinct from `string`?

olegat: Done https://github.com/ag-grid/ag-charts/pull/4031/commits/285ebf92b2da0c9bfd17fc21817335abf5278382

alantreadway (approved):
LGTM

PR #4041, opened 2025-04-29
AG-14664 Reset zoom in all axes of same direction when double clickin…

jacobp100:
…g axis

https://ag-grid.atlassian.net/browse/AG-14664


alantreadway (approved):
LGTM

PR #4043, opened 2025-04-29
AG-14734 - Improve example styles

taktran:
# Changes

* Move example styles to a separate file
* Don't show `index.html` in the example runner
* Only add e2e check script for example in new tab
* Move body padding code to be a flag for framework template


markdurrant (approved):
Looks great to me 👍

PR #4048, opened 2025-04-29
AG-8876 Leftside submenus pop out (when out of room on right side)

olegat:


packages/ag-charts-community/src/dom/theme.css
```diff
@@ -26,6 +26,7 @@
 
     --ag-charts-input-background-color: #fff;
     --ag-charts-input-text-color: #464646;
+    --ag-charts-input-text-color-dimmed: rgba(70, 70, 70, 0.5);
```
olegat: > [!NOTE]
> This is the same colour as `--ag-charts-input-text-color`, but with the alpha channel set to 50%.
>
> Do we know whether the `--ag-charts-input-text-color` has a different value in dark mode?
>
> If so, then the `--ag-charts-input-text-color-dimmed` colour will also need to be changed in dark mode to reflect this.

packages/ag-charts-website/e2e/context-menu.spec.ts-snapshots/zoom-contextmenu-chromium-linux.png
olegat, line 1: The final font colour of the disabled "Pan to here" has subtly changed. It's so hard to notice that I think it's acceptable.

packages/ag-charts-community/src/dom/theme.css
alantreadway: Generally we've been using color mixing to achieve this sort of effect, both in other CSS cases and in the new computed (JS) themes - maybe just run this past @markdurrant and/or @lsjroberts?

lsjroberts (approved):
LGTM

packages/ag-charts-community/src/dom/theme.css
olegat: Discussed with @markdurrant and @lsjroberts on Slack. The `color-mix` achieves the same desired result. Therefore, we'll go with that approach because it the closest thing to the existing `opacity: 0.5` in contextMenuStyle.css. Pending CI checks.

markdurrant (approved):
CSS stuff LGTM 👍

alantreadway (approved):
LGTM

PR #4055, opened 2025-04-30
AG-11249: Allow parent level column label rotation on the group-category axis

iMoses:
https://ag-grid.atlassian.net/browse/AG-11249


alantreadway (approved):
LGTM

PR #4059, opened 2025-04-30
AG-13561 Secondary axis improvements

jacobp100:
- Removes 40 as a nice interval for secondary axes
    - This was done to accommodate zooming as if you had a step of 5 on the primary axis and a step of 100 on the secondary, when you zoom the primary axis to a step of 2, you'd need a step of 40 on the secondary axis
    - We now generate nice secondary ticks at the zoomed out axis, and apply the zoom which can result in the step of 40
    - We no longer need an explicit 40 step
- If it's possible to make the domain a multiple of the step, it will do so


alantreadway (approved):
LGTM

packages/ag-charts-website/e2e/gallery-examples.spec.ts-snapshots/gallery-grouped-bar-line-combination-chromium-linux.png
jacobp100, line 1: The blue 'employees' bars are in their own axis for some reason, so this is expected

PR #4065, opened 2025-05-01
AG-14799 - Add validator support for type union defaulting + fix Sparklines axis.type defaulting.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-14799


packages/ag-charts-core/src/utils/validation.ts
```diff
@@ -321,8 +327,17 @@ export const optionsDefs = <T>(defs: OptionsDefs<T>, description = 'an object'):
         return { valid, cleared: result.cleared, invalid: result.invalid };
     }, description);
 
-export const typeUnion = <T extends { type: string }>(defs: TypeUnionDefs<T, T['type']>, description = 'an object') =>
-    ({ ...defs, [descriptionSymbol]: description, [unionSymbol]: true }) as OptionsDefs<T>;
+export const typeUnion = <T extends { type: string }>(
+    defs: TypeUnionDefs<T, T['type']>,
+    defaultType?: T['type'],
+    description = 'an object'
```
iMoses: Although description is optional we do enforce setting it everywhere, so I'd argue it shouldn't be optional thus their order should be reversed.

alantreadway: Ok, I'll take a look

iMoses (approved):
LGTM

PR #4072, opened 2025-05-02
AG-14809 Hide "Toggle Other Series" on pies, donuts, pyramids and funnels

olegat:
https://ag-grid.atlassian.net/browse/AG-14809

### DIFF
--- a/packages/ag-charts-community/src/chart/legend/legend.ts
+++ b/packages/ag-charts-community/src/chart/legend/legend.ts
@@ -869,6 +869,8 @@ export class Legend extends BaseProperties {
             this.ctx.contextMenuRegistry.builtins.items['toggle-series-visibility'].enable = true;
         }
 
+        this.ctx.contextMenuRegistry.setVisible('toggle-other-series', this.ctx.chartService.series.length > 1);
+
         const { offsetX, offsetY } = sourceEvent;
         const { x: canvasX, y: canvasY } = Transformable.toCanvasPoint(node, offsetX, offsetY);
         this.ctx.contextMenuRegistry.dispatchContext('legend-item', { widgetEvent, canvasX, canvasY }, { legendItem });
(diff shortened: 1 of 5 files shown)


packages/ag-charts-community/src/chart/legend/legendDatum.ts
```diff
@@ -41,6 +41,7 @@ export interface CategoryLegendDatum extends BaseChartLegendDatum {
     };
     skipAnimations?: boolean;
     isFixed?: boolean;
+    hideToggleOtherSeries?: true; // used to hide "Toggle Other Series" for Multi-Donut Charts.
```
olegat: See JIRA ticket for further explanation.

packages/ag-charts-community/src/chart/legend/legend.ts
```diff
@@ -869,6 +869,11 @@ export class Legend extends BaseProperties {
             this.ctx.contextMenuRegistry.builtins.items['toggle-series-visibility'].enable = true;
         }
 
+        const toggleOtherSeriesVisible =
+            this.ctx.chartService.series.length > 1 &&
+            this.ctx.chartService.series[0]?.getLegendData('category')[0]?.hideToggleOtherSeries !== true;
```
alantreadway, line 874: Donuts and pies can be used in combination too, so:
- should `pieSeries.ts` also have this behaviour?
- should the check here use a `.some()` predicate to be slightly more robust in multi-series cases?

olegat: * `pieSeries.ts` should also set `hideToggleOtherSeries: true` to fix the pie/donut combo case.
* we could use `.some()`, but I'm not sure if we gain much from that. `hideToggleOtherSeries` is always either `true` for all series or `false` for all series. `legend.ts` does not have a mechanism to hide/show the `"Toggle Other Series"` menuitem on a legend-item basis; it's either always shown or always hidden.

olegat: Perhaps instead of using `.some` we could add a method `Series.hidesToggleOtherSeries(): boolean` (returning `false`) which `PieSeries` and `DonutSeries` can override. LMK.

alantreadway: Ok thanks - lets roll with this, I don't see any other series combinations being an issue in this respect.

alantreadway (approved):
LGTM

PR #4076, opened 2025-05-02
AG-13979 - Fix duplicate sync listeners + disabled tooltip handling.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-13979


packages/ag-charts-enterprise/src/features/sync/chartSync.ts
alantreadway, line 1: The actual fixes are in this file, everything else is supporting these or improving tests.

iMoses (approved):

PR #4083, opened 2025-05-02
AG-14822 - Fix zoom toolbar never being hidden due to incorrect CSS selector.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-14822


lsjroberts (approved):
LGTM

PR #4088, opened 2025-05-06
AG-14800 Add time unit shorthands for time intervals

jacobp100:
https://ag-grid.atlassian.net/browse/AG-14800

- Adds a new utility file for times, which will eventually replace the time util file


packages/ag-charts-website/src/middleware/index.ts
```diff
@@ -1,5 +1,8 @@
+import { getIsDev } from '@utils/env';
```
jacobp100, line 1: This is kind of irrelevant but I was getting Astro errors but while looking at it, this came out as a way to have plunker examples be formatted in dev

alantreadway (approved):
LGTM

PR #4090, opened 2025-05-06
AG-8876 TC14 Infer `type: 'submenu'` from items array length

olegat:


alantreadway (approved):
LGTM

PR #4094, opened 2025-05-06
Ag-14759: Legend toggle state is reset when theme is updated at runtime

iMoses:
https://ag-grid.atlassian.net/browse/AG-14759


alantreadway (approved):
LGTM

PR #4100, opened 2025-05-07
AG-14273 Add google font loading and font family arrays

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14273


packages/ag-charts-community/src/chart/fonts/fontManager.ts
```diff
@@ -0,0 +1,52 @@
...
+        this.domManager.addChild('canvas-container', `font-check-${encodeURIComponent(font)}`, fontCheckElement);
+
+        // Observe changes to the element size as a proxy for the font loading
+        const fontCheckObserver = new ResizeObserver((entries) => {
+            const width = entries?.at(0)?.contentBoxSize.at(0)?.inlineSize;
+            if (width != null && width > 0) {
+                this.updateService.update(ChartUpdateType.PERFORM_LAYOUT);
+            }
+        });
+        fontCheckObserver.observe(fontCheckElement);
```
lsjroberts, line 59: The element and observer are not destroyed because there may be multiple fonts in the string, some of which may be available locally and others need to be loaded. So the observer will trigger multiple times.

I'll do some more extensive testing to see if this can be checked and destroyed after either a known number of times or (not preferable) a timeout.

alantreadway: Could we at least ensure that things are completed cleaned up on `chart.destroy()` to avoid a memory leak?

lsjroberts: Added `destroy()` for the observers, elements are destroyed by the `DOMManager`.

alantreadway (approved):
LGTM

PR #4114, opened 2025-05-08
AG-14273 Fix font swapping layout measuring in captions and legend

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14273


lsjroberts:
Note: not targetting latest, so check this build - https://github.com/ag-grid/ag-charts/actions/runs/14903553779

PR #4117, opened 2025-05-08
git subrepo pull external/ag-website-shared

markdurrant:
- Users can see the buy now buttons whenever the comparison table is visible


taktran (approved):
LGTM 👍🏼

PR #4152, opened 2025-05-12
CRT-801 Set `role="presentation"` on context menu item labels

olegat:
https://ag-grid.atlassian.net/browse/CRT-801


alantreadway (approved):
LGTM

PR #4157, opened 2025-05-12
CRT-778 Change shared tooltip to hide rows with no value

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-778

### DIFF
--- a/packages/ag-charts-community/src/chart/tooltip/tooltipContent.ts
+++ b/packages/ag-charts-community/src/chart/tooltip/tooltipContent.ts
@@ -96,6 +96,10 @@ function dataHtml(label: string | undefined, value: string, inline: boolean) {
 function tooltipRowContentHtml(content: GroupedStructuredContent['items'][0]) {
     let html = '';
 
+    if (content.data?.length && content.data.every((datum) => datum.value == null || datum.value === '')) {
+        return html;
+    }
+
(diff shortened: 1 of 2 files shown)


packages/ag-charts-community/src/chart/tooltip/tooltipContent.ts
```diff
@@ -96,6 +96,10 @@ function dataHtml(label: string | undefined, value: string, inline: boolean) {
 function tooltipRowContentHtml(content: GroupedStructuredContent['items'][0]) {
     let html = '';
 
+    if (content.data?.length && content.data.every((datum) => !datum.value)) {
```
alantreadway: Maybe to avoid `0` not being rendered?
```suggestion
    if (content.data?.length && content.data.every((datum) => datum.value != null)) {
```

lsjroberts: Updated to `datum.value != null && datum.value !== ''` to prevent empty strings but allow if a `0` sneaks in.

lsjroberts: That logic was inverted 🤦‍♂️, fixed.

alantreadway (approved):
LGTM

PR #4189, opened 2025-05-14
Refactor TimeInterval instances to object format

jacobp100:
https://ag-grid.atlassian.net/browse/AG-14829


packages/ag-charts-community/src/util/time/index.ts
alantreadway, line 1: We discussed keeping these symbols for backwards compatibility, has that changed?

alantreadway (approved):
LGTM

PR #4205, opened 2025-05-15
AG-14263 Fix `seriesId` param of `errorBar.itemStyler`

olegat:
https://ag-grid.atlassian.net/browse/AG-14263

### DIFF
--- a/packages/ag-charts-enterprise/src/features/error-bar/errorBarNode.ts
+++ b/packages/ag-charts-enterprise/src/features/error-bar/errorBarNode.ts
@@ -88,7 +88,7 @@ export class ErrorBarNode extends _ModuleSupport.Group {
         return {
             ...(style as Required<AgErrorBarThemeableOptions>),
             datum: datum.datum,
-            seriesId: datum.datum.seriesId,
+            seriesId: datum.series.id,
             xKey: datum.xKey,
             yKey: datum.yKey,
             xLowerKey,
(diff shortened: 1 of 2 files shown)


packages/ag-charts-enterprise/src/features/error-bar/errorBarNode.ts
```diff
@@ -88,7 +88,7 @@ export class ErrorBarNode extends _ModuleSupport.Group {
         return {
             ...(style as Required<AgErrorBarThemeableOptions>),
             datum: datum.datum,
-            seriesId: datum.datum.seriesId,
+            seriesId: datum.series.id,
```
olegat, line 91: Food for thought: Should we change the type of `SeriesNodeDatum.datum` from `any` to `unknown` to stop these typos slipping through the cracks?
https://github.com/ag-grid/ag-charts/blob/ea5870264efe1ae9937eeddd4f725a25462f5813/packages/ag-charts-community/src/chart/series/seriesTypes.ts#L65

alantreadway: Seems reasonable - `unknown` wasn't part of the version of TS we were using when the types were written, so I think it's reasonable for us to make that change now, especially since there is a pattern of bugs emerging around this.

alantreadway (approved):
LGTM

PR #4210, opened 2025-05-15
AG-14931 Fix bug where hover events can hide the context menu

olegat:
https://ag-grid.atlassian.net/browse/AG-14931


alantreadway (approved):
LGTM

PR #4217, opened 2025-05-16
Fix click interaction

jacobp100:


packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
```diff
@@ -445,7 +445,9 @@ export class SeriesAreaManager extends BaseManager {
 
         this.focusIndicator?.overrideFocusVisible(false);
 
-        this.onHoverLikeEvent(event);
+        if (current === this.chart.ctx.widgets.seriesWidget) {
+            this.onHoverLikeEvent(event);
+        }
```
olegat, line 450: The check makes sense. But I think that we might want to move it further down into the `onHoverLikeEvent` method itself to avoid undesirable side effects like breaking keyboard/mouse hot-swapping or not calling `preventDefault` on touch event.s

olegat (changes requested):

alantreadway:
Closed in favour of https://github.com/ag-grid/ag-charts/pull/4515

PR #4230, opened 2025-05-20
Large Dataset Perf

jacobp100:
* Removes all `dirty` flags from `Node`, moves them to `Group`
* Skips all child iteration on non-dirty groups

Scene will need the root to be a `Group` (or maybe better - a `Root` subclass) to check the dirty flags to restore the no-op rendering


packages/ag-charts-community/src/scene/shape/shape.ts
```diff
@@ -201,10 +201,12 @@ export abstract class Shape<D = any> extends Node<D> {
 
     private cachedDefaultGradientFillBBox?: BBox;
 
+    override markDirty(property?: string): void {
+        super.markDirty(property);
+        this.cachedDefaultGradientFillBBox = undefined;
+    }
+
     override preRender(renderCtx: RenderContext, thisComplexity?: number) {
```
alantreadway: Nit: Just remove this method as the override is now a no-op?

packages/ag-charts-community/src/scene/node.ts
```diff
@@ -423,32 +403,17 @@ export abstract class Node<D = any> {
...
-        if (!this._dirty) return;
-
-        this._dirty = false;
-        this.debugDirtyProperties();
-
-        for (const child of this.children()) {
-            child.markClean();
-        }
+        this.cachedBBox = undefined;
+        this.parentNode?.markDirty();
```
alantreadway, line 418: The original intent was to avoid percolating `markDirty()` calls upwards if this has happened previously, so we need to be careful that we haven't reduced performance in other cases by causing a lot more calls upwards.

packages/ag-charts-community/src/scene/group.ts
```diff
@@ -219,19 +228,6 @@ export class Group<D = any> extends Node<D> {
         super.render(childRenderCtx); // Calls markClean().
     }
 
-    private skipRender(childRenderCtx: RenderContext) {
-        const { stats } = childRenderCtx;
-
-        for (const child of this.children()) {
-            // Skip invisible children, but make sure their dirty flag is reset.
-            child.markClean();
```
alantreadway, line 227: The intent is that all child `dirty` flags are reset at this point, as we have had a recurring class of bug where unrendered nodes never had their dirty flags reset to `false` - we need to be sure we're not allowing this class of issue to be re-introduced.

jacobp100: If something gets marked, all its parents get marked dirty. Only groups track dirty state. Groups may skip rendering if they're not dirty - which means none of their children are dirty. A group's dirty flag is reset on render

packages/ag-charts-community/src/scene/node.ts
jacobp100: I can't figure out how to do this. I tried not doing it, but it breaks tests. There's probably more performance to be won here - but it's at least not a regression

packages/ag-charts-community/src/scene/shape/shape.ts
jacobp100: Updated

alantreadway (approved):
LGTM

PR #4237, opened 2025-05-21
BBox cleanup

iMoses:


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -560,7 +560,7 @@ export abstract class Chart extends Observable implements ModuleInstance, ChartS
     private runningUpdateType: ChartUpdateType = ChartUpdateType.NONE;
 
     private updateShortcutCount = 0;
-    private readonly seriesToUpdate: Set<ISeries<any, any, any>> = new Set();
```
alantreadway: We want to collect the super-set of `seriesToUpdate` across potentially multiple calls, so the changes here don't align with that.

packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
alantreadway, line 1: @olegat could you review the changes here?

iMoses: Changed were removed, will be part of a separate PR

packages/ag-charts-core/src/utils/boundingBox.ts
```diff
@@ -0,0 +1,22 @@
+export interface BoundingBox {
```
olegat: nit: I'm wondering whether we can up with a better name than `BoundingBox`, because we already have `BBox` which sounds identical (but isn't). Some possible alternatives:

- `BBoxLike`
- `BBoxSpec`
- `BBoxParams`
- `BBoxProps`
- `BBoxMixin`
- `PlainBBox`
- `SimpleBBox`
- `BBoxInterface`

olegat:
I think that these changes are good. Thanks Ido!

The only nit is that the new typename `BoundingBox` sounds too uncomfortably similar to `BBox`. But we can always rename that in a follow-up.

alantreadway (approved):
LGTM, naming might need more work, but let's roll with this.

PR #4244, opened 2025-05-21
AG-11754 API Reference Generator improvements

olegat:
https://ag-grid.atlassian.net/browse/AG-11754

- Remove obsolete build target `ag-charts-community:docs-debug-interfaces`.
- Remove obsolete build target `ag-charts-community:docs-resolved-interfaces`.
- Set `"strictNullChecks": true` on `ag-charts-generate-code-reference-files` project
- Improve error reporting of `ag-charts-generate-code-reference-files`

### DIFF
--- a/plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts
+++ b/plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts
@@ -471,7 +471,7 @@ function getJsDoc(node: ts.Node & { jsDoc?: { getFullText(): string }[] }) {
 }
 
 export function printNode(node: ts.Node | undefined) {
-    if (node == null) return null;
+    if (node == null) throw new Error(`unexpected ${node} node`);
     try {
         return tsPrinter.printNode(ts.EmitHint.Unspecified, node, node.getSourceFile()).replace(/\n\s*/g, ' ');
     } catch (e) {
--- a/plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts
+++ b/plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts
@@ -3,6 +3,7 @@ import { inputGlob, parseFile } from 'ag-shared/plugin-utils';
 import * as ts from 'typescript';
 
 type NodeType = any;
+// TODO: AG-14962 tighten HeritageType using a union.
 type HeritageType =
     | { kind?: string; type: any; typeParams: any[]; typeArguments?: any[]; members?: TypingMapItem[] }
     | string;
@@ -141,11 +142,11 @@ export class TypeMapper {
             } else if (h.type === 'Omit' || h.type === 'Pick' || h.type === 'Required') {
                 const n = this.resolveTypeRef(h);
                 node.members.push(...n.members);
-            } else if (h.type === 'Readonly' && h.typeArguments) {
+            } else if (h.type === 'Readonly' && h.typeArguments /* TODO: AG-14962 remove null-check */) {
                 const n = this.resolveType({ kind: 'typeAlias', type: h.typeArguments[0] });
                 node.members.push(...n.members);
             } else if (h.kind === 'typeLiteral') {
-                if (h.members) {
+                if (h.members /* TODO: AG-14962 remove null-check */) {
                     node.members.push(...h.members);
                 }
             } else {
@@ -384,7 +385,7 @@ export function formatNode(node: ts.Node | undefined) {
 
     if (ts.isTypeReferenceNode(node)) {
         const nodeType = formatNode(node.typeName);
-        if (nodeType === 'Array' && node.typeArguments) {
+        if (nodeType === 'Array' && node.typeArguments /* TODO: AG-14962 remove null-check */) {
             return {
                 kind: 'array',
                 type:
--- a/plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts
+++ b/plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts
@@ -141,11 +141,13 @@ export class TypeMapper {
             } else if (h.type === 'Omit' || h.type === 'Pick' || h.type === 'Required') {
                 const n = this.resolveTypeRef(h);
                 node.members.push(...n.members);
-            } else if (h.type === 'Readonly') {
+            } else if (h.type === 'Readonly' && h.typeArguments) {
                 const n = this.resolveType({ kind: 'typeAlias', type: h.typeArguments[0] });
                 node.members.push(...n.members);
             } else if (h.kind === 'typeLiteral') {
-                node.members.push(...h.members);
+                if (h.members) {
+                    node.members.push(...h.members);
+                }
             } else {
                 console.warn(`Unhandled type "${h.type}" on ${node.name}`, h);
                 throw Error(`Unhandled type "${h.type}" on ${node.name}`);
@@ -237,7 +239,9 @@ export class TypeMapper {
     }
 }
 
-export function formatNode(node: ts.Node) {
+export function formatNode(node: ts.Node | undefined) {
+    if (node == null) return undefined;
+
     if (ts.isUnionTypeNode(node)) {
         return {
             kind: 'union',
@@ -322,7 +326,7 @@ export function formatNode(node: ts.Node) {
             name: formatNode(node.name),
             members: node.members.map((n) => {
                 let memberDocs = getJsDoc(n);
-                const matchDefault = memberDocs?.at(-1).match(/^\s*Default:\s*`([^`]+)`\s*$/);
+                const matchDefault = memberDocs?.at(-1)?.match(/^\s*Default:\s*`([^`]+)`\s*$/);
                 let defaultValue: string | undefined;
                 if (matchDefault) {
                     defaultValue = matchDefault[1];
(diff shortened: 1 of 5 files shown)


alantreadway (approved):
Nx changes LGTM, @iMoses you are probably most familar with the other changes?

plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts
```diff
@@ -141,11 +141,13 @@ export class TypeMapper {
             } else if (h.type === 'Omit' || h.type === 'Pick' || h.type === 'Required') {
                 const n = this.resolveTypeRef(h);
                 node.members.push(...n.members);
-            } else if (h.type === 'Readonly') {
+            } else if (h.type === 'Readonly' && h.typeArguments) {
```
iMoses: nit: was it necessary? I can't think of a case in which you'll use `Readonly` without it's argument, TS won't allow it far before this script will be executed.

plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts
```diff
@@ -141,11 +141,13 @@ export class TypeMapper {
...
                 const n = this.resolveType({ kind: 'typeAlias', type: h.typeArguments[0] });
                 node.members.push(...n.members);
             } else if (h.kind === 'typeLiteral') {
-                node.members.push(...h.members);
+                if (h.members) {
+                    node.members.push(...h.members);
+                }
```
iMoses: If we have a `typeLiteral` without members then it mean something went wrong.
If this is a real case then it should be prevented at a higher level, non-interface types shouldn't make it into the `heritage` property.

plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts
```diff
@@ -380,7 +384,7 @@ export function formatNode(node: ts.Node) {
 
     if (ts.isTypeReferenceNode(node)) {
         const nodeType = formatNode(node.typeName);
-        if (nodeType === 'Array') {
+        if (nodeType === 'Array' && node.typeArguments) {
```
iMoses: nit: same as with `Readonly`, `Array` will always have an argument otherwise it's a TS error.

plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts
```diff
@@ -322,7 +326,7 @@ export function formatNode(node: ts.Node) {
             name: formatNode(node.name),
             members: node.members.map((n) => {
                 let memberDocs = getJsDoc(n);
-                const matchDefault = memberDocs?.at(-1).match(/^\s*Default:\s*`([^`]+)`\s*$/);
+                const matchDefault = memberDocs?.at(-1)?.match(/^\s*Default:\s*`([^`]+)`\s*$/);
```
iMoses, line 330: This basically is allowing empty docs? `/** */`
I think it should be fixed at a higher level, e.g. the docs extractor shouldn't set docs in case what returned was empty. This will keep types simpler here to work with, without certain arguments becoming possibly `null`

plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts
```diff
@@ -475,5 +480,6 @@ export function printNode(node: ts.Node) {
 }
 
 function trimArray(array?: string[]): string[] {
-    return array?.join('\n').trim().split('\n');
+    if (array == null) return [];
+    return array.join('\n').trim().split('\n');
```
iMoses: If we make sure not to have empty docs then this change is unnecessary

plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts
```diff
@@ -466,7 +470,8 @@ function getJsDoc(node: ts.Node & { jsDoc?: { getFullText(): string }[] }) {
     );
 }
 
-export function printNode(node: ts.Node) {
+export function printNode(node: ts.Node | undefined) {
```
iMoses: Ideally `formatNode` and `printNode` should never get an undefined node, so my initial thought is that it shouldn't be silenced. What was the issue here that led to the change?

plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts
iMoses: To clarify, this could only happen when using `typeAlias` intersection with both interfaces and primitives, which we internally convert into an interface, but this should either change or at least the types which we cannot inherit from should be dropped.

plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts
olegat: Yes, it's necessary because this is a consequence of enabling `strictNullChecks: true`. The `typeArguments` property is always optional, see:
https://github.com/ag-grid/ag-charts/blob/5f05a935d3274bffb35739b118d80dd18fad96a5/plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts#L6

To get rid of this check, we would need to tighten the declaration of `HeritageType` to something like:
```ts
type HeritageType =
    | { kind: 'Readonly'; type: any; typeParams: any[]; typeArguments: any[]; members?: TypingMapItem[] }
    | { kind: 'Omit' /** ... and so on ... */ }
    | string;
```

But this falls out of the scope of this PR.

plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts
olegat: This is same problem as with `h.kind === 'Readonly'`. I agree that the check shouldn't be necessary, but it is necessary as consequence of enabling `strictNullChecks: true` because the `HeritageType.members` is always optional regardless of what `h.kind` is.

plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts
olegat: Sure, I can add a check before this line and throw an error if the `memberDocs` array is empty.

plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts
olegat: Agreed. I think that we can change `array?: string[]` to be required, and fix `getJsDoc()` to throw an error if it find an empty doc string.

plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts
olegat: Yes, enabling `strictNullChecks: true` revealed these errors:

```
plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts:145:71 - error TS18048: 'h.typeArguments' is possibly 'undefined'.
  143 |                 node.members.push(...n.members);
  144 |             } else if (h.type === 'Readonly') {
> 145 |                 const n = this.resolveType({ kind: 'typeAlias', type: h.typeArguments[0] });
      |                                                                       ^
  146 |                 node.members.push(...n.members);
  147 |             } else if (h.kind === 'typeLiteral') {
  148 |                 if (h.members) {

plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts:263:35 - error TS2345: Argument of type 'TypeNode | undefined' is not assignable to parameter of type 'Node'.
  Type 'undefined' is not assignable to type 'Node'.
  261 |             kind: 'typeParam',
  262 |             name: printNode(node.name),
> 263 |             constraint: printNode(node.constraint),
      |                                   ^
  264 |             default: printNode(node.default),
  265 |         };
  266 |     }

plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts:264:32 - error TS2345: Argument of type 'TypeNode | undefined' is not assignable to parameter of type 'Node'.
  Type 'undefined' is not assignable to type 'Node'.
  262 |             name: printNode(node.name),
  263 |             constraint: printNode(node.constraint),
> 264 |             default: printNode(node.default),
      |                                ^
  265 |         };
  266 |     }
  267 | 

plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts:315:33 - error TS2345: Argument of type 'PropertyName | undefined' is not assignable to parameter of type 'Node'.
  Type 'undefined' is not assignable to type 'Node'.
  313 |                 kind: 'member',
  314 |                 docs: getJsDoc(n),
> 315 |                 name: printNode(n.name),
      |                                 ^
  316 |                 type: formatNode(n),
  317 |                 optional: !!n.questionToken,
  318 |             })),
Found 4 errors.
```

These callers should be fixed, but to keep this PR simple I'll change `printNode` to throw an error for now.

plugins/ag-charts-generate-code-reference-files/src/doc-interfaces/types-utils.ts
olegat: Yes. Same thing. I'll make a tech-debt ticket to address this is a followup.

olegat:
Action Items
- [x] Add `TODO` comments to tech-debt ticket AG-14962 (when `kind: 'Readonly' | 'typeLiteral' | 'Array'`).
- [ ] ~Throw error in `getJsDoc` when docs are empty.~
- [ ] ~Throw error in `printNode` when `node == null`.~
- [x] Remove `?` from `trimArray`.

alantreadway (approved):
LGTM

PR #4246, opened 2025-05-21
Scene perf 2

jacobp100:


packages/ag-charts-community/src/scene/node.ts
```diff
@@ -245,88 +220,23 @@ export abstract class Node<D = any> {
...
-        this.markDirtyChildrenOrder();
-        this.markDirty();
-    }
-
-    appendChild<T extends Node>(node: T): T {
-        this.append(node);
-        return node;
-    }
-
     removeChild(node: Node) {
```
alantreadway, line 253: Seems odd to retain `removeChild()` here - but I guess this is due to  the dependency from `remove()`?

Perhaps we could change `parentNode` to be of an interface type, and have `Group` implement that interface to break the dependency cycle whilst removing group-like behaviours from `Node`?

packages/ag-charts-community/src/scene/node.test.ts
```diff
@@ -1,9 +1,10 @@
 import { describe, expect, it } from '@jest/globals';
 
 import { BBox } from './bbox';
+import { Group } from './group';
 import { Node } from './node';
 
-class TestNode<D = any> extends Node<D> {
+class TestNode<D = any> extends Group<D> {
```
alantreadway, line 7: Nit: Could we move tests for `Group` behaviours to `group.test.ts` rather than exercising them here still?

alantreadway (approved):
LGTM

PR #4253, opened 2025-05-22
CleanupRegistry normalisation

iMoses:
also adds `attachListener` that returns a cleanup callback.


alantreadway (approved):

packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
```diff
@@ -654,30 +654,27 @@ export class SeriesAreaManager extends BaseManager {
...
             focus.seriesIndex = pick.otherIndex;
         }
-        focus.datumIndex = pick.datumIndex;
-        focus.datum = datum;
 
         if (this.focusIndicator?.isFocusVisible()) {
             this.chart.ctx.animationManager.reset();
-        }
 
-        if (this.focusIndicator?.isFocusVisible()) {
```
olegat, line 673: LGTM - this is safe. The second condition use to include an `&& seriesRect`. See https://github.com/ag-grid/ag-charts/commit/516341f7ede0d7f0b5a353a10966c5615e77274e

packages/ag-charts-core/src/utils/dom/domEvents.ts
olegat, line 1: Using function overloads like we did in `keynavUtil.ts` is preferable. This new implementation includes the use of the `as` keyword which introduces some risk.

Also, the caller is permitted to use a union type such as `T = Window | Document`. This is implementation, this could lead to unexpected errors because `EventMapFor<T>` would expand to `WindowEventMap` and not `WindowEventMap | DocumentEventMap` as expected.

packages/ag-charts-core/src/utils/dom/domEvents.ts
```diff
@@ -0,0 +1,24 @@
...
+    element: T,
+    eventName: K & string,
+    handler: (this: T, event: EventMapFor<T>[K]) => void,
+    options?: boolean | AddEventListenerOptions
+): () => void;
+
+export function attachListener(
+    element: Document | HTMLElement | Window,
+    eventName: string,
+    handler: (event: Event) => void,
```
olegat, line 19: nit: I'm surprised that this compiles... shouldn't it be something like `handler: (this: object, event: Event) => void`? Otherwise LGTM.

olegat (approved):

packages/ag-charts-core/src/utils/dom/domEvents.ts
iMoses: TypeScript allows this because the implementation signature only needs to be compatible with the overloads. The `this` type is enforced at the call site via the overloads, so it's safe (and common) to omit it in the implementation.

alantreadway (approved):
LGTM

PR #4256, opened 2025-05-22
AG-8436 - Axis highlight band

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-8436


packages/ag-charts-community/src/chart/axis/axisManager.ts
alantreadway, line 1: I'm not sure if `AxisManager` is the right place for the responsibilities being added here - since most of the interaction handling relates to the series-area, `SeriesAreaManager` might be a better spot?

alantreadway (approved):
LGTM

iMoses (approved):
LGTM 👍

PR #4260, opened 2025-05-23
AG-14976 - Fix zero height gradient bbox breaking rendering.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-14976

### DIFF
--- a/packages/ag-charts-community/src/scene/canvas/hdpiCanvas.ts
+++ b/packages/ag-charts-community/src/scene/canvas/hdpiCanvas.ts
@@ -92,6 +92,10 @@ export class HdpiCanvas {
         Object.freeze(this);
     }
 
+    reset() {
+        this.context.resetTransform();
+    }
+
     private onEnabledChange() {
         if (this.element) {
             this.element.style.display = this.enabled ? '' : 'none';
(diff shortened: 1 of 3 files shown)


packages/ag-charts-community/src/scene/canvas/hdpiCanvas.ts
```diff
@@ -92,6 +92,10 @@ export class HdpiCanvas {
         Object.freeze(this);
     }
 
+    reset() {
+        this.context.resetTransform();
```
alantreadway, line 96: We should be doing more than this (e.g. popping all saved state, creating a replacement clean context, or resetting all attributes manually) - but for a patch this is low risk and actually recoverable for the failure case the main fix addresses.

packages/ag-charts-community/src/scene/gradient/linearGradient.ts
```diff
@@ -48,6 +48,9 @@ export class LinearGradient extends Gradient {
 
     protected override createCanvasGradient(ctx: CanvasRenderingContext2D, bbox: BBox): CanvasGradient | undefined {
         const { x0, y0, x1, y1 } = this.getGradientPoints(bbox);
+        if (isNaN(x0) || isNaN(y0) || isNaN(x1) || isNaN(y1)) {
+            return undefined;
+        }
```
alantreadway, line 53: This is the actual fix.

jacobp100 (approved):

PR #4263, opened 2025-05-23
AG-14976 - Improved gradient invalid bbox handling.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-14976

### DIFF
--- a/packages/ag-charts-community/src/scene/gradient/gradient.ts
+++ b/packages/ag-charts-community/src/scene/gradient/gradient.ts
@@ -36,7 +36,7 @@ export abstract class Gradient {
     ): CanvasGradient | string | undefined {
         const bbox = this.bbox ?? shapeBbox;
 
-        if (isNaN(bbox.x) || isNaN(bbox.y) || !bbox.isFinite()) {
+        if (!bbox.isFinite()) {
             return;
         }
 
--- a/packages/ag-charts-community/src/scene/gradient/gradient.ts
+++ b/packages/ag-charts-community/src/scene/gradient/gradient.ts
@@ -36,7 +36,7 @@ export abstract class Gradient {
     ): CanvasGradient | string | undefined {
         const bbox = this.bbox ?? shapeBbox;
 
-        if (isNaN(bbox.x) || isNaN(bbox.y)) {
+        if (isNaN(bbox.x) || isNaN(bbox.y) || !bbox.isFinite()) {
             return;
         }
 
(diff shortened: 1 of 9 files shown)


packages/ag-charts-community/src/scene/shape/path.ts
```diff
@@ -134,20 +134,22 @@ export class Path<D = any> extends Shape<D> implements DistantObject {
...
-            this._clipPath ??= new ExtendedPath2D();
-            this._clipPath.clear();
-            this._clipPath.rect(-margin, -margin, this._clipX + margin, this._clipY + margin + margin);
-
-            // Bound the shape rendered to the clipping path.
-            ctx.clip(this._clipPath?.getPath2D());
-
-            if (this._clipX > 0 && this._clipY > 0) {
-                this.drawPath(ctx);
+            try {
```
jacobp100, line 137: This will have a performance impact on things like a large scatter plot - is this try/catch definitely needed

packages/ag-charts-community/src/scene/transformable.ts
```diff
@@ -131,16 +131,18 @@ function MatrixTransform<N extends Node>(Parent: Constructor<N>) {
...
-                ctx.save();
-                performRestore = true;
-                matrix.toContext(ctx);
-            }
-
-            super.render(renderCtx);
-
-            if (performRestore) {
-                ctx.restore();
+            try {
```
jacobp100, line 134: Again this will have a performance impact

packages/ag-charts-community/src/scene/shape/path.ts
alantreadway: Verified offline, no performance impact.

packages/ag-charts-community/src/scene/transformable.ts
alantreadway: Not visible in `/benchmarks` - I agree that in the event of errors being thrown there is a performance hit, but for the normal non-error case there should be negligible overhead for this construct.

jacobp100 (approved):

packages/ag-charts-community/src/scene/gradient/gradient.ts
```diff
@@ -36,7 +36,7 @@ export abstract class Gradient {
     ): CanvasGradient | string | undefined {
         const bbox = this.bbox ?? shapeBbox;
 
-        if (isNaN(bbox.x) || isNaN(bbox.y)) {
+        if (isNaN(bbox.x) || isNaN(bbox.y) || !bbox.isFinite()) {
```
jacobp100:
```suggestion
        if (!bbox.isFinite()) {
```

jacobp100 (approved):

PR #4273, opened 2025-05-27
AG-14982 - Fix crosshair labels violating style CSP constraints.

alantreadway:


packages/ag-charts-enterprise/src/features/crosshair/crosshairLabel.ts
```diff
@@ -69,10 +70,13 @@ export class CrosshairLabel extends BaseProperties {
         this.toggle(true);
     }
 
-    setLabelHtml(html?: string) {
+    setLabelHtml({ html, styles }: { html?: string; styles?: Record<string, StyleValue> }) {
         if (html !== undefined) {
             this.element.innerHTML = html;
         }
+        if (styles !== undefined) {
```
jacobp100, line 77: Does this have a case that if anything gets unset in styles, it won't be unset in the html?

jacobp100 (approved):

packages/ag-charts-enterprise/src/features/crosshair/crosshairLabel.ts
alantreadway: Agreed, the `html` case is also suspect, but I don't want to touch more than necessary for a patch.

PR #4281, opened 2025-05-28
AG-NONE `PerKeyMap` interface

olegat:
Per-key typing variation of `Map<K, V>`

### DIFF
--- a/packages/ag-charts-core/src/classes/eventEmitter.ts
+++ b/packages/ag-charts-core/src/classes/eventEmitter.ts
@@ -1,9 +1,9 @@
-import type { PerKeyMap } from "../interfaces/strictMap";
+import type { PerKeyMap } from '../interfaces/strictMap';
 
 export type EventListener<T> = (event: T) => void;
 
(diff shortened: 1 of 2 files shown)


packages/ag-charts-core/src/classes/eventEmitter.ts
```diff
@@ -1,7 +1,9 @@
+import type { PerKeyMap } from '../interfaces/strictMap';
+
 export type EventListener<T> = (event: T) => void;
 
 export class EventEmitter<EventMap extends object> {
-    private readonly events = new Map<keyof EventMap, Set<EventListener<any>>>();
+    private readonly events: PerKeyMap<{ [K in keyof EventMap]: Set<EventListener<EventMap[K]>> }> = new Map();
```
olegat, line 6: This is version of `eventEmitter.ts` is more type-safe, because `this.events.get` and `this.events.set` now receive & return values of the correct typing that matches the input key.

iMoses (approved):

PR #4287, opened 2025-05-28
AG-14990 Add focusout/mouseout listeners on interactive tooltips

olegat:
https://ag-grid.atlassian.net/browse/AG-14990

The `SeriesAreaManager` class ignores event for type `'leave'` (for mouse users) and `'blur'` (for keyboard users) to allow the user to interact with the tooltip.

However, this means that the tooltip can incorrectly remain shown when the mouse moves or the focus changes.

Therefore, add custom event listeners to interactive tooltip to address is anomaly.


alantreadway (approved):
LGTM

PR #4292, opened 2025-05-28
AG-11754 Fix `aglint/require-explicit-generic` warnings

olegat:
https://ag-grid.atlassian.net/browse/AG-11754

* Fix all `aglint/require-explicit-generic` violations.
* Promote the `aglint/require-explicit-generic` rule from warning to error.

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -209,7 +209,7 @@
       "executor": "nx:run-commands",
       "options": {
         "cwd": "{projectRoot}",
-        "command": "npx eslint --quiet"
+        "command": "npx eslint"
       },
       "cache": true,
       "inputs": [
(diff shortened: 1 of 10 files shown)


nx.json
```diff
@@ -209,7 +209,7 @@
       "executor": "nx:run-commands",
       "options": {
         "cwd": "{projectRoot}",
-        "command": "npx eslint --quiet"
+        "command": "npx eslint"
```
olegat, line 212: Open question: There is some debate as to whether we want to keep warnings muted from `nx lint`. Perhaps we can an option for it like `nx lint -c quiet`. LMKWYT.

alantreadway: Let's see what happens on this PR once CI passes - my guess is that ALL the lint warnings will get reported as annotations on this PR, which would be a bit much - if not then it's fine.

alantreadway (commented):
LGTM

PR #4293, opened 2025-05-28
AG-15005 - Switch to `skia-canvas`.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-15005


libraries/ag-charts-test/src/canvas/mock-canvas.ts
```diff
@@ -223,7 +171,8 @@ export function setup(opts: { width?: number; height?: number; document?: Docume
 export function teardown(mockContext: MockContext) {
     mockContext.document.createElement = mockContext.realCreateElement!;
     if (typeof window !== 'undefined') {
-        window.OffscreenCanvas = mockContext.realOffscreenCanvas;
+        // @ts-expect-error types don't exactly align
```
jacobp100: Is this right? It's meant to restore the original global

packages/ag-charts-community/src/scene/shape/text.ts
```diff
@@ -247,6 +247,10 @@ export class Text<D = any> extends Shape<D> {
     ): void {
         const { lines, x, y } = this;
 
+        if (isNaN(x) || isNaN(y) || isNaN(offsetY) || isNaN(lineHeight)) {
```
jacobp100:
```suggestion
        if (!Number.isFinite(x) || !Number.isFinite(y) || !Number.isFinite(offsetY) || !Number.isFinite(lineHeight)) {
```

Since I'd assume the canvas won't handle infinite values too

Also just a note that the global isNaN will coerce a string to a number, so `isNaN('f')` returns true - but `Number.isNaN` will only return true for a true NaN value

packages/ag-charts-core/src/utils/numbers.ts
```diff
@@ -1,4 +1,8 @@
 export function clamp(min: number, value: number, max: number) {
+    if (value == null) {
```
jacobp100: I think we should not do this check here. It also means the (inferred) type is incorrect - you're returning `number | null | undefined` in a function that's marked as returning `number`.

Do we know what was passing in nil values?

libraries/ag-charts-test/src/canvas/mock-canvas.ts
alantreadway: Fair point - reverted and there are no snapshot changes or error, so I agree lets keep as before.

packages/ag-charts-community/src/scene/shape/text.ts
alantreadway: Updated.

packages/ag-charts-core/src/utils/numbers.ts
alantreadway: I've updated this to `throw Error()` in this situation, and found one spot in `coneFunnelUtil.ts` that was causing issues. This was showing up as a problem more widely before, so perhaps your suggestion for `text.ts` improved things too.

packages/ag-charts-community/src/scene/shape/text.ts
alantreadway: Change applied, thanks.

libraries/ag-charts-test/src/canvas/mock-canvas.ts
```diff
@@ -189,13 +139,13 @@ export function setup(opts: { width?: number; height?: number; document?: Docume
...
-            const nextCanvas = mockCtx.canvasStack.shift() ?? createCanvas(width, height);
+            const nextCanvas = mockCtx.canvasStack.shift() ?? new ConfiguredCanvas(width, height);
             mockCtx.registerCanvasInstance(nextCanvas);
 
             proxyGetContext2D(mockCtx, nextCanvas, mockedElement);
 
-            mockedElement.toDataURL = (mimeType?: 'image/png') => {
-                return nextCanvas.toDataURL(mimeType ?? 'image/png');
+            mockedElement.toDataURL = (mimeType?: 'png') => {
+                return nextCanvas.toDataURLSync(mimeType ?? 'png');
```
jacobp100: Their type is incorrect here

```suggestion
            mockedElement.toDataURL = (mimeType: 'image/png') => {
                return nextCanvas.toDataURLSync(mimeType.split('/')[1]);
```

package.json
```diff
@@ -64,7 +63,7 @@
     "@vitest/coverage-v8": "1.5.3",
     "@vitest/ui": "1.5.3",
     "algoliasearch": "^4.18.0",
-    "canvas": "^3.1.0",
+    "skia-canvas": "^2.0.2",
```
jacobp100: Minor - but this isn't alphabetical (think this is my fault)

jacobp100 (approved):

libraries/ag-charts-test/src/canvas/mock-canvas.ts
alantreadway: Good spot, applied

package.json
alantreadway: Fixed

packages/ag-charts-core/src/utils/numbers.ts
alantreadway: Revert this change in favor of defaulting using `??` at the call site for now.

alantreadway:
Included as part of #4307

PR #4296, opened 2025-05-29
AG-15009 - Update to enable Angular 20 compatibility and testing.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-15009


packages/ag-charts-website/src/content/docs/supported-frameworks/index.mdoc
```diff
@@ -23,6 +23,7 @@ There are no version compatibility requirements for AG Charts when using the Jav
 
 | AG Charts Version | Angular Version |
 | ----------------- | --------------- |
+| 12.x              | 17 - 20         |
```
taktran, line 26: Might want to consider dropping Angular 17 support here, as [Angular have dropped support for it](https://angular.dev/reference/releases#actively-supported-versions) (Grid is doing the same)

```suggestion
| 12.x              | 18 - 20         |
```

Same would have to happen on the `package.json`'s

taktran (approved):
LGTM 👍🏼 Just a comment about version support

packages/ag-charts-website/src/content/docs/supported-frameworks/index.mdoc
alantreadway: Tracking distinctly in https://ag-grid.atlassian.net/browse/AG-15017 - if we patch this PR to `b11.3.x` we don't want to also couple it with that change.

PR #4298, opened 2025-05-29
AG-15014: Validate that axis is an array of at least 2 objects

iMoses:
https://ag-grid.atlassian.net/browse/AG-15014


alantreadway (approved):
LGTM

PR #4309, opened 2025-05-30
Gallery examples for global formatter

jacobp100:
https://ag-grid.atlassian.net/browse/AG-13777


packages/ag-charts-website/src/content/gallery/_examples/bubble-with-images/main.ts
```diff
@@ -81,14 +74,24 @@ const options: AgCartesianChartOptions = {
...
-                formatter: ({ value }) => (value === 0 ? `Easy` : value === 10 ? `Difficult` : value),
-            },
             interval: {
                 values: [0, 10],
             },
         },
     ],
+    formatter: {
+        x: (params) => {
+            const value = params.value as number;
```
alantreadway, line 84: Can we avoid this cast? (Just contrasting with the previous formatters, this looks like an extra step?)

alantreadway: Discussed offline, `value` isn't directly derived from `TDatum` so it's not really possible to do better here.

alantreadway (approved):
LGTM

PR #4318, opened 2025-06-02
AG-15013 Allow navigator min & max values to be equal

olegat:
https://ag-grid.atlassian.net/browse/AG-15013

This fixes an odd visual glitch that occasionally occurs on some displays.


alantreadway (approved):
LGTM

PR #4324, opened 2025-06-03
Reformat HTML to make it easier to copy localhost script tags over to…

jacobp100:
… other examples

Moves the `@vite/client` script near the AG Charts script, so you can copy and paste both in one go

image


taktran (approved):
LGTM 👍🏼

alantreadway (approved):
LGTM

PR #4344, opened 2025-06-05
AG-14737 Add options graph

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14737


packages/ag-charts-community/src/chart/axis/__image_snapshots__/axis-grid-line-test-ts-axis-grid-line-do-not-draw-empty-styles-1-snap.png
lsjroberts, line 1: These axis grid line snapshots were broken by this change https://github.com/ag-grid/ag-charts/commit/e5e059a5e4cb189af730bce72845c0681e30ffda which reduced domain of the chart. I've added an extra datum to ensure the lines are drawn.

packages/ag-charts-community/src/chart/chartHeapMemory.test.ts
```diff
@@ -122,7 +122,7 @@ describe.skip('Chart Heap Memory', () => {
             const heapProportionChange = Math.abs(endingHeap - startingHeap) / startingHeap;
 
             // console.log({ startingHeap, endingHeap, heapProportionChange });
-            expect(heapProportionChange).toBeLessThan(0.2);
+            expect(heapProportionChange).toBeLessThan(0.3);
```
lsjroberts: There is an amount of extra memory used, typically up to about `0.25` on my machine. The graph vertices and edges are all cleared after it is resolved. This is not unexpected.

packages/ag-charts-community/src/chart/themes/chartTheme.ts
```diff
@@ -506,38 +517,10 @@ export class ChartTheme {
         this.params = mergeDefaults(params, this.getPublicParameters());
 
         this.config = deepFreeze(this.templateTheme(defaults));
+        this.overrides = deepFreeze(overrides);
         this.presets = deepFreeze(presets);
     }
 
-    private mergeOverrides(defaults: AgChartThemeOverrides, presets: AgPresetOverrides, overrides: AgThemeOverrides) {
```
lsjroberts, line 512: The merging of overrides is now handled within the graph. There are two discussion points on this:

- The removal of `zoom` and `navigator` from non-cartesian chart types is done directly in the graph with a `without()` call. It was a bit awkward trying to re-use the current `CHART_TYPE_SPECIFIC_COMMON_OPTIONS` object. Though I've noted it is something that could be improved.
- The `isPresetOverridesType()` call is no longer made. It seems like it is no longer necessary and the examples all seem to run fine without it. But I don't fully understand the problem it was solving.

packages/ag-charts-community/src/module/__snapshots__/optionsModule.test.ts.snap
lsjroberts, line 1: The changes here are expected. It is the removal of `theme` from the user options that get snapshotted. The user options `theme` object is already merged into the config by the `ChartTheme`. The graph uses this merged config and does not need to do the additional processing of the lingering user options `theme`.

packages/ag-charts-community/src/chart/chartHeapMemory.test.ts
alantreadway: I also have a change that should improve stability of this test (it will get run in series rather than parallel with other tests): https://github.com/ag-grid/ag-charts/pull/4339

packages/ag-charts-community/src/module/optionsGraphOperations.ts
```diff
@@ -0,0 +1,741 @@
...
+    const backgroundColor = graph.getParamValue('backgroundColor');
+
+    if (typeof foregroundColor === 'string' && typeof backgroundColor === 'string' && isRatio(backgroundRatio)) {
+        return Color.mix(
+            Color.fromString(foregroundColor),
+            Color.fromString(backgroundColor),
+            backgroundRatio
+        ).toString();
+    }
+    // Logger.warnOnce(
```
alantreadway: Nit: Needed?

packages/ag-charts-community/src/module/optionsGraphOperations.ts
```diff
@@ -0,0 +1,741 @@
...
+        isRatio(backgroundRatio) &&
+        isRatio(accentRatio)
+    ) {
+        return Color.mix(
+            Color.mix(Color.fromString(foregroundColor), Color.fromString(backgroundColor), backgroundRatio),
+            Color.fromString(accentColor),
+            accentRatio
+        ).toString();
+    }
+    // Logger.warnOnce(
```
alantreadway: Nit: Needed?

packages/ag-charts-community/src/module/optionsGraphOperations.ts
alantreadway: If we want these warnings to alert us that we have internal inconsistencies, we can use `Debug.inDevelopmentMode()` to add these warnings to cause tests to fail, without needing to affect runtime execution for real users?

packages/ag-charts-community/src/module/optionsGraphOperations.ts
```diff
@@ -0,0 +1,741 @@
...
+
+function remOperation(graph: OptionsGraphInterface, _vertex: VertexInterface, values: Set<VertexInterface>) {
+    const [valueVertex] = values;
+    const value = graph.getVertexValue(valueVertex);
+    const fontSize = graph.getParamValue('fontSize');
+
+    if (typeof fontSize === 'number' && typeof value === 'number') {
+        return Math.round(value * fontSize);
+    }
+    // Logger.warnOnce(`\`$rem\` json operation failed on [${String(a)}] at [${path.join('.')}], expecting a number.`);
```
alantreadway: Nit needed?

packages/ag-charts-community/src/module/optionsGraphOperations.ts
```diff
@@ -0,0 +1,741 @@
...
+// --- LOGIC ---
+
+enum LogicOperation {
+    And = '$and',
+    Eq = '$eq',
+    GreaterThan = '$greaterThan',
+    If = '$if',
+    Not = '$not',
+    Or = '$or',
+    // Switch = '$switch',
```
alantreadway, line 298: Nit: needed?

packages/ag-charts-community/src/util/graph.ts
alantreadway, line 1: Maybe we can move this into `ag-charts-core` since i's a generic implementation with low number of dependencies?

packages/ag-charts-community/src/chart/chartHeapMemory.test.ts
lsjroberts: Yep, that fixed it, thanks.

packages/ag-charts-community/src/module/optionsGraphOperations.ts
lsjroberts: Added these in with the dev mode check. More of these sorts of warnings could be added in the future, but not high priority.

packages/ag-charts-community/src/util/graph.ts
lsjroberts: Agreed. Done.

alantreadway (approved):
LGTM

PR #4345, opened 2025-06-05
AG-15047 - Update gallery examples to use TDatum.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-15047


packages/ag-charts-website/src/content/gallery/_examples/area-with-labels/main.ts
```diff
@@ -51,7 +51,7 @@ const options: AgChartOptions = {
     ],
     formatter: {
         y(params) {
-            if (params.source === 'series-label' && !labelDates.includes(params.datum.date.valueOf())) {
+            if (params.source === 'series-label' && !labelDates.includes(params.datum?.date.valueOf() ?? 0)) {
```
olegat, line 54: Check that the output JS code is easy to read. IIRC, either the `?` or `??` operator adds some wrapper that makes the final JS more difficult to read.

packages/ag-charts-website/src/content/gallery/_examples/customised-funnel/main.ts
```diff
@@ -26,7 +24,7 @@ const options: AgChartOptions = {
             strokeOpacity: 0,
             itemStyler: ({ datum, valueKey, stroke }) => ({
                 fill: stroke,
-                fillOpacity: getOpacity(datum[valueKey], valueKey as ValueKey, 0.4, 1),
+                fillOpacity: getOpacity(datum, valueKey as DataNumberKey, 0.4, 1),
```
olegat, line 27: This is fine for this PR.

But, perhaps as a follow-up, it would be nice if the `as` wasn't necessary. Although this will involve some more sophisticated typing:
```
diff --git a/packages/ag-charts-types/src/series/cartesian/funnelOptions.ts b/packages/ag-charts-types/src/series/cartesian/funnelOptions.ts
index e64c9ddf4f..41d7a26544 100644
--- a/packages/ag-charts-types/src/series/cartesian/funnelOptions.ts
+++ b/packages/ag-charts-types/src/series/cartesian/funnelOptions.ts
@@ -67,11 +67,11 @@ export interface AgFunnelSeriesThemeableOptions<TDatum = TDatumDefault>
     itemStyler?: Styler<AgFunnelSeriesItemStylerParams<TDatum>, AgFunnelSeriesStyle>;
 }
 
-export interface AgFunnelSeriesOptionsKeys {
+export interface AgFunnelSeriesOptionsKeys<TDatum> {
     /** The key to use to retrieve stage values from the data. */
     stageKey: string;
     /** The key to use to retrieve values from the data. */
-    valueKey: string;
+    valueKey: TDatum extends object ? keyof TDatum & string : string;
 }
 
 export interface AgFunnelSeriesOptionsNames {}
```

packages/ag-charts-website/src/content/gallery/_examples/horizontal-range-bar-scatter-combination/main.ts
```diff
@@ -50,17 +55,13 @@ const options: AgChartOptions = {
...
-                        style: 'percent',
-                        maximumFractionDigits: 0,
-                    });
+                renderer({ datum }) {
                     return {
-                        title: datum[yKey],
+                        title: datum.smartphone,
                         data: [
                             {
-                                label: xName!,
```
olegat, line 62: I think that this line can stay. It will ensure that the tooltip label matches the series `xName` value.

olegat (approved):
some nits, but mostly LGTM

packages/ag-charts-website/src/content/gallery/_examples/area-with-labels/main.ts
alantreadway: I'll check, IIRC we uplifted the TS target output version at some point and that stopped being a thing.

alantreadway: Screenshot 2025-06-06 at 09 39 01

^ looks ok in the generated output, so this is no longer an issue

packages/ag-charts-website/src/content/gallery/_examples/customised-funnel/main.ts
alantreadway: Agreed, I was hoping this PR would spark the discussion - I'm not sure exactly how involved that update would be?

packages/ag-charts-website/src/content/gallery/_examples/horizontal-range-bar-scatter-combination/main.ts
alantreadway: Fair, reverted

packages/ag-charts-website/src/content/gallery/_examples/customised-funnel/main.ts
olegat: I think that the main obstacle would be the API docs generator, but we already patch the API interface for the `axes` and `series` options. So perhaps it won't be difficult to also patch the interface parser to find & replace `TDatum extends object ? keyof TDatum & string : string` to just `string`.

PR #4354, opened 2025-06-06
AG-15047 - Update community examples to use TDatum.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-15047


olegat (approved):
LGTM - We'll have to how the changes to properties like `xKey` affect the docs generator, but we can address that in a follow-up. Nicely done!

alantreadway:
> LGTM - We'll have to how the changes to properties like `xKey` affect the docs generator, but we can address that in a follow-up. Nicely done!

Checked vs. staging and there is no difference in renderings, however the API rendering is broken for other reasons at the moment

PR #4367, opened 2025-06-10
AG-15062 - Remove range from install versions

markdurrant:
Only show specific version references without a range.


alantreadway (approved):
LGTM

PR #4387, opened 2025-06-11
AG-14916 Change navigator default height to 40 if miniChart enabled

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14916


alantreadway (approved):
LGTM

PR #4414, opened 2025-06-13
AG-14226 Add default strokeWidth when stroke is user defined

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14226

Removes a few `stroke` values from tests that expected `strokeWidth: 0` to avoid snapshot updates.


alantreadway (approved):
LGTM

PR #4416, opened 2025-06-13
Improve retrospective benchmark execution, and update benchmark results.

alantreadway:
- Better seperates the benchmark tests from the chart implementation, which should improve robustness over time.
  - Moves many test fixture setup + mock event generation to `ag-charts-test` for re-usability outside the scope of the main library (makes time-travel easier!).
- Updates historic benchmarks to take recent benchmark improvements into account.
- Unifies manual & CI historic run tooling.


libraries/ag-charts-test/src/canvas/to-match-image.ts
```diff
@@ -0,0 +1,31 @@
...
+export function toMatchImage(this: any, actual: ImageData, expected: ImageData, { writeDiff = true } = {}) {
+    // Grab values from enclosing Jest scope.
+    const { testPath, currentTestName } = this;
+
+    const width = CANVAS_WIDTH;
+    const height = CANVAS_HEIGHT;
+    const diff = new PNG({ width, height });
+    const result = pixelmatch(actual.data, expected.data, diff.data, width, height, { threshold: 0.01 });
+
+    const diffOutputFilename = `${testPath.substring(
```
jacobp100: Is this just `path.dirname(testPath)`?

jacobp100 (approved):

libraries/ag-charts-test/src/canvas/to-match-image.ts
alantreadway: Good shout, fixed.

PR #4418, opened 2025-06-16
Pre-12.0.0 optimisations Part 1

alantreadway:
- **Improve large-scale-multi-series load time by about 25%.**
- **Cleanup `readonly` properties.**


lsjroberts (approved):
LGTM. Though as mentioned we may be able to completely ditch `_edgesByEdge` and `edges()` as the `OptionsGraph` does not use it. This would however limit future uses of the generic graph. A future optimisation may instead define the type of edge storing based on the particular graph's need.

lsjroberts (approved):

PR #4422, opened 2025-06-16
AG-15138 Fix priority of series type overrides over common overrides

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-15138


alantreadway (approved):
LGTM

PR #4427, opened 2025-06-17
Pre-12.0.0 optimisations pt2

alantreadway:


packages/ag-charts-community/src/scale/continuousScale.ts
```diff
@@ -55,42 +54,42 @@ export abstract class ContinuousScale<D extends number | Date, I = number> exten
...
 
+        const { transform } = this;
         const clamp = options?.clamp ?? this.defaultClamp;
-        const d0 = Number(this.transform(domain[0]));
-        const d1 = Number(this.transform(domain[1]));
-        const x = Number(this.transform(value));
-
-        const { range } = this;
-        const [r0, r1] = range;
+        const d0 = (transform ? transform(domain[0])?.valueOf() : domain[0]?.valueOf()) ?? NaN;
```
jacobp100: You could do `this.transform?.(...)` rather than binding

packages/ag-charts-community/src/scale/logScale.ts
```diff
@@ -33,6 +33,18 @@ function pow(base: number, domain: number[], x: number) {
     return start >= 0 ? fn(base, x) : -fn(base, -x);
 }
 
+function transform(this: LogScale, x: number) {
+    const [min, max] = findMinMax(this.domain);
```
jacobp100: Or - if you took the domain in here, we could avoid `this` references

jacobp100 (approved):

packages/ag-charts-community/src/scale/continuousScale.ts
alantreadway: Reworked this aspect, thanks for the feedback.

packages/ag-charts-community/src/scale/logScale.ts
alantreadway: Ditto

PR #4441, opened 2025-06-18
AG-15046 Use `update()` instead of `updateDelta()`

olegat:
Fixes the Y-axis not updating (only fixes the example for JS).


alantreadway (approved):
LGTM

PR #4445, opened 2025-06-18
AG-15162 - Fix axis removal updates.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-15162


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1606,6 +1606,7 @@ export abstract class Chart extends Observable implements ModuleInstance, ChartS
         skip: string[] = []
     ) {
         if (!('axes' in options) || !options.axes) {
+            chart.axes = [];
```
alantreadway: This is the actual fix; everything else is just updating/adding test coverage.

lsjroberts (approved):
LGTM

PR #4458, opened 2025-06-19
AG-14226 Add default marker strokeWidth when stroke is user defined

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14226


packages/ag-charts-community-examples/src/charts-overview/examples/simple-line/main.ts
```diff
@@ -39,6 +39,7 @@ const options: AgChartOptions = {
             marker: {
                 stroke: '#01c185',
                 fill: '#01c185',
+                strokeWidth: 0
```
lsjroberts, line 42: Ideally this would be:

```diff
-stroke: '#01c185,
fill: '#01c185'
```

However then the highlighting cases fail since highlighting applies the `highlightSeries.strokeWidth` to both the line and the marker. And the marker would then have the default colour, not the custom colours seen here.

The `strokeWidth: 0` is required to therefore not fail all the other snapshots.

packages/ag-charts-community/src/module/optionsModule.test.ts
lsjroberts, line 1: Updated these since the `strokeWidth` was a new failure.

alantreadway (approved):
LGTM

PR #4485, opened 2025-06-20
adding band highlight to crosshair page

AG-DavidG:


packages/ag-charts-website/src/content/versions/ag-charts-versions.json
```diff
@@ -4,22 +4,28 @@
         "date": "June 24th, 2024",
         "highlights": [
             {
-                "text": "Axes Label Collision Avoidance",
+                "text": "Axis Band Highlight",
```
jacobp100, line 7: This is a historical changelog

AG-DavidG: Good catch!

PR #4488, opened 2025-06-20
Dg/ag 15163

AG-DavidG:


packages/ag-charts-website/src/content/docs/context/index.mdoc
```diff
@@ -3,40 +3,37 @@ title: Context API
...
+The Context Object is accessible via the `context` property in all callback and event handler parameters.
 
-## Context Object Overrides
+## Context Object
 
 The Context Object can be set using the `context` property either at the root of
-the chart options, or on individual series and axes. The options
-`series[].context` and `axes[].context` will be used for callbacks related to
-the target series or axis, but will fallback to the root `context` property if
+the chart options, or on an individual series or axes. The
```
olegat: nit: I'm not 100% sure that the grammar is correct here. AFAIK, I think it should be:
`on individual series or axes`
or
`on an individual series or axis`

packages/ag-charts-website/src/content/docs/context/index.mdoc
```diff
@@ -64,31 +61,30 @@ not set.
...
     title="Currency Converter"
     name="currency-converter"
     type="generated"
     options={ "enterprise": true }
 /%}
+
+Note that the Context Object is used by the following callbacks:
+
+-   The Y-axis [Label Formatter](./axes-labels/#formatter), to convert USD values to the preferred User Currency.
+-   The [Tooltip Renderer](./tooltips//#modifying-content), to render both the stock prices in USD and the User
```
olegat: nit: shouldn't this be `./tooltips/#modifying-content` (remove double-slash).

olegat (approved):
LGTM, just minor nits

PR #4493, opened 2025-06-20
Pre-12.0.0 optimisations pt. 6

alantreadway:
More graph.ts optimisations - biggest win is removing cost of lookups in `AdjacencyListGraph.{_cachedNeighbours,_edges}`.

### DIFF
--- a/packages/ag-charts-core/src/utils/graph.ts
+++ b/packages/ag-charts-core/src/utils/graph.ts
@@ -59,23 +59,18 @@ export class AdjacencyListGraph<V, E = undefined> {
         for (const [_edge, adjacentVertices] of edges) {
             this._vertexCount -= adjacentVertices.length;
         }
-        vertex.edges.clear();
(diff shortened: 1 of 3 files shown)


packages/ag-charts-core/src/utils/graph.ts
```diff
@@ -179,57 +159,20 @@ export class AdjacencyListGraph<V, E = undefined> {
...
-        for (const [edge, tos] of edges) {
-            for (const to of tos) {
-                let t: any = this.getVertexValue(to);
-                if (isObject(t)) t = `{${Object.keys(t).join(',')}}`;
-                string += `  -> ${edge as any} -> ${t}\n`;
-            }
-        }
-
-        return string;
-    }
```
lsjroberts, line 227: Can we leave these here, I use them all the time.

lsjroberts (approved):
LGTM

PR #4495, opened 2025-06-20
Pre 12.0.0 optimisations pt7.

alantreadway:
- **Optimise object accesses in setPathSafe()**
- **Optimise getOperation() data accesses.**
- **Optimise setProperties() by removing dead path.**
- **Optimise ESBuild CJS output to avoid ES-module proxying.**


packages/ag-charts-community/src/module/optionsGraphOperations.ts
alantreadway, line 1: Changes here significantly reduce key-iteration overhead and checking if the operation is a known type.

PR #4516, opened 2025-06-23
CRT-834 Add $shallow operation to allow user to replace array entirely

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-834


alantreadway (approved):
LGTM

PR #4520, opened 2025-06-23
CRT-838 - Skip no-op path animations, fixing range-area example.

alantreadway:


packages/ag-charts-community/src/chart/interaction/animationManager.ts
alantreadway, line 1: NOTE: This is mostly an indentation change (turn on ignore whitespace to make it easier to review).

packages/ag-charts-community/src/motion/pathMotion.ts
alantreadway, line 1: This is the core of the change - if a path-based animation function yields no change from start-to-finish, allow it to be collapsable.

jacobp100 (approved):

PR #4522, opened 2025-06-23
CRT-855: Callback params ref shouldn't display "required" in members

iMoses:
https://ag-grid.atlassian.net/browse/CRT-855


alantreadway (approved):
LGTM

PR #4534, opened 2025-06-23
CRT-858 - Force hardware acceleration to fix Safari example borders i…

markdurrant:
Force hardware acceleration to fix Safari example borders issue


taktran (approved):
👍🏼

PR #4543, opened 2025-06-24
CRT-809: Set `maxThicknessRatio` default to 0.5 on `grouped-category`

iMoses:
https://ag-grid.atlassian.net/browse/CRT-809


alantreadway (approved):
LGTM

PR #4553, opened 2025-06-24
Fix chart context type on the docs

iMoses:
+ rename `TDatumDefault` and `TContextDefault` to `DatumDefault` and `ContextDefault`


alantreadway (approved):
LGTM

PR #4554, opened 2025-06-24
CRT-860: `grouped` option is missing on range-bar series

iMoses:
https://ag-grid.atlassian.net/browse/CRT-860


alantreadway (approved):
LGTM

PR #4562, opened 2025-06-24
AG-14318 Convert change-detection lint script to ESLint Rule

olegat:
https://ag-grid.atlassian.net/browse/AG-14318


alantreadway (commented):
I'd strongly prefer it if we didn't have tooling-related code in `ag-charts-community`.

alantreadway (approved):
LGTM - thanks for making those updates.

PR #4563, opened 2025-06-25
AG-15218 - Optimise memory churn.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-15218

### DIFF
--- a/packages/ag-charts-community/src/module/optionsGraphOperations.ts
+++ b/packages/ag-charts-community/src/module/optionsGraphOperations.ts
@@ -685,7 +685,7 @@ function applyOperation(graph: OptionsGraphInterface, vertex: VertexInterface, v
     if (!isPlainObject(object)) return;
 
     const defaultValue = defaultValueVertex ? graph.getVertexValue(defaultValueVertex) : undefined;
-    const children = graph.neighboursWithEdgeValue(vertex, PATH_EDGE);
+    const children = graph.neighboursWithEdgeValue(vertex, PATH_EDGE) ?? [];
     const overridesPath1 = overridesPathVertex1
         ? (graph.resolveVertexValue(vertex, overridesPathVertex1) as Array<string>)
         : undefined;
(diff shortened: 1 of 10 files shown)


packages/ag-charts-community/src/module/optionsGraphOperations.ts
```diff
@@ -50,11 +50,11 @@ type OperationResolver = (
     values: Array<VertexInterface>
 ) => unknown;
 
-export function getOperation(value: unknown) {
+export function getOperation(value: unknown, keys?: Array<string>) {
     if (value == null || typeof value !== 'object' || Array.isArray(value)) return;
 
     // Get the first key more efficiently than Object.keys()
-    const keys = Object.getOwnPropertyNames(value);
+    keys ??= Object.keys(value);
```
lsjroberts, line 56: The comment above is out of date? Why was this change needed?

packages/ag-charts-community/src/module/optionsGraphOperations.ts
```diff
@@ -701,13 +704,16 @@ function applyOperation(graph: OptionsGraphInterface, vertex: VertexInterface, v
...
     }
 
-    for (const child of children) {
-        if (graph.neighboursWithEdgeValue(child, PATH_EDGE).length === 0) {
-            // Add a stub if we are applying to a child object with no keys, e.g. `gridLine: { styles: [{}] }`
-            const stubVertex = graph.addVertex({});
-            graph.addEdge(child, stubVertex, DEFAULTS_EDGE);
-        } else {
-            graph.graftObject(child, object, [overridesPath1, overridesPath2]);
+    if (children) {
```
lsjroberts: perhaps `if (!children) return RESOLVED_TO_BRANCH;`?

lsjroberts (approved):
LGTM

packages/ag-charts-community/src/module/optionsGraphOperations.ts
alantreadway: `Object.getOwnPropertyNames(value)` as an optimisation was an AI hallucination it turned out :P

The improvement here is now from the use of `??=` avoiding to make the call at all in some cases.

packages/ag-charts-community/src/module/optionsGraphOperations.ts
alantreadway: Fixed

packages/ag-charts-community/src/module/optionsGraphOperations.ts
alantreadway: Removed comment as it's redundant.

PR #4565, opened 2025-06-25
AG-15010 Ordinal time axis parent level

jacobp100:
https://ag-grid.atlassian.net/browse/AG-15010

https://github.com/ag-grid/ag-charts/pull/4567 - contains first part of this PR (to make it easier to review)


packages/ag-charts-community/src/scale/__snapshots__/ordinalTimeScale.test.ts.snap
```diff
@@ -16,10 +16,10 @@ exports[`OrdinalTimeScale should create nice ticks when reversed 1`] = `
 {
   "count": undefined,
   "ticks": [
-    2024-03-06T00:00:00.000Z,
-    2024-03-04T00:00:00.000Z,
-    2024-02-29T00:00:00.000Z,
     2024-02-27T00:00:00.000Z,
+    2024-02-29T00:00:00.000Z,
+    2024-03-04T00:00:00.000Z,
+    2024-03-06T00:00:00.000Z,
```
jacobp100, line 22: This just reversed the ticks (but we never offered guarantees on this)

packages/ag-charts-community/src/chart/crossline/__image_snapshots__/cross-line-test-ts-cross-line-create-for-range-flipped-outside-domain-min-crosslines-it-should-render-to-canvas-as-expected-1-snap.png
jacobp100, line 1: Crosslines range that didn't match a date in the unit and ordinal would always either round down or up, respectively. This has changed so the crossline grows to include the dates - which is what you see here

alantreadway (approved):
LGTM

PR #4570, opened 2025-06-25
CRT-831: Array union inside a type union resolves incorrectly

iMoses:
https://ag-grid.atlassian.net/browse/CRT-831


alantreadway (approved):
LGTM

PR #4571, opened 2025-06-25
CRT-852: union with primitives and interfaces cannot be expanded

iMoses:
+ remove empty doc strings
https://ag-grid.atlassian.net/browse/CRT-852


alantreadway (approved):
LGTM

PR #4574, opened 2025-06-25
AG-15173 Always call `preventDefault()` on navigator `<input>` elems

olegat:
https://ag-grid.atlassian.net/browse/AG-15173


alantreadway (approved):
LGTM

PR #4578, opened 2025-06-26
AG-14174 Improve logic around zooming ordinal time axes where dates d…

jacobp100:
…o not exist

https://ag-grid.atlassian.net/browse/AG-14174


alantreadway (approved):
LGTM

PR #4585, opened 2025-06-26
AG-15292: Grouped Category Axis does not display labels below a depth…

iMoses:
… that has `enabled: false`

https://ag-grid.atlassian.net/browse/AG-15292


alantreadway (approved):
LGTM

PR #4587, opened 2025-06-26
AG-15303 `aglint/require-explicit-generic` test and bugfix

olegat:
https://ag-grid.atlassian.net/browse/AG-15303


libraries/ag-charts-eslint-rules/rules/require-explicit-generic.mjs
```diff
@@ -44,19 +45,19 @@ export default {
...
                     if (symbol) {
-                        genericTypeNames.add(symbol.getName());
+                        genericTypeNames.set(symbol.getName(), node.typeParameters.length);
                     }
                 }
             });
         }
 
         // Helper to check and report if type ref is missing generics
         function checkTypeReference(node, typeName, typeArguments) {
```
olegat, line 56: The `checkTypeReference` function had this bug:

```ts
type Foo<A = number, B = number> = { a: A, b: B };
let x: Foo;          // lint error (expected).
let y: Foo<string>;  // no lint error (not expected).
```

This changes fixes this by ensuring that the type reference `typeArguments.length` is equal to the type declaration `typeParameters.length`.

alantreadway (approved):
LGTM

PR #4593, opened 2025-06-27
AG-15228 A11y bugs with overlays and async data loads

olegat:
https://ag-grid.atlassian.net/browse/AG-15228

### DIFF
--- a/packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
+++ b/packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
@@ -616,7 +616,7 @@ export class SeriesAreaManager extends BaseManager {
             if (this.handleSeriesFocus(seriesIndexDelta, datumIndexDelta) === PickedFocusStatus.SUCCESS) {
                 this.announceMode = 'when-changed';
             } else {
-                // As a safe-guard, always announce the next focus-change if this currenct focus-change failed.
+                // As a safe-guard, always announce the next focus-change if this current focus-change failed.
                 this.announceMode = 'always';
             }
         } else {
--- a/packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
+++ b/packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
@@ -785,9 +785,9 @@ export class SeriesAreaManager extends BaseManager {
         tooltipContent: TooltipContent[]
     ) {
         const { focus } = this;
-        let mode = this.announceMode;
+        let mode: 'always' | 'never';
 
-        if (mode === 'when-changed') {
+        if (this.announceMode === 'when-changed') {
             // AG-13874 If all deltas are 0, it means that we're tabbing in (always announce). Otherwise, announce
             // the datum pick only if the indices have changed.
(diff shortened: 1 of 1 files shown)


packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
```diff
@@ -593,20 +613,26 @@ export class SeriesAreaManager extends BaseManager {
...
-            this.handleSeriesFocus(seriesIndexDelta, datumIndexDelta);
+            if (this.handleSeriesFocus(seriesIndexDelta, datumIndexDelta) === PickedFocusStatus.SUCCESS) {
+                this.announceMode = 'when-changed';
+            } else {
+                // As a safe-guard, always announce the next focus-change if this currenct focus-change failed.
+                this.announceMode = 'always';
+            }
         } else {
             this.focusIndicator?.update(overlayFocus.rect, this.seriesRect, false);
+            this.swapChain.update(overlayFocus.text);
```
olegat, line 624: > [!IMPORTANT]
> This is the initial bugfix. It ensures that the aria-label is updated whenever the overlay focused or updated.

packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
```diff
@@ -735,18 +761,49 @@ export class SeriesAreaManager extends BaseManager {
...
+                oldDatumIndex !== pick.datumIndex ||
+                oldOtherIndex !== (pick.otherIndex ?? focus.seriesIndex);
+            if (shouldAnnouncePick) {
+                mode = 'always';
+            } else {
+                mode = 'never';
+            }
+        }
+
+        if ((mode satisfies 'always' | 'never') === 'always') {
```
alantreadway: Nit: Is the `satisfies`/casting needed here? (Seems like it should be derivable automatically?)

alantreadway (approved):

packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
olegat: It's not needed, but it's a useful sanity-check.

It tells compiler to make sure that the `if (mode === 'when-changed')` branch changes the type of `mode` from `'when-changed' | 'always' | 'never'` to `'always' | 'never'` (in other words, it checks that `mode` is correctly reassigned to either `'always'` or `'never'`).

This will catch faulty changes such as adding string values to the `FocusAnnounceMode` union or accidentally removing the `if (mode === 'when-changed')` branch.

alantreadway: A more conventional way to encode this would be to create a variable `let mode: 'always' | 'never';` and assign to that in the `if` block - add a new `else` clause do `mode = this.announceMode` which will flag up the mismatch without harming readability.

olegat: The `satisfies`-way is perfectly readable, but happy to change it to a typed variable declaration if that's preferred.

alantreadway (approved):
LGTM, thank for the updates.

PR #4594, opened 2025-06-30
AG-3390 - Exclude mkcert in tests

taktran:
Port over changes from https://github.com/ag-grid/ag-grid/pull/11175 to prevent CI failures


alantreadway (approved):
LGTM

PR #4599, opened 2025-07-01
AG-13179 High performance bar pt 1

jacobp100:


packages/ag-charts-community/src/chart/series/cartesian/barShape.ts
alantreadway, line 1: I'd prefer to keep this with the `scene/shape/` folder as it isn't series specific.

alantreadway (approved):
LGTM

PR #4611, opened 2025-07-02
AG-15233 Set role="textbox" on annotations text input elems

olegat:
https://ag-grid.atlassian.net/browse/AG-15233

This fixes bugs with screenreaders not announcing key presses on text annotations edits.


alantreadway (approved):
LGTM

PR #4612, opened 2025-07-02
AG-15367 Improve heatmap highlight performance

jacobp100:
[AG-15367](https://ag-grid.atlassian.net/browse/AG-15367)

### DIFF
--- a/packages/ag-charts-community/src/chart/series/cartesian/abstractBarSeries.ts
+++ b/packages/ag-charts-community/src/chart/series/cartesian/abstractBarSeries.ts
@@ -175,6 +175,11 @@ export abstract class AbstractBarSeries<
         addHitTestersToQuadtree(quadtree, this.datumNodesIter());
     }
 
+    protected override pickNodesExactShape(point: Point): SeriesNodeDatum<unknown>[] {
+        const item = findQuadtreeMatch(this, point);
+        return item != null && item.distance <= 0 ? [item.datum] : [];
+    }
+
     protected override pickNodeClosestDatum(point: Point): SeriesNodePickMatch | undefined {
         return findQuadtreeMatch(this, point);
     }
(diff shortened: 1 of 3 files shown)


olegat (approved):
The logic is correct, but the quad-tree internally uses distances-squared to avoid computing square roots (which is slow). Therefore, `findQuadtreeMatch().distance` is calculated using [`Math.sqrt()`](https://github.com/ag-grid/ag-charts/blob/6f633f967dffb0edecbcf63e2e6eda830e551176/packages/ag-charts-community/src/chart/series/cartesian/quadtreeUtil.ts#L39), which is wasteful because it is being compared to ≤ 0.

Although performance impact of this wasteful `Math.sqrt` is almost certainly negligible, consider adding a `findQuadtreeExactMatch` variation of `findQuadtreeMatch` that checks whether the distance-squared is 0.

packages/ag-charts-community/src/chart/series/cartesian/abstractBarSeries.ts
```diff
@@ -175,6 +175,11 @@ export abstract class AbstractBarSeries<
         addHitTestersToQuadtree(quadtree, this.datumNodesIter());
     }
 
+    protected override pickNodesExactShape(point: Point): SeriesNodeDatum<unknown>[] {
```
alantreadway, line 178: This seems to impact all bar series, not just heatmap?

jacobp100: It already had the `pickClosest` - so I think this was just an omission

alantreadway:
> Although performance impact of this wasteful `Math.sqrt` is almost certainly negligible, consider adding a `findQuadtreeExactMatch` variation of `findQuadtreeMatch` that checks whether the distance-squared is 0.

Discussed offline with @jacobp100, and this is O(1) so I wouldn't be concerned about it at this juncture.

alantreadway (approved):
LGTM

PR #4615, opened 2025-07-02
AG-13179 Data model perf

jacobp100:


packages/ag-charts-community/src/chart/data/dataModel.ts
```diff
@@ -941,13 +938,17 @@ export class DataModel<
     }
 
     private groupData(data: UngroupedData<D>, groupingFn?: GroupingFn<D>): GroupedData<D> {
-        type Group = { keys: unknown[]; datumIndices: number[][]; validScopes: Set<string> };
-        const groups = new Map<string, Group>();
+        type Group = { keys: unknown[]; datumIndices: number[][]; aggregation: any[]; validScopes: Set<string> };
```
jacobp100, line 941: Set `aggregation` so we can avoid duplicating the object just to add this property

packages/ag-charts-community/src/chart/data/dataModel.ts
```diff
@@ -977,17 +978,21 @@ export class DataModel<
...
                         validScopes: allScopes,
-                    } satisfies Group);
-                if (!groups.has(groupStr)) {
-                    groups.set(groupStr, outputGroup);
+                    };
+
+                    groups?.set(groupStr!, outputGroup);
+
+                    resultGroups.push(outputGroup.keys);
+                    resultData.push(outputGroup);
```
jacobp100, line 995: Avoid looping over the `groups` (which may not exist) - just append them as they're created

packages/ag-charts-community/src/chart/data/dataModel.ts
```diff
@@ -977,17 +978,21 @@ export class DataModel<
                 }
 
                 const group = groupingFn?.(keys) ?? keys;
-                const groupStr = toKeyString(group);
+                const groupStr = groups != null ? toKeyString(group) : undefined;
```
jacobp100, line 981: Skip stringifying if it's not needed

alantreadway (approved):
LGTM

PR #4616, opened 2025-07-02
AG-14555 $isUserOption and $isThemeOverride

lsjroberts:
Part of https://ag-grid.atlassian.net/browse/AG-14555 ported out to support https://ag-grid.atlassian.net/browse/AG-9514 and added `$isThemeOverride`.

### DIFF
--- a/packages/ag-charts-community/src/module/optionsGraph.ts
+++ b/packages/ag-charts-community/src/module/optionsGraph.ts
@@ -234,6 +234,25 @@ export class OptionsGraph extends AdjacencyListGraph<unknown, string> implements
         return hasPathSafe(this.userOptions, path);
     }
 
+    hasThemeOverride(path: Array<string>) {
+        if (path[0] === 'axes' && path.length > 1) {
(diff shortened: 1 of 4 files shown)


packages/ag-charts-community/src/module/optionsGraph.ts
```diff
@@ -231,6 +230,29 @@ export class OptionsGraph extends AdjacencyListGraph<unknown, string> implements
         return vertex;
     }
 
+    hasUserOption(path: Array<string>) {
+        return hasPathSafe(this.userOptions, path);
+    }
+
+    hasThemeOverride(path: Array<string>) {
```
alantreadway, line 237: Micro-optimisation: How often is this being invoked? Might be worth making the constant paths here declared statically if so?

lsjroberts: It's not in this PR. I believe it will be used in the highlight defaults, so once or twice per series.

alantreadway (approved):

PR #4622, opened 2025-07-03
AG-13179 Remove need for flag to make bar series high performance

jacobp100:


packages/ag-charts-website/src/content/docs/benchmarks/_examples/summary/data.ts
alantreadway, line 1: Please drop these, I only want to keep a permanent record of `latest` and release branches here, from one consistent piece of hardware (my M4 MacBook Pro at the moment).

jacobp100: DOne

alantreadway (approved):

PR #4623, opened 2025-07-03
AG-15379 use `preventScroll: true` when clicking the series-area

olegat:
https://ag-grid.atlassian.net/browse/AG-15379


alantreadway (approved):
LGTM

PR #4625, opened 2025-07-03
AG-15250 - highlight changes check

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-15250

### DIFF
--- a/packages/ag-charts-community/src/chart/interaction/highlightManager.ts
+++ b/packages/ag-charts-community/src/chart/interaction/highlightManager.ts
@@ -7,16 +7,13 @@ import type { SeriesNodeDatum } from '../series/seriesTypes';
  * distinct dependents and handles conflicting highlight requests.
  */
 export class HighlightManager {
-    private readonly highlightStates = new StateTracker<{
-        highlightedDatum?: HighlightNodeDatum;
-        previousHighlightedDatum?: HighlightNodeDatum;
-    }>();
+    private readonly highlightStates = new StateTracker<HighlightNodeDatum>();
 
     constructor(private readonly eventsHub: EventsHub) {}
 
     public updateHighlight(callerId: string, highlightedDatum?: HighlightNodeDatum): void {
         const previousHighlight = this.getActiveHighlight();
-        this.highlightStates.set(callerId, { highlightedDatum, previousHighlightedDatum: previousHighlight });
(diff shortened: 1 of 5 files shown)


packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts
```diff
@@ -371,6 +371,15 @@ export abstract class CartesianSeries<
         return series === this || (legendItemName != null && legendItemName === activeLegendItemName);
     }
 
+    protected strokewidthChange() {
+        // TODO: might need to check marker.strokeWidth
+        return (
+            'strokeWidth' in this.properties &&
+            this.properties.strokeWidth != null &&
+            this.properties.strokeWidth > this.properties.highlight.highlightedItem?.strokeWidth
```
alantreadway:
```
3. Potential Null Reference (cartesianSeries.ts:379)

  this.properties.strokeWidth > this.properties.highlight.highlightedItem?.strokeWidth

  Issue: If highlightedItem is undefined, accessing strokeWidth will be undefined, and comparing a number
  to undefined may not behave as expected.

  Fix: Should handle the undefined case explicitly:
  this.properties.strokeWidth > (this.properties.highlight.highlightedItem?.strokeWidth ?? 0)
```

packages/ag-charts-community/src/chart/interaction/highlightManager.ts
```diff
@@ -7,21 +7,28 @@ import type { SeriesNodeDatum } from '../series/seriesTypes';
  * distinct dependents and handles conflicting highlight requests.
  */
 export class HighlightManager {
-    private readonly highlightStates = new StateTracker<HighlightNodeDatum>();
+    private readonly highlightStates = new StateTracker<{
+        highlightedDatum?: HighlightNodeDatum;
+        previousHighlightedDatum?: HighlightNodeDatum;
+    }>();
```
alantreadway: On reflection, this is the wrong pattern to use, `StateTracker` has a current state that is a combination of multiple inputs - conflating 'previous' state into the current state doesn't make logical sense.

Instead we should make use of the `highlight:change` event below to know when this has happened.

alantreadway (approved):
LGTM, thanks!

PR #4629, opened 2025-07-07
AG-13179 Bar series perf

jacobp100:

- Don't animate >1k data points
- Don't do animation validation on >1k data points
- Don't fill/stroke shapes if the {fill,stroke}Opacity is 0


alantreadway (approved):
LGTM

PR #4631, opened 2025-07-07
AG-14981 Fix zoom events firing with previous rangeX and rangeY

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14981


alantreadway (approved):
LGTM

PR #4635, opened 2025-07-07
AG-15393 Improve tick generation performance

jacobp100:
About 10% of the bottleneck was down to formatting all ticks - even though in many cases we only needed the rawTicks. The  remaining 90% was caching of callbacks. We do expect in 90+% of cases that the previous iteration's cache will be invalid, because things like the domain will change, so it's unlikely that removing the cache entirely will have much real impact (other than a performance win)

### DIFF
--- a/packages/ag-charts-community/src/chart/axis/axis.ts
+++ b/packages/ag-charts-community/src/chart/axis/axis.ts
@@ -651,6 +651,7 @@ export abstract class Axis<
             truncateDate = tickFormatParams.truncateDate;
         }
 
+        // The serialization required for caching is too slow for large category domains
         const f = this.uncachedCallWithContext.bind(this);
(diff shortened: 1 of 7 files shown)


packages/ag-charts-community/src/chart/axis/axis.ts
```diff
@@ -651,7 +651,7 @@ export abstract class Axis<
             truncateDate = tickFormatParams.truncateDate;
         }
 
-        const f = this.callWithContext.bind(this);
+        const f = this.uncachedCallWithContext.bind(this);
```
alantreadway, line 655: Nit: Could you add comments about why this is needed for future reference?

alantreadway (commented):
/benchmarks

PR #4636, opened 2025-07-07
AG-15221 Add zoom axisDraggingMode for axis panning

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-15221

Also refactors the `onDrag[N]` handlers to separate out series-area and axis logic into `onSeriesAreaDrag[N]` and `onAxisDrag[N]` handlers. There is a little duplication of a few lines but simplifies the logic.


packages/ag-charts-enterprise/src/features/zoom/zoom.ts
```diff
@@ -129,6 +129,15 @@ export class Zoom extends _ModuleSupport.BaseModuleInstance implements _ModuleSu
         this.ctx.zoomManager.setAutoScaleYAxis(newValue.enabled, newValue.padding);
     });
 
+    @ActionOnSet<Zoom>({
+        changeValue(newValue) {
+            if (!this.domProxy) return;
```
alantreadway, line 134: Nit: Maybe just use a `?` on the next line?

alantreadway (approved):
LGTM, @olegat could you review from an interactions perspective?

olegat (approved):
The clean-up in `zoom.ts` to distinguish between series-area and axes dragging is very good.

I've only skimmed through it though, so LGTM assuming that all tests still pass.

PR #4649, opened 2025-07-08
AG-8290/AG-15452 Add `UnknownSeries` type alias

olegat:
https://ag-grid.atlassian.net/browse/AG-8290
https://ag-grid.atlassian.net/browse/AG-15452

### DIFF
--- a/packages/ag-charts-community/src/chart/polarChart.ts
+++ b/packages/ag-charts-community/src/chart/polarChart.ts
@@ -10,7 +10,7 @@ import type { TransferableResources } from './chart';
 import { Chart } from './chart';
 import { ChartAxisDirection } from './chartAxisDirection';
 import type { SeriesArea } from './series-area/seriesArea';
-import { PolarSeries } from './series/polar/polarSeries';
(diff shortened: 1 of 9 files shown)


packages/ag-charts-community/src/chart/polarChart.ts
```diff
@@ -78,7 +78,8 @@ export class PolarChart extends Chart {
     }
 
     private async computeCircle(seriesBox: BBox) {
-        const polarSeries = this.series.filter(isPolarSeries);
+        const anySeries: any[] = this.series;
```
alantreadway: Nit: Could we use `UnknownSeries[]` instead of `any[]` to cleanup `any` use?

alantreadway (commented):
Nice, just one nit

packages/ag-charts-community/src/chart/polarChart.ts
olegat: The `any[]` isn't actually necessary here. So I've removed it.

alantreadway (approved):
LGTM

PR #4650, opened 2025-07-09
AG-7695 Fix legend container

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-7695

### DIFF
--- a/packages/ag-charts-community/src/scene/group.ts
+++ b/packages/ag-charts-community/src/scene/group.ts
@@ -416,10 +416,10 @@ export class Group<D = any> extends Node<D> {
 
     *excludeChildren(exclude: { instance?: any; name?: string }) {
         for (const child of this.children()) {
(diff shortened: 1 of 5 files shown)


packages/ag-charts-community/src/scene/group.ts
```diff
@@ -414,6 +414,17 @@ export class Group<D = any> extends Node<D> {
...
     }
 
+    *excludeChildren(exclude: { instance?: any; name?: string }) {
+        for (const child of this.children()) {
+            if (exclude.instance && !(child instanceof exclude.instance)) {
+                yield child;
+            }
+            if (exclude.name && child.name !== exclude.name) {
+                yield child;
+            }
```
alantreadway: Nit: Should the 2nd if be an `else if` to avoid double yielding the same child?

alantreadway (approved):
LGTM

PR #4653, opened 2025-07-09
AG-15239 - Add zoom info trial form update

taktran:
Same as https://github.com/ag-grid/ag-grid/pull/11274


markdurrant (approved):
LGTM 👍

PR #4655, opened 2025-07-10
AG-15295 Update legend border and padding

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-15295

- Also adds `$isUserOption: [['./path/one', './path/two'], true, false]` for considering multiple paths at the same time.


alantreadway (approved):
LGTM

PR #4665, opened 2025-07-11
AG-15385 Update annotation options and values

alantreadway:
https://ag-grid.atlassian.net/browse/AG-15385


packages/ag-charts-enterprise/src/features/annotations/utils/values.ts
```diff
@@ -59,8 +60,7 @@ export function invert(
     const width = scale.bandwidth === 0 ? scale.step : scale.bandwidth ?? 0;
     const bandStart = scale.convert(value);
     const bandEnd = bandStart + width;
-    const position = clampArray(n, scale.range);
```
manapeirov, line 62: Is it not possible for n to be outside of the scale range?

alantreadway: See TC1 on the linked JIRA

manapeirov (approved):

PR #4666, opened 2025-07-11
AG-15363 Add `legend.floating` option

olegat:
https://ag-grid.atlassian.net/browse/AG-15363

### DIFF
--- a/packages/ag-charts-enterprise/src/gradient-legend/gradientLegend.ts
+++ b/packages/ag-charts-enterprise/src/gradient-legend/gradientLegend.ts
@@ -59,6 +59,9 @@ export class GradientLegend {
     position: AgChartLegendPosition = 'bottom';
 
     @Property
+    floating: boolean = false;
+
+    @Property
     reverseOrder: boolean = false;
 
     @Property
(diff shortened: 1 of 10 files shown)


packages/ag-charts-enterprise/src/gradient-legend/gradientLegend.ts
```diff
@@ -274,29 +277,31 @@ export class GradientLegend {
...
-                shrinkRect.shrink(height + this.spacing, 'top');
-                break;
-            case 'bottom':
-            case 'bottom-left':
-            case 'bottom-right':
-                shrinkRect.shrink(height + this.spacing, 'bottom');
-                break;
-            default:
-                unreachable(this.position);
+        if (!this.floating) {
```
alantreadway, line 280: Nit: Maybe just short-circuit here and remove a nested block with:
```suggestion
        if (this.floating) return { top, left };
```

alantreadway (commented):
One nit, otherwise LGTM

packages/ag-charts-enterprise/src/gradient-legend/gradientLegend.ts
olegat: Let's leave this as-is for now. Some additional computation will be needed to add `xOffset` and `yOffset` so `getMeasurement()` will may need to be broken-up to simply it.

alantreadway (approved):
LGTM

PR #4672, opened 2025-07-14
AG-15473 Fix visible item calculation for ohlc, candlestick

jacobp100:
https://ag-grid.atlassian.net/browse/AG-15473


alantreadway (approved):
LGTM

PR #4687, opened 2025-07-15
AG-15481 Fix console warning when updating axes with interval property

alantreadway:
https://ag-grid.atlassian.net/browse/AG-15481


packages/ag-charts-community/src/util/json.ts
```diff
@@ -264,7 +264,11 @@ export function jsonApply<Target extends object, Source extends DeepPartial<Targ
             }
 
             if (isProperties(currentValue)) {
-                targetAny[property].set(newValue);
+                if (newValue === undefined) {
```
alantreadway, line 267: This is the key change.

iMoses (approved):
LGTM

PR #4692, opened 2025-07-15
AG-14639 QA Feedback 2. Continuous Time

olegat:
https://ag-grid.atlassian.net/browse/AG-14639

AG-14639 Do not draw text boxing on empty text


alantreadway (approved):
LGTM

PR #4707, opened 2025-07-17
AG-15509 - Add security contact to docs and footer link

taktran:


markdurrant (approved):
LGTM 👍

packages/ag-charts-website/src/content/footer/footer.json
```diff
@@ -40,6 +40,10 @@
                 "name": "Support via Zendesk",
                 "url": "https://ag-grid.zendesk.com/",
                 "newTab": true
+            },
+            {
+                "name": "Security",
+                "url": "/r/security"
```
seanlandsman, line 46: is that right? url": "/r/security"

seanlandsman (changes requested):

packages/ag-charts-website/src/content/footer/footer.json
taktran: Yep, it's a redirect link, which goes to the framework the user has currently selected

seanlandsman (approved):
lgtm

PR #4718, opened 2025-07-18
AG-15360 Avoid laggy tooltip caused by tooltip update interferences

olegat:
https://ag-grid.atlassian.net/browse/AG-15360


alantreadway (approved):
LGTM

PR #4726, opened 2025-07-21
AG-9686: Text segments, fix spacing bug + add donut innerLabels support

iMoses:
https://ag-grid.atlassian.net/browse/AG-9686


alantreadway (approved):
LGTM

PR #4727, opened 2025-07-21
AG-15371: First and last labels on number axis can be hidden even if …

iMoses:
…`avoidCollisions: false`
https://ag-grid.atlassian.net/browse/AG-15371


alantreadway (approved):
LGTM

PR #4738, opened 2025-07-21
AG-14639 Include boxing padding in text BBox computation

olegat:
https://ag-grid.atlassian.net/browse/AG-14639

Depends on
- https://github.com/ag-grid/ag-charts/pull/4729


packages/ag-charts-community/src/chart/label.ts
```diff
@@ -124,6 +125,22 @@ export class Label<TParams = never, TDatum = any>
     }
 }
 
+type LabelBoxingMixin = { border?: { enabled?: boolean; stroke?: string }; fill?: unknown; padding?: Padding };
+export function expandLabelPadding(label: LabelBoxingMixin | undefined): Required<PaddingOptions> {
+    const { enabled: borderEnabled = false, stroke: borderStroke = undefined } = label?.border ?? {};
```
iMoses:
```suggestion
    const { enabled: borderEnabled = false, stroke: borderStroke } = label?.border ?? {};
```
nit: A default value only ever applies when the value is undefined, so assigning it a default value of undefined is just redundant.

packages/ag-charts-community/src/chart/axis/__image_snapshots__/grouped-category-axis-test-ts-grouped-category-axis-examples-ag-14639-label-boxing-1-snap.png
iMoses, line 1: It seems like now the box is considered as part of the collision detection, which is good, but also that spacing was added between the axis line and that labels, and that's a bug.

packages/ag-charts-enterprise/src/axes/__image_snapshots__/time-axis-test-ts-time-axis-examples-ag-14639-should-apply-label-options-to-parent-label-1-snap.png
iMoses, line 1: Isn't it a regression? Looks like it was previously better

packages/ag-charts-community/src/scene/shape/text.ts
```diff
@@ -172,13 +172,14 @@ export class Text<D = any> extends Shape<D> {
...
         const { x, y, lines, textBaseline, textAlign, lineHeight } = this;
+        const padding = this.boxing == null ? {} : this.boxPadding;
         return Text.computeBBox(
             lines,
             x,
             y,
             { font: this, textBaseline, textAlign, lineHeight },
             useGlyphIndependentMeasurements
-        );
+        ).grow(padding);
```
iMoses: Text bbox calculation are being made at a very large scale and very often.
I think it's better if we only execute these operations when needed, and try to avoid creating garbage, such as empty object and arrays.
IMO something like this is much better in these delicate places:
```js
const bbox = Text.computeBBox(...);
if (this.boxing) bbox.grow(this.boxPadding);
return bbox;
```

iMoses (commented):
Looks good, left a few notes

packages/ag-charts-enterprise/src/axes/__image_snapshots__/time-axis-test-ts-time-axis-examples-ag-14639-should-apply-label-options-to-parent-label-1-snap.png
olegat: I think this change is expected. The previous snapshot looks better by fluke because the `padding: 5` is very small. If I edit the example to larger value `padding: 20` then it looks very ugly:
Screenshot 2025-07-24 at 12 15 13

packages/ag-charts-community/src/chart/axis/__image_snapshots__/grouped-category-axis-test-ts-grouped-category-axis-examples-ag-14639-label-boxing-1-snap.png
olegat: Well spotted!

AFAICT, this isn't a bug. It just so happens that the `label.padding` and `label.spacing` values both default to `10`. So the new snapshot is the expected output.

`axes[].label.spacing: 0`
Screenshot 2025-07-24 at 12 22 44

`axes[].label.spacing: undefined (default)`
Screenshot 2025-07-24 at 12 23 02

packages/ag-charts-community/src/scene/shape/text.ts
olegat: Done.

packages/ag-charts-enterprise/src/axes/__image_snapshots__/time-axis-test-ts-time-axis-examples-ag-14639-should-apply-label-options-to-parent-label-1-snap.png
alantreadway: Ok, so I see that this example has `padding: 5` set as mentioned, and we need to take padding into account to do a better job here; since the scope is limited (verified by no other snapshot changes) then I think it's ok to proceed.

packages/ag-charts-community/src/chart/axis/__image_snapshots__/grouped-category-axis-test-ts-grouped-category-axis-examples-ag-14639-label-boxing-1-snap.png
iMoses: I see. Make sense. 👍

iMoses (approved):

packages/ag-charts-community/src/chart/series/cartesian/__image_snapshots__/scatter-series-test-ts-scatter-series-gradient-fill-should-render-scatter-series-with-a-vertical-linear-gradient-fill-1-snap.png
olegat, line 1: I do not know why this file has changed, but the diff is very small anyway so I think it's safe.
[Failed Tests.html.zip](https://github.com/user-attachments/files/21412944/Failed.Tests.html.zip)

olegat: This failure is no longer reproducible. Closing.

PR #4752, opened 2025-07-22
AG-15363 Apply `legend.spacing` on floating legends.

olegat:
https://ag-grid.atlassian.net/browse/AG-15363


olegat:
Closing. We've agreed to keep the `legend.spacing` behave as-is. I will create another PR to update the documentation.

PR #4765, opened 2025-07-23
AG-8003 Fix category axis interval between first tick when zoomed

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-8003


alantreadway (approved):
LGTM

PR #4772, opened 2025-07-23
AG-9686: Fix captions positioning

iMoses:


alantreadway (approved):
LGTM, spot-checked a subset of image diffs.

PR #4774, opened 2025-07-23
AG-15540 Fix zoom state restore start value less than end

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-15540


alantreadway (approved):
LGTM

PR #4812, opened 2025-07-25
Update 'Fills & Borders' example for darkmode

markdurrant:
Fixes minor darkmode issues with main 'Fills & Borders' example


AG-DavidG (approved):
LGTM

PR #4821, opened 2025-07-28
CRT-872 Fix image fill background when the image is not loaded

jacobp100:
https://ag-grid.atlassian.net/browse/CRT-872


alantreadway (approved):
LGTM

PR #4822, opened 2025-07-28
CRT-861 - Fix non-floating legend background colour backwards compat.

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-861

### DIFF
--- a/packages/ag-charts-community/src/chart/themes/chartTheme.ts
+++ b/packages/ag-charts-community/src/chart/themes/chartTheme.ts
@@ -456,11 +456,7 @@ export class ChartTheme {
                     label: { color: { $ref: 'textColor' } },
                 },
                 fill: {
-                    $if: [
-                        { $path: ['./position/floating', false] },
-                        { $ref: 'chartBackgroundColor' },
-                        'transparent',
-                    ],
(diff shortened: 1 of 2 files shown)


packages/ag-charts-community/src/chart/themes/chartTheme.ts
```diff
@@ -455,7 +455,13 @@ export class ChartTheme {
                     highlightStyle: { fill: { $ref: 'foregroundColor' } },
                     label: { color: { $ref: 'textColor' } },
                 },
-                fill: { $ref: 'chartBackgroundColor' },
+                fill: {
+                    $if: [
+                        { $eq: [{ $path: './position/floating' }, true] },
+                        { $ref: 'chartBackgroundColor' },
+                        'transparent',
+                    ],
```
lsjroberts: I think this should work and save an operation (the 2nd item in `$path` is a fallback value if the path is not found):

```suggestion
                    $if: [
                        { $path: ['./position/floating', false] },
                        { $ref: 'chartBackgroundColor' },
                        'transparent',
                    ],
```

lsjroberts (approved):
LGTM otherwise

PR #4832, opened 2025-07-28
RTI-2977 - Fix zoom enable disable toggle.

alantreadway:
https://ag-grid.atlassian.net/browse/RTI-2977

### DIFF
--- a/packages/ag-charts-enterprise/src/features/zoom/zoom.ts
+++ b/packages/ag-charts-enterprise/src/features/zoom/zoom.ts
@@ -251,6 +251,7 @@ export class Zoom extends _ModuleSupport.BaseModuleInstance implements _ModuleSu
         this.ctx.zoomManager.setZoomModuleEnabled(false);
         this.buttons.destroy();
         this.destroyContextMenuActions?.();
+        this.domProxy.destroy();
     }
 
     private onEnabledChange(enabled: boolean) {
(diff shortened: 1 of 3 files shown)


packages/ag-charts-enterprise/src/features/zoom/zoom.ts
```diff
@@ -251,6 +251,7 @@ export class Zoom extends _ModuleSupport.BaseModuleInstance implements _ModuleSu
         this.ctx.zoomManager.setZoomModuleEnabled(false);
         this.buttons.destroy();
         this.destroyContextMenuActions?.();
+        this.domProxy.destroy();
```
alantreadway, line 254: This is the primary fix, but this fix then reveals that toggling on->off->on also doesn't work since the `DOMManager` doesn't allow a re-used of the element ID, requiring the `Widget.destroyListener()` workaround.

olegat: The addition of the `destroyListener?` callback is unusual, but it's fine for now. Please create a backlog ticket (assigned to me) to refactor `DOMManager` to follow the same parent-child mechanism as the `Widget` class.

olegat (approved):

packages/ag-charts-enterprise/src/features/zoom/zoom.ts
alantreadway: :+1: Raised https://ag-grid.atlassian.net/browse/AG-11978 to follow-up

PR #4838, opened 2025-07-28
CRT-893 Fix legend item marker widths to be calculated per item

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-893


alantreadway (approved):
LGTM

PR #4841, opened 2025-07-28
CRT-868 - Fix node selection on category-like axis when error-bars enabled.

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-868


manapeirov (approved):
LGTM thank you :)

PR #4844, opened 2025-07-28
CRT-906 Fix zoom button tooltip default

olegat:
https://ag-grid.atlassian.net/browse/CRT-906

### DIFF
--- a/packages/ag-charts-enterprise/src/features/zoom/zoomToolbar.ts
+++ b/packages/ag-charts-enterprise/src/features/zoom/zoomToolbar.ts
@@ -135,6 +135,21 @@ export class ZoomToolbar extends BaseProperties {
         const { buttons, container } = this;
         const { rect } = event.series;
 
+        // CRT-906 Fix the tooltip defaults when using custom zoom toolbars
+        for (const b of buttons) {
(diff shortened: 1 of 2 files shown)


packages/ag-charts-enterprise/src/features/zoom/zoomToolbar.ts
```diff
@@ -135,6 +135,21 @@ export class ZoomToolbar extends BaseProperties {
...
+                    'pan-left': 'toolbarZoomPanLeft',
+                    'pan-right': 'toolbarZoomPanRight',
+                    'pan-start': 'toolbarZoomPanStart',
+                    'zoom-in': 'toolbarZoomZoomIn',
+                    'zoom-out': 'toolbarZoomZoomOut',
+                    reset: 'toolbarZoomReset',
+                } as const;
+                b.tooltip = map[b.value];
+            }
+        }
```
olegat, line 152: > [!NOTE] 
> I tried to implement this by editing the `themeTemplate` of the zoom module to use the `$map` graph operator, but I couldn't get it to work correctly. For now, let's fix the tooltip defaults using a regular for-loop to ensure that the `value` and `tooltip` properties match.

alantreadway (approved):

PR #4849, opened 2025-07-29
CRT-904: Axis labels aren't removed when there are collisions

iMoses:
https://ag-grid.atlassian.net/browse/CRT-904


alantreadway (approved):
LGTM

PR #4851, opened 2025-07-29
AG-15595 Use localised aria label "Target" for gauge targets

olegat:
https://ag-grid.atlassian.net/browse/AG-15595


alantreadway (approved):
LGTM

PR #4853, opened 2025-07-29
AG-15607 Fix focus indicator on Simple Bullet gallery example

olegat:
https://ag-grid.atlassian.net/browse/AG-15607


packages/ag-charts-enterprise/src/series/linear-gauge/linearGaugeSeries.ts
olegat, line 1: > [!IMPORTANT]
> This file includes the entirety of the fix. The rest of the PR is all related to e2e tests.

packages/ag-charts-website/e2e/keyboard-nav.spec.ts-snapshots/linear-gauge-corners-segmented-container-chromium-linux.png
olegat, line 1: I've verified that this snapshot update is correct

> These new snapshots are correct because the old snapshots computed the width as
> `last.x + last.width`, which is incorrect because `first.x` is 1 in this
> example. The correct width computation should have been:
> `last.x + last.width - first.x`
> 
> -- https://github.com/ag-grid/ag-charts/commit/f19532fffc64145376e50b1f17ddd023ded9a97a

alantreadway (approved):
LGTM

PR #4867, opened 2025-07-30
AG-3390 - Patch website

taktran:
## Changes

* Update changelog default sorting
* Fix redirect with search params
* Update privacy with new address

Similar to https://github.com/ag-grid/ag-grid/pull/11565


markdurrant (approved):
LGTM 👍

PR #4868, opened 2025-07-30
AG-13731 Fix visible range calculation and aggregation for bubble, sc…

jacobp100:
…atter

https://ag-grid.atlassian.net/browse/AG-13731


alantreadway (approved):
LGTM

PR #4874, opened 2025-07-31
AG-15597 A11y: Announce both title&value when using on Font Size

olegat:
https://ag-grid.atlassian.net/browse/AG-15597


olegat:
This bug report is invalid. Closing

PR #4935, opened 2025-08-08
AG-14511 Fix donut animations

jacobp100:
https://ag-grid.atlassian.net/browse/AG-14511
https://ag-grid.atlassian.net/browse/AG-15716


alantreadway (approved):
LGTM, but lets follow-up with tests around the cluster of issues we've had related to `DonutSeries.getDatumId()`.

PR #4939, opened 2025-08-08
AG-15592 Part 6: Close sibling buttons in toolbars on `expand-widget`

olegat:
https://ag-grid.atlassian.net/browse/AG-15592

> [!IMPORTANT]
> **Key Changes**
> 1. Add `expanded?` property to `toolbar.ts`
> 2. Call `expanded?.collapse({ mode: CollapseMode.SIBLING_OPENED })` when expanding a button (similar to how `MenuWidget` closes sibling popups).

### DIFF
--- a/packages/ag-charts-community/src/components/toolbar/toolbar.ts
+++ b/packages/ag-charts-community/src/components/toolbar/toolbar.ts
@@ -6,7 +6,8 @@ import type { ModuleContext } from '../../module/moduleContext';
 import { BBox } from '../../scene/bbox';
 import { Listeners } from '../../util/listeners';
 import { BaseProperties } from '../../util/properties';
-import type { ExpansionControllerWidget } from '../../widget/expandableWidget';
+import { CollapseMode } from '../../widget/collapseMode';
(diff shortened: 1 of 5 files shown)


packages/ag-charts-community/src/components/toolbar/toolbar.ts
```diff
@@ -179,6 +182,14 @@ export abstract class BaseToolbar<
             const params: ToolbarEventMap<ButtonOptions>['button-focused'] = { button: { index } };
             this.events.dispatch('button-focused', params);
         });
+        buttonWidget.addListener('expand-controlled-widget', (e) => {
+            this.expanded?.collapse({ mode: CollapseMode.SIDLING_OPENED });
+            this.expanded = e.controlled;
+            const removeListener = this.expanded.addListener('collapse-widget', () => {
+                this.expanded = undefined;
+                removeListener();
```
olegat, line 190: > [!NOTE]
> I think that this code is fine for now. But if this pattern becomes very common then we may want to consider adding a `{once: true}` option to `addListener`.

alantreadway (approved):
LGTM

PR #4946, opened 2025-08-12
AG-11673 Create bar-series level stylers

olegat:


packages/ag-charts-community/src/chart/series/cartesian/barSeries.ts
```diff
@@ -704,43 +707,136 @@ export class BarSeries extends AbstractBarSeries<
...
+        return {
+            cornerRadius: stylerResult.cornerRadius ?? cornerRadius,
+            fill: stylerResult.fill ?? fill,
+            fillOpacity: stylerResult.fillOpacity ?? fillOpacity,
+            lineDash: stylerResult.lineDash ?? lineDash,
+            lineDashOffset: stylerResult.lineDashOffset ?? lineDashOffset,
+            opacity: 1,
+            stroke: stylerResult.stroke ?? stroke,
+            strokeOpacity: stylerResult.strokeOpacity ?? strokeOpacity,
+            strokeWidth: stylerResult.strokeWidth ?? strokeWidth,
```
alantreadway, line 810: We should be able to avoid needing to encode the fallback `??` here, the theme processing can wrap the callback function and perform fallback handling to these values (we're using this approach in a bunch of cases to try and reduce complexity in the series) - maybe sync with @lsjroberts about how to do that?

lsjroberts (approved):
LGTM

PR #4947, opened 2025-08-12
Fix rect clipping code

jacobp100:
- Fixes Rect clipping when no corner radius is applied to correctly clip the rect
- Fixes `BBox#intersection` with infinite BBoxes
- Fixes linear gauge calculations which were incorrect but masked by incorrect rect clipping code


alantreadway (approved):
LGTM

PR #4949, opened 2025-08-12
AG-14510 - Load homepage gallery examples on the page without iframes

taktran:
## Changes

* Build a `/homepage/examples.js` file that combines all the gallery examples, with the create method swapped out for a global update function. Load it on the homepage along with the charts library, and implement the global update function to switch out the chart
* Remove old example runner component
* Create `/debug/homepage-examples.json` for debugging homepage examples


plugins/ag-charts-generate-example-files/src/executors/generate/generator/utils/getDarkModeSnippet.ts
```diff
@@ -63,8 +64,9 @@ if (darkmode) {
     }
 }
 window.addEventListener('message', (event) => {
-    if (event.data && event.data.type === 'color-scheme-change') {
-        darkmode = event.data.darkmode;
+    const data = event.data || event.detail;
```
taktran, line 67: Updated, so the message works on the same page and through an iframe

plugins/ag-charts-generate-example-files/src/executors/generate/generator/utils/getDarkModeSnippet.ts
```diff
@@ -1,3 +1,4 @@
+// NOTE: This is duplicated from `external/ag-website-shared/src/utils/getDarkModeSnippet.ts`
```
taktran, line 1: I needed to duplicate this in `ag-website-shared` because I can't load files outside of plugins here

alantreadway (approved):
LGTM

PR #4952, opened 2025-08-12
AG-15565 Add crosshair test

jacobp100:
https://ag-grid.atlassian.net/browse/AG-15565


packages/ag-charts-enterprise/src/features/crosshair/crosshair.test.ts
```diff
@@ -330,4 +330,55 @@ describe('Crosshair', () => {
         await hoverAction(300, 300)(chart);
         await compare();
     });
+
+    it('AG-15565', async () => {
```
alantreadway: Nit: Can you also include a human-readable description so the expectation for this test-case are clearer?

alantreadway (approved):
LGTM

PR #4959, opened 2025-08-13
AG-14555 Fix itemStylers with partial resolve on options graph

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14555


alantreadway (approved):
LGTM

PR #4960, opened 2025-08-13
AG-15641 - Spruce up more examples.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-15641


packages/ag-charts-website/src/components/gallery/utils/modifiedExamples.ts
```diff
@@ -0,0 +1,37 @@
+import { getIsDev } from '@utils/env';
+import { execSync } from 'child_process';
+
+/**
+ * Gets a set of gallery example names that have been modified compared to the main branch
+ */
+export function getModifiedGalleryExamples(): Set<string> {
+    if (!getIsDev()) {
```
taktran: Cool - so the `isModified` dot should only be shown in dev right?

taktran (approved):
LGTM 👍🏼 Just a minor q

packages/ag-charts-website/src/components/gallery/utils/modifiedExamples.ts
alantreadway: Correct

alantreadway: I've broken this change out into https://github.com/ag-grid/ag-charts/pull/4999 since we're having to tackle review in batches, and that is the first batch to actually be ready for merge.

PR #4961, opened 2025-08-14
AG-11673 Create area-series level stylers

olegat:


packages/ag-charts-community/src/chart/optionsGraphService.ts
```diff
@@ -1,15 +1,25 @@
 import type { PlainObject } from 'ag-charts-core';
 
-export type ResolvePartialCallback = (path: Array<string>, partialOptions?: PlainObject) => any;
+type Resolved = Pick<PlainObject, string> | undefined;
+type ResolvePartialOpts = { pick?: boolean };
+
+interface Graph {
+    resolvePartial(path: Array<string>, partialOptions?: PlainObject, opts?: ResolvePartialOpts): Resolved;
+    hasUserOption(path: Array<string>): boolean;
```
lsjroberts: As discussed - let's remove this option as its risky to expose and hopefully not needed with our bug fixes.

packages/ag-charts-community/src/module/optionsGraph.ts
```diff
@@ -246,7 +246,7 @@ export class OptionsGraph extends AdjacencyListGraph<unknown, string> implements
      * Resolve partial options against the existing graph at a given path without overriding the existing user values.
      * Returns an object with only those keys that were also present within `partialOptions`.
      */
-    resolvePartial(path: Array<string>, partialOptions?: PlainObject) {
+    resolvePartial(path: Array<string>, partialOptions?: PlainObject, opts?: { pick?: boolean }) {
```
lsjroberts: `pick` option makes sense. For context, this solves the issue with `strokeWidth` being dependent on `stroke` and needing to get back all the changed values in some uses. Compared to the other cases which wanted to not override the existing values.

packages/ag-charts-community/src/module/optionsGraph.ts
```diff
@@ -305,7 +312,10 @@ export class OptionsGraph extends AdjacencyListGraph<unknown, string> implements
         // the original object and must be found in the graph.
         const pathVertex = this.findVertexAtPath(path);
         if (pathVertex) {
-            return this.findNeighbour(pathVertex, USER_OPTIONS_EDGE) != null;
+            return (
+                this.findNeighbour(pathVertex, USER_OPTIONS_EDGE) != null ||
+                this.findNeighbour(pathVertex, USER_PARTIAL_OPTIONS_EDGE) != null
+            );
```
lsjroberts: As discussed - this bug should be fixed by the similar changes made here https://github.com/ag-grid/ag-charts/pull/4959/files#diff-27b8479ea6fa70d9276fa0fb48c97885b34220e6c9169f5b4fafb808d98b4757R414-R417

lsjroberts (approved):
OptionsGraph changes LGTM with changes as discussed.

packages/ag-charts-community/src/module/optionsGraph.ts
olegat: This change was reverted in favour of the existing fix in the other PR

packages/ag-charts-community/src/chart/optionsGraphService.ts
olegat: Done. `hasUserOption` has been removed. Instead, `AreaSeries.updateDatumSelection` resolves the `marker.enabled` property using the graph service https://github.com/ag-grid/ag-charts/commit/d005bbf86d8fccb45f3a1ce2330a26e7c6f15e7f

PR #4965, opened 2025-08-14
Improve benchmark memory reliability

alantreadway:


libraries/ag-charts-test/src/benchmarks/timing.ts
```diff
@@ -90,13 +106,42 @@ function getTotalMemoryUsage(memoryStats: NonNullable<BenchmarkMeasurement['memo
     );
 }
 
-function getRelativeMemoryUsage(memoryStats: NonNullable<BenchmarkMeasurement['memory']>): number {
-    const jsHeapSize = Math.max(0, memoryStats.after.heapUsed - memoryStats.before.heapUsed);
```
alantreadway, line 94: This was probably the source of the mis-calculations, as negative memory swings were `0`ed here.

packages/ag-charts-enterprise/benchmarks/area.test.ts
```diff
@@ -15,7 +15,7 @@ describe('area series benchmark', () => {
     benchmark(
         'initial load',
         ctx,
-        { expectedRelativeMB: 230, expectedCanvasCount: 4, autoSnapshot: false },
+        { expectedRelativeMB: 7, expectedCanvasCount: 4, autoSnapshot: false },
```
jacobp100: Are these high perf tests run with a warm up cycle?

jacobp100 (approved):

packages/ag-charts-enterprise/benchmarks/area.test.ts
alantreadway: Good shout, investigating.

alantreadway: Digging in, yes these are now run with a single warmup, then we take a measurement - I wonder if we were also measuring the example dataset prior to loading into a chart, let me have a deeper investigate.

PR #4981, opened 2025-08-19
Refactor tick generation (Part 1)

iMoses:


packages/ag-charts-community/src/chart/label.ts
```diff
@@ -293,28 +293,23 @@ export function timeIntervalMaxLabelSize(
...
-        const width = metrics.width + xPadding;
-        const height = metrics.height + yPadding;
+        const width = textMetrics.width + xPadding;
+        const height = textMetrics.height + yPadding;
+
         labelData.push({
-            point: { x, y },
-            label: { text, width, height },
+            label: tickLabel,
+            bounds: { x, y, width, height },
```
jacobp100, line 312: Why not a bbox?

jacobp100 (approved):

alantreadway (approved):

packages/ag-charts-community/src/chart/label.ts
iMoses: There's just no need to initiate an instance, we don't really need the BBOX implementation for a simple collision detection

PR #5028, opened 2025-08-26
AG-15594 Fix vanilla theme docs to state immutable theme

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-15594


AG-DavidG (approved):
LGTM

PR #5046, opened 2025-08-27
AG-11672 Add Pie/Donut `calloutLine.itemStyler` property

olegat:
https://ag-grid.atlassian.net/browse/AG-11672


lsjroberts (approved):
LGTM

PR #5048, opened 2025-08-28
AG-15798 - boxplot series segmented styles

manapeirov:
https://ag-grid.atlassian.net/browse/AG-15798


packages/ag-charts-enterprise/src/series/box-plot/boxPlotNode.ts
```diff
@@ -224,8 +224,11 @@ export class BoxPlotNode extends Scalable(Path) {
         }
     }
 
-    override drawPath(ctx: CanvasRenderingContext2D | OffscreenCanvasRenderingContext2D) {
-        super.drawPath(ctx);
+    override drawPath(
+        ctx: CanvasRenderingContext2D | OffscreenCanvasRenderingContext2D,
+        path: Path2D = this.path.getPath2D()
```
olegat: Optional: it's unclear whether this `path` parameter is still necessary

The 2-param `path: child.path.getPath2D()` usage in `segmentedGroup.ts` has been removed and the 2-params usage in `segmentedPath.ts` uses the same value the default `path: this.path.getPath2D()`.

Consider removing the 2nd parameter `path` if it's no longer needed

olegat (approved):
LGTM

PR #5056, opened 2025-08-28
Gha/snapshots ag 15784 upgrade astro

taktran:
NOTE: The snapshot is incorrect, but merging to unblock for now


jacobp100 (approved):
LGTM

PR #5061, opened 2025-08-29
AG-15824 Improve scene stats for high perf examples

jacobp100:
https://ag-grid.atlassian.net/browse/AG-15824

### DIFF
--- a/packages/ag-charts-community/src/scene/sceneDebug.ts
+++ b/packages/ag-charts-community/src/scene/sceneDebug.ts
@@ -42,8 +42,6 @@ function memoryUsage() {
 }
 
 function debugFps(debugSplitTimes: Record<string, number>, ctx: CanvasRenderingContext2D, seriesRect = BBox.zero) {
-    if (!Debug.check(DebugSelectors.SCENE_STATS, DebugSelectors.SCENE_FPS)) return;
-
     const end = performance.now();
     const { start } = debugSplitTimes;
     const duration = Math.round(end - start);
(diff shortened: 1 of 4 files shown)


iMoses (approved):
LGTM

packages/ag-charts-community/src/scene/sceneDebug.ts
```diff
@@ -48,6 +85,11 @@ export function debugStats(
     extraDebugStats = {},
     seriesRect = BBox.zero
 ) {
+    if (!Debug.check(DebugSelectors.SCENE_STATS, DebugSelectors.SCENE_FPS)) {
```
lsjroberts, line 86: This condition is the same as at the top of `debugFps()` so will it not always just return out of that too?

jacobp100: Yep good catch. Forgot to remove that after refactoring

packages/ag-charts-community/src/scene/sceneDebug.ts
```diff
@@ -40,6 +41,42 @@ function memoryUsage() {
     return `Heap ${result.join(' / ')}`;
 }
 
+function debugFps(debugSplitTimes: Record<string, number>, ctx: CanvasRenderingContext2D, seriesRect = BBox.zero) {
+    if (!Debug.check(DebugSelectors.SCENE_STATS, DebugSelectors.SCENE_FPS)) return;
+
+    const end = performance.now();
```
jacobp100:
```suggestion
    const end = performance.now();
```

lsjroberts (approved):
LGTM

PR #5077, opened 2025-09-02
AG-15798 - Range area and range bar series segmentation

manapeirov:
https://ag-grid.atlassian.net/browse/AG-15798


packages/ag-charts-enterprise/src/series/range-area/rangeArea.ts
```diff
@@ -346,6 +347,15 @@ export class RangeAreaSeries extends _ModuleSupport.CartesianSeries<
             return interpolatePoints(lowPoints, interpolation);
         });
 
+        const segments = calculateSegments(
+            this.properties.segmentation,
+            xAxis,
+            yAxis,
+            this.chart!.seriesRect!,
```
olegat: consider checking `this.chart.seriesRect!=null` at the start of `createNodeData()` instead of using `!` here.

packages/ag-charts-enterprise/src/series/range-bar/rangeBarSeries.ts
```diff
@@ -250,6 +251,14 @@ export class RangeBarSeries extends _ModuleSupport.AbstractBarSeries<
 
         const itemId = `${yLowKey}-${yHighKey}`;
 
+        const segments = calculateSegments(
+            this.properties.segmentation,
+            xAxis,
+            yAxis,
+            this.chart!.seriesRect!,
```
olegat: consider checking `this.chart.seriesRect!=null` at the start of `createNodeData()` instead of using `!` here.

olegat (approved):
LGTM, with optional nits.

PR #5082, opened 2025-09-02
AG-11673 QA Feedback (TC3)

olegat:

> [!IMPORTANT]
> **Reviewer Notes** - Key Changes:
> -   Change `hasItemStylers()` to return `true` when `styler` callbacks are defined.
> -   Uses `cachedCallWithContext` instead of `callWithContext` for calling `styler` callbacks.
> -   Updates test snapshots.


olegat:
> [!IMPORTANT]
> **Reviewer Notes** - Key Changes:
> -   Removes the `styles` property (type `SeriesNodeStyleContext`) from `_contextNodeData`
> -   Uses `cachedCallWithContext` instead of `callWithContext` for calling `styler` callbacks.

packages/ag-charts-community/src/chart/series/series.ts
```diff
@@ -153,6 +153,7 @@ export type SeriesNodeDataContext<I extends DatumIndexType, S = SeriesNodeDatum<
     labelData: L[];
 };
 
+// TODO(AG-15782) remove this type (still used by Enterprise series)x
```
olegat: > [!CAUTION]
> Removing the use of `SeriesNodeStyleContext` could have a significant impact on performance. Some of the changes of this PR may need to be reverted.

olegat: Obsolete - this change has been reverted.

manapeirov (approved):
LGTM thank you Oli :)

PR #5090, opened 2025-09-02
first draft of large dataset docs

AG-DavidG:


packages/ag-charts-website/src/content/docs/large-dataset-interactivity/_examples/ordered-data/main.ts
```diff
@@ -0,0 +1,164 @@
...
+                    yKey: 'y',
+                    fillOpacity: 0.2,
+                    strokeOpacity: 0.2,
+                },
+            ];
+            options.data = getBubbleData(datapoints);
+            options.axes = numberAxes;
+            break;
+        case 'bubble':
+            series = [
```
manapeirov, line 126: why don't we do options.series in these blocks too?

manapeirov (approved):
LGTM

PR #5092, opened 2025-09-03
AG-15644 Fix axis prediction when mixed of some and no prediction

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-15644


manapeirov (approved):
LGTM

PR #5094, opened 2025-09-03
AG-15765 Fix zoom min vis items for horizontal bar

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-15765

### DIFF
--- a/packages/ag-charts-enterprise/src/features/__snapshots__/features.test.ts.snap
+++ b/packages/ag-charts-enterprise/src/features/__snapshots__/features.test.ts.snap
@@ -209,7 +209,7 @@ HTMLCollection [
         />
       </div>
       <div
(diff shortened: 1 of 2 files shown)


packages/ag-charts-enterprise/src/features/__snapshots__/features.test.ts.snap
```diff
@@ -209,7 +209,7 @@ HTMLCollection [
         />
       </div>
       <div
-        aria-disabled="true"
+        aria-disabled="false"
```
lsjroberts, line 212: Manual testing confirms this was a bug, and it now enables/disables the 'zoom to here' option when it should depending on the location right-clicked and the current zoom.

manapeirov (approved):
LGTM

PR #5096, opened 2025-09-03
AG-15675 Fix horizontal bars to explicitly disable autoscale

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-15765


packages/ag-charts-enterprise/src/features/zoom/__image_snapshots__/zoom-test-ts-zoom-flipped-axes-should-zoom-to-minimum-visible-items-1-snap.png
olegat, line 1: Can you confirm whether the added Y-offset at the origin is intentional?

olegat (approved):
LGTM, assuming that the updated snapshot is intentional

packages/ag-charts-enterprise/src/features/zoom/__image_snapshots__/zoom-test-ts-zoom-flipped-axes-should-zoom-to-minimum-visible-items-1-snap.png
lsjroberts: I _believe_ previously it was using the wrong dimension of the chart to calculate the pixel visible range against. The width when it should have been using the height, and it coincidentally matched up.

PR #5098, opened 2025-09-03
AG-15644 Fix axis prediction when mixed and first datum is different

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-15644


iMoses (approved):
LGTM

PR #5100, opened 2025-09-03
AG-15580: Allow multiple font styles in series and axis labels (fixes)

iMoses:
https://ag-grid.atlassian.net/browse/AG-15580

### DIFF
--- a/packages/ag-charts-core/src/utils/math/shapeUtils.ts
+++ b/packages/ag-charts-core/src/utils/math/shapeUtils.ts
@@ -42,6 +42,35 @@ export function getMaxInnerRectSize(
     };
 }
 
+/**
+ * Calculates the minimum axis-aligned outer rectangle that fully contains
+ * an inner rectangle of the given size when rotated by a specified angle.
+ *
+ * @param rotationDeg - Rotation angle in degrees.
+ * @param innerWidth - Width of the inner rectangle.
+ * @param innerHeight - Height of the inner rectangle (defaults to Infinity).
+ * @returns The smallest outer rectangle that contains the rotated inner rectangle.
+ */
+export function getMinOuterRectSize(rotationDeg: number, innerWidth: number, innerHeight: number = Infinity): Size {
+    const w = innerWidth;
+    const h = innerHeight;
+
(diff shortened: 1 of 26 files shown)


packages/ag-charts-core/src/utils/math/shapeUtils.ts
```diff
@@ -42,6 +42,35 @@ export function getMaxInnerRectSize(
...
+export function getMinOuterRectSize(rotationDeg: number, innerWidth: number, innerHeight: number = Infinity): Size {
+    const w = innerWidth;
+    const h = innerHeight;
+
+    // Periodic every 180 degrees
+    const angle = (rotationDeg % 180) * (Math.PI / 180);
+    const sin = Math.abs(Math.sin(angle));
+    const cos = Math.abs(Math.cos(angle));
+
+    // Special cases
```
olegat, line 63: The special cases seem equivalent to default result
https://www.desmos.com/calculator/bkmtvfavss

1. $\theta = 0, \pi$

$$
width = w * cos + h * sin = w * 1 + h * 0 = w
$$

$$
height = w * sin + h * cos = w * 0 + h * 1 = h
$$

1. $\theta = \frac{\pi}{2}$

$$
width = w * cos + h * sin = w * 0 + h * 1 = h
$$

$$
height = w * sin + h * cos = w * 1 + h * 0 = w
$$

If this change is required due to floating-point errors (e.g. `sin=0.9999997` instead of `1`) then could you please specify that in the comment?

Otherwise, these special cases can just be removed.

olegat (approved):
LGTM.

Non-critical suggestion: Please just update the comment in `getMinOuterRectSize` or remove the special case (whichever is applicable)

packages/ag-charts-core/src/utils/math/shapeUtils.ts
iMoses: Similar to `getMaxInnerRectSize`, we add short-circuits to deal with simple/no rotation cases faster and with less calculations. I agree that in this case it may be a little redundant, as there are not much calculations done after the checks, but I kept it for consistency.

PR #5121, opened 2025-09-05
AG-15582: Fix text segments wrapping

iMoses:
https://ag-grid.atlassian.net/browse/AG-15582


olegat (approved):
LGTM. Tested the changes on localhost and they seem OK.

PR #5131, opened 2025-09-05
AG-15866 Fix `label.itemStyler` highlight updates

olegat:
https://ag-grid.atlassian.net/browse/AG-15922


packages/ag-charts-community/src/chart/series/series.ts
olegat, line 1: > [!IMPORTANT]
> This file contains the key change.

The early `if (currentHighlightState === previousHighlightState)` return-statement was the root cause of the bug.

The rest of the PR just updates the `hasItemStyler` overrides to check for `label.itemStyler != null`.

alantreadway (approved):
LGTM

PR #5141, opened 2025-09-08
CRT-937 - Stop deep clone on large dataset example.

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-937


lsjroberts (approved):
LGTM

PR #5149, opened 2025-09-08
change label font size to account for new measurement

AG-DavidG:
https://ag-grid.atlassian.net/browse/CRT-949


alantreadway (approved):
LGTM

PR #5154, opened 2025-09-08
CRT-938 Change empty truncated text to always include ellipsis

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-938


alantreadway (approved):
LGTM

PR #5160, opened 2025-09-08
CRT-950 - Fix bar series add/remove animation and add regression tests.

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-950


packages/ag-charts-community/src/chart/series/cartesian/__image_snapshots__/bar-series-test-ts-bar-series-initial-animation-for-bar-number-x-axis-number-y-axis-should-animate-at-25-1-snap.png
lsjroberts, line 1: Is this right? Shouldn't a single series start animating immediately?

alantreadway: This is an unfortunate side-effect of the way the tests are setup - there is no phase skipping going on, only fixed ratios of time per phase, so this isn't entirely representative of the end-user experience.

lsjroberts (approved):
LGTM

PR #5165, opened 2025-09-09
CRT-943 Do not call preventDefault on wheel handler when zoom is capped

olegat:
https://ag-grid.atlassian.net/browse/CRT-943


lsjroberts (approved):
LGTM

PR #5170, opened 2025-09-09
CRT-950 - Fix bar series initial vs add phase animation.

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-950


lsjroberts (approved):
LGTM

PR #5181, opened 2025-09-10
AG-14836 Add quick measure annotation into event

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14836


alantreadway (approved):
LGTM

PR #5197, opened 2025-09-10
Fix annotations buttons to hide defaults when custom provided

lsjroberts:
Same issue as https://ag-grid.atlassian.net/browse/AG-15691


alantreadway (approved):
LGTM

PR #5198, opened 2025-09-10
AG-15828 Fix polar series to use shared tooltip for 2 or 3 series

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-15828


alantreadway (approved):
LGTM

PR #5203, opened 2025-09-11
AG-15708 Update zoom to debounce isValid checks on dom elements

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-15708

### DIFF
--- a/packages/ag-charts-enterprise/src/features/zoom/zoom.ts
+++ b/packages/ag-charts-enterprise/src/features/zoom/zoom.ts
@@ -469,7 +469,9 @@ export class Zoom extends _ModuleSupport.BaseModuleInstance implements _ModuleSu
             const axisZoom = zoomManager.getAxisZoom(axisId);
             const newZoom = axisDragger.update(event, direction, anchor, seriesRect, zoom, axisZoom);
             zoomManager.setAxisManuallyAdjusted('zoom', axisId);
-            this.updateAxisZoom(axisId, direction as _ModuleSupport.CartesianAxisDirection, newZoom);
+            this.updateAxisZoom(axisId, direction as _ModuleSupport.CartesianAxisDirection, newZoom, {
+                directional: true,
+            });
         }
 
         tooltipManager.updateTooltip(TOOLTIP_ID);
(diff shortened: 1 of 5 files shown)


packages/ag-charts-enterprise/src/features/zoom/zoom.ts
```diff
@@ -748,24 +750,49 @@ export class Zoom extends _ModuleSupport.BaseModuleInstance implements _ModuleSu
         return constrainedZoom ?? newZoom;
     }
 
-    private isZoomValid(newZoom: DefinedZoomState) {
+    private previousZoomValid = true;
```
alantreadway, line 755: Do we need to reset this (and/or `previousAxisZoomValid`) under certain circumstances (perhaps in `resetZoom()`) to avoid stale state?

packages/ag-charts-enterprise/src/features/zoom/zoomToolbar.ts
```diff
@@ -194,6 +194,10 @@ export class ZoomToolbar extends BaseProperties {
             : `translateY(${container.getBounds().height + verticalSpacing}px)`;
     }
 
+    private readonly toggleButtonsDebounced = debounce(this.toggleButtons.bind(this), 300, {
```
alantreadway: Nit: Maybe extract a constant for the `300`ms debounce? Are we debouncing user-events anywhere else, and should the delay be consistent with those cases?

packages/ag-charts-website/e2e/zoom.spec.ts
```diff
@@ -69,6 +77,7 @@ test.describe('zoom', () => {
         // 6. Drag the x-axis twice with the navigator hidden again to zoom out
         await dragCanvas(page, withoutNavigatorXAxisLeft, withoutNavigatorXAxisRight);
         await dragCanvas(page, withoutNavigatorXAxisLeft, withoutNavigatorXAxisRight);
+        await delay(300); // Delay due to debounce in ZoomToolbar
```
alantreadway: Nit: Use constant once added?

packages/ag-charts-enterprise/src/features/zoom/zoomToolbar.ts
lsjroberts: Extracted to a constant as it is used twice within zoom.

The only other debounce I'm aware of is to help with capturing/ignoring wheel scroll when at max zoom or not, which is has a different purpose so I don't think they should necessarily be the same time.

packages/ag-charts-website/e2e/zoom.spec.ts
lsjroberts: As far as I can see, we don't have util type exports for enterprise we can use in other packages. Not sure it makes sense to add it for this one only? So I've added a reference to the constant here.

alantreadway (approved):
LGTM

PR #5213, opened 2025-09-15
format snippet

AG-DavidG:
https://ag-grid.atlassian.net/browse/AG-15952

### DIFF
--- a/packages/ag-charts-website/src/content/docs/stylers/index.mdoc
+++ b/packages/ag-charts-website/src/content/docs/stylers/index.mdoc
@@ -103,21 +103,19 @@ theme: {
                 styler: (params) => {
                     if (params.yKey.includes('benchmark')) {
                         return { fill: 'lightgray' };
-                    }
-                    else return {fill: '#5090DC'}
-                },
+                    }  else return {fill: '#5090DC'};
                 },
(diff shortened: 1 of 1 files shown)


packages/ag-charts-website/src/content/docs/stylers/index.mdoc
```diff
@@ -103,21 +103,19 @@ theme: {
                 styler: (params) => {
                     if (params.yKey.includes('benchmark')) {
                         return { fill: 'lightgray' };
-                    }
-                    else return {fill: '#5090DC'}
-                },
+                    }  else return {fill: '#5090DC'};
```
lsjroberts, line 106: It should probably be:

```
} else {
    return { fill: '5090DC' };
}
```

lsjroberts: Or actually, just ditch the else statement as it isn't needed and do

```
if (params.yKey.includes('benchmark')) {
    return { fill: 'lightgray' };
}
return { fill: '#5090DC' };
```

alantreadway:
Superceeded by #5223 - @lsjroberts I'll apply your feedback there

PR #5224, opened 2025-09-17
AG-15916 Fix zoom scroll change direction prevent default

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-15916


alantreadway (approved):
LGTM

PR #5233, opened 2025-09-18
AG-3390 - Ignore dark mode for examples in new tab

taktran:
Allow dark mode to be ignored based on the rendering page. In this case, we want to ignore dark mode rendering on the new tabs page.


markdurrant (approved):
LGTM 👍

PR #5252, opened 2025-09-24
AG-15782 Restrict `callWithContext` to single-arg functions

olegat:
https://ag-grid.atlassian.net/browse/AG-15782

The `callWithContext` function is only ever called with one argument, and wouldn't work when being called with multiple arguments anyway. See function `maybeSetContext` in `callbackCache.ts`; it deferences `params[0]`.

This change cleans up & simplifies the code, but also helps to workaround an error when trying to generalise the `getStyle` implementation of `RadarAreaSeries` and `RadarLineSeries`.

See commit https://github.com/ag-grid/ag-charts/commit/19b712feadd104dff8dc0f84b92073a82d0dede8. The return type of the abstract `RadarSeries.makeStylerParams` was conflicting the expected type for `cachedCallWithContext` at line 742

### DIFF
--- a/packages/ag-charts-community/src/chart/axis/axis.ts
+++ b/packages/ag-charts-community/src/chart/axis/axis.ts
@@ -1,5 +1,6 @@
 import {
-    type AnyFn,
+    type Callback,
+    type CallbackParam,
     CleanupRegistry,
     type Point,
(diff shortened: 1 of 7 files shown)


alantreadway (approved):
LGTM

PR #5253, opened 2025-09-25
Use permalinks in the `snapshot-notify` GitHub Action

olegat:


alantreadway (changes requested):
The link generated here will just be a diff of the PR branch itself at the point the snapshots are generated, but excluding the snapshots - you maybe highlighting there are two cases to consider here though.

We need (distinctly):
- A link to allow the snapshots generated to be easily merged back into the PR branch (the intent of the current approach).
- A permanent snapshot diff link that can be reviewed later (this is the case you are going for?)

Edit: I've pushed changes to add these two distinct cases, just testing now.

PR #5273, opened 2025-10-01
AG-15977: Unify our module registries

iMoses:
https://ag-grid.atlassian.net/browse/AG-15977


packages/ag-charts-community/src/module/optionsGraph.ts
```diff
@@ -1105,4 +1115,59 @@ export class OptionsGraph extends AdjacencyListGraph<unknown, string> implements
         this.rollbackEdgesValue = [];
         this.isRollingBack = false;
     }
+
+    private predictAxes(seriesType: SeriesType, userSeriesOptions?: any, data?: DatumDefault[]) {
```
alantreadway, line 1119: Could we easily extract this to a utility function, this doesn't feel like the best place for this logic to live?

packages/ag-charts-enterprise/src/charts/topologyChart.ts
```diff
@@ -52,7 +52,8 @@ export class TopologyChart extends Chart {
         const { seriesRoot, annotationRoot } = this;
         const { layoutBox } = ctx;
 
-        const seriesRect = layoutBox.clone().shrink(this.modulesManager.getModule<any>('seriesArea').getPadding());
+        const seriesAreaModule = this.modulesManager.getModule('seriesArea') as any;
```
alantreadway: Are the `as any` casts a temporary measure?

packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1738,7 +1738,7 @@ export abstract class Chart extends Observable implements ModuleInstance, ChartS
 
         for (let index = 0; index < options.length; index++) {
             const axisOptions = options[index];
-            const axis = axisRegistry.create(axisOptions.type, moduleContext);
+            const axis = ModuleRegistry.getAxisModule(axisOptions.type)!.create(moduleContext) as any;
```
alantreadway, line 1760: Are the `as any` casts a temporary measure?

packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1676,7 +1675,8 @@ export abstract class Chart extends Observable implements ModuleInstance, ChartS
     }
 
     private createSeries(seriesOptions: SeriesOptionsTypes): UnknownSeries {
-        const seriesInstance = seriesRegistry.create(seriesOptions.type, this.getModuleContext()) as UnknownSeries;
+        const seriesModule = ModuleRegistry.getSeriesModule(seriesOptions.type);
+        const seriesInstance = seriesModule!.create(this.getModuleContext()) as UnknownSeries;
```
alantreadway, line 1698: Do we need the `!` here and below?

packages/ag-charts-community/src/module/optionsGraph.ts
iMoses: Sure. I started by making it a standalone function, but this was the only part of the code that actually uses it, and I didn't want to breakdown the optionsGraph approach.

packages/ag-charts-community/src/chart/chart.ts
iMoses: Yes, it's temporary, though we should probably find a better way to handle this.

packages/ag-charts-community/src/chart/chart.ts
iMoses: Yes, because `seriesModule` can also be `undefined`.
We could change the logic to throw in such cases, to simplify the types, but it's a bigger change.

packages/ag-charts-enterprise/src/charts/topologyChart.ts
iMoses: In this specific case you can see that it was already case to `any` in here: `getModule<any>('seriesArea')`
We can obviously improve on it, but not sure if it'll be in this iteration.

alantreadway:
Nit: Could you link the related JIRA in the PR description and use the JIRA ticket number in the subject?

packages/ag-charts-community/src/chart/factory/setupModules.ts
```diff
@@ -1,60 +1,43 @@
...
-            for (const axisType of m.axisTypes) {
-                const axisTypeTheme = axisRegistry.getThemeTemplate(axisType);
-                const theme = mergeDefaults(m.themeTemplate, axisTypeTheme);
-                axisRegistry.setThemeTemplate(axisType, theme);
+    // if (ModuleRegistry.hasEnterpriseModules()) {
+    //     const expectedButUnused = getUnusedExpectedModules();
+    //     if (expectedButUnused.size > 0) {
+    //         Logger.errorOnce('Enterprise modules expected but not registered: ', joinFormatted([...expectedButUnused]));
+    //     }
+    // }
```
alantreadway: Still needed?

packages/ag-charts-community/src/chart/factory/setupModules.ts
```diff
@@ -1,60 +1,43 @@
...
-    }
+    // function getUnusedExpectedModules() {
+    //     const unusedExpectedModules = new Set<string>();
+    //     for (const s of ExpectedModules) {
+    //         if (s.enterprise && !verifiedModules.has(s.name)) {
+    //             unusedExpectedModules.add(s.name);
+    //         }
+    //     }
+    //     return unusedExpectedModules;
+    // }
```
alantreadway: Ditto?

packages/ag-charts-core/src/globals/debug-logger.ts
alantreadway, line 1: Could we preserve the git history of some of these moves?

packages/ag-charts-core/src/interfaces/moduleDefinition.ts
```diff
@@ -1,39 +1,84 @@
...
+}
+
+export interface SeriesModuleInstance extends ModuleInstance {}
+
+export interface PluginModuleInstance extends ModuleInstance {
+    processData?(dataController: any): Promise<void> | void;
+}
+
+export type PickNodeDatumResult = { datum: NonNullable<any>; distanceSquared: number } | undefined;
+// export type PickNodeDatumResult = { datum: SeriesNodeDatum<DatumIndexType>; distanceSquared: number } | undefined;
```
lsjroberts, line 44: still needed?

packages/ag-charts-enterprise/src/setup.ts
```diff
@@ -1,112 +1,28 @@
...
-        SharedToolbarModule,
-        SunburstModule,
-        SyncModule,
-        TreemapModule,
-        WaterfallModule,
-        ZoomModule
-    );
+    ModuleRegistry.registerMany(AllEnterpriseModules, VERSION);
+
+    // _ModuleSupport.moduleRegistry.register(
```
alantreadway: Needed?

packages/ag-charts-community/src/chart/factory/expectedModules.ts
```diff
@@ -0,0 +1,101 @@
...
+    { type: 'plugin', name: 'chartToolbar', chartType: 'cartesian', enterprise: true },
+    { type: 'plugin', name: 'contextMenu', enterprise: true },
+    { type: 'plugin', name: 'statusBar', chartType: 'cartesian', enterprise: true },
+    { type: 'plugin', name: 'dataSource', enterprise: true },
+    { type: 'plugin', name: 'sync', chartType: 'cartesian', enterprise: true },
+    { type: 'plugin', name: 'ranges', chartType: 'cartesian', enterprise: true },
+    {
+        type: 'plugin',
+        name: 'zoom',
+        // chartTypes: ['cartesian', 'topology']
```
lsjroberts: is this needed? zoom is still functionally limited to these chart types I believe

packages/ag-charts-enterprise/project.json
```diff
@@ -126,6 +126,14 @@
       "options": {
         "commands": ["node tools/package/sanity-check-package.js dist/packages/contents/ag-charts-enterprise/package"]
       }
+    },
+    "validate-size": {
```
alantreadway, line 130: Nit: Could we have this as `test:validate-size` and link it from the `test` target perhaps?

Should we also have this in the `ag-charts-community` package too?

alantreadway (approved):
LGTM, one nit for post-merge

packages/ag-charts-enterprise/project.json
iMoses: it exists in both packages.
we should talk about enabling it as part of CI as well.

PR #5281, opened 2025-10-02
AG-5877- Add cartesian axis crossAt options to position axes in serie…

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-5877


packages/ag-charts-community/src/chart/axis/axis.ts
```diff
@@ -454,9 +454,7 @@ export abstract class Axis<
         this.updatePosition();
         this.updateSelections();
 
-        this.tickLineGroup.visible = this.tick.enabled;
```
alantreadway, line 457: Is this an intentional change? If it's a bug being fixed, we should make sure to capture that in JIRA for David to review.

Do the non-cartesian axes types still behave correctly after this change?

packages/ag-charts-community/src/chart/cartesianChart.ts
```diff
@@ -314,9 +319,109 @@ export class CartesianChart extends Chart {
             });
         }
 
+        this.applyAxisCrossing(seriesRect, crossPositions);
+
         return { clipSeries, seriesRect, axisAreaWidths: newAxisAreaWidths, overflows };
     }
 
+    private calculateAxesCrossPositions(axisWidths: Map<string, number>, seriesRect: BBox): Map<string, number> {
```
alantreadway: It might be better to short-circuit early if no axes have a `crossAt?.value` - e.g. I would make the first step here to find all axes configured in that way, before we get into searching for X and Y axes.

If you could do this check earlier outside of iterating on the axis layout, it would be even better to minimise redundant checks on each iteration.

packages/ag-charts-community/src/chart/cartesianChart.ts
```diff
@@ -314,9 +319,109 @@ export class CartesianChart extends Chart {
...
+        const crossPositions = new Map<string, number>();
+        const primaryXAxis = this.axes.find((axis) => axis.direction === ChartAxisDirection.X);
+        const primaryYAxis = this.axes.find((axis) => axis.direction === ChartAxisDirection.Y);
+
+        for (const axis of this.axes) {
+            if (!(axis instanceof CartesianAxis && axis.crossAt?.value != null)) continue;
```
iMoses: nit: it makes more sense IMO to change the type than to ensure everywhere that you axis is cartesian.
In a cartesian chart we cannot have anything other than cartesian axes.

packages/ag-charts-community/src/chart/axis/axis.ts
manapeirov: It was being overriden in cartesianAxis update function but I have updated polarAxis to have it so there are no regression, none of the tests flagged an issue though

alantreadway (approved):
LGTM

PR #5282, opened 2025-10-02
AG-15953 - High frequency examples and plumbing.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-15953


packages/ag-charts-website/src/content/docs/high-frequency-data-test/_examples/high-freq-area/main.ts
```diff
@@ -0,0 +1,332 @@
...
+function use(newMethod: string) {
+    method = newMethod;
+    console.log(method);
+}
+
+const updateCallback = async () => {
+    const startTime = performance.now();
+    updateCount++;
+    const isPrepend = method.includes('prepend');
+    const isRemove = method.includes('remove');
```
lsjroberts, line 227: Are there examples that combine remove/prepend/append at the same time?

packages/ag-charts-community/src/chart/data/dataSet.ts
```diff
@@ -0,0 +1,143 @@
...
+        return value instanceof DataSet;
+    }
+
+    /**
+     * Returns the net size of the data after applying all pending transactions.
+     * This calculates the final data length without mutating the underlying array.
+     *
+     * Note: This assumes transactions don't have overlapping updates (i.e., removes
+     * don't target items in append/prepend arrays). This is a safe assumption for
+     * typical usage where transactions operate on existing data.
```
lsjroberts: How does `DataSet` handle the case when a user incorrectly removes items they are also appending?

packages/ag-charts-community/src/chart/data/transactionUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+                if (writeIndex !== readIndex) {
+                    data[writeIndex] = data[readIndex];
+                }
+                writeIndex++;
+            }
+        }
+        // Truncate array to new length
+        data.length = writeIndex;
+        return data;
+    } else {
```
lsjroberts: `else` branch not needed since you return above

packages/ag-charts-community/src/chart/data/transactionUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Removes items from an array by reference equality.
+ * Optimized for performance with minimal memory allocation.
+ *
+ * @param data - The array to remove items from (must be non-null array)
+ * @param removeRefs - Array of references to remove (must be non-null array)
+ * @param mutate - If true, mutates data in-place; if false, returns new array
+ * @returns The array with items removed (either mutated original or new array)
+ */
+export function applyRemoveByReference<T>(data: T[], removeRefs: readonly T[], mutate: boolean): T[] {
```
lsjroberts: This function appears to only be called once with `mutate = true`? Can the other case be ditched and the argument removed?

packages/ag-charts-community/src/chart/series/series.ts
```diff
@@ -351,12 +354,12 @@ export abstract class Series<
...
     }
 
-    setOptionsData(input: unknown[]) {
+    setOptionsData(input: DataSet | undefined) {
         this._data = input;
         this.onDataChange();
     }
 
-    setChartData(input: unknown[]) {
+    setChartData(input: DataSet | undefined) {
```
lsjroberts, line 362: Could it take the array as before and call `DataSet.wrap()` itself, to save repeating it elsewhere?

packages/ag-charts-website/src/content/docs/high-frequency-data-test/_examples/high-freq-area/main.ts
alantreadway: The `high-freq-stacked-line` and `high-freq-multi-chart` examples combine the most common `remove` and `append` operations for a rolling window of data, which is likely the most common use-case for this feature - I don't see a common case for all three in one, although perhaps `remove` and `prepend` (for right-to-left scrolling) might be a thing, we could amend the examples to cover that too.

packages/ag-charts-community/src/chart/data/dataSet.ts
alantreadway: I've clarified the comment, unless the user is doing something daft like adding or removing the same item multiple times, we can handle this scenario perfectly fine assuming we apply adds before removes.

packages/ag-charts-community/src/chart/series/series.ts
alantreadway: No, we need the `DataSet` reference to be identical for deduplication purposes in `DataController`.

packages/ag-charts-community/src/chart/data/transactionUtils.ts
alantreadway: Good spot, cleaned up.

packages/ag-charts-website/src/content/docs/high-frequency-data-test/_examples/high-freq-area/main.ts
lsjroberts: Ok cool, sounds good as is.

lsjroberts (approved):
LGTM

PR #5285, opened 2025-10-02
AG-15773 Make `SeriesNodeDatum.itemId` type stricter than any

olegat:
https://ag-grid.atlassian.net/browse/AG-15773


alantreadway (approved):
LGTM

PR #5286, opened 2025-10-02
AG-15773 Add unit test for range-area itemStyler itemId

olegat:
https://ag-grid.atlassian.net/browse/AG-15773


alantreadway (approved):
LGTM

PR #5287, opened 2025-10-03
Optimise GHA Nx cache size.

alantreadway:
- **Prevent @ag-option-extract examples from being cached on CI.**
- **Automate Nx cache reset and optimise GHA cache use.**


seanlandsman (approved):
lgtm

PR #5292, opened 2025-10-06
AG-9002 Remove stroke-strokeWidth-defaults test

olegat:

This test is not helpful for the range-area series, it's specific to the area series.


alantreadway (approved):
LGTM

PR #5309, opened 2025-10-08
AG-12944 Add sankey verticalAlignment and sort options

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12944


alantreadway (approved):
LGTM

PR #5314, opened 2025-10-09
AG-9002 Fix range-area deprecated options

olegat:

### DIFF
--- a/packages/ag-charts-enterprise/src/series/range-area/rangeArea.ts
+++ b/packages/ag-charts-enterprise/src/series/range-area/rangeArea.ts
@@ -95,6 +95,15 @@ interface RangeAreaSpanPointDatum {
     low: _ModuleSupport.LineSpanPointDatum;
 }
 
+type BaseUpdate = _ModuleSupport.CartesianSeries<
+    _ModuleSupport.Marker,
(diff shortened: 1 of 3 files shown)


packages/ag-charts-enterprise/src/series/range-area/rangeArea.ts
```diff
@@ -492,6 +501,130 @@ export class RangeAreaSeries extends _ModuleSupport.CartesianSeries<
         return low.marker.isDirty() || high.marker.isDirty();
     }
 
+    public applyDeprecations() {
```
olegat, line 504: See comment below this line.

The implementation of `/series/$index/marker/fill` (and possibly `/series/$index/marker/enabled`) will probably need to stay in here.

However, most of the deprecated options like `/series/$index/stroke` could implemented straight within the graph configuration in `rangeAreaThemes.ts`. We can do this if desired, I have no stronger preferences about this. But bear in mind that keeping all the deprecation-handling in `applyDeprecation` may make it easier to remove & clean up in the future.

olegat:
Closing, I found a better approach: https://github.com/ag-grid/ag-charts/pull/5316

PR #5336, opened 2025-10-13
AG-16097 - Cleanup new sonarcloud issues pt6

alantreadway:
https://ag-grid.atlassian.net/browse/AG-16097


packages/ag-charts-community/src/chart/interaction/contextMenuRegistry.ts
```diff
@@ -54,11 +54,11 @@ export class ContextMenuRegistry {
...
-            this.hiddenActions.delete(id);
-        } else {
-            this.hiddenActions.add(id);
-        }
+    public show(id: AgContextMenuItemLiteral) {
+        this.hiddenActions.delete(id);
+    }
+
+    public hide(id: AgContextMenuItemLiteral) {
+        this.hiddenActions.add(id);
```
iMoses: why not keep a third `toggle`/`setVisible` method?
it seems that other than the constructor that's the actual use case.
what's sonar issues with it?

packages/ag-charts-community/src/chart/themes/constants.ts
```diff
@@ -10,12 +10,12 @@ export enum FONT_SIZE {
...
-    SMALL = FONT_SIZE.SMALL / BASE_FONT_SIZE,
-    MEDIUM = FONT_SIZE.MEDIUM / BASE_FONT_SIZE,
-    LARGE = FONT_SIZE.LARGE / BASE_FONT_SIZE,
-    LARGEST = FONT_SIZE.LARGEST / BASE_FONT_SIZE,
+    SMALLEST = 0.6666666666666666, // FONT_SIZE.SMALLEST / BASE_FONT_SIZE (8 / 12)
+    SMALLER = 0.8333333333333334, // FONT_SIZE.SMALLER / BASE_FONT_SIZE (10 / 12)
+    SMALL = 1, // FONT_SIZE.SMALL / BASE_FONT_SIZE (12 / 12)
+    MEDIUM = 1.0833333333333333, // FONT_SIZE.MEDIUM / BASE_FONT_SIZE (13 / 12)
+    LARGE = 1.1666666666666667, // FONT_SIZE.LARGE / BASE_FONT_SIZE (14 / 12)
+    LARGEST = 1.4166666666666667, // FONT_SIZE.LARGEST / BASE_FONT_SIZE (17 / 12)
```
iMoses: I think that when it comes to static vs dynamic values in such cases I'd rather have them dynamic.
what's sonar's issues with it?

packages/ag-charts-community/src/chart/interaction/contextMenuRegistry.ts
alantreadway: https://sonarcloud.io/organizations/ag-grid/rules?open=typescript%3AS2301&rule_key=typescript%3AS2301

packages/ag-charts-community/src/chart/themes/constants.ts
alantreadway: https://sonarcloud.io/organizations/ag-grid/rules?open=typescript%3AS6550&rule_key=typescript%3AS6550

alantreadway: In this particular case I would propose that it shouldn't be `enum`, perhaps `s/enum/const/` would work better?

alantreadway: I've converted these to the `const` solution which looks like a more reasonable substitution.

packages/ag-charts-community/src/chart/interaction/contextMenuRegistry.ts
iMoses: maybe it's worth considering something like:
```js
    public toggle(id: AgContextMenuItemLiteral, action?: 'show' | 'hide') {
        action ??= this.isVisible(id) ? 'hide' : 'show';
        switch (action) {
            case 'show':
                return this.show(id);
            case 'hide':
                return this.hide(id);
        }
    }
```

alantreadway: That'll get flagged too - the rule is about having a different implementation based upon a parameter

alantreadway: Good suggestion, fixed.

iMoses (approved):
LGTM

PR #5339, opened 2025-10-13
AG-9002 Fix highlighting on range-area gradient markers

olegat:


packages/ag-charts-enterprise/src/series/range-area/rangeArea.ts
olegat, line 1: Based off of https://github.com/ag-grid/ag-charts/commit/ba02c7a4194af8faa032af6c573c6affdceefae0. Overrides `CartesianSeries.strokewidthChanged` to check both the low & high strokeWidth.

alantreadway (approved):
LGTM

PR #5353, opened 2025-10-16
AG-16097 - Cleanup new SonarCloud issues pt7

alantreadway:
https://ag-grid.atlassian.net/browse/AG-16097


lsjroberts (approved):
LGTM

iMoses (changes requested):
Don't change `main-modules.ts`
Add this: `/* eslint-disable unicorn/prefer-export-from */`

IMO this rule make sense generally, but not in the case of an entry file.

alantreadway:
> Don't change `main-modules.ts` Add this: `/* eslint-disable unicorn/prefer-export-from */`

Fixed

iMoses (approved):
LGTM

PR #5372, opened 2025-10-17
AG-15955 - Split website build into distinct job.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-15955


AG-DavidG (approved):
LGTM
Let's what improvements we can see on CI

PR #5373, opened 2025-10-17
AG-9002 Warn if item.(low|high).marker.itemStyler option is set

olegat:


packages/ag-charts-enterprise/src/series/range-area/rangeAreaThemes.ts
olegat, line 1: This must be removed because removing `itemStyler` from the validation def also raises multiple (somewhat duplicate) warnings from the navigator styling.

I've moved the copy-logic into a `getMarkerStyle` method override which is where the `itemStyler` callback is required.

alantreadway (approved):
LGTM

PR #5378, opened 2025-10-17
AG-14960 Remove minSize option and fix layout of nodes/links

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-14940


alantreadway (approved):
LGTM

PR #5388, opened 2025-10-20
CRT-964 Fix zoom buttons show immediately when visible=zoomed

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-964


alantreadway (approved):
LGTM

PR #5397, opened 2025-10-20
CRT-957: Docs - Combo-box does not include separators in frameworks

iMoses:
https://ag-grid.atlassian.net/browse/CRT-957


alantreadway (approved):
LGTM

PR #5399, opened 2025-10-20
CRT-960 Make the "en-US" locale explicit

olegat:
https://ag-grid.atlassian.net/browse/CRT-960

Is it not possible to format a date without specifying a locale, because date formats are inherently locale-dependent.

The AG Charts library internally uses the "en-US" locale, but does so implicitly. This PR makes this assumption explicit.


olegat:
Closing. See https://ag-grid.atlassian.net/browse/AG-13220

PR #5402, opened 2025-10-21
AG-16050 Fix parentLevel.tick.enabled=false to hide primary ticks lines

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-16050


alantreadway (approved):
LGTM

PR #5417, opened 2025-10-22
AG-15926 Fix zoom fast wheel scroll when autoScaling y-axis

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-15926


alantreadway (approved):
LGTM

PR #5424, opened 2025-10-23
AG-3390 - Add CodeQL SAST tool

taktran:

PR #5425, opened 2025-10-23
AG-3390 - Add CodeQL SAST tool

taktran:


taktran:
WIP from https://github.com/ag-grid/ag-charts/pull/5424 - not working

PR #5427, opened 2025-10-23
AG-16165 - Increase TLD limit for trial license form

taktran:
Based on max character limit of TLDs.

Based on https://github.com/ag-grid/ag-grid/pull/12311


markdurrant (approved):
LGTM 👍

PR #5440, opened 2025-10-27
AG-3390 - Add OSS Scorecard results check

taktran:


alantreadway (commented):
Why isn't this just part of `scorecard.yml`?

seanlandsman (approved):

PR #5443, opened 2025-10-27
AG-15601 Fix ranges active button border transition

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-15601

Snapshot updates as expected. Previously the buttons appeared too narrow due to the `margin-right: -1px`.


alantreadway (approved):
LGTM

PR #5448, opened 2025-10-27
Revert "AG-15926 Fix zoom fast wheel scroll when autoScaling y-axis"

alantreadway:
Reverts ag-grid/ag-charts#5417


iMoses (approved):
LGTM

PR #5436, opened 2025-10-24
Refactor DataModel.

alantreadway:

PR #5460, opened 2025-10-28
Refactor dataModel.test.ts

alantreadway:
Decompose dataModel.test.ts in alignment with DataModel refactor #5436


lsjroberts (approved):
LGTM.

Spotted a few places could use the `verifyDomain()` util that aren't, but not needed as part of this PR.

PR #5461, opened 2025-10-28
AG-16168 - Default changes for 13.0.0 (Pt.2)

alantreadway:
https://ag-grid.atlassian.net/browse/AG-16168


packages/ag-charts-community/src/chart/axis/axisInterval.ts
```diff
@@ -3,7 +3,7 @@ import type { TickInterval } from './axisTick';
 
 export class AxisInterval<S> extends BaseProperties {
     @Property
-    placement?: 'on' | 'between';
+    placement?: 'on' | 'between' = 'between';
```
lsjroberts:
```suggestion
    placement: 'on' | 'between' = 'between';
```

lsjroberts (approved):

packages/ag-charts-community/src/chart/axis/axisInterval.ts
alantreadway: I tried applying this change, but it causes typescript compilation issues in linearGaugeSeries.ts which can't be easily worked around.

PR #5472, opened 2025-10-29
AG-8627 Split `zoom:change` into 2 event types

olegat:

-   `zoom:change-requested` (same as old `zoom:change`).
-   `zoom:change-completed` (emitted by zoom.ts when change is done).

### DIFF
--- a/packages/ag-charts-community/src/core/eventsHub.ts
+++ b/packages/ag-charts-community/src/core/eventsHub.ts
@@ -53,8 +53,8 @@ export interface EventsHubMap {
     'series-area:click': SeriesAreaClickEvent;
     'series:redo': null;
     'series:undo': null;
-    'zoom:change-requested': ZoomChangeRequestedEvent;
-    'zoom:change-completed': null;
+    'zoom:change-request': ZoomChangeRequestedEvent;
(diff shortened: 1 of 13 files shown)


alantreadway (approved):
LGTM, but lets run past @lsjroberts too?

packages/ag-charts-community/src/core/eventsHub.ts
```diff
@@ -53,7 +53,8 @@ export interface EventsHubMap {
     'series-area:click': SeriesAreaClickEvent;
     'series:redo': null;
     'series:undo': null;
-    'zoom:change': ZoomChangeEvent;
+    'zoom:change-requested': ZoomChangeRequestedEvent;
+    'zoom:change-completed': null;
```
lsjroberts: To match the tense of the other events these should be:

```
zoom:change-request
zoom:change-complete
```

lsjroberts (approved):
LGTM with one minor point. Much better fix than my previous one with the 'bound fire once' event!

packages/ag-charts-community/src/core/eventsHub.ts
olegat: Done

PR #5481, opened 2025-10-30
AG-16222 - Fix no data overlay bug.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-16222

### DIFF
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -1121,7 +1121,7 @@ export abstract class Chart extends Observable implements ModuleInstance, ChartS
         if (this.mode !== 'integrated') {
             // Validate each series that shares a legend item label uses the same fill colour
             const seriesMarkerFills: { [key: string]: Map<TextOrSegments, AgColorType | undefined> } = {};
-            const seriesTypeMap = new Map(this.series.map((s) => [s.id, s.type]));
+            const seriesMap = new Map(this.series.map((s) => [s.id, s]));
 
             for (const {
                 seriesId,
@@ -1130,7 +1130,10 @@ export abstract class Chart extends Observable implements ModuleInstance, ChartS
             } of legendData.filter((d) => !d.hideInLegend)) {
                 if (marker.fill == null) continue;
 
-                const seriesType = seriesTypeMap.get(seriesId)!;
+                const series = seriesMap.get(seriesId);
+                if (!series?.hasData) continue;
+
+                const seriesType = series.type;
                 const markerFill = (seriesMarkerFills[seriesType] ??= new Map());
 
                 if (markerFill.has(label.text)) {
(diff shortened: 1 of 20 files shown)


packages/ag-charts-community/src/chart/chart.ts
alantreadway, line 1: Fixes a tangential issue with legend processing when the no-data use-case happens.

packages/ag-charts-community/src/chart/overlay/overlay.ts
alantreadway, line 1: Mostly to assist with test detection of the overlay.

packages/ag-charts-community/src/chart/series/dataModelSeries.ts
alantreadway, line 1: This is the main part of the fix - just use the data-model size/invalid/missing counts to determine available data-set size for the series.

Consequently we don't need series-specific handling in sub-classes.

packages/ag-charts-community/src/chart/overlay/overlay.test.ts
```diff
@@ -212,5 +212,115 @@ HTMLCollection [
...
+                test('should not show no data overlay when first series has some invalid data', async () => {
+                    chart = await createChart({
+                        data: [
+                            { quarter: "Q1'18", iphone: null },
+                            { quarter: "Q2'18", iphone: 16 },
+                        ],
+                        series: [{ type: seriesType, xKey: 'quarter', yKey: 'iphone' }],
+                    });
+
+                    // Check that no data overlay IS shown
```
lsjroberts: comment doesn't match case

packages/ag-charts-community/src/chart/series/dataModelSeries.ts
```diff
@@ -48,21 +48,16 @@ export abstract class DataModelSeries<
...
-            }
-        }
+    missingDataCount() {
+        return this.dataModel?.resolveMissingDataCount(this) ?? 0;
+    }
 
-        return false;
+    override get hasData() {
+        return Math.max(0, this.dataCount() - this.invalidDataCount() - this.missingDataCount()) > 0;
     }
```
lsjroberts, line 61: The tests suggest that it should be classed as having data even if some of that data is invalid.

If that is correct, should the `invalidDataCount()` **not** be removed from the count here?

lsjroberts (approved):
LGTM

packages/ag-charts-community/src/chart/overlay/overlay.test.ts
alantreadway: fixed

packages/ag-charts-community/src/chart/series/dataModelSeries.ts
alantreadway: Discussed offline, datum should only be one status of valid / invalid / missing, so it should be safe to simply subtract.

PR #5486, opened 2025-10-31
Disable CodeQL trap caching - generating 520MB cache entries on EVERY run.

alantreadway:
See https://github.com/github/codeql-action/issues/2030


taktran (approved):
LGTM

PR #5517, opened 2025-11-04
AG-8627 Use branded AxisID string

olegat:

This enforce better type-safety (general-purpose strings cannot be passed as params to functions that require an axis ID).


alantreadway (approved):
LGTM

PR #5522, opened 2025-11-04
AG-16258 Fix Sparkline axes bug.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-16258

### DIFF
--- a/packages/ag-charts-enterprise/src/features/zoom/zoomDOMProxy.ts
+++ b/packages/ag-charts-enterprise/src/features/zoom/zoomDOMProxy.ts
@@ -53,10 +53,11 @@ export class ZoomDOMProxy {
     ) {
         this.seriesRect = seriesRect;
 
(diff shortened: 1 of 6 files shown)


packages/ag-charts-enterprise/src/features/zoom/zoomDOMProxy.ts
```diff
@@ -55,7 +56,7 @@ export class ZoomDOMProxy {
         for (const ax of this.axes) {
             ax.div.setHidden(!enableAxisDragging && !enableAxisScrolling);
```
olegat: This `hidden` status should also be updated to `!enabled || (!enableAxisDragging && !enableAxisScrolling)`

olegat (approved):
LGTM – just make sure to also update the visibility-state of the HTML elements.

alantreadway:
> LGTM – just make sure to also update the visibility-state of the HTML elements.

Good point, updated

iMoses (approved):
LGTM

PR #5529, opened 2025-11-05
AG-8627 Migrate Zoom `autoScaling` to enterprise

olegat:

## Brief
The autoScaling feature implementation is being moved from `ZoomManager.ts` to a new file `ZoomAutoScaler.ts`

## Summary of changes

- Add `ZoomAutoScaler.ts` (code from `ZoomManager.ts`) in the enterprise module.
- Add memento events
  - These events are used for `ZoomManager.ts` & `ZoomAutoScaler.ts` to collaborate the memento creation/restoration.
  - `zoom:save-memento` for creating new ZoomMemento objects.
  - `zoom:load-memento` for restoring an old ZoomMemento object.
- Add `changeType: 'sync'`
  - Used by `ChartSync` to avoid infinite recursion on `zoomModule.updateSyncZoom`.
- Change the `zoom:change-complete` event behaviour
  - Most module now listen for this event instead of `zoom:change-request`
  - "No-op" zoom changes do not dispatch a completion-events (which avoids infinite loops).
  - Add `ZoomChangeCompleteEvent`, which includes `changeType` (used by chart-sync) and `zoom.x` (used by navigator).
- Update the `zoom:change-request` event
  - Includes a `constrainZoom()` method – used by the autoScaler to adjust the Y-axis scaling when a zoom change is requested.


alantreadway (approved):
LGTM, @lsjroberts ?

packages/ag-charts-enterprise/src/features/zoom/zoom.ts
```diff
@@ -949,7 +907,9 @@ export class Zoom extends AbstractModuleInstance {
     }
 
     public updateSyncZoom(zoom: DefinedZoomState) {
+        this.isSyncing = true;
         this.updateZoom(zoom);
+        this.isSyncing = false;
```
lsjroberts, line 920: nit: Could the syncing state be passed as an option, e.g. `this.updateZoom(zoom, { isSyncing: true })`?

lsjroberts (approved):
LGTM

packages/ag-charts-enterprise/src/features/zoom/zoom.ts
olegat: Yes, I thought of that. But that would mean adding parameter to most `Zoom.updateThing` methods.

I think that this change is fine for now – I envision that `zoom.ts` will undergo a similar "update-flow" standardisation like `zoomManager.ts`. We could have a general-use `Zoom.updateZoom(zoom, options)` method instead of like lots of separate methods like `updateUnifiedZoom` and so on.

PR #5539, opened 2025-11-06
AG-16272 - Improve shared tooltip missing-value handling.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-16272


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -461,7 +462,9 @@ export abstract class Chart extends Observable implements ModuleInstance, ChartS
             const seriesDatumIndex = s.datumIndexForCategoryValue(categoryValue);
             const seriesTooltipContent =
                 seriesDatumIndex == null ? undefined : s.getTooltipContent(seriesDatumIndex, undefined);
-            return seriesTooltipContent == null ? [] : [seriesTooltipContent];
+            // Exclude series with all missing data from shared tooltips
+            if (seriesTooltipContent == null || hasAllMissingData(seriesTooltipContent)) return [];
```
jacobp100: I think this should be handled solely in tooltip content rather than having this logic spread out

packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
```diff
@@ -1206,6 +1206,7 @@ export class AreaSeries extends CartesianSeries<
                         label: yName,
                         fallbackLabel: yKey,
                         value: this.getAxisValueText(yAxis, 'tooltip', yValue, datum, yKey, legendItemName),
+                        missing: yValue == null,
```
jacobp100: Would this also be if the y value was not finite?

packages/ag-charts-community/src/chart/chart.ts
alantreadway: Good shout - done

packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
alantreadway: Extended the guard to cover `NaN` and `Infinity` cases - I don't think the chart would render correctly in these cases anyway, but agree if there were some way to reach that state these are probably illegal tooltip values.

jacobp100 (approved):

alantreadway:
Snapshots are showing that some `Date` and `string` cases aren't being handled correctly, I'll see if I can address this too.

PR #4988, opened 2025-08-19
Fix enterprise zoom tests

iMoses:

PR #5555, opened 2025-11-10
AG-16316 Fix grey (invalid) zoom selection behaviour

olegat:
https://ag-grid.atlassian.net/browse/AG-16316

This reverts 71fbd9851f4041548d157eca2ae1176c436db397

This restores the `customSnapshotIdentifier` parameter in `zoom.test.ts`, which allows to test authors to explicitly ensure that image-snapshots of different test-units match (such as `'reset'`).

This fixes the false-positive test:
- `zoom.test.ts -> Zoom -> select -> invalid -> should not zoom on mouseup`

The false-positive test regression happened here:
- https://github.com/ag-grid/ag-charts/pull/4988/files#r2510196504


packages/ag-charts-enterprise/src/features/zoom/zoom.test.ts
```diff
@@ -329,7 +333,7 @@ describe('Zoom', () => {
 
             it('should not zoom on mouseup', async () => {
                 await mouseUpAction(b.x, b.y)(chart);
-                await compare();
```
olegat, line 332: > [!IMPORTANT]
> This is the false-positive unit test that the PR fixes.

alantreadway (approved):
Feedback left offline, we need further discussion about this style of test (overlapping snapshots).

LGTM

PR #5559, opened 2025-11-10
AG-16270 Fix sankey weighting to better group nodes by their neighbours

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-16270


alantreadway (approved):
LGTM

PR #5564, opened 2025-11-10
AG-8627 Add option `zoom.onDataChange.stragegy: 'preserveDomain'`

olegat:


olegat:
Closing. Let's implement `strategy: 'reset'` first – it's much simpler because we only need to call the existing `ZoomManager.resetZoom` method

PR #5578, opened 2025-11-12
AG-16291: Change formatter return `type` to allow numbers and dates

iMoses:
https://ag-grid.atlassian.net/browse/AG-16291


alantreadway (approved):
LGTM

PR #5587, opened 2025-11-13
AG-16239 - Bar series performance improvements pt.1

alantreadway:

### DIFF
--- a/packages/ag-charts-community/src/chart/caption.ts
+++ b/packages/ag-charts-community/src/chart/caption.ts
@@ -82,6 +82,7 @@ export class Caption extends BaseProperties implements CaptionLike {
 
     private truncated = false;
     private proxyText?: BoundedTextWidget;
+    private proxyTextListeners?: Array<() => void>;
 
     registerInteraction(moduleCtx: ModuleContext, where: 'beforebegin' | 'afterend') {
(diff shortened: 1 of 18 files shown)


packages/ag-charts-website/src/content/docs/high-frequency-data-test/_examples/high-freq-high-volume/main.ts
alantreadway, line 1: Updates here are mostly QoL improvements so that reloading the example (e.g. after an incremental build) leaves it in the same state. Also adding support for different series types I need to exercise.

packages/ag-charts-community/src/chart/caption.ts
alantreadway, line 1: There was an actual event listener leak here, caption event listeners were potentially being added on every layout, which showed up as a top performance sink on numerous profiling runs.

packages/ag-charts-community/src/chart/series/cartesian/barSeries.ts
alantreadway, line 1: Actual bar-series high-volume performance optimisations (setting `xNeedsValueOf` and `yNeedsValueOf` into the accumulation functions).

alantreadway:
@codex PTAL

lsjroberts (approved):
LGTM

PR #5603, opened 2025-11-18
AG-16333: Top level `label.truncate` not in options for grouped-categ…

iMoses:
https://ag-grid.atlassian.net/browse/AG-16333


alantreadway (approved):
LGTM

PR #5609, opened 2025-11-19
AG-16337 - Fix listeners reset when `undefined` specified.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-16337


alantreadway:
@codex PTAL

lsjroberts (approved):
LGTM, but may be worth adding a test to check the `chartClick` etc mocks are not called after being cleared, to ensure the `destroy()` is connected up.

PR #5611, opened 2025-11-19
AG-16168 - Remove highlighted callback parameters

alantreadway:
https://ag-grid.atlassian.net/browse/AG-16168


lsjroberts (approved):
LGTM, though you may get a snapshot conflict with https://github.com/ag-grid/ag-charts/pull/5610

PR #5621, opened 2025-11-21
AG-16239 - BarSeries performance improvements Pt.3b; add aggregation tests

alantreadway:

### DIFF
--- a/packages/ag-charts-community/src/chart/series/cartesian/barAggregation.test.ts
+++ b/packages/ag-charts-community/src/chart/series/cartesian/barAggregation.test.ts
@@ -1,0 +1,419 @@
+import { computeBarAggregation } from './barAggregation';
+
+describe('computeBarAggregation', () => {
+    describe('threshold behavior', () => {
+        it('should return undefined for datasets below threshold (< 1000 points)', () => {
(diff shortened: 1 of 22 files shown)


packages/ag-charts-community/src/chart/series/cartesian/barAggregation.test.ts
```diff
@@ -0,0 +1,383 @@
...
+                smallestKeyInterval: undefined,
+                xNeedsValueOf: false,
+                yNeedsValueOf: false,
+            });
+
+            expect(result).toBeDefined();
+
+            for (const filter of result!) {
+                expect(filter.positiveIndices.length).toBeGreaterThan(0);
+                expect(Array.isArray(filter.negativeIndices)).toBe(true);
```
lsjroberts, line 142: Is the idea that `filter.negativeIndices.length` is `0`? Is it worth having that as the expectation instead or just being an array?

lsjroberts (approved):
LGTM

packages/ag-charts-community/src/chart/series/cartesian/barAggregation.test.ts
alantreadway: Fixed.

PR #5629, opened 2025-11-24
AG-16314 Honor `initialState` in the initial zoom event

olegat:
https://ag-grid.atlassian.net/browse/AG-16314


olegat:
Closing. Ticket was reassigned and resolved.

PR #5630, opened 2025-11-24
AG-16328 Fix the expansion of the Range Area `styler` property

olegat:
https://ag-grid.atlassian.net/browse/AG-16328


alantreadway (approved):
LGTM

PR #5631, opened 2025-11-24
Preserve caller stack traces in logger by binding console methods

alantreadway:
## Summary

Adopt AG Dash approach of binding console methods to preserve stack traces. This ensures stack traces point to the original caller instead of the logger wrapper functions, improving debugging experience.

## Changes

- Added bound console function constants (, , , etc.) at the top level
- Refactored all logger functions to use the bound references
- Preserves existing AG Charts-specific prefixes and behavior

## Reference

Based on AG Dash implementation: https://github.com/ag-grid/ag-dash/blob/300bb39442fac4c7c37890db8689bb13640f9a18/packages/ag-dash/src/shared/util/logger.ts

## Testing

- No linting errors
- Public API unchanged
- All existing functionality preserved


alantreadway:
This is not viable without a bigger rework of mocked console calls for tests, parking for now

PR #5659, opened 2025-11-26
ESLint ModuleRegistry: correct false positives for partial axis definitions

alantreadway:
## Problem

The `validate-module-registration` ESLint rule was incorrectly flagging `OrdinalTimeAxisModule` as "registered but not used" when users defined only some axes (e.g., `axes: { y: { type: 'number' } }`) with a candlestick series.

The issue occurred because:
1. The rule used a boolean `hasExplicitAxes` flag that was set to `true` when ANY `axes` property was found
2. `applyDefaultAxes()` would skip entirely when `hasExplicitAxes` was `true`
3. The default x-axis (`ordinal-time`) for candlestick was never added to required modules
4. The registered `OrdinalTimeAxisModule` was then incorrectly flagged as unnecessary

## Solution

- Replace boolean `hasExplicitAxes` flag with a `Set` called `explicitAxes` that tracks which specific axes ('x', 'y', 'angle', 'radius') are explicitly defined
- Update `processAxes()` to record which axis keys are explicitly defined:
  - For object format (`axes: { x: {...}, y: {...} }`): tracks the property names
  - For array format (`axes: [{ position: 'left' }, ...]`): maps positions to axes (bottom/top → x, left/right → y)
- Modify `applyDefaultAxes()` to only apply default axes for axes that were NOT explicitly defined

## Changes

- `libraries/ag-charts-eslint-rules/rules/validate-module-registration.mjs`: Core fix implementation
- `libraries/ag-charts-eslint-rules/src/lint-validate-module-registration.data.ts`: Added comprehensive test cases
- `libraries/ag-charts-eslint-rules/src/__snapshots__/eslint-rules.test.ts.snap`: Updated test snapshots
- Auto-fixed two example files that were missing required default axis modules

## Testing

✅ All ESLint rule tests pass
✅ The original reported examples (simple-candlestick, annotations-customisation) now validate without false positives
✅ No errors about `OrdinalTimeAxisModule` being "registered but not used" for partial axis configurations
✅ The rule still correctly catches genuinely missing modules

Fixes the false positive issue reported where candlestick charts with only y-axis defined were incorrectly flagged.

### DIFF
--- a/libraries/ag-charts-eslint-rules/rules/validate-module-registration.mjs
+++ b/libraries/ag-charts-eslint-rules/rules/validate-module-registration.mjs
@@ -62,7 +62,7 @@ export default {
         let registeredModulesArrayNode = null;
         let requiredModules = new Map(); // moduleId -> { reason, node }
         let isEnterprise = false;
-        let hasExplicitAxes = false;
+        let explicitAxes = new Set(); // tracks 'x', 'y', 'angle', 'radius'
         let seriesTypes = [];
 
         // Import tracking for auto-fix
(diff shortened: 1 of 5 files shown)


lsjroberts (approved):
LGTM

PR #5667, opened 2025-11-27
AG-16360 CR Use `update` instead of `updateDelta`

olegat:
https://ag-grid.atlassian.net/browse/AG-16360


packages/ag-charts-community/src/module/optionsModule.ts
```diff
@@ -177,6 +178,7 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
             });
             this.specialOverrides = this.specialOverridesDefaults({ ...specialOverrides });
         }
+        this.findSeriesWithUserVisiblity(newUserOptions, deltaOptions);
```
olegat, line 183: > [!WARNING]
> This may hinder the fast-setup. The performance impact of `findSeriesWithUserVisibility` is low risk (loops through an array of series, at looks up the `visible` key). We may need to retest sparklines and fine-tune this.

packages/ag-charts-community/src/module/optionsModule.ts
```diff
@@ -223,6 +228,28 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
         Debug.inDevelopmentMode(() => deepFreeze(this));
     }
 
+    private findSeriesWithUserVisiblity(newUserOptions: T, deltaOptions: DeepPartial<T> | null | undefined) {
+        for (const o of [newUserOptions, deltaOptions]) {
+            const series = o?.series;
+            if (Array.isArray(series)) {
```
alantreadway: Nit: Make this a short-circuit in the negative case to reduce nesting here?
```suggestion
            if (!Array.isArray(series)) continue;
```

packages/ag-charts-community/src/module/optionsModule.ts
```diff
@@ -223,6 +228,28 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
...
+                for (let index = 0; index < series.length; index++) {
+                    const s = series[index];
+                    if ('visible' in s) {
```
alantreadway: Nit: Likewise
```suggestion
                    if (!('visible' in s)) continue;
```

packages/ag-charts-community/src/module/optionsModule.ts
alantreadway: It should be ok, it looks to be O(2n), and sane charts are probably n<10 (although we've seen come users with ~150, the charts would be unreadable, and there rendering time for that many series would dominate the update time).

alantreadway (approved):
LGTM

PR #5669, opened 2025-11-27
AG-16254 Fix dataSource.getData undefined params when zoomed out

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-16254


alantreadway (approved):
LGTM

PR #5671, opened 2025-11-27
AG-16239 - BarSeries performance improvements Pt. 3f

alantreadway:


packages/ag-charts-community/src/scene/node.ts
```diff
@@ -287,9 +287,9 @@ export abstract class Node<TDatum = unknown> {
         }
     }
 
-    setProperties<T extends Node>(this: T, styles: { [K in keyof T]?: T[K] }) {
+    setProperties<T extends Node>(this: T, styles: { [K in keyof T]?: T[K] }, keys?: readonly string[]) {
```
lsjroberts: Why not use the `setPropertiesWithKeys()` that already exists?

packages/ag-charts-community/src/scene/shape/shape.ts
```diff
@@ -188,7 +188,7 @@ export abstract class Shape<TDatum = unknown> extends Node<TDatum> {
     }
 
     protected fillStroke(ctx: CanvasContext, path?: Path2D) {
-        if (this.drawingMode === 'cutout') {
+        if ((this as any).__drawingMode === 'cutout') {
```
lsjroberts: These `as any` casts seem a little sketchy, losing quite a bit of type safety in shape, rect and barShape.

lsjroberts (approved):

packages/ag-charts-community/src/scene/shape/shape.ts
alantreadway: Fixed by use of improved decorators + typings requiring the fields to be declared :tada:

packages/ag-charts-community/src/scene/node.ts
alantreadway: Fixed, good spot

lsjroberts (approved):
LGTM, great stuff.

PR #5683, opened 2025-11-28
AG-15405 Add test for aglint/validate-module-registration `flashOnUpdate`

olegat:
https://ag-grid.atlassian.net/browse/AG-15406

### DIFF
--- a/libraries/ag-charts-eslint-rules/rules/validate-module-registration.mjs
+++ b/libraries/ag-charts-eslint-rules/rules/validate-module-registration.mjs
@@ -91,9 +91,9 @@ export default {
         /**
          * Add a required module with reason
          */
-        function requireModule(moduleId, reason, node) {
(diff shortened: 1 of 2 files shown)


olegat:
Closing – The bug is that the `validate-module-registration.mjs` rule only cares about the last `ModuleRegistry.registerModules` call, because the `CallExpression(node)` callback overwrites the `registeredModules` each time the callback is called (therefore only the last one matters in the `'Program:exit'(node)` callback).
* https://github.com/ag-grid/ag-charts/blob/682b6391bd89115c3a472a3467465a030b6d7219/libraries/ag-charts-eslint-rules/rules/validate-module-registration.mjs#L627

It doesn't look so straightforward to fix without significantly rewriting the rule.

I've logged the Tech-Debt ticket for this bug 
https://ag-grid.atlassian.net/browse/AG-16456

PR #5690, opened 2025-12-01
Add benchmark counts and full tables to Slack notifications

alantreadway:
This PR enhances the benchmark CI Slack notifications by:

- **Adding +ve/-ve counts**: The summary header now shows counts of benchmarks that improved vs regressed for both timing and memory
- **Attaching full tables**: Complete timing and memory benchmark tables are now attached as Slack attachments (secondary content below the main message)
- **Improved formatting**: Added proper percentage formatting with + signs and N/A for null values

The main message body still serves as a quick glance summary (top/bottom 5), while the full data is available in the attachments for detailed analysis.

### DIFF
--- a/tools/benchmark/format-slack-message.js
+++ b/tools/benchmark/format-slack-message.js
@@ -82,8 +82,90 @@ function formatSection(data, headers, headerLabels) {
     return table;
 }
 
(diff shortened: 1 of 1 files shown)


manapeirov (approved):
LGTM

PR #5697, opened 2025-12-01
AG-16314 Fix zoom initial event rangeX.min

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-16314

The `layout:complete event` is fired even when an update is shortcut by another update. This shortcut causes the axis scale to not be fully initialised with a domain and so the pixel extents of the scale are unknown.

The update service `update-complete` event allows zoom to check if the update was shortcut and only fire the user's listener after the axis scale domain has been initialised.


alantreadway (approved):
LGTM

PR #5700, opened 2025-12-01
AG-16074 - Improve applyTransaction() update-case + zoom performance.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-16074

PR #5712, opened 2025-12-02
AG-16074 - Address PR review feedback from #5700

alantreadway:
This PR addresses the review feedback from PR #5700:

1. **incrementalProcessor.ts**: Removed redundant `preserveDomainRanges` variable that was always set to `false` in all branches. Domain ranges are now always invalidated directly.

2. **cartesianSeries.ts**: Restored the early-return pattern in `domainForVisibleRange` method - checking `sortOrder != null` first and returning early, keeping the longer fallback block unindented for better readability.

Addresses comments from @lsjroberts on PR #5700.


lsjroberts (approved):
LGTM

PR #5724, opened 2025-12-03
AG-16140 Axes

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-16140


packages/ag-charts-website/src/content/docs/range-bar-series/_examples/multiple-horizontal-range-bars/main.ts
lsjroberts, line 1: This example isn't used in the docs and it covers the same testing ground as the gallery horizontal range-bars + scatter example. So it failed for the same reason.

alantreadway (approved):
LGTM

PR #5731, opened 2025-12-03
AG-16074 - Handle Selection mode transitions to prevent ghost nodes.

alantreadway:


manapeirov (approved):
LGTM

PR #5725, opened 2025-12-03
AG-16074 - Optimise discrete value data processing.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-16074

PR #5732, opened 2025-12-03
AG-16074 - Optimise discrete value data processing Pt. 2

alantreadway:

Depends on https://github.com/ag-grid/ag-charts/pull/5725

### DIFF
--- a/packages/ag-charts-core/src/interfaces/scaleTypes.ts
+++ b/packages/ag-charts-core/src/interfaces/scaleTypes.ts
@@ -2,6 +2,43 @@ import type { AgTimeInterval } from 'ag-charts-types';
 
 export type ScaleType = 'number' | 'log' | 'time' | 'unit-time' | 'ordinal-time' | 'category' | 'mercator' | 'color';
 
+/**
+ * Metadata about a domain's sort characteristics.
+ * Used to skip expensive sort order detection in scales.
+ */
+export interface DomainSortMetadata {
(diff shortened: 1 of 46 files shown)


packages/ag-charts-core/src/interfaces/scaleTypes.ts
```diff
@@ -2,6 +2,43 @@ import type { AgTimeInterval } from 'ag-charts-types';
...
+ * A domain array with optional metadata for optimization.
+ * When metadata is present, scales can skip expensive scans.
+ */
+export interface DomainWithMetadata<D> {
+    domain: D[];
+    sortMetadata?: DomainSortMetadata;
+}
+
+/** Input type for normalizeDomains - either a plain array or annotated domain. */
+export type DomainInput<D> = D[] | DomainWithMetadata<D>;
```
lsjroberts, line 26: Would it simplify things if we only used `DomainWithMetadata`? So we don't need to check the type?

lsjroberts (approved):
LGTM

packages/ag-charts-core/src/interfaces/scaleTypes.ts
alantreadway: Let me see if I can tweak that

alantreadway: Having some issues integrating the changes into my follow-on change, I'll make them on top of those changes instead in that PR.

PR #5739, opened 2025-12-03
AG-16074 - Optimise discrete data processing Pt. 3

alantreadway:


packages/ag-charts-community/src/scale/discreteTimeScale.ts
```diff
@@ -3,6 +3,53 @@ import type { AgTimeInterval, AgTimeIntervalUnit } from 'ag-charts-types';
...
+    endIdx: number = bands.length - 1
+): UniformityCheck {
+    const n = endIdx - startIdx + 1;
+    if (n < 2) return { isUniform: false };
+
+    // Sample SAMPLE_POINTS points evenly spaced within the range
+    const indices = Array.from(
+        { length: SAMPLE_POINTS },
+        (_, i) => startIdx + Math.floor((i * (n - 1)) / (SAMPLE_POINTS - 1))
+    );
```
lsjroberts, line 31: Not sure if it matters, but this will create duplicate indices if `n < SAMPLE_POINTS`. If it does, there's the (unoptimised) `distribute()` fn in `arrays.ts` that is used for sampling data for the axis prediction to solve this issue.

But I imagine this is only used when far greater than 20 datums?

lsjroberts (approved):
LGTM

packages/ag-charts-community/src/scale/discreteTimeScale.ts
alantreadway: Yes, this is only in play if there are at least 1000 data-points in view (or in total).

PR #5743, opened 2025-12-04
AG-8627 Change default strategy to `preserveDomain`

olegat:

### DIFF
--- a/packages/ag-charts-enterprise/src/features/zoom/zoomOnDataChange.ts
+++ b/packages/ag-charts-enterprise/src/features/zoom/zoomOnDataChange.ts
@@ -128,7 +128,7 @@ export class ZoomOnDataChange {
      */
     private computeDomainMinMax(axisId: AxisID): DomainMinMax | undefined {
         const ctx = this.ctx.axisManager.getAxisIdContext(axisId);
-        if (!ctx) return;
+        if (!ctx?.continuous) return;
 
         // Integrated Charts wrappers datum values in a ChartValueWrapper object.
         // `domainAt` reads `ChartValueWrapper.value` when applicable.
--- a/packages/ag-charts-enterprise/src/features/zoom/zoomOnDataChange.ts
+++ b/packages/ag-charts-enterprise/src/features/zoom/zoomOnDataChange.ts
@@ -71,8 +71,7 @@ function fromVisibleMinMax(domainMinMax: DomainMinMax, visibleMinMax: VisibleMin
 // `chart.zoom.onDataChange` options
 export class ZoomOnDataChangeProperties extends BaseProperties implements DeepRequired<AgZoomOnDataChange> {
     @Property
-    // TODO(olegat): change default to 'preserveDomain'
-    strategy: AgZoomOnDataChangeStrategy = 'preserveRatios';
+    strategy: AgZoomOnDataChangeStrategy = 'preserveDomain';
 
     @Property
     // TODO(olegat): change default to 'true'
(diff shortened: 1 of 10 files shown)


packages/ag-charts-enterprise/src/features/zoom/zoomOnDataChange.ts
```diff
@@ -129,10 +128,32 @@ export class ZoomOnDataChange {
      */
     private computeDomainMinMax(axisId: AxisID): DomainMinMax | undefined {
         const ctx = this.ctx.axisManager.getAxisIdContext(axisId);
-        if (!ctx) return;
+        if (!ctx?.continuous) return;
```
olegat, line 131: > [!NOTE]
> Fixes TC11

packages/ag-charts-enterprise/src/features/zoom/zoomOnDataChange.ts
```diff
@@ -129,10 +128,31 @@ export class ZoomOnDataChange {
...
+            if (isChartValueWrapper(elem)) {
+                return elem.value;
+            }
+            return elem;
+        }
 
-        const min: unknown = ctx.scale.domain.at(0);
-        const max: unknown = ctx.scale.domain.at(-1);
+        const min: unknown = domainAt(0);
+        const max: unknown = domainAt(-1);
```
olegat: > [!NOTE]
> Fixes test case:
> ```
> zoom.test.ts > data changes during zoom > should update the data and zoom as expected
> ```

Up for discussion – Should `domainAt/isChartValueWrapper` be part of the `Scale` object? I feel like we should have a built-in mechanism for this (if one doesn't exist already).

packages/ag-charts-enterprise/src/features/zoom/zoomOnDataChange.ts
```diff
@@ -129,10 +128,31 @@ export class ZoomOnDataChange {
...
+                    'value' in o &&
+                    'toString' in o &&
+                    typeof o.id === 'number' &&
+                    o.toString instanceof Function
+                );
+            }
+            const elem: unknown = ctx?.scale.domain.at(index);
+            if (isChartValueWrapper(elem)) {
+                return elem.value;
+            }
```
alantreadway: We have some utilities to help with these cases here:
https://github.com/ag-grid/ag-charts/blob/latest/packages/ag-charts-core/src/utils/value.ts#L30

Might be best to add/integrate as appropriate there?

packages/ag-charts-enterprise/src/features/zoom/zoomOnDataChange.ts
alantreadway: Agreed - not sure exactly how trivial or not that is, but I'd support pushing the concept of `scale.getDomainRange()` to normalise this more - I do wonder if for Integrated cases though that should still return the Integrated object though, since it is actually the datum.

olegat: I think that we can have both (e.g. `scale.getRawDomainRange()` and `scale.getDomainRange()`). But for now I'll move this logic and keep `scale.domain` as-is.

olegat: Done. Added `Scale.getDomainMinMax()`

packages/ag-charts-enterprise/src/features/zoom/zoomOnDataChange.ts
olegat: I've moved the `isChartValueWrapper` logic over there. I've also simplified it.

alantreadway (approved):
LGTM

PR #5747, opened 2025-12-04
AG-16389 - Disable series labels by default to fix mini-chart performance regression.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-16389


manapeirov (approved):
LGTM

PR #5769, opened 2025-12-05
AG-16140 Fix horizontal axes predicting primary axes and positions

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-16140


alantreadway (approved):
LGTM

PR #5795, opened 2025-12-08
CRT-984 - Fix navigator disable not resetting zoom when ZoomModule not registered.

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-984


iMoses (approved):
LGTM

PR #5801, opened 2025-12-08
CRT-1006 Fix angular prefix instance method calls to exclude strings

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-1006


alantreadway (approved):
LGTM

PR #5805, opened 2025-12-08
CRT-1003 Fix angular examples binding functions called within options

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-1003


alantreadway (approved):
LGTM

PR #5806, opened 2025-12-08
CRT-997 - Force clipping for histogram series to fix animations breaching series area.

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-997


lsjroberts (approved):
LGTM

PR #5819, opened 2025-12-08
CRT-1018 Fix function-signature type mismatch

olegat:
https://ag-grid.atlassian.net/browse/CRT-1018

### DIFF
--- a/packages/ag-charts-community/src/api/agCharts.ts
+++ b/packages/ag-charts-community/src/api/agCharts.ts
@@ -286,7 +286,7 @@ class AgChartsInternal {
     }
 
     // CRT-1018 Use `Parameters` and `unknown` to strictly enforce type-safety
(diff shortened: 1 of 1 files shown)


packages/ag-charts-community/src/api/agCharts.ts
```diff
@@ -284,9 +285,16 @@ class AgChartsInternal {
         return proxy;
     }
 
-    private static markRemovedProperties(this: void, node: any, _: unknown, modified = false) {
```
olegat, line 287: > [!IMPORTANT]
> The `_parallelNode` parameter was missing. I do not know why errors did not appears (perhaps because of `any`, or because of optional params, or a combination of both).

 I've added much stricter typing to guard against `jsonWalk` refactoring.

alantreadway (approved):
LGTM

PR #5823, opened 2025-12-09
CRT-1002 Fix radar chart marker size extending pointer interactions

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-1002


alantreadway (approved):
LGTM

PR #5828, opened 2025-12-09
CRT-991 Fix map pin shape artefacts from large stroke widths

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-991


alantreadway (approved):
LGTM

PR #5836, opened 2025-12-09
CRT-1002 Fix radar series closest node pick to use radius of nodes

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-1002


alantreadway (approved):
LGTM

PR #5838, opened 2025-12-09
CRT-1007 - Fix small angle sector rendering when innerRadius > 0.

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-1007


manapeirov (approved):
LGTM

PR #5850, opened 2025-12-11
AG-16426 TC2 Use more explicit Y-axis config in minichart tests

olegat:
https://ag-grid.atlassian.net/browse/AG-16426

These tests are verifying the navigator + X-axis work and stay in-sync with one another. Changes in the Y-axis are pretty much irrelevant, so disable autoScaling and use very explicity min/max Y axis config to make the tests more concise.


alantreadway (approved):
LGTM

PR #5855, opened 2025-12-12
/docs-review fixes

alantreadway:


packages/ag-charts-website/src/content/docs/annotations/index.mdoc
```diff
@@ -27,7 +27,7 @@ For more details about these, and for the full set of annotations, see [Financia
 
 ## Toolbar
 
-The Annotations Toolbar offers menu options for annotating charts. By default, the toolbar is enabled, but it can be disabled by setting `toolbar.enabled` to `false`.
+The Annotations Toolbar offers menu options for annotating charts. The toolbar can be enabled by setting `annotations.enabled` to `true`, or disabled by setting `toolbar.enabled` to `false`.
```
alantreadway: 🤔

AG-DavidG: Our style is actually

The Annotations Toolbar offers menu options for annotating charts and is enabled by default. Use `toolbar.enabled: false` to disable.

packages/ag-charts-website/src/content/docs/context-menu/index.mdoc
```diff
@@ -190,6 +190,7 @@ In this example:
 
 -   The Context Menu on Bar Nodes includes a console `Log Datum` Menu Item,
     whose text is dynamically derived from the Context Menu Event parameters.
+-   When the callback returns `undefined` (e.g., for non-series-node contexts), the context menu falls back to the `items` property configuration. This allows selective customization based on context.
```
AG-DavidG: Reject this change.
This is too complex a case to document. We don't recommend using both these mechanisms together, however the behaviour is in line with convention that returning 'undefined' falls back to the defined options

packages/ag-charts-website/src/content/docs/bar-series/index.mdoc
```diff
@@ -48,7 +48,6 @@ To show a Horizontal Bar Series, set `direction: 'horizontal'`.
 ```
 
 When the `direction` is `'horizontal'` the `xKey` values will be plotted on the default `y` axis, while the `yKey` values will be plotted on the default `x` axis.
-values along the x-axis.
```
AG-DavidG, line 51: 😱
Good catch!

packages/ag-charts-website/src/content/docs/pyramid-series/index.mdoc
```diff
@@ -82,6 +82,18 @@ In this configuration:
 
 -   `aspectRatio` is set such that width will grow `3px` for every `2px` of height.
```
AG-DavidG, line 84: We could leave this in but
- We should add an interactive example
- Remove mention of defaults within docs

packages/ag-charts-website/src/content/docs/radial-column-series/index.mdoc
```diff
@@ -65,7 +65,7 @@ This is changed via the `innerRadiusRatio` option on the Radius Number Axis.
 }
 ```
 
-Any value between `0` and `1` will set the inner radius as a proportion of the overall radius.
+The default value is `0.5`. Any value between `0` and `1` will set the inner radius as a proportion of the overall radius.
```
AG-DavidG: Reject. 
Remove mention of defaults from the docs as they are determined by theme.

packages/ag-charts-website/src/content/docs/range-area-series/_examples/range-area-labels/main.ts
AG-DavidG, line 1: OK. 
Does the same thing in this scenario but if you wanted to add a border you'd use `spacing` for it to look correct.

packages/ag-charts-website/src/content/docs/range-area-series/index.mdoc
AG-DavidG, line 1: Ditto

packages/ag-charts-website/src/content/docs/series-highlighting/_examples/lines/data.ts
AG-DavidG, line 1: Delete or rename?

packages/ag-charts-website/src/content/docs/series-highlighting/_examples/lines/index.html
AG-DavidG, line 1: ditto

packages/ag-charts-website/src/content/docs/series-highlighting/_examples/lines/main.ts
AG-DavidG, line 1: ditto

packages/ag-charts-website/src/content/docs/series-highlighting/index.mdoc
```diff
@@ -4,7 +4,7 @@ title: 'Series Highlighting'
 
 Highlighting the hovered data item or series allows for easier differentiation, especially in charts with many series and data points.
 
-{% chartExampleRunner title="Overlapping Series" name="lines" type="generated" /%}
+{% chartExampleRunner title="Overlapping Series" name="stacked-bars" type="generated" /%}
```
AG-DavidG: Update the 'title' too - there's no overlapping.

packages/ag-charts-website/src/content/docs/supported-frameworks/index.mdoc
AG-DavidG, line 1: Reject?
As 13 > 12, this is not necessary. It only adds overhead when we come to future releases.

AG-DavidG (changes requested):

packages/ag-charts-website/src/content/docs/supported-frameworks/index.mdoc
alantreadway: Lets take it, this is debatable, but updating make the page feel a bit fresher in my mind

AG-DavidG (approved):
LGTM

PR #5857, opened 2025-12-12
feat(axis-modules): add compatibility mapping for axis modules and update default axis requirements

alantreadway:


AG-DavidG (approved):
LGTM - thanks

PR #5860, opened 2025-12-12
AG-NONE Rename Zoom Types

olegat:
The current names are sort of confusing.

* `DefinedZoomState` sounds like a defined version of `ZoomState` (but it's not, it's defined version of `AxisZoomState`).
* `definedZoomState()` sounds like it takes a `ZoomState` type input (but it doesn't, it takes a `AxisZoomState`).
* `AxisZoomState` sounds like the state of a specific axis (but it's not, it's the state of both axes).
* `ZoomState` sounds like the full state of the viewport (but it's not, it's only the state of a specific axis).

New Type Names:
* `ZoomState` -> `ZoomMinMax`
* `ZoomStateDirection` -> `ZoomMinMaxDirection`
* `DefinedZoomState` -> _unchanged_
* `AxisZoomState` -> `ZoomState`

### DIFF
--- a/packages/ag-charts-community/src/chart/interaction/zoomManager.ts
+++ b/packages/ag-charts-community/src/chart/interaction/zoomManager.ts
@@ -22,7 +22,7 @@ import type {
     OptionsDefs,
     RequireOptional,
     Scale,
-    ZoomState,
+    ZoomMinMax,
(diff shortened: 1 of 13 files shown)


lsjroberts (approved):
Yep that's clearer, LGTM!

PR #5865, opened 2025-12-15
AG-16461 Fix zoom custom buttons should replace default buttons

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-16461


alantreadway (approved):
LGTM

PR #5881, opened 2025-12-18
Fix CI benchmark failures on macOS

alantreadway:
- Replace `git restore` with `git checkout` to handle case-sensitive
  filename renames (e.g., geoJson.ts → geojson.ts) on macOS
  case-insensitive filesystem
- Fix release branch selection to use strict regex pattern, excluding
  branches like `test/b13.0.0-crt-regression-tests`


taktran (approved):
LGTM

PR #5897, opened 2026-01-02
Fix e2e test failures due caused by new year

olegat:
For most examples, using `new Date()` is pretty harmless. But for automated tests, this will cause an unusual breakage every time the year changes.

This problem first became apparent when I rebased another PR that I was working on:
* https://github.com/ag-grid/ag-charts/actions/runs/20659233524/job/59318195697


AG-DavidG (approved):
LGTM

PR #5901, opened 2026-01-05
AG-15850 Add `hover-focus-api` example

olegat:

Adds an example to `bar-series-test` with some controls for testing


alantreadway (approved):
LGTM

PR #5908, opened 2026-01-07
AG-15850 Add & implement abstract `Series.findNodeDatum`

olegat:

This will be necessary for implementing the `setState()` behaviour for picked datums.


alantreadway (approved):
LGTM

PR #5951, opened 2026-01-14
AG-7918 - Allow adding a scrollbar to chart axes

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-7918

### DIFF
--- a/packages/ag-charts-community/src/widget/sliderWidget.ts
+++ b/packages/ag-charts-community/src/widget/sliderWidget.ts
@@ -80,6 +80,7 @@ export class SliderWidget extends Widget<HTMLInputElement> {
         const elem = this.getElement();
         elem.value = `${value}`;
         elem.ariaValueText = ariaValueText;
+        elem.ariaValueNow = `${value}`;
     }
 
     getValueRatio() {
--- a/packages/ag-charts-enterprise/src/features/scrollbar/scrollbarDOMProxy.ts
+++ b/packages/ag-charts-enterprise/src/features/scrollbar/scrollbarDOMProxy.ts
@@ -1,0 +1,142 @@
+import { _ModuleSupport } from 'ag-charts-community';
+import { clamp } from 'ag-charts-core';
+import type { BoxBounds } from 'ag-charts-core';
+
+const { SliderWidget } = _ModuleSupport;
+
+type ScrollbarDOMProxyCtx = Pick<_ModuleSupport.ModuleContext, 'proxyInteractionService' | 'localeManager'>;
+
+export class ScrollbarDOMProxy {
+    private readonly toolbar: _ModuleSupport.ToolbarWidget;
+    private readonly slider: _ModuleSupport.SliderWidget;
+
+    private span = 1;
+    private dragStartRatio = 0;
+    private dragMoved = false;
+    private isDragging = false;
+
+    constructor(
+        private readonly ctx: ScrollbarDOMProxyCtx,
+        private readonly orientation: 'horizontal' | 'vertical',
+        private readonly onChange: (min: number, max: number) => void,
+        private readonly onHoverChange: (hovered: boolean) => void
+    ) {
+        this.toolbar = ctx.proxyInteractionService.createProxyContainer({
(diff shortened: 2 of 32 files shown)


packages/ag-charts-community/src/widget/sliderWidget.ts
```diff
@@ -80,6 +80,7 @@ export class SliderWidget extends Widget<HTMLInputElement> {
         const elem = this.getElement();
         elem.value = `${value}`;
         elem.ariaValueText = ariaValueText;
+        elem.ariaValueNow = `${value}`;
```
manapeirov, line 83: @olegat what do you think of this? will it break anything

packages/ag-charts-enterprise/src/features/zoom/zoomAutoScale.ts
```diff
@@ -61,6 +61,8 @@ export class ZoomAutoScaler {
     }
 
     private onChangeRequest(event: _ModuleSupport.ZoomChangeRequestEvent) {
+        this.manuallyAdjusted = event.manualAdjustment ?? this.manuallyAdjusted;
```
olegat: It seems that adding `ZoomChangeRequestEvent.manualAdjustment` duplicates the intended purposes of `source` and `sourceDetail`.

Understandably, this PR follows is following the same pattern as `ZoomChangeRequestEvent.isReset`, but that's tech-debt rather than a desired outcome. Previously, `ZoomAutoScale.onChangeRequest` used `source == 'reset'` instead of `isReset` (see https://github.com/ag-grid/ag-charts/commit/ea138e71f284d744262da71cee62bc72172467e3), which is ideally what we'd want.

It would be nice to address this tech-debt with `isReset` at some point. But for now I think that checking `sourceDetail === 'scrollbar'` here is sufficient.

packages/ag-charts-community/src/widget/sliderWidget.ts
olegat: Let spotted! This property is actually required for both the `slider` and `scrollbar` roles. It should indeed also be a decimal number (not-localised). So this looks correct ✅

Spec:
* https://w3c.github.io/aria/#slider
* https://w3c.github.io/aria/#scrollbar
* https://w3c.github.io/aria/#aria-valuenow

packages/ag-charts-enterprise/src/features/scrollbar/scrollbarDOMProxy.ts
```diff
@@ -0,0 +1,142 @@
...
+            classList: ['ag-charts-proxy-scrollbar', `ag-charts-proxy-scrollbar-${orientation}`],
+            orientation,
+            ariaLabel: { id: 'ariaLabelNavigatorRange' },
+        });
+
+        this.slider = ctx.proxyInteractionService.createProxyElement({
+            type: 'slider',
+            domIndex: 0,
+            ariaLabel: { id: 'ariaLabelNavigatorRange' },
+            role: 'scrollbar',
```
olegat: Stick with `role="slider"` for now. I'll add more info in the JIRA ticket.

packages/ag-charts-enterprise/src/features/scrollbar/scrollbarDOMProxy.ts
```diff
@@ -0,0 +1,142 @@
...
+    private dragMoved = false;
+    private isDragging = false;
+
+    constructor(
+        private readonly ctx: ScrollbarDOMProxyCtx,
+        private readonly orientation: 'horizontal' | 'vertical',
+        private readonly onChange: (min: number, max: number) => void,
+        private readonly onHoverChange: (hovered: boolean) => void
+    ) {
+        this.toolbar = ctx.proxyInteractionService.createProxyContainer({
```
olegat: I don't think that you need `role="toolbar"` in this case. The navigator uses a toolbar because it has three sliders. Having a toolbar for `this.slider` just adds an unnecessary intermediate element in the Accessibility Tree.

alantreadway (approved):
LGTM

PR #5953, opened 2026-01-14
AG-13929 - SSR package implementation.

alantreadway:


packages/ag-charts-server-side/src/agChartsServerSide.ts
```diff
@@ -0,0 +1,147 @@
...
+                height,
+                overrideDevicePixelRatio: pixelRatio,
+            } as any);
+
+            await this.waitWithTimeout(chart.waitForUpdate(), timeout);
+
+            const exportOptions = format === 'jpeg' && quality !== undefined ? { quality: quality / 100 } : undefined;
+            const buffer = mainCanvas.toBufferSync(format, exportOptions);
+
+            chart.destroy();
```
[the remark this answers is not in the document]
alantreadway: Fixed

packages/ag-charts-server-side/src/agChartsServerSide.ts
```diff
@@ -0,0 +1,147 @@
...
+
+            const container = env.document.getElementById('container')!;
+
+            // Note: as any is required because AgCharts API methods have incompatible union types
+            // that TypeScript cannot narrow properly when api is a union of method names
+            const chart = AgCharts[api]({
+                ...options,
+                container,
+                document: env.document,
+                window: env.window,
```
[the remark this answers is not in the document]
alantreadway: Added mutex to avoid race conditions/interference.

alantreadway:
@codex

packages/ag-charts-server-side/src/environment.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export function createIsolatedEnvironment(): IsolatedEnvironment {
+    const dom = new JSDOM('<!DOCTYPE html><html><body><div id="container"></div></body></html>', {
+        url: 'http://localhost/',
+    });
+
+    const win = dom.window as unknown as ServerSideWindow;
+
+    win.requestAnimationFrame = (cb: FrameRequestCallback) => setTimeout(cb, 0) as unknown as number;
```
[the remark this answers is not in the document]
alantreadway: Fixed

packages/ag-charts-server-side/src/agChartsServerSide.ts
```diff
@@ -0,0 +1,135 @@
...
+        try {
+            patchDocumentCreateElement(env.document, {
+                getCanvas: () => canvasStack.shift() ?? new NodeCanvas(width * pixelRatio, height * pixelRatio),
+            });
+
+            const container = env.document.getElementById('container')!;
+
+            // Note: as any is required because AgCharts API methods have incompatible union types
+            // that TypeScript cannot narrow properly when api is a union of method names
+            const createdChart = AgCharts[api]({
```
lsjroberts, line 100: Could alternatively switch on `api` and call the methods directly?

packages/ag-charts-server-side/src/canvasConfig.ts
```diff
@@ -0,0 +1,77 @@
...
+        return bitmap;
+    }
+}
+
+let patchesApplied = false;
+export function applySkiaPatches(): void {
+    if (patchesApplied) return;
+    patchesApplied = true;
+
+    // https://github.com/samizdatco/skia-canvas/issues/241
```
lsjroberts, line 43: This issue appears to have been fixed? https://github.com/loveloki/skia-canvas/commit/bbf3094825f85706373b64c9a5ce573e5ccf49e9

Or are we using an older version?

packages/ag-charts-server-side/src/types.ts
```diff
@@ -0,0 +1,42 @@
...
+    pixelRatio?: number;
+    /** JPEG quality 0-100 (only applies when format is 'jpeg') */
+    quality?: number;
+    /** Render timeout in milliseconds (default: 30000) */
+    timeout?: number;
+}
+
+/** Generic render options parameterized by chart options type */
+export interface ChartRenderOptions<T> extends BaseRenderOptions {
+    options: T;
```
lsjroberts: Should this omit `width` and `height` from `T`? How would that be handled alongside the render options of width and height?

lsjroberts (approved):
LGTM, some minor points and questions

packages/ag-charts-server-side/src/canvasConfig.ts
alantreadway: We're on an older version, I'll take a note to look at bumping us up as a follow up as that might be a bigger change :-/

packages/ag-charts-server-side/src/types.ts
alantreadway: I'll make the adjustment, David and I offline discussed also removing `container`, although I guess we need to be careful not to bust type compatibility with `AgChartOptions`.

packages/ag-charts-server-side/src/agChartsServerSide.ts
alantreadway: Although that is technically possible, the relationship between `api` type and options type would need to be captured and correctly managed here to make that work (in the function parameters + generics), and I think the complexity of capturing that is probably worse than using an `any` here.

packages/ag-charts-server-side/src/types.ts
alantreadway: Fix pushed

PR #5955, opened 2026-01-14
AG-13929 - Add tests for ag-charts-server-side package

alantreadway:


alantreadway:
@codex

packages/ag-charts-server-side/src/agChartsServerSide.test.ts
```diff
@@ -0,0 +1,293 @@
...
+            );
+
+            const buffers = await Promise.all(promises);
+
+            expect(buffers).toHaveLength(5);
+
+            // Each buffer should match its own unique snapshot
+            for (const buffer of buffers) {
+                expect(buffer).toMatchImageSnapshot(IMAGE_SNAPSHOT_OPTIONS);
+            }
```
lsjroberts, line 216: Is it possible to verify and correlate each snapshot to each set of options to ensure no bleed across them?

lsjroberts (approved):
LGTM

packages/ag-charts-server-side/src/agChartsServerSide.test.ts
alantreadway: Discussed offline, as the order of cases matches the `buffers` order, this is sufficient for now.

PR #5954, opened 2026-01-14
Restore graceful ag-charts-prompts handling in setup-prompts.sh

alantreadway:
## Summary
- Restore functionality for handling `ag-charts-prompts` repo that was removed in the recent rulesync refactor
- Add graceful error handling so missing prompts repo doesn't block `yarn install`
- Support interactive prompts to clone/update the prompts repo
- Handle git worktrees correctly (prompts dir adjacent to main repo, not worktree)

PR #5960, opened 2026-01-15
Handle rulesync failures gracefully in setup-prompts.sh

alantreadway:
## Summary
- Handle rulesync failures gracefully instead of failing `yarn postinstall`
- When rulesync encounters broken symlinks (e.g., from missing `external/prompts`), warn but continue
- Show the error message so users understand what's missing

## Context
Follow-up to #5954. Users without `ag-charts-prompts` cloned were still experiencing failures because rulesync would fail when trying to read broken symlinks in `.rulesync/rules/`.

## Test plan
- [x] Verified with broken symlink - script warns but exits with code 0
- [x] Verified normal operation still works correctly


taktran (approved):
LGTM

manapeirov (approved):
LGTM

PR #5966, opened 2026-01-16
AG-13929 - Add SSR package tests + cleanup.

alantreadway:


packages/ag-charts-core/jest.jsdom-env.cjs
```diff
@@ -1,8 +1,8 @@
 const { TestEnvironment } = require('jest-environment-jsdom');
-const { DOMMatrix, Image } = require('skia-canvas');
+const { DOMMatrix, Image, Canvas } = require('skia-canvas');
 const timezoneMock = require('timezone-mock');
 
-const { mockCanvas } = require('ag-charts-test');
+const { ConfiguredCanvasMixin } = require('ag-charts-core');
```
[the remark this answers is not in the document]
alantreadway: Fixed

libraries/ag-charts-test/src/canvas/mock-canvas.ts
alantreadway, line 1: I've moved this to `ag-charts-core` as otherwise we'd have duplicate code in `ag-charts-test` and `ag-charts-server-side` or cyclic dependencies between `ag-charts-community`, `ag-charts-test` and `ag-charts-server-side`.

The code in `ag-charts-core` doesn't have a direct dependency on `skia-canvas`, and is a standalone function/class, so should be treeshakable easily.

packages/ag-charts-core/src/state/memento.test.ts
alantreadway, line 1: Removing this dependency from `ag-charts-core` to `ag-charts-test` as it was the only dependency and helps straighten out the build dependency chain.

packages/ag-charts-core/src/utils/configuredCanvasMixin.ts
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        transferToImageBitmap(): CanvasLike {
+            const { width, height } = this;
+            const bitmap = new ConfiguredCanvasClass(Math.max(1, width), Math.max(1, height));
+            if (width > 0 && height > 0) {
+                try {
+                    (bitmap.getContext('2d') as any).drawCanvas(this, 0, 0, width, height);
+                } catch {
+                    // Skia-canvas can throw dimensionless errors even when width/height checks pass
```
lsjroberts: What happens in this case? An empty image and silent error?

packages/ag-charts-server-side/src/test/constants.ts
```diff
@@ -1,6 +1 @@
-export const CANVAS_WIDTH = 800;
-export const CANVAS_HEIGHT = 600;
-
-export const CANVAS_TO_BUFFER_DEFAULTS = {
-    quality: 1, // Disable MSAA for consistent output
-};
+export { CANVAS_WIDTH, CANVAS_HEIGHT, CANVAS_TO_BUFFER_DEFAULTS } from 'ag-charts-core';
```
lsjroberts: Do we need to re-export these, or can it just be imported from core where it is used?

lsjroberts (approved):
LGTM

packages/ag-charts-core/src/utils/configuredCanvasMixin.ts
alantreadway: This is how the code looked in the previous release looking at the history of where this originally came from - I agree it should probably be fail-fast rather than silent.

alantreadway: Fixed, now fails-fast.

packages/ag-charts-server-side/src/test/constants.ts
alantreadway: Removed - good call.

PR #5973, opened 2026-01-19
AG-15850 Move `class PickManager` into a separate file

olegat:


olegat:
@codex review

PR #5979, opened 2026-01-19
AG-13929 - Better align server-side package tests.

alantreadway:


tests/shared/run.sh
```diff
@@ -62,8 +63,13 @@ while getopts ":eniupc" opt; do
...
 
-version=$1
-project=${2:-/project}
+if [[ ${test_type} == "ssr" ]] ; then
+    version=${1:-}
+    project=${2:-/project}
+else
+    version=$1
+    project=${2:-/project}
+fi
```
lsjroberts, line 71: nit: `project=${2:-/project}` looks like it can be moved outside this if/else block

lsjroberts (approved):
LGTM

tests/shared/run.sh
alantreadway: Fixed, thanks

PR #5991, opened 2026-01-21
AG-16611 - Fix SSR watermark.

alantreadway:
https://ag-grid.atlassian.net/browse/AG-16611


packages/ag-charts-enterprise/src/license/licenseManager.ts
```diff
@@ -182,6 +184,35 @@ export class LicenseManager {
...
+            return undefined;
+        }
+        return this.buildWatermarkConfig(message);
+    }
+
+    public getWatermarkForegroundConfigForBrowser(): object | undefined {
+        if (!this.isDisplayWatermark()) {
+            return undefined;
+        }
+        return this.buildWatermarkConfig(this.getWatermarkMessage());
```
lsjroberts: This one doesn't check for `!message`, does that matter?

tests/server-side-package-tests/e2e/basic-snapshots/enterprise-linear-gauge.png
lsjroberts, line 1: Should a gauge not have a watermark?

lsjroberts (approved):
LGTM

tests/server-side-package-tests/e2e/basic-snapshots/enterprise-linear-gauge.png
alantreadway: Fixed, thanks, not sure why I didn't spot that

packages/ag-charts-enterprise/src/license/licenseManager.ts
alantreadway: There is an edge case where `#forcedWatermark` can be used in-browser to force the watermark display with no message (for debugging) so we need to keep this as an allowed case for in-browser testing.

PR #6002, opened 2026-01-22
AG-6511 Add zoom when fixed width bars exceed chart width

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-6511


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1260,6 +1304,15 @@ export abstract class Chart extends Observable implements ModuleInstance, ChartS
         }
     }
 
+    protected preSeriesUpdate() {
+        const { _requiredWidth } = this;
+        const seriesRectWidth = this.seriesRect?.width ?? 0;
+        const requiredWidthRatio = _requiredWidth / seriesRectWidth;
+
```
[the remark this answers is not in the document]
lsjroberts: There doesn't seem any reasonable scenario this will happen, but adding a skip just to be safe.

packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -782,6 +784,13 @@ export abstract class Chart extends Observable implements ModuleInstance, ChartS
                 this.updateSplits('⛰️');
             // fallthrough
 
+            case ChartUpdateType.PROCESS_RANGE:
+                if (this.checkUpdateShortcut(ChartUpdateType.PROCESS_DOMAIN)) break;
```
alantreadway:
```suggestion
                if (this.checkUpdateShortcut(ChartUpdateType.PROCESS_RANGE)) break;
```
?

packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -790,6 +799,13 @@ export abstract class Chart extends Observable implements ModuleInstance, ChartS
                 this.updateSplits('⌖');
             // fallthrough
 
+            case ChartUpdateType.PRE_SERIES_UPDATE:
+                if (this.checkUpdateShortcut(ChartUpdateType.PERFORM_LAYOUT)) break;
```
alantreadway:
```suggestion
                if (this.checkUpdateShortcut(ChartUpdateType.PRE_SERIES_UPDATE)) break;
```
?

packages/ag-charts-community/src/chart/interaction/zoomManager.ts
alantreadway, line 1: Maybe worth having @olegat review here?

olegat: I think it's good for now.

We may want to consider migrating this "required-width" logic into it's own submodule in the enterprise package (like `zoomOnDataChange.ts` and `zoomAutoScale.ts`) to avoid delegating too much responsibility to `ZoomManager`. But that's not essential, it may be done later.

olegat (approved):

PR #6027, opened 2026-01-26
Fix scheduled benchmark workflow for b13.0.x branches

alantreadway:
## Summary
- Add special case handling for b13.0.x series (where x >= 1) to fall back to b13.0.0 as the base for scheduled benchmarks
- The b13.0.1 branch has a circular dependency issue that prevents benchmark data collection, causing the overnight benchmark job to fail

## Test plan
- [x] Verified the regex logic handles edge cases correctly:
  - `b13.0.1` → `b13.0.0` (falls back)
  - `b13.0.0` → `b13.0.0` (unchanged)
  - `b13.1.0` → `b13.1.0` (unchanged)
  - `b12.3.1` → `b12.3.1` (unchanged)
- [ ] Manually trigger the benchmark workflow via workflow_dispatch with `compare-base: b13.0.0` to confirm it works
- [ ] Wait for next scheduled run to verify


taktran (approved):
LGTM

PR #6028, opened 2026-01-26
AG-16535 Add band alignment for category axis

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-16535


packages/ag-charts-community/src/chart/axis/axis.ts
```diff
@@ -239,10 +239,11 @@ export abstract class Axis<
 
     layoutConstraints: ChartAxis['layoutConstraints'] = {
         stacked: true,
-        align: 'start',
+        align: 'justify',
```
lsjroberts, line 242: This is only used internally, so it's not a breaking change. It only comes into effect for axes that have a set `width`. Which previously was only the volume y-axis on the priceVolume preset. For axes with the default 100% width it has no impact. And `justify` is what we are treating as our current default.

alantreadway (approved):
LGTM

PR #6029, opened 2026-01-26
AG-15850 Fix active state `itemId` on bar series, and add tests

olegat:


alantreadway (approved):
LGTM

PR #6036, opened 2026-01-27
AG-7918 - Focus indicator should be around the thumb not around the t…

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-7918


alantreadway (approved):
LGTM

PR #6078, opened 2026-02-02
AG-16605- Fix voiceover announcements and keyboard tab navigation

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-16605

### DIFF
--- a/packages/ag-charts-community/src/dom/proxyInteractionService.ts
+++ b/packages/ag-charts-community/src/dom/proxyInteractionService.ts
@@ -35,7 +35,7 @@ type ContainerParams<T extends ProxyContainerType> = {
     readonly type: T;
     readonly domManagerId: string;
     readonly classList: string[];
-    readonly ariaLabel?: TranslationKey;
+    readonly ariaLabel: TranslationKey | undefined;
     readonly role?: string;
 };
 
--- a/packages/ag-charts-community/src/dom/proxyInteractionService.ts
+++ b/packages/ag-charts-community/src/dom/proxyInteractionService.ts
@@ -35,7 +35,8 @@ type ContainerParams<T extends ProxyContainerType> = {
     readonly type: T;
     readonly domManagerId: string;
     readonly classList: string[];
-    readonly ariaLabel: TranslationKey;
+    readonly ariaLabel?: TranslationKey;
+    readonly role?: string;
 };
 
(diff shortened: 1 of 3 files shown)


packages/ag-charts-community/src/dom/proxyInteractionService.ts
```diff
@@ -35,7 +35,8 @@ type ContainerParams<T extends ProxyContainerType> = {
     readonly type: T;
     readonly domManagerId: string;
     readonly classList: string[];
-    readonly ariaLabel: TranslationKey;
+    readonly ariaLabel?: TranslationKey;
```
olegat:
```suggestion
    readonly ariaLabel: TranslationKey | undefined;
```
Let's keep `ariaLabel` explicit, we want to avoid new calls to `createProxyContainer()` from accidentally forgetting to specify the label.

We could potentially even take it one step further and make ContainersParams a union where ariaLabel is always required if `role!=='presentation'`, and always disallowed when `role==='presentation'`, but that is a bit more overkill:

```ts
type ContainerParams<T extends ProxyContainerType> = {
    readonly type: T;
    readonly domManagerId: string;
    readonly classList: string[];
    readonly ariaLabel: TranslationKey;
    readonly role?: never;
} |  {
    readonly type: T;
    readonly domManagerId: string;
    readonly classList: string[];
    readonly ariaLabel?: never;
    readonly role: 'presentation';
} 
```

packages/ag-charts-enterprise/src/features/scrollbar/scrollbarDOMProxy.ts
```diff
@@ -165,13 +165,16 @@ export class ScrollbarDOMProxy {
             type: 'group',
             domManagerId: `scrollbar-${orientation}`,
             classList: ['ag-charts-proxy-scrollbar', `ag-charts-proxy-scrollbar-${orientation}`],
-            ariaLabel: { id: 'ariaLabelNavigatorRange' },
+            role: 'presentation',
```
olegat, line 169:
```suggestion
            ariaLabel: undefined,
            role: 'presentation',
```

olegat (approved):
LGTM: Just an optional nit to improve the type-safety of `ContainerParams`

alantreadway (approved):
LGTM

PR #6080, opened 2026-02-02
AG-3390 - Replace mdoc prettier patch with vscode markdown association

taktran:
There is already configuration for mdoc in `.prettierrc`


alantreadway (approved):
LGTM

PR #6085, opened 2026-02-02
AG-15850 Remove `itemId` from box-plot node datum

olegat:
We're now using numeric itemIds instead of xKey strings

The culprit is that `xValues[number]` is of type `any` (which annoyingly is assignable to type `never`).


alantreadway (approved):
LGTM

PR #6099, opened 2026-02-04
AG-15850 TC9 Fix tooltip not shown when call `setState({active})` on maps

olegat:


alantreadway (approved):
LGTM

PR #6100, opened 2026-02-04
Ag 16449/marker drawing modes

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-16449


packages/ag-charts-community/src/chart/series/cartesian/bubbleSeries.ts
```diff
@@ -1375,6 +1372,11 @@ export class BubbleSeries extends CartesianSeries<BubbleSeriesTypes> {
         return !!(styler ?? itemStyler ?? marker.itemStyler ?? label.itemStyler);
     }
 
+    protected override shouldIgnoreDefaultMarkerFillOpacity(): boolean {
+        const seriesPath = ['series', `${this.declarationOrder}`];
+        return !this.ctx.optionsGraphService.hasUserOption([...seriesPath, 'fillOpacity']);
```
lsjroberts: Can this be determined through the graph instead? For example `drawingMode: { $isUserOption: ['./fillOpacity', true, false] }` as an undocumented option.

packages/ag-charts-community/src/chart/optionsGraphService.ts
```diff
@@ -15,12 +16,18 @@ type ResolvePartialOpts = {
...
         this.resolvePartialCallback = resolvePartialCallback;
+        this.hasUserOptionCallback = hasUserOptionCallback;
     }
 
     resolvePartial(path: Array<string>, partialOptions?: PlainObject, resolveOptions?: ResolvePartialOpts) {
         return this.resolvePartialCallback?.(path, partialOptions, resolveOptions);
     }
+
+    hasUserOption(path: Array<string>) {
+        return this.hasUserOptionCallback?.(path) ?? false;
```
lsjroberts: We should not expose the options graph without requiring calls to be processed through the graph itself. Otherwise this will lead to bugs where the order of node dependency resolution is not correct.

packages/ag-charts-community/src/module/optionsGraph.ts
```diff
@@ -433,6 +433,7 @@ export class OptionsGraph extends AdjacencyListGraph<unknown, string> implements
     hasUserOption(path: Array<string>) {
         const hasUserOptionSimple = hasPathSafe(this.userOptions, path);
         if (hasUserOptionSimple) return true;
+        if (!this.root) return false;
```
lsjroberts: The reason this is needed is because the graph is deleted after it is resolved. This a reason why we shouldn't make direct calls into the graph externally, but instead should utilise graph operations, i.e. `$isUserOption`.

So, we shouldn't need this line.

packages/ag-charts-community/src/chart/series/cartesian/bubbleSeries.ts
manapeirov: actually that might be a really good idea

alantreadway (approved):
LGTM

PR #6134, opened 2026-02-06
AG-16496 - Fix donut sector label highlight style

alantreadway:
## Summary
- Fixes donut/pie sector label styling during highlighting to use per-datum highlight state instead of selection-level boolean
- Previously all sector labels were styled as "highlighted" when any datum was highlighted, because `updateSelection` used a blanket `isHighlight` flag based on which selection layer was being updated
- Now each label resolves its own highlight state from the highlight manager, so only the actually highlighted datum's label gets the highlighted style

## Test plan
- [x] Updated image snapshots for highlight dimming (donut, pie, polar series)
- [x] Updated `donutSeries.test.ts` snapshot — removed spurious `sectorLabel.itemStyler` calls with `highlightState: "highlighted-item"` that were incorrectly firing for all datums


manapeirov (approved):
LGTM

PR #6141, opened 2026-02-06
AG-16713 Listen for `active:update` events in SeriesAreaManager

olegat:
https://ag-grid.atlassian.net/browse/AG-16713

Bugfix: Clears series-area highlight state on legend highlight state change

PR #6142, opened 2026-02-06
docs tweaks

AG-DavidG:


packages/ag-charts-website/src/content/docs/events/index.mdoc
```diff
@@ -67,7 +67,16 @@ In this example:
 
 ### activeChange
 
-This event is fired when the [active](./api-state/#active) state (highlight) is changed.
+This event is fired when the [active](./api-state/#active) state is changed.
+This occurs when a user interaction (mouse, touch, keyboard) on a series node or legend causes a highlight or tooltip change.
```
olegat, line 71: Nit: The `activeChange` callback is also fired when calling `setState` (with `source: 'state-change'`)

olegat (approved):

packages/ag-charts-website/src/content/docs/api-state/_examples/active-save-restore/main.ts
```diff
@@ -43,13 +44,21 @@ const options: AgCartesianChartOptions<DataType> = {
     },
     legend: {
         position: 'bottom',
+        toggleSeries: false,
```
olegat, line 47: Do we want the legend to be non-interactive? When testing this option locally, the legend becomes non-interactive. I've got a PR in progress to fix the bugs that we're seeing with the legend https://github.com/ag-grid/ag-charts/pull/6141

packages/ag-charts-website/src/content/docs/events/index.mdoc
AG-DavidG: This matches the other event descriptions on the page.
I'd argue that technically `setState()` changes the state.

packages/ag-charts-website/src/content/docs/api-state/_examples/active-save-restore/main.ts
AG-DavidG: Discussed and agreed that this allows clicking on the legend to save the active state and then restore it without confusing the user.

PR #6166, opened 2026-02-09
CRT-1033 Fix navigator.enabled & zoom.enabled toggling

olegat:
https://ag-grid.atlassian.net/browse/CRT-1033

The navigator.ts module does not need to call ZoomManager.updateZoom when navigator.enabled is `false`, the chart.ts will automatically reset the zoom if it detects that both navigator.enabled and zoom.enabled are `false`.

Also, chart.ts was only resetting the X axis instead of the X and Y axis.


alantreadway (approved):
LGTM

PR #6170, opened 2026-02-09
Crt-1048: Axes label collision for waterfall example

iMoses:
revert previous fix, fixed collision detection logic


alantreadway (approved):
LGTM

PR #6179, opened 2026-02-10
CRT-1038 fix tooltip freezing during click-drag

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-1038


olegat (approved):
LGTM - assuming that dragging with `zoom.enabled: true` still works as intended.

I suspect that the CI would have failed if that's not the case, but I'm not 100% certain if we have reliable automated test for zoom-dragging/tooltip. Specifically, watch out for tooltip "flickering" we've had in the past (and which the automated tests might not notice).

PR #6183, opened 2026-02-10
CRT-1054 fix range-area-with-labels example animation

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-1054

Fix duplicate Date keys in the range-area-with-labels gallery example data (Nov listed twice per year instead of Nov+Dec), which caused `uniqueKeys` validation to fail and disabled all animation. Also adds `AnimationModule` registration and removes unused `ContextMenuModule` import.


lsjroberts (approved):
LGTM

PR #6186, opened 2026-02-10
CRT-1031 Fix annotations text dropping last character to newline

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-1031

This was only seen in one example on firefox, but theoretically may occur in other browsers depending on floating point math.


alantreadway (approved):
LGTM

PR #6194, opened 2026-02-11
CRT-1047 Add e2e test for legend item focus with font family change

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-1047


packages/ag-charts-website/e2e/keyboard-nav.spec.ts-snapshots/CRT-1047-after-change-font-chromium-linux.png
lsjroberts, line 1: Previously, this focus indicator would have been in it's old position, so offset to the right and a different size.

alantreadway (approved):
LGTM

PR #6200, opened 2026-02-11
AG-16745: Heatmap - Highlighting shows inconsistent border lines

iMoses:
https://ag-grid.atlassian.net/browse/AG-16745


alantreadway (approved):
LGTM

packages/ag-charts-website/e2e/toolbar.spec.ts-snapshots/callout-6-change-fill-color-chromium-linux.png
lsjroberts, line 1: If I was being very picky, I'd say the 8th box from the right looks less centred now. Fine as is, just worth checking it doesn't flicker in motion when zooming & panning.

lsjroberts (approved):
LGTM

packages/ag-charts-website/e2e/toolbar.spec.ts-snapshots/callout-6-change-fill-color-chromium-linux.png
iMoses: I looked into it, measured the changes. Before spacing was 2px on the left and 3px on the right, and now it's the opposite. I don't think it matters much considering it's a very specific edge cas, and that it does solve the empty line bug.

PR #6209, opened 2026-02-17
AG-16519: Chart interactions fail when rendered in popup window

iMoses:
Minimal changes to support rendering a chart inside a child window.


packages/ag-charts-community/src/chart/tooltip/tooltip.ts
```diff
@@ -444,8 +443,14 @@ export class Tooltip extends BaseProperties {
         this.element.togglePopover(visible);
```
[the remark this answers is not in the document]
iMoses: This feature became available on all evergreen browsers on April 2024.
We don't need to safeguard it anymore.

packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -412,7 +419,12 @@ export abstract class Chart extends Observable implements ModuleInstance, ChartS
             () => this.subtitle.destroy(),
             () => this.footnote.destroy(),
 
-            Widget.addWindowEvent('page-left', () => this.destroy()),
+            this.ctx.agDocument.attachListener('pagehide', (event: PageTransitionEvent) => {
```
alantreadway, line 422: Is `pagehide` equivelant to `page-left`?

iMoses: These is no page left in that sense. 
Logic wasn't changes, I just removed a layer of abstraction.
The event we listen to was always page hide.

iMoses: See previous implementation: https://github.com/ag-grid/ag-charts/pull/6209/changes#diff-260ea17de25587206a642950a1e49d944a31969693e0214c2e5e8f5f57eb3001L311

alantreadway (approved):
LGTM

PR #6210, opened 2026-02-17
Add rulesync agentic tooling crib-sheet

alantreadway:
Quick-reference README for all AI agent commands, skills, sub-agents, and rules available in the repo. Documents provenance (local vs shared), activation patterns, and workflow groupings.


AG-DavidG (approved):
LGTM

PR #6214, opened 2026-02-18
AG-16727 Add skipNullBars widthRatio tests

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-16727


packages/ag-charts-community/src/chart/series/cartesian/__image_snapshots__/bar-series-test-ts-bar-series-skip-null-bars-width-ratio-grouped-1-snap.png
lsjroberts, line 1: The gaps between bars here is expected since the `widthRatio` only affects the rendered bar, not the space it takes up. Unlike `width`, which does reduce the occupied space.

alantreadway (approved):
LGTM

PR #6219, opened 2026-02-18
Consolidate memory commands into /remember and /recall

alantreadway:
Merges four separate agentic memory commands (`/context-remember`, `/git-branch-save-context`, `/git-branch-load-context`) into two intuitive commands:

- **`/remember`** — save branch context or project learnings as memory
- **`/recall`** — load branch context and browse project memories

`/optimise-context` stays separate as a maintenance/audit tool.

**Cross-product note:** AG Grid will need its `.rulesync/commands/` symlinks updated similarly (remove 3 old, add 2 new) after `git subrepo push`.


AG-DavidG (approved):
LGTM

PR #6221, opened 2026-02-19
AG-16704 Cleanup unnecessary types in `pickManager.ts`

olegat:
https://ag-grid.atlassian.net/browse/AG-16704

* Remove `IPickManager` interface (we can just use `PickManager` directly).
* Remove unhelpful `PickedNode` intermediate type


alantreadway (approved):
LGTM

PR #6242, opened 2026-02-24
AG-16787 Fix chart download crash when legend hidden due to small size

alantreadway:
https://ag-grid.atlassian.net/browse/AG-16787


manapeirov (approved):
LGTM

PR #6248, opened 2026-02-24
AG-16720 Bubble all events

olegat:
https://ag-grid.atlassian.net/browse/AG-16736
https://ag-grid.atlassian.net/browse/AG-16376


packages/ag-charts-community/src/chart/legend/__image_snapshots__/legend-event-test-ts-legend-event-click-and-visibility-change-events-legend-item-click-prevent-default-1-snap.png
olegat, line 1: The new image-snapshot looks correct. Verified by manually testing the test-case in Chrome:
https://plnkr.co/edit/5n4ypTHBNw43s1Ex?open=main.js

I suspect that the old image-snapshot was faulty because the Jest unit-tests do not have proper browser-based event bubbling support (although we attempt to partially emulate event-bubbling).

PR #6249, opened 2026-02-25
Ag 15406/flash on update

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-15406


packages/ag-charts-enterprise/src/features/flash-on-update/flashOnUpdate.ts
```diff
@@ -42,130 +60,269 @@ export class FlashOnUpdate extends BaseProperties implements ModuleInstance, AgF
...
+        opts?: { onPlay?: () => void }
+    ): void {
+        if (rects.length === 0) return;
+
+        const { opacity } = this;
+        const { animationManager } = this.ctx;
+        const timing = this.getCustomTiming();
+        const phaseDuration = _ModuleSupport.PHASE_METADATA[phase].animationDuration;
+        const duration = timing ? timing.duration * phaseDuration : undefined;
+        const ease = timing?.ease;
```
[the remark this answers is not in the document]
alantreadway: This is acceptable for now as long as we document this in the option JSDoc

packages/ag-charts-enterprise/src/features/flash-on-update/flashOnUpdate.ts
```diff
@@ -42,130 +60,260 @@ export class FlashOnUpdate extends BaseProperties implements ModuleInstance, AgF
...
+            else updateRects.push(rect);
+        }
+
+        this.animate(removeRects, 'remove');
+        this.animate(updateRects, 'update', {
+            onPlay: () => this.refreshRectBounds(updateRects),
+        });
+        this.animate(addRects, 'add', {
+            onPlay: () => this.refreshRectBounds(addRects),
+        });
```
alantreadway, line 333: For high-volume cases, are we consolidating bands together (so continuous added/remove/updated flashes are actually just a single `Rect` rather than one per rendered node)?

packages/ag-charts-enterprise/src/features/flash-on-update/flashOnUpdate.test.ts
alantreadway, line 1: Nice coverage - is there a way we could also programmatically assert in a non-visual way that the flash is present or not as expected at the point of each snapshot?

I'd assert non-zero area of the rect(s) and opacity/fill perhaps?

Maybe this is even a class of test-case we haven't added yet, which can assert against the scene-graph state?

alantreadway (approved):
LGTM, thanks for the updates

PR #6252, opened 2026-02-26
Remove direnv configuration

alantreadway:
Remove `.envrc` and the `opt_enable_direnv()` function from the cloud install script. The direnv setup was for personal multi-account Claude configuration and isn't used by the team.


AG-DavidG (approved):
LGTM

PR #6253, opened 2026-02-26
Update examples rules with integrated charts and API guidelines

AG-DavidG:
Adds guidance for AG Grid integrated charts setup (script load order, chartDataType) and AG Charts initialState zoom structure (rangeX/rangeY, AgStateSerializableDate format). Also adds a note about using the plunker skill for Plunker operations.


alantreadway (approved):
LGTM

PR #6256, opened 2026-02-27
AG-3390 Handle concatenated JSON output from newer Snyk CLI versions

taktran:
## Summary

- Newer Snyk CLI versions (via the floating `snyk/snyk:node` Docker image used by `snyk/actions/node@master`) append workspace error objects as separate JSON documents after the main results array, rather than embedding them within it
- This caused `JSON.parse` to fail with `SyntaxError: Unexpected non-whitespace character after JSON` when the root workspace path (`/github/workspace`) returned a `"Forbidden"` error from the Snyk API
- Fixed by catching the parse error, extracting the byte position from the error message, and re-parsing only the first (valid) JSON document

## Test plan

- [x] Verify the security scan CI job passes on next scheduled run
- [x] Confirm the warning annotation appears in the job log when the truncation fallback is triggered

### DIFF
--- a/external/ag-shared/.gitrepo
+++ b/external/ag-shared/.gitrepo
@@ -6,7 +6,7 @@
 [subrepo]
 	remote = https://github.com/ag-grid/ag-shared.git
 	branch = latest
-	commit = f999ddabe0d121bf65607f7c2bdab710335fab06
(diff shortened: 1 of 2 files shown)


taktran:
Confirmed in https://github.com/ag-grid/ag-charts/actions/runs/22482757537

seanlandsman (approved):

PR #6257, opened 2026-02-27
AG-16720 Refactor widgetEvents.ts

olegat:

This rewrites the file in a Single-Source-Of-Truth (SSOT) style. Previously, adding new event types involved modifying too many different parts of the file. Now, we only need to add/remove entries from the `WIDGET_META` constant, and all other types are derived from that using meta-programming.


alantreadway (approved):
LGTM

PR #6259, opened 2026-02-27
AG-15840 Eliminate forced synchronous relayouts via DOM read deferral and write caching

alantreadway:
https://ag-grid.atlassian.net/browse/AG-15840

## AG-15840: Eliminate Forced Synchronous Relayout

**Problem:** The chart was triggering forced synchronous browser reflows by reading DOM geometry (`getBoundingClientRect`, `offsetWidth`, etc.) immediately after writes — a classic layout-thrashing pattern causing significant re-layout performance overhead.

Fixes:
- Defer DOM writes on hot-paths for `crosshair`, `tooltip`, `overlay`.
- Adds caching of canvas size with `ResizeObserver`-based invalidation.
- Strategically sets root canvas font to reduce thrashing of font during render (avoiding forced style recalculate).
  - Broke titles into their own layer to avoid font thrashing for non-initial-load cases. 
- Use of cache invalidation listeners on various browser events to ensure consistency on demand.
- Addition of e2e tests to detect forced reflows for common cases.


packages/ag-charts-community/src/dom/domManager.ts
```diff
@@ -417,7 +428,8 @@ export class DOMManager extends BaseManager {
 
     /** Get the main chart area client bound rect. */
     getBoundingClientRect() {
-        return this.rootElements['canvas'].element.getBoundingClientRect();
+        this._cachedCanvasRect ??= this.rootElements['canvas'].element.getBoundingClientRect();
+        return this._cachedCanvasRect;
```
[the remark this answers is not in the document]
alantreadway: Added ResizeObserver to invalidate the cached value.

packages/ag-charts-community/src/chart/tooltip/tooltip.ts
```diff
@@ -190,6 +187,8 @@ export class Tooltip extends BaseProperties {
     private arrowPosition: ArrowPosition | undefined = undefined;
     private _visible = false;
 
+    private dom?: DOMElementProxy;
```
lsjroberts: Renaming `element` to `dom` is confusing, since we are not referring to the dom here, only an element. Even if it is via a proxy.

e.g. calling `dom.toggleClass()` implies we are toggling the class on the html/body tag or similar

packages/ag-charts-community/src/dom/domElementProxy.ts
```diff
@@ -0,0 +1,197 @@
...
+ *
+ * - Compare-before-write optimisation for hot paths (style properties, classes,
+ *   attributes, innerHTML): a flat key→value cache with `===` comparison skips
+ *   redundant writes. Callers must serialise objects before comparison.
+ * - Cache lifecycle: `invalidate()` clears a single key; `reset()` clears all
+ *   cached values on hide/destroy transitions.
+ * - Delegated operations (events, structural mutations) pass through to the
+ *   underlying element without caching.
+ */
+export class DOMElementProxy {
```
lsjroberts, line 51: Proxying and caching all the html elements is a bit concerning. This seems like it would introduce subtle bugs, like stale caches and particularly with end-user css/js.

lsjroberts (approved):
LGTM (+ rename of dom)

packages/ag-charts-community/src/dom/domElementProxy.ts
lsjroberts: Discussed offline

alantreadway: Discussed offline - render pipeline will flush at the end, other cases not involving a re-render cycle need to manually `flush()`. I'll update the docs here to cover this off.

packages/ag-charts-community/src/chart/tooltip/tooltip.ts
alantreadway: I'll try `elementProxy` for size.

packages/ag-charts-community/src/dom/domElementProxy.ts
```diff
@@ -0,0 +1,128 @@
...
+    private cache: Record<string, unknown> = {};
```
olegat: Optional: consider using a stricter type like:
```ts
'innerHTML' | 'contentStyles' | `${'p' | 'a' | 'c'}:${string}`
```

packages/ag-charts-community/src/dom/domElementProxy.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    /** Delegates to setting element.innerText. Invalidates innerHTML cache. */
+    set innerText(text: string) {
+        this.element.innerText = text;
+        this.invalidate('innerHTML');
+    }
+
+    /** Clear all cached values. Call on hide/destroy transitions. */
+    reset(): void {
+        this.cache = {};
```
olegat, line 232: Optional: Might be worth comparing the performance difference between `Record` and `Map`. Although it's probably OK, I would expect the difference to be negligible for small mappings.

packages/ag-charts-community/src/dom/domManager.ts
```diff
@@ -654,6 +696,34 @@ export class DOMManager extends BaseManager {
...
+
+    private setupGlobalListeners() {
+        const win = this.element.ownerDocument.defaultView;
+        if (win == null) return;
+
+        const invalidateRects = () => this.invalidateRectCaches();
+        const invalidateAll = () => this.invalidateAllCaches();
+
+        win.addEventListener('scroll', invalidateRects, { capture: true, passive: true });
+        win.addEventListener('resize', invalidateRects, { passive: true });
```
olegat: Do we also invalidate the cache when only the chart resizes? There could be a cases where that could happen:
1. User interactions: if the chart occupies 100% of a resizable container, which the user resizes.
2. Programmatically: if the chart (or one of its ancestors) gets a `style` update that changes its size. An obvious case could be changing the chart container's `display: none` to `display: block` – we want to make sure that we're not reading outdated widths/heights of 0px.

packages/ag-charts-community/src/chart/tooltip/tooltip.ts
alantreadway: Renamed.

packages/ag-charts-community/src/dom/domElementProxy.ts
alantreadway: I've added JSDoc to explain these points to future readers.

packages/ag-charts-community/src/dom/domElementProxy.ts
olegat, line 1: LGTM - Nicely done!

Optional Improvement: The type-safety of `setProperty` and `setAttr` could be improved a bit. The `attributeUtil.ts` utility has been helpful at catching careless typos of attribute names & values (like `elem.setAttribute('aria-labeledby', myId)` instead of  `elem.setAttribute('aria-labelledby', myId)`).
https://github.com/ag-grid/ag-charts/blob/e95092e6501f406cb3ba4aa03eb29457f5181d26/packages/ag-charts-core/src/utils/dom/attributeUtil.ts

So `setProperty` / `setAttr` could benefit from using  `setElementStyle` / `setAttribute` internally.

packages/ag-charts-community/src/dom/domManager.ts
olegat: Potentially, the same goes for scrolling too. Does the cache is correctly cleared when the chart is inside a scrollable container? I would expect the `win` scroll listener to take care of that, but we'll need to verify that.

olegat (commented):
My only main concern is that the scroll and resize event listeners in `DOMManager.setupGlobalListener` on...:

```ts
const win = this.element.ownerDocument.defaultView
```

... might be too shallow. This needs to be verified this with a chart with a scroll/resizeable container to ensure that the cache is correctly cleared. Example: https://plnkr.co/edit/jcZXmbHkQALfYpjN?open=main.js

packages/ag-charts-community/src/dom/domElementProxy.ts
alantreadway: Fair, done

packages/ag-charts-community/src/dom/domElementProxy.ts
alantreadway: For the number of operation (probably ~O(10)) this is trivial and should be fine.

alantreadway:
> Example: https://plnkr.co/edit/jcZXmbHkQALfYpjN?open=main.js

Thanks for the example. I've tested it with the changes, I don't see any odd behaviour with tooltips or crosshair labels after scrolling and resizing the container, so practically this looks to be ok

packages/ag-charts-community/src/dom/domManager.ts
alantreadway: Thanks, I've tried your example and verified that tooltip + crosshairs honour both the enclosing viewport bounds, and the window bounds if the viewport is wider than the window bounds.

packages/ag-charts-community/src/dom/domElementProxy.ts
alantreadway: I've added stricter typings 👍

PR #6277, opened 2026-03-03
Route chart update requests through eventsHub

alantreadway:
Removes `UpdateService.update()` and routes all chart update requests through `eventsHub.emit('chart:request-update', ...)` instead. A single centralised listener in `chart.ts` handles the event. `UpdateService` is retained for its lifecycle dispatch methods (`pre-series-update`, `update-complete`, etc.).


lsjroberts (approved):
LGTM

PR #6291, opened 2026-03-04
AG-16852: FIX - Comment and Callout annotations do not resize when te…

iMoses:
…xt is updated by setState

https://ag-grid.atlassian.net/browse/AG-16852


alantreadway (approved):
LGTM

PR #6296, opened 2026-03-04
AG-16825 TC4 Fix PgUp/PgDown on reversed X axes

olegat:
https://ag-grid.atlassian.net/browse/AG-16825

### DIFF
--- a/packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts
+++ b/packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts
@@ -854,30 +854,55 @@ export abstract class CartesianSeries<TTypes extends CartesianSeriesTypes> exten
         let left: number = 0;
         let mid: number;
         let right: number = this.contextNodeData.nodeData.length - 1;
+        const reverse: boolean = this.axes.x?.reverse ?? false;
 
         const shrinkBounds: (focusBBox: Readonly<BoxBounds>) => void = (function initShrinkBounds() {
             if (where === 'viewport-start') {
-                return function shrinkViewportStart(focusBBox: Readonly<BoxBounds>): void {
-                    // Looking for smallest index whose left edge is inside viewport
-                    if (focusBBox.x >= hoverRect.x) {
-                        resultIndex = mid;
-                        right = mid - 1;
-                    } else {
-                        left = mid + 1;
-                    }
-                };
+                // Looking for smallest index whose left edge is inside viewport
+                if (reverse) {
+                    return function shrinkViewportStartReverse(focusBBox: Readonly<BoxBounds>): void {
+                        const viewportRight = hoverRect.x + hoverRect.width;
+                        const nodeRight = focusBBox.x + focusBBox.width;
+                        if (nodeRight <= viewportRight) {
+                            resultIndex = mid;
+                            right = mid - 1;
+                        } else {
+                            left = mid + 1;
+                        }
+                    };
+                } else {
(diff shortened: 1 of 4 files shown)


packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts
```diff
@@ -849,40 +849,71 @@ export abstract class CartesianSeries<TTypes extends CartesianSeriesTypes> exten
...
+        let result: PickFocusOutputs | undefined = undefined;
 
         let left: number = 0;
         let mid: number;
         let right: number = this.contextNodeData.nodeData.length - 1;
+        const reverse: boolean = this.axes.x?.reverse ?? false;
 
-        const shrinkBounds: (focusBBox: Readonly<BoxBounds>) => void = (function initShrinkBounds() {
+        // Generate a binary-search iterator (`where` and `reverse`)
+        const iterate: (pickedFocus: PickFocusOutputs) => void = (function initIterator() {
```
alantreadway: Nit: The cyclomatic complexity is quite high here, and there seems to be some surface level similarity in the 4 nested functions, could we factor these out to improve readability here?

olegat: The algorithm basically needs to ask two questions:

Which boundary are we testing?
1.  left predicate: `nodeLeft >= viewportLeft`
2.  right predicate: `nodeRight <= viewportRight`

Which direction does the index move when the predicate is true?
1.  move left: `if(predicate) { result = pickedFocus; left = mid + 1; } else { right = mid - 1 } `
2.  move right: `if(predicate) { result = pickedFocus; right = mid - 1; } else { left = mid + 1 }`

So there are two predicates × two search directions = four permutations that the iterator needs to support.

These permutations depend on both `where` and `reverse`:

|                      | `reverse = true` | `reverse = false` |
|----------------------|------------------|-------------------|
| `'viewport-start'`   | right predicate<br>move right  | left predicate<br>move right    |
| `'viewport-end'`     | left predicate<br>move left  | right predicate<br>move left   |

You could in theory derive the predicate using an XOR operation:

```ts
const predicate = ((where === 'viewport-end') ^ reverse) ? rightPredicate : leftPredicate;
if (where === 'viewport-start') {
   moveRight(predicate);
} else if (where === 'viewport-end') {
   moveLeft(predicate)
}
```

However, I'm not a big fan of this approach. The flattened condition makes it look like there are only 2 cases, when in reality there are still 4 logical permutations. That tends to make the code harder to reason about later.

If we really wanted to avoid the nested if-statements, another option would be to encode the truth table explicitly, for example:


```ts
const truthTable = {
  'viewport-start': {
    move: moveRight,
    predicate: leftPredicate,
    reversePredicate: rightPredicate,
  },
  'viewport-end': {
    move: moveLeft,
    predicate: rightPredicate,
    reversePredicate: leftPredicate,
  },
}

const iterate = () => {
    const { move, predicate, reversePredicate } = truthTable[where];
    move( reverse ? reversePredicate : predicate );
}
```

Personally I find the nested version a bit more direct, since it maps discreetly to the four concrete cases in the table above. That said, if the truth-table approach reads better to others, I'm happy to refactor it.

alantreadway: Sure, could you at least address the cyclomatic complexity here and avoid the double nested functions on top of these nested conditional blocks (factor this logic out to a distinct method please)?

alantreadway (approved):
LGTM

PR #6298, opened 2026-03-04
AG-16843 Only set clipped when min/max narrows the domain

alantreadway:
https://ag-grid.atlassian.net/browse/AG-16843


manapeirov (approved):
LGTM

PR #6299, opened 2026-03-05
Add validate-prompts skill with path exception rules

alantreadway:
Adds the `validate-prompts` skill (`/validate-prompts`) for checking `.rulesync/` source files for path reference hygiene.

**Skill capabilities:**
- Detects bare `skills/`/`rules/` references missing the `.rulesync/` prefix
- Flags `external/` paths that should use `.rulesync/` equivalents
- Detects `.claude/` references in source files (generated output paths)
- Checks for stale symlinks and orphaned generated files
- Cross-validates backtick file path references for existence

**Exception rules added** to avoid false positives:
- `.rulesync/README.md` excluded from Steps 2 & 3 (structural overview referencing both conventions by design)
- `external/prompts/**/_*.md` underscore-prefixed partials excluded from Step 2 (no `.rulesync/` equivalent exists)
- `external/prompts/technical-review*/**` files excluded from Step 2 (live in `external/` by design)

All excluded paths are still validated for existence in Step 6.

Also includes:
- README entry for the new skill
- `NX_DAEMON=false` note in `ag-charts.md` rules

### DIFF
--- a/.rulesync/README.md
+++ b/.rulesync/README.md
@@ -74,6 +74,12 @@ Quick-reference for all AI agent commands, skills, sub-agents, and rules availab
 | Agent   | 🟠 `technical-research-analyst`    | Auto                                 | In-depth technical research with citations         |
 | Agent   | 🔵 `nx-expert`                     | Auto                                 | Nx monorepo configuration and build optimisation   |
 
+## Prompt Hygiene
+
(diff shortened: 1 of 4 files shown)


AG-DavidG (approved):
LGTM

PR #6302, opened 2026-03-05
Reduce watch/build reload time by ~1.4s

alantreadway:
Four low-risk changes that cut fixed dead time out of the dev watch pipeline:

- **Quiet period debounce:** 1000ms → 50ms (`constants.js`). IDE multi-file saves complete well under 50ms; git operations are already handled by `isBuildBlocked()`.
- **HMR plugin debounce:** 300ms → 10ms (`agHotModuleReload.ts`). The watch script already batches builds and touches the signal file exactly once per cycle.
- **Skip minification in watch config** (`esbuild.config.cjs`). The dev server consumes unminified ESM/CJS; `.min.js` outputs are not needed during watch.
- **Combined Nx invocations:** a single `nx run-many -t build:umd build` replaces 2–3 separate spawns (~280–620ms saving). Uses `NX_FORCE_REUSE_CACHED_GRAPH` and a `nx run` fast path for single-project changes.
- **Remove `build:package` from `build:umd` critical path** (`community/project.json`, `enterprise/project.json`). Both UMD builds resolve all imports from source via `tsconfig.base.json` paths — the inherited `dependsOn: ["build:package", "^build:package"]` was dead weight (~200–600ms for core changes). Validated by removing all `dist/package/` output and confirming `build:umd` still succeeds. The `build` noop still depends on all three sub-targets, so production builds are unaffected.

Also adds `WATCH.md` (pipeline design doc) and inline comments to all three `*Watch.config.js` files explaining ordering, fan-out, and reload-gating decisions.

**Expected combined saving:** ~0.7–2.0s per rebuild cycle for a core source change (cached rebuilds: ~0.3–1.1s vs ~1.8–3.0s before).

### DIFF
--- a/esbuild.config.cjs
+++ b/esbuild.config.cjs
@@ -152,7 +152,9 @@ if (process.env.NX_TASK_TARGET_TARGET?.endsWith('umd')) {
     };
 }
 
-plugins.push(postBuildMinificationPlugin);
+if (process.env.NX_TASK_TARGET_CONFIGURATION !== 'watch') {
+    plugins.push(postBuildMinificationPlugin);
+}
 
 /** @type {import('esbuild').BuildOptions} */
 const options = {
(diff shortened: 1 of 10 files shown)


taktran (approved):
LGTM

PR #6304, opened 2026-03-05
AG-16825 Add upperbound safety check in binary search

olegat:
https://ag-grid.atlassian.net/browse/AG-16825

As far as I can tell, this while-loop is not the root cause of the infinite loop bug that we're seeing. The bug seems to be in DataModelSeries.pickFocus(), which can return the wrong node-datum if you press Home/End/PgUp/PgDown.


alantreadway (approved):
LGTM

PR #6315, opened 2026-03-06
AG-16795 Add range buttons dropdown

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-16795

Targets https://github.com/ag-grid/ag-charts/pull/6315


lsjroberts:
I believe codex is suggesting not re-using the toolbar buttons and instead having a separate dropdown button. This is a reasonable suggestion for accessibility, so I will look into it. Though perhaps in a follow-up PR.

alantreadway (approved):
LGTM

PR #6319, opened 2026-03-06
AG-13929 Iterate on SSR feedback

alantreadway:
Addresses review feedback on the SSR documentation and package setup:

- Make `ag-charts-enterprise` a runtime dependency of `ag-charts-server-side` (removes need for separate install)
- Always import enterprise for watermark support without requiring `isEnterprise()` guard
- Add Codespaces forwarded-port URL note
- Hide framework switcher on SSR docs page (Node.js only)
- Clean up stale comments and remove redundant "Enterprise features" docs section

Fix AG-13929


AG-DavidG (approved):
LGTM

PR #6337, opened 2026-03-10
Use CoW clone for .nx directory instead of symlink in worktrees

alantreadway:
## Summary
- Replace symlink + partial copy of `.nx` with a single CoW (`cp -cR`) clone of the entire directory, matching the existing `node_modules` approach
- Each worktree gets an independent `.nx` that starts warm with the root's cache but doesn't compete with the main repo's Nx process


iMoses (approved):
LGTM

PR #6338, opened 2026-03-10
AG-16049: Fix unit-time axis with initialState zoom initial render

iMoses:
https://ag-grid.atlassian.net/browse/AG-16049


alantreadway (approved):
LGTM

PR #6347, opened 2026-03-11
AG-16807 Add ZoomInteraction and AxisDOMProxy modules to allow scrollbar scrolling

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-16807


packages/ag-charts-enterprise/.size-limit.js
alantreadway, line 1: I think @iMoses just bumped these so you maybe able to drop this

alantreadway (commented):
@olegat would you mind reviewing too?

Here's an AI-generated architectural overview:
https://gist.github.com/alantreadway/c1caad6cb03ab5a2d64090e30e87c5e3

olegat (approved):
LGTM. Left some feedback on JIRA, but no further actions/improvements are needed at this point.

PR #6351, opened 2026-03-12
AG-16783 Fix calendar range buttons month intervals for varying lengths

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-16783

PR #6354, opened 2026-03-12
AG-12033 Fix range buttons out of range

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-12033

Targets https://github.com/ag-grid/ag-charts/pull/6351


alantreadway (approved):
LGTM

PR #6357, opened 2026-03-12
Add single-track-value rule and UMD module registration note

AG-DavidG:
## Summary
- **JIRA skill**: Add note that each ticket takes exactly one track value — prevents setting multiple tracks on a single ticket
- **Plunker skill**: Clarify that vanilla JS UMD bundles auto-register all modules, so `setupModules()` / `ModuleRegistry.register()` calls are unnecessary

## Test plan
- [x] Ran skill-creator eval workflow with 3 test cases (multi-track bug, UMD module reg, dual-track feature)
- [x] Eval 3 (dual-track feature) confirmed the old skill set two tracks; new skill correctly flags the constraint
- [x] Eval 2 confirmed both old and new skill handle UMD correctly (new note reinforces existing behaviour)


alantreadway (approved):
LGTM

PR #6358, opened 2026-03-12
AG-16729 Decouple highlight.enabled from data flow to crosshairs

alantreadway:
[AG-16729](https://ag-grid.atlassian.net/browse/AG-16729)

## Summary

Decouples `highlight.enabled` from the data processing pipeline so that crosshairs, tooltips, and other non-visual consumers continue to work when highlight is disabled on a series.

**Problem:** Setting `highlight.enabled: false` on a series previously suppressed the highlight datum at source (`HighlightManager.updateHighlight`), which blocked crosshairs from snapping to data points and prevented cursor updates.

**Approach:** Remove the upstream gate in `HighlightManager` and the early-exit in `Series.pickNodeClosestDatum`, then add downstream visual guards:

- `Series.getHighlightState()` returns `None` when the source series has highlight disabled
- `Series.hasHighlightOpacity()` returns `false` when the source series has highlight disabled
- `ErrorBar.onHighlightChange()` checks `isHighlightEnabled()` before applying highlight styles
- `HighlightChangeEvent` gains a `highlightSuppressed` flag, set centrally in `HighlightManager`, so consumers can check a single event property
- `SeriesAreaManager.changeHighlightDatum()` skips `SERIES_UPDATE` dispatch when highlight is suppressed on both previous and current series
- `Crosshair.onHighlightChange()` requests its own `SCENE_RENDER` instead of relying on `SERIES_UPDATE` as a side-effect


olegat (commented):
LGTM. Thanks for making these changes.

The snapshot update for `highlight-disabled-series-focus-chromium-linux.png` looks suspicious; the crosshairs shouldn't be enabled in that example. Might need more investigation, but perhaps it's simply a matter of explicitly setting `crosshair.enabled: false` in the example's options.

alantreadway:
> The snapshot update for `highlight-disabled-series-focus-chromium-linux.png` looks suspicious; the crosshairs shouldn't be enabled in that example. Might need more investigation, but perhaps it's simply a matter of explicitly setting `crosshair.enabled: false` in the example's options.

Fixed 🤞

packages/ag-charts-website/e2e/keyboard-nav.spec.ts-snapshots/highlight-disabled-series-other-series-focus-chromium-linux.png
olegat, line 1: Something about this image-snapshot update still smells fishy. The removal of the crosshair is expected, but why is the Bananas series (green) opaque? I wouldn't have expected the series opacity to be affected like this.

alantreadway: Should be fixed now, only the crosshair is remove in-alignment with example changes.

olegat: The snapshots look OK now. The crosshair is removed (expected) but opacities are unchanged 👍

olegat (approved):

PR #6368, opened 2026-03-13
AG-16076 Support dataIdKey in activeItem state and events

alantreadway:
https://ag-grid.atlassian.net/browse/AG-16076

Fix `refreshSetState()` to clear active state when the active datum is removed by a transaction or data update. Add integration tests for `dataIdKey` + active state flows.


iMoses (approved):
LGTM

PR #6372, opened 2026-03-13
Add plan workflow to JIRA skill for plan-mode support

alantreadway:
## Summary

- Adds a `workflows/plan.md` to the JIRA skill so it triggers during plan mode and embeds correct field defaults (Track values, component mappings, description templates, Cloud ID) into plans
- Updates skill description to trigger on planning-related keywords
- Fixes incorrect claim that the MCP API doesn't support `createIssueLink`

**Problem:** When planning JIRA ticket creation, the skill wasn't triggered. Plans were authored with wrong defaults (e.g. "Charts" as a Track value, free-form `##` headers instead of TC templates) that failed at execution time.

**Eval results:** 100% assertion pass rate with skill vs 23% without across 3 test cases (bug, feature, tech-debt).

## Test plan

- [x] Eval 1: Bug plan uses correct Track=Bug, Component=Charts, TC template, UUID Cloud ID
- [x] Eval 2: Feature plan uses Track=Feature Request, Type=Task, 12-section numbered template
- [x] Eval 3: Tech-debt plan uses Track=Housekeeping, Type=Task, correct Component and Cloud ID

### DIFF
--- a/external/ag-shared/prompts/skills/jira/SKILL.md
+++ b/external/ag-shared/prompts/skills/jira/SKILL.md
@@ -4,13 +4,15 @@ name: jira
 description: >-
   Whenever the user asks to create a JIRA ticket, file a bug, log an issue,
   write up a ticket, estimate a ticket, size effort, analyse a JIRA issue, do
-  product analysis, or link tickets — ALWAYS invoke this skill first. It provides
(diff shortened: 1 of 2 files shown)


AG-DavidG (approved):
LGTM

PR #6390, opened 2026-03-16
Improve rulesync skill description for better triggering

alantreadway:
Rewrites the rulesync skill description to improve auto-triggering accuracy and ensure it co-loads with the skill-creator plugin.

Key changes:
- Leads with imperative action verbs (creating, editing, registering, troubleshooting)
- Enumerates specific artifacts (SKILL.md, frontmatter fields, setup-prompts.sh, verify-rulesync.sh)
- Explicitly ties to skill-creator as a co-dependency
- Mentions the external/ content tier system
- States the constraint that generated directories must not be edited directly


AG-DavidG (approved):
LGTM

PR #6391, opened 2026-03-16
AG-16919 Fix line animation crash when switching between datasets of different lengths

alantreadway:
https://ag-grid.atlassian.net/browse/AG-16919


manapeirov (approved):
LGTM

PR #6399, opened 2026-03-17
AG-16910: Flip direction within groups when RTL is enabled

iMoses:
https://ag-grid.atlassian.net/browse/AG-16910


alantreadway (approved):
LGTM

PR #6402, opened 2026-03-17
AG-16729 Hide gradient legend arrow when highlight is disabled

alantreadway:
https://ag-grid.atlassian.net/browse/AG-16729

When a series has `highlight: { enabled: false }`, the gradient legend arrow should not appear on hover. Previously `updateArrow()` only checked for a null `colorValue` — it now also checks `isHighlightEnabled()` on the highlighted series.


manapeirov (approved):
LGTM

PR #6417, opened 2026-03-18
AG-16783 Fix ranges calendar intervals with unit-time axes

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-16783


packages/ag-charts-website/e2e/ranges.spec.ts-snapshots/range-buttons-actions-calendar-chromium-linux.png
lsjroberts, line 1: Diff expected as this fix no longer artificially clamps a unit-time to the end of the domain.

alantreadway (approved):
LGTM

PR #6421, opened 2026-03-18
AG-16887 Focus on navigator when Mouse-Left-Down triggers

olegat:
https://ag-grid.atlassian.net/browse/AG-16887


alantreadway (approved):
LGTM

PR #6430, opened 2026-03-18
AG-16820 Fix cone funnels to use user fills and strokes

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-16820


alantreadway (approved):
LGTM

PR #6432, opened 2026-03-18
AG-16883 Fix linear gauge labels tick text color

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-16883


alantreadway (approved):
LGTM

PR #6433, opened 2026-03-18
Fix flaky E2E snapshot tests for ranges and fonts

alantreadway:
Screenshots in `ranges.spec.ts` were captured before charts finished re-rendering because `getByText().click()` bypasses the stability proxy (which only wraps `page.mouse`/`keyboard`/`touchscreen`). Added `waitForAllChartUpdates()` after each click.

The `fonts.spec.ts` test used a hard-coded 200ms delay for the font-load → ResizeObserver → re-render chain. Replaced with polling for `data-scene-renders` increment.

These two issues caused three consecutive CI failures on `latest`.


lsjroberts (approved):
LGTM

PR #6440, opened 2026-03-19
AG-16861 Fix frozen datum showing/hiding when toggling legend-items

olegat:
https://ag-grid.atlassian.net/browse/AG-16861


alantreadway (approved):
LGTM

PR #6441, opened 2026-03-19
AG-16934 Fix Omit type display in SSR API reference docs

alantreadway:
https://ag-grid.atlassian.net/browse/AG-16934

The docs generator rendered `Omit<T, K>` utility types as bare `Omit;` (no type arguments) in the API reference for server-side rendering interfaces. Two root causes:

1. Private (non-exported) type aliases used as the second arg to `Omit` were preserved in `.d.ts` output but unresolvable by the docs toolchain
2. `normalizeType()` unconditionally stripped type arguments from all `TypeReferenceNode`s

**Fix:**
- Inline the literal unions in `Omit` second args (removes `AgRenderExcludedKeys` and `AgCartesianAxisThemeSpecialOptions` private aliases)
- Show type arguments for `Omit` and `Pick` in `normalizeType()`
- Update `applyGenericsToMember()` to check the raw AST node for Omit detection instead of the normalised string


iMoses (approved):
LGTM

PR #6450, opened 2026-03-19
AG-16915 Guard bubble series against NaN size values

alantreadway:
## Summary

- Skip rendering bubbles when `sizeKey` is configured but the value is NaN/invalid, preventing NaN propagation through marker sizes, coordinate ranges, and hit-testing
- Guard `xCoordinateRange`/`yCoordinateRange` against NaN size values, falling back to default marker size
- Skip NaN entries during aggregation bucketing using the self-inequality idiom (`value !== value`) for hot-loop performance
- Add test covering bubble charts with NaN and undefined size data

Fix AG-16915

## Test plan

- [x] `yarn nx build:types ag-charts-community` — passes
- [x] `yarn nx test ag-charts-community --testPathPattern bubbleSeries` — 54/54 tests pass
- [x] New snapshot written for NaN size values test


manapeirov (approved):
LGTM

PR #6452, opened 2026-03-19
AG-16936 Document img-src CSP requirement for SVG icons

alantreadway:
## Summary

Fixes AG-16936 — CSP error with SVG icons on `default-src` directive.

The root cause: AG Charts uses `data:image/svg+xml` URIs in CSS `mask-image` for toolbar/context menu icons. When `default-src 'self'` is set without a separate `img-src` directive, the browser blocks these data URIs.

**Fix:** Documentation — explicitly require `img-src 'self' data:` in CSP policies. This matches AG Grid's approach (AG Grid has the same icon pattern and the same documented requirement).

An initial blob URL conversion approach was attempted and reverted — `blob:` URLs are also blocked by strict `default-src 'self'` policies, providing no benefit.

### Changes

- **Security docs** — `img-src data:` documented as required (not optional), with a callout explaining the `default-src` fallback pitfall and a locked-down CSP example
- **CSP test example** — new `strict-csp-icons` example with zoom toolbar and context menu under strict CSP (`img-src 'self' data:`)
- **Blob URL code removed** — reverted as ineffective

## Test Plan

- [x] Type checks pass (`ag-charts-community`, `ag-charts-enterprise`)
- [x] Example validation passes (`validate-examples`)
- [x] Manual: strict-csp-icons example renders zoom icons with no CSP console errors
- [x] Manual: inline SVG spike confirmed working under strict CSP (future option if needed)

Fix AG-16936

### DIFF
--- a/packages/ag-charts-website/src/content/docs/security-test/_examples/strict-csp-icons/head.html
+++ b/packages/ag-charts-website/src/content/docs/security-test/_examples/strict-csp-icons/head.html
@@ -1,0 +1,4 @@
+<meta
+    http-equiv="Content-Security-Policy"
+    content="default-src 'self' https://cdn.jsdelivr.net 'nonce-123123'; style-src 'self' 'nonce-a1b2c3d4' 'nonce-123123'; img-src 'self' data:"
+/>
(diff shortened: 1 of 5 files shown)


lsjroberts (approved):
LGTM

AG-DavidG (approved):
LGTM

PR #6459, opened 2026-03-19
AG-16915 Treat NaN as minimum size in aggregation bucketing

alantreadway:
## Summary

- Treat NaN-size data points as minimum size in the aggregation bucketing path, consistent with the non-aggregated rendering path (merged in #6450)
- Use the self-inequality idiom (`value !== value`) for the NaN check in this hot loop

Follow-up to #6450 which was merged before this consistency fix was added.

Fix AG-16915

## Test plan

- [x] `yarn nx build:types ag-charts-community` — passes
- [x] `yarn nx test ag-charts-community --testPathPattern bubbleSeries` — 54/54 tests pass
- [x] `yarn nx test ag-charts-community --testPathPattern tooltip` — 38/38 tests pass


manapeirov (approved):
LGTM

PR #6464, opened 2026-03-20
AG-16882 Fix ranges buttons border colors with high stroke widths

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-16882


alantreadway (approved):
LGTM

PR #6466, opened 2026-03-20
AG-16076: Eagerly validate dataIdKey duplicates on initial load

alantreadway:
## Summary

- Trigger duplicate-ID and missing-key warnings at `DataSet` construction time rather than deferring until the first transaction
- Adds a dedicated test for the eager validation on initial load

Addresses QA feedback: duplicate `dataIdKey` values now warn immediately when the chart is created, not only when the first transaction happens to need the ID index.

Fix AG-16076


iMoses (approved):
LGTM

PR #6467, opened 2026-03-20
AG-16947 Override :focus-visible to false on slider mousedown

olegat:
https://ag-grid.atlassian.net/browse/AG-16947


alantreadway (approved):
LGTM

PR #6468, opened 2026-03-20
AG-16889 Add initial range controls docs

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-16889


packages/ag-charts-enterprise/src/features/ranges/rangesTheme.ts
```diff
@@ -119,7 +119,7 @@ export const rangesTheme: WithThemeParams<AgRangesOptions> = {
             {
                 label: 'toolbarRangeAll',
                 ariaLabel: 'toolbarRangeAllAria',
-                value: undefined, // Reset zoom
+                value: [undefined, undefined], // Reset zoom
```
lsjroberts, line 122: Fixes a bug where the `All` button would only reset to the initial state, rather than change fully zoomed out.

PR #6482, opened 2026-03-23
CRT-1064 Check that `closestDatum():any` returns an object

olegat:
https://ag-grid.atlassian.net/browse/CRT-1064


alantreadway (approved):
LGTM

PR #6488, opened 2026-03-23
CRT-1063 Fix UnitTimeScale tick generation performance regression

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-1063

## Summary
- Fix performance regression in `UnitTimeScale.ticks()` introduced by `defaultTickCount = Infinity` change
- Use cached numeric timestamps (`numericBands`) for band comparisons instead of materialising all bands as Date objects
- Cache `calculateBands` results for the default domain case
- Add domain value equality check to preserve caches across `withTemporaryDomain` calls

## Benchmark result
High-perf area benchmark (1M data points): **366ms** avg initial load, down from **1463ms** (regressed) and faster than the **~500ms** baseline in 13.1.0.

### DIFF
--- a/packages/ag-charts-community/src/scale/unitTimeScale.ts
+++ b/packages/ag-charts-community/src/scale/unitTimeScale.ts
@@ -58,6 +58,16 @@ export class UnitTimeScale extends DiscreteTimeScale {
     override set domain(domain: Date[]) {
         if (domain === this._domain) return;
 
+        if (
+            domain.length === this._domain.length &&
+            domain.length >= 2 &&
+            domain[0].valueOf() === this._domain[0].valueOf() &&
+            domain[1].valueOf() === this._domain[1].valueOf()
+        ) {
+            this._domain = domain;
+            return;
+        }
+
         this._domain = domain;
         this._bands = undefined;
         this._numericBands = undefined;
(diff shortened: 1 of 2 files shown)


packages/ag-charts-community/src/scale/unitTimeScale.ts
```diff
@@ -407,8 +492,8 @@ export class UnitTimeScale extends DiscreteTimeScale {
             intervalEndIndex = intervalTicks.length - 1;
         } else {
             const i0 = bandsSliceIndices ? bandsSliceIndices[0] : 0;
-            const i1 = bandsSliceIndices ? bandsSliceIndices[1] : bands.length - 1;
-            intervalTicks = bands; // Could be large array - avoid copying
+            const i1 = bandsSliceIndices ? bandsSliceIndices[1] - 1 : bands.length - 1;
```
lsjroberts, line 495: Is this `- 1` a behaviour change or part of the performance fixes?

lsjroberts (approved):
LGTM, besides offline discussion.

packages/ag-charts-community/src/scale/unitTimeScale.ts
alantreadway: Re: the `- 1` on line 495 — this is a **latent bug fix**, not a performance change.

`visibleTickSliceIndices` returns an exclusive upper bound (designed for `.slice()`), but the original code passed it directly to `findMaxIndex` as an inclusive index — which would access `bands[bands.length]` (`undefined`). The fast path (`ticksFromNumericBands`) already had `- 1`; this aligns the fallback path to match.

In practice the affected code path was unreachable before this PR (the `isOwnDomain && interval != null` case now routes to the fast path), but it's correct to fix the fallback in case routing changes in the future.

PR #6490, opened 2026-03-23
CRT-1073 Fix interactive tooltip flickering

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-1073

## Summary

- Fix interactive tooltip flickering when moving the pointer around data points
- Root cause: `popInteractiveLeaveCallback` performed **immediate** tooltip removal, causing a synchronous HIDE followed by a deferred SHOW (~28ms gap = 2 frames of visible flicker)
- Secondary fix: `isManagedChildDOMElement` guard checked `canvas-overlay` but the tooltip moved to `tooltip-container` in v12.0.0, so the guard never matched

## Test plan

- [ ] Open [tooltip interaction example](https://ag-grid.com/charts/react/tooltips/#example-tooltip-interaction) and hover over bars — tooltip should follow smoothly without flickering
- [ ] Verify tooltip still dismisses when pointer leaves the chart entirely
- [ ] Verify non-interactive tooltips are unaffected
- [ ] CI passes

### DIFF
--- a/packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
+++ b/packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
@@ -1217,7 +1217,11 @@ export class SeriesAreaManager extends BaseManager {
         if (
             targetElement &&
             this.chart.tooltip.interactive &&
-            this.chart.ctx.domManager.isManagedChildDOMElement(targetElement, 'canvas-overlay', DEFAULT_TOOLTIP_CLASS)
+            this.chart.ctx.domManager.isManagedChildDOMElement(
+                targetElement,
+                'tooltip-container',
+                DEFAULT_TOOLTIP_CLASS
+            )
         ) {
             // Skip tooltip update if tooltip is interactive, and the source event was for a tooltip HTML element.
             return;
@@ -1260,7 +1264,7 @@ export class SeriesAreaManager extends BaseManager {
         return this.chart.tooltip.maybeEnterInteractiveTooltip(event, () => {
             this.pickManager.maybeActivate(undefined, (): void => {
                 this.tooltip.lastHover = undefined;
-                this.chart.ctx.tooltipManager.removeTooltip(this.id);
+                this.chart.ctx.tooltipManager.removeTooltip(this.id, undefined, true); // true = delayed
                 this.chart.ctx.highlightManager.updateHighlight(this.id, undefined, true); // true = delayed
             });
         });
--- a/packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
+++ b/packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
@@ -1217,11 +1217,7 @@ export class SeriesAreaManager extends BaseManager {
         if (
             targetElement &&
             this.chart.tooltip.interactive &&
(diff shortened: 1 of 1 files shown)


packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
```diff
@@ -1217,7 +1217,11 @@ export class SeriesAreaManager extends BaseManager {
         if (
             targetElement &&
             this.chart.tooltip.interactive &&
-            this.chart.ctx.domManager.isManagedChildDOMElement(targetElement, 'canvas-overlay', DEFAULT_TOOLTIP_CLASS)
```
olegat, line 1220: This does not look necessary. AFAICT `canvas-overlay` and `tooltip-container` are the same
https://github.com/ag-grid/ag-charts/blob/4d0d5c5d26e7d7ee15a967e0dbc0ef5c23915da4/packages/ag-charts-community/src/dom/domLayout.html#L9

packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
```diff
@@ -1260,7 +1264,7 @@ export class SeriesAreaManager extends BaseManager {
         return this.chart.tooltip.maybeEnterInteractiveTooltip(event, () => {
             this.pickManager.maybeActivate(undefined, (): void => {
                 this.tooltip.lastHover = undefined;
-                this.chart.ctx.tooltipManager.removeTooltip(this.id);
+                this.chart.ctx.tooltipManager.removeTooltip(this.id, undefined, true); // true = delayed
```
olegat, line 1263: LGTM for this release.

Although this feels like a hack. The `tooltip.maybeEnterInteractiveTooltip` should only run the callback when the cursor leaves the tooltip element AND leaves the bounds of the series-area.

The code in this callback is _mostly_ unnecessary, because `SeriesAreaManager` already knows when to update or not-update the highlight/tooltip when it receives `mousemove` / `mouseleave` for the series-area bounds. The edge case that's of concern here in this part of the code is when:
1. The cursor enters an interactive tooltip, then...
2. The cursor leaves the series-area bounds; `SeriesAreaManager` ignores the `mouseleave` event in this case because the tooltip/highlight must remain visible. Then...
3. The cursor leaves the interactive tooltip and **DOES NOT** re-enter the series-area bound; `SeriesAreaManager` receives no events and does nothing.
4. This callback is run to compensate the fact the `SeriesAreaManager` does not clear the highlight/tooltip.

olegat (approved):

packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
alantreadway: Fair - reverted

packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
alantreadway: Lets take this for now (I can demonstrate this fixes the flicker, and seems pretty narrowly scoped) and follow-up post release with a wider change?

olegat: SGTM.

PR #6492, opened 2026-03-23
CRT-1067 Fix annotation options toolbar menu pixel perfect positioning

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-1067


packages/ag-charts-community/src/components/toolbar/toolbar.ts
```diff
@@ -144,7 +144,18 @@ export abstract class BaseToolbar<
     protected getButtonWidgetBounds(buttonWidget: ButtonWidget) {
         const parent = this.getBounds();
         const bounds = buttonWidget.getBounds();
-        return new BBox(bounds.x + parent.x, bounds.y + parent.y, bounds.width, bounds.height);
+
+        // Adjust the relative position by the fractional part of the absolute position for a more precise placement.
+        const preciseBounds = buttonWidget.getBoundingClientRect();
```
alantreadway, line 149: We need to be careful with calling this, it can trigger forced reflows - how often is this being called?

lsjroberts: Will address caching post-release.

alantreadway (approved):

PR #6493, opened 2026-03-23
CRT-1084 Fix crossline offset with secondary/stacked axes

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-1084

Crosslines were positioned relative to the axis edge instead of the series area boundary when secondary or stacked axes were present. Propagates `gridPadding` to crosslines, mirroring the existing grid line padding pattern.


packages/ag-charts-community/src/chart/crossline/cartesianCrossLine.ts
alantreadway, line 1: This is the actual fix

manapeirov (approved):
LGTM

PR #6494, opened 2026-03-23
CRT-1083 - Fix stale animation on set legend state

manapeirov:
PR for https://ag-grid.atlassian.net/browse/CRT-1083

When the animation batch is skipped (e.g. due to layout changes), updateSelections() early-returns for hidden series without updating _contextNodeData. This leaves stale visible: true data from before the series was hidden. On the next update where the batch isn't skipped, the animation system compares the stale previousContextData (visible: true) against the new contextNodeData (visible: false) and plays a spurious hide animation from the wrong state.

Fix: When the early return skips createNodeData() for a hidden series, update _contextNodeData.visible = false so subsequent updates don't see stale visibility state.


alantreadway (approved):
LGTM

PR #6498, opened 2026-03-23
CRT-1074 Always check isFocusVisible() in updateComplete()

olegat:
https://ag-grid.atlassian.net/browse/CRT-1074

It is incorrect to assume that `getHoverDevice() == 'pointer'` implies `isFocusVisible() === false`.

The hoverDevice (pointer | keyboard | setState) is the device that is currently responsible for updating the highlight/tooltip (i.e., active). The focus indicator is not the same thing as the active state; they cannot be assumed to be the same. We MUST update the focus element SVG/HTML path/bounds regardless of what the current highlight/tooltip is.

However, we only need to compute the CSS style when receiving a 'focus' event, because that's the only time we cannot determine whether the event came from a keyboard user or a mouse user. In all other cases, we can listen to 'click' and 'keydown' events (i.e., overrideFocusVisible) to determine whether the focus indicator is visible.

### DIFF
--- a/packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
+++ b/packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
@@ -346,12 +346,7 @@ export class SeriesAreaManager extends BaseManager {
             } finally {
                 this.focus.pendingViewportFocus = undefined;
             }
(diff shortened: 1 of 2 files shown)


alantreadway (approved):
LGTM

PR #6503, opened 2026-03-24
AG-16970 Explicit setFocusOverride(true) on slider keydown event

olegat:
https://ag-grid.atlassian.net/browse/AG-16970

This fixes the focus indicator visibility for mouse->keyboard hotswapping in Firefox.


alantreadway (approved):
LGTM

PR #6504, opened 2026-03-24
AG-16865 Fix annotations toolbar to allow top level item buttons

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-16865


alantreadway (approved):
LGTM

PR #6505, opened 2026-03-24
Add missing regression tests for CRT-1083, CRT-1078, CRT-705

alantreadway:
## Summary

Adds regression tests for four b13.2.0 fixes that shipped without test coverage. Each test creates a real chart, triggers the exact failure scenario, and asserts internal state. All four were verified with the fix-revert protocol (fails without fix, passes with fix).

- **CRT-1083** (`lineSeries.test.ts`): Stale `_contextNodeData.visible` when series hidden during animation skip — uses line series (not bar) because bar has `animationAlwaysUpdateSelections: true` which bypasses the affected path
- **CRT-1078** (`mapMarkerSeries.test.ts`): Legend hover crash when `highlightedDatum.datum == null` — exercises `getHighlightedDatum()` with an undefined datum highlight
- **CRT-705** (`ranges.test.ts`): `isDropdown` state not reset when ranges module disabled — verifies state resets to undefined on disable

## Test plan

- [x] CRT-1083 test passes with fix, fails when fix reverted
- [x] CRT-1078 test passes with fix, fails when fix reverted
- [x] CRT-705 test passes with fix, fails when fix reverted
- [ ] CI green on all test suites

### DIFF
--- a/packages/ag-charts-enterprise/src/features/ranges/ranges.test.ts
+++ b/packages/ag-charts-enterprise/src/features/ranges/ranges.test.ts
@@ -2,7 +2,7 @@ import { afterEach, describe, expect, it } from '@jest/globals';
 
 import type { AgCartesianChartOptions, AgChartInstance } from 'ag-charts-community';
 import { AgCharts } from 'ag-charts-community';
-import { setupMockCanvas, setupMockConsole, waitForChartStability } from 'ag-charts-community-test';
+import { deproxy, setupMockCanvas, setupMockConsole, waitForChartStability } from 'ag-charts-community-test';
 import type { AgRangesButtonValueSource } from 'ag-charts-types';
 
 import { prepareEnterpriseTestOptions } from '../../test/utils';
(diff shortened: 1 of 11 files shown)


packages/ag-charts-enterprise/src/features/ranges/ranges.test.ts
```diff
@@ -19,6 +19,45 @@ describe('Ranges', () => {
         }
     });
 
+    // CRT-705: When the ranges module is disabled, the `isDropdown` state must be reset so
+    // that re-enabling the module starts from a clean state. Without this, the button toolbar
+    // may remain hidden if the module was in dropdown mode when disabled.
+    describe('enable/disable state reset (CRT-705)', () => {
+        it('should reset isDropdown state when ranges is disabled', async () => {
```
lsjroberts: `isDropdown` is an implementation detail, it's a bit odd to test that directly. It'd make more sense to do an e2e test that checks the buttons are hidden and re-shown when toggling `enabled`.

packages/ag-charts-enterprise/src/series/map-marker/mapMarkerSeries.test.ts
```diff
@@ -100,6 +100,33 @@ describe('MapMarkerSeries', () => {
...
+            const highlightManager = (chart as Chart).ctx.highlightManager;
+
+            // Simulate the legend hover scenario: highlight with datum == null.
+            // This occurs when hovering a legend item that has no direct datum association.
+            highlightManager.updateHighlight(chart.id, {
+                series: mapMarkerSeries,
+                datum: undefined,
+            } as any);
+
+            // getHighlightedDatum should return undefined for null-datum highlights
```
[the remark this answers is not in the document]
alantreadway: Already addressed — commit 6badee15 switched this test from direct highlightManager injection to `hoverAction` via `computeLegendBBox`, which exercises the real legend-hover code path and produces whatever datum value the production code naturally creates.

packages/ag-charts-enterprise/src/features/ranges/ranges.test.ts
alantreadway: Agreed — replaced with an e2e test that verifies the button toolbar disappears when `ranges.enabled` is set to `false` and reappears when re-enabled. Unit test removed.

lsjroberts (approved):
LGTM pending snapshot updates

packages/ag-charts-website/src/content/docs/range-buttons-test/_examples/e2e-enable-disable/main.ts
```diff
@@ -0,0 +1,35 @@
...
+            { label: 'Half', value: ['Cat 0', 'Cat 9'] },
+            { label: 'All', value: undefined },
+        ],
+    },
+};
+
+const chart = AgCharts.create(options);
+(window as any).chart = chart;
+(window as any).AgCharts = AgCharts;
+(window as any).chartOptions = options;
```
alantreadway: These are unecessary

alantreadway: Rework to add buttons to the example to trigger the necessary state changes - then just push the buttons from the e2e tests.

alantreadway: Removed — now using `AgCharts.getInstance()` + `updateDelta()` from the e2e test instead.

PR #6516, opened 2026-03-25
AG-NONE Split "styler highlights sequenced" into multiple snapshot comparisons

olegat:
This makes is much easier to debug failures with these tests, because Jest will report exactly which `hover()` interaction is causing snapshot mismatches.


alantreadway (approved):
LGTM

PR #6517, opened 2026-03-25
AG-16796 Fix range buttons styling when gapped

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-16976

The test for this would be part of the skipped flaky test.


alantreadway (approved):
LGTM

PR #6528, opened 2026-03-27
AG-3390 - Sync up nx prettier patch

taktran:
Instead of allowing for all extensions, only allow mdoc and astro


alantreadway (approved):
LGTM

PR #6538, opened 2026-03-27
AG-17008 Fix scrollbar not appearing when barWidth changes at runtime

alantreadway:
## Summary

- Fix `restoreRequiredRangeIterations` counter in `zoomManager` not resetting between independent user-initiated updates, causing the zoom update to be silently skipped when `barWidth` changes at runtime
- Pass the raw `_requiredRange` (pixel value) through the `pre-series-update` event so `restoreRequiredRange` can distinguish genuine option changes from layout oscillation (AG-16803)
- Add regression tests verifying scrollbar appears after `barWidth` increase and across successive changes

## Test plan

- [x] `yarn nx test ag-charts-community` — 3229 passed
- [x] `yarn nx test ag-charts-enterprise` — 1809 passed (includes 2 new regression tests)
- [x] `yarn nx lint ag-charts-community` / `ag-charts-enterprise` — clean
- [ ] Manual verification with TC1 plunker (set Width:10 then Width:40 at small chart size)
- [ ] Manual verification with TC2 plunker (scrollbar at small chart with axis labels)

Fix AG-17008


manapeirov (approved):
LGTM

PR #6554, opened 2026-03-31
Simplify UnitTimeScale tick generation and interval type checks

iMoses:
Normalise calendar interval types early via `toTimeInterval()`, replace `isPlainObject || typeof === 'string'` checks with `typeof !== 'number'`, and remove extend parameter from `UnitTimeScale` by splitting the hierarchical tick case.


alantreadway (approved):
LGTM

PR #6555, opened 2026-03-31
AG-16512 Add loading state option

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-16512

Targets https://github.com/ag-grid/ag-charts/pull/6550


packages/ag-charts-community/src/chart/update/overlaysProcessor.ts
lsjroberts, line 1: There were some conflicts between the loading and no-data overlays, where no-data hid the loading overlay when `this.hideOverlay(this.overlays.noData)` was called. So I've refactored it to use a state.

alantreadway (approved):
LGTM

PR #6556, opened 2026-03-31
Initial implementation of ChartState - Use ChartState with the legend module

iMoses:
https://ag-grid.atlassian.net/browse/AG-16871


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1632,20 +1636,9 @@ export abstract class Chart extends Observable implements ModuleInstance, ChartS
...
-            const legendListeners = deltaOptions.legend.listeners;
-            if (legendListeners) {
-                Object.assign((this as any).legend.listeners, legendListeners);
-            } else {
-                // Clear legend listeners when set to undefined
-                (this as any).legend.listeners.clear();
-            }
-        }
+
+        this.ctx.chartState.setValue('options', newChartOptions.processedOptions as ChartState['options']);
```
olegat, line 1636: It feels like `processedOptions` should really be of type `ResolvedChartOptions`. This would avoid the need for the `as` assertion which mutes current and future errors.

```diff
diff --git a/packages/ag-charts-community/src/chart/chartState.ts b/packages/ag-charts-community/src/chart/chartState.ts
index 8871ee821e..9f6d06e803 100644
--- a/packages/ag-charts-community/src/chart/chartState.ts
+++ b/packages/ag-charts-community/src/chart/chartState.ts
@@ -2,10 +2,10 @@ import type { AgChartOptions, NormalisedLegendOptions } from 'ag-charts-types';
 
 import type { CategoryLegendDatum } from './legend/legendDatum';
 
-export type ResolvedChartOptions = Omit<AgChartOptions, 'legend'> & { legend: NormalisedLegendOptions };
+export type ResolvedChartOptions<T extends AgChartOptions> = Omit<T, 'legend'> & { legend: NormalisedLegendOptions };
 
 export interface ChartState {
-    options: ResolvedChartOptions;
+    options: ResolvedChartOptions<AgChartOptions>;
     legendData: Record<string, CategoryLegendDatum[]>;
     legendVisible: boolean;
 }
diff --git a/packages/ag-charts-community/src/module/optionsModule.ts b/packages/ag-charts-community/src/module/optionsModule.ts
index ae6b0d013f..b35059e36a 100644
--- a/packages/ag-charts-community/src/module/optionsModule.ts
+++ b/packages/ag-charts-community/src/module/optionsModule.ts
@@ -56,6 +56,7 @@ import { getChartTheme } from '../chart/mapping/themes';
 import { detectChartType } from '../chart/mapping/types';
 import { ChartTheme } from '../chart/themes/chartTheme';
 import { type OptionsGraphAccessor, createOptionsGraph } from './optionsGraph';
+import type { ResolvedChartOptions } from '../chart/chartState';
 
 export interface ChartSpecialOverrides {
     document: Document;
@@ -128,7 +129,7 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
     }
 
     activeTheme: ChartTheme;
-    processedOptions: T;
+    processedOptions: ResolvedChartOptions<T>;
     userOptions: Partial<T>;
     processedOverrides: Partial<T>;
     specialOverrides: ChartSpecialOverrides;
```

Are they any other major obstacles to this approach?

packages/ag-charts-core/src/types/normalised-options/normalise.ts
olegat, line 1: I've not convinced that `ag-charts-types/src/internal/` is the best location for these.

Are these internal-types going to be referenced by any other types within the `ag-charts-types` package? If not, keeping these types as private within the `ag-charts-community` package might be a better option.

olegat (commented):
LGTM.

Some optional improvements could be made, but overall this looks very promising. Well done!

packages/ag-charts-community/src/chart/chart.ts
iMoses: Good point. I'll test it and update if there are any issues to address to make it so.

packages/ag-charts-core/src/types/normalised-options/normalise.ts
iMoses: I agree that it's unclear if this is the right location. I think it's worth talking about at our next tech debt sync, with Alan involved.

I'll add that I think that if we do decide not to have it as part of `ag-charts-types` then I personally think it should be in the `ag-charts-core` package instead.

packages/ag-charts-community/src/chart/chart.ts
alantreadway: Discussed offline - we'll try and tend towards this, but it may not be practical to apply this immediately

packages/ag-charts-core/src/types/normalised-options/normalise.ts
alantreadway: Discussed offline: Agreed that `ag-charts-core` is probably the better place for this, as for several release process checks we assume any changes in `ag-charts-types` are user-facing contract changes that we need to scrutinise.

alantreadway (approved):
LGTM

PR #6563, opened 2026-04-01
AG-16826 Test that only the first focus event reads initialFocus

olegat:
https://ag-grid.atlassian.net/browse/AG-16826


alantreadway (approved):
LGTM

PR #6566, opened 2026-04-01
Fix legend hover clearing active state after setState restore

alantreadway:
## Summary

- Remove `setHoverDevice('pointer')` from `seriesAreaManager.onLegendHover()` to fix 4 failing E2E tests (`frozen-legendtoggle` thawed variants)
- The `onLegendHover` handler (added in #6174) was changing `hoverDevice` from `'setState'` to `'pointer'`, which altered downstream active state management during chart updates, causing a spurious `activeChange` event that cleared the active item after a legend visibility toggle

## Test plan

- [x] `yarn nx test ag-charts-community` — 3241 passed
- [x] `yarn nx test ag-charts-enterprise` — 1843 passed
- [ ] CI E2E `frozen-legendtoggle` tests should pass (shard 16/16)

### DIFF
--- a/packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
+++ b/packages/ag-charts-community/src/chart/series/seriesAreaManager.ts
@@ -1426,7 +1426,6 @@ export class SeriesAreaManager extends BaseManager {
 
     private onLegendHover(_event: null): void {
         if (!this.isState(InteractionState.Hoverable)) return;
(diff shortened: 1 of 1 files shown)


alantreadway:
Closing — removing setHoverDevice('pointer') fixes the frozen-legendtoggle tests but breaks the AG-16744 legend hover tests from the same PR (#6174). Will revert #6174 instead.

PR #6588, opened 2026-04-13
Suggested bundling changed to fix tree-shaking

iMoses:


iMoses:
Decided to go with a solution that wouldn't change the dist structure

PR #6598, opened 2026-04-15
Zoom module to use new ChartState

iMoses:


alantreadway (approved):
LGTM, @lsjroberts and @olegat should take a look too

packages/ag-charts-enterprise/src/features/zoom/zoom.ts
```diff
@@ -814,35 +789,31 @@ export class Zoom extends AbstractModuleInstance {
...
-        if (this.axes === 'xy') return true;
-        return this.shouldFlipXY ? this.axes === 'y' : this.axes === 'x';
-    }
-
-    private isScalingY() {
-        if (this.axes === 'xy') return true;
-        return this.shouldFlipXY ? this.axes === 'x' : this.axes === 'y';
+    private isScalingX(axes: ZoomOpts['axes']) {
+        if (axes === 'xy') return true;
+        return this.shouldFlipXY ? axes === 'y' : axes === 'x';
```
lsjroberts, line 798: Why not just `this.opts.axes` here instead of passing it in?

packages/ag-charts-enterprise/src/features/zoom/zoom.ts
```diff
@@ -814,35 +789,31 @@ export class Zoom extends AbstractModuleInstance {
...
     }
 
-    private getAnchorPointX() {
```
lsjroberts, line 827: Similarly, why not keep this function?

lsjroberts (approved):
LGTM, great stuff! Some very minor comments.

packages/ag-charts-community/src/chart/chartContext.ts
```diff
@@ -158,7 +158,7 @@ export class ChartContext implements ModuleContext {
         this.animationManager = new AnimationManager(this.agDocument, this.interactionManager, updateMutex);
         this.dataService = new DataService<any>(this.eventsHub, chart, this.animationManager);
         this.tooltipManager = new TooltipManager(this.eventsHub, this.localeManager, this.domManager, chart.tooltip);
-        this.zoomManager = new ZoomManager(this.eventsHub, fireEvent);
+        this.zoomManager = new ZoomManager(this, fireEvent);
```
olegat, line 161: This introduces a construction-order hazard. `this` is passed to `ZoomManager` before `ChartContext` is fully initialised, so it does not yet satisfy the `ModuleContext` contract.

If `ZoomManager` accesses properties initialised later (e.g. `collapsedManager`), this will fail at runtime.

Prefer explicit dependencies, which provide clear ordering guarantees and stronger type safety:
```
this.zoomManager = new ZoomManager(this.chartState, this.eventsHub, fireEvent);
```

packages/ag-charts-community/.dependency-cruiser.js
```diff
@@ -11,6 +11,7 @@ module.exports = {
             from: {},
             to: {
                 circular: true,
+                dependencyTypesNot: ['type-only'],
```
olegat, line 14: Is this because `zoomManager.ts` imports `moduleContext.ts` and `moduleContext.ts` imports `zoomManager.ts`? This might no longer be necessary if the construction-order hazard is addressed.

packages/ag-charts-enterprise/src/features/zoom/zoom.ts
iMoses: Alan expressed some concern about performance if each options fetching required triggering the `getValue` method on the chart-state, so the skill I created is trained to try and avoid fetching the options multiple time and instead pass them down to internal methods. That's the gist of it.
It could be that in certain places it'll make less sense, I guess we'll have to judge these case by case.

packages/ag-charts-community/src/chart/interaction/zoomManager.ts
```diff
@@ -461,7 +456,8 @@ export class ZoomManager extends BaseManager implements MementoOriginator<ZoomMe
 
     // Fire this event to signal to listeners that the view is changing through a zoom and/or pan change.
     public fireZoomPanStartEvent(callerId: 'navigator' | 'zoom') {
-        this.eventsHub.emit('zoom:pan-start', { callerId });
+        this.ctx.chartState.setValue('zoomPanning', true);
```
olegat: I'm not seeing why `zoomPanning` is needed. It appears to be set but never read. Presumably you intend to use this sometime in a follow-up PR?

packages/ag-charts-enterprise/src/features/zoom/zoom.ts
iMoses: IIUC it's a similar case, because `anchorPointY` and `anchorPointX` are part of the options.
If you are not satisfied by this explanation then let's talk offline to try and resolve this.

packages/ag-charts-community/src/chart/chartContext.ts
iMoses: This was make as a pre-step before improving the context itself to support lazy-loading and resolving of registers services. True that currently it is problematic but once the change as been fully applied it'll make sense, normalising the constructors signature and allowing us to automate the creation process.

packages/ag-charts-community/.dependency-cruiser.js
iMoses: As said previously, it's a pre-step towards to context changes. If you want to push back on it, let's discuss it offline.

packages/ag-charts-community/src/chart/interaction/zoomManager.ts
iMoses: You are correct. It might be a left over from previous iteration. I'll investigate and remove if deemed unnecessary.

packages/ag-charts-enterprise/src/features/zoom/zoom.ts
lsjroberts: Ah right, I hadn't realised it would retrigger that.

packages/ag-charts-enterprise/src/features/zoom/zoom.ts
lsjroberts: All good.

packages/ag-charts-community/src/chart/update/dataWindowProcessor.ts
```diff
@@ -69,7 +78,7 @@ export class DataWindowProcessor<D extends object> implements UpdateProcessor {
         let shouldRefresh = true;
```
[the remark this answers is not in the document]
iMoses: False positive — `dataWindowProcessor` only reads the primary X axis via `chart.axes.find(({ direction }) => direction === ChartAxisDirection.X)` (line 75), and `toZoomState` populates `chartState.zoom.x` from the first X entry in insertion order, i.e. the same primary axis. So `pickDirectionZoom(chartState.zoom, 'x')` resolves to exactly what `zoomManager.getAxisZoom(primaryXId)` did before. The call site was never per-axis aware — it only runs for lazy data services, and always for the primary X.

packages/ag-charts-core/src/types/normalised-options/normalisedZoomOptions.ts
```diff
@@ -0,0 +1,38 @@
...
+    | 'deceleration'
+    | 'minVisibleItems'
+    | 'panKey'
+    | 'scrollingStep'
+    | 'anchorPointX'
+    | 'anchorPointY'
+    | 'axes'
+    | 'autoScaling'
+    | 'onDataChange'
+    | 'buttons',
```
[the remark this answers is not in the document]
iMoses: False positive — the old `@Property axisDraggingMode = 'zoom'` initialiser wasn't the effective default. The theme template in `zoomModule.ts` never specified `axisDraggingMode` / `scrollingMode` / `keepAspectRatio` either, so in both old and new code these come from user options or stay undefined. Every read site handles undefined correctly and matches the prior defaults:

- `axisDraggingMode`: `zoom.ts:421`, `:463`, `:533` all branch on `=== 'pan'` with the zoom path in the `else`, matching the old `'zoom'` default.
- `scrollingMode`: `zoom.ts:584` (`=== 'pan' || isHorizontalScrolling`) and `:602` (`= 'zoom'` destructure default).
- `keepAspectRatio`: `zoom.ts:1001` (`?? false`).

`NormalisedZoomOptions` could still be tightened to list these three as required and the theme template updated to match — a fair robustness improvement — but there's no runtime regression.

PR #6612, opened 2026-04-16
AG-17082(6) Cancel drag-selection when Escape is pressed

olegat:


alantreadway (approved):
LGTM

PR #6616, opened 2026-04-17
AG-16048 Gate scatter/bubble colour scale as enterprise

alantreadway:
https://ag-grid.atlassian.net/browse/AG-16048

Follow-up PR addressing David's review feedback on the scatter/bubble colour scale work shipped in #6558.

## Changes

- **`enterprise()` validator wrapper** — new wrapper in `ag-charts-core/src/state/validation.ts`, mirroring the existing `undocumented()` / `required()` pattern. Applied to `colorKey` / `colorName` / `colorScale` on scatter and bubble option defs: user-supplied values are stripped and a console warning is emitted when `ag-charts-enterprise` is not registered, so the series instance never sees them and the feature stays dormant for community consumers.
- **`$enterprise` options-graph DSL operator** — new primitive in `optionsGraphOperations.ts` (mirrors `$if`). Theme templates can declare `{ $enterprise: [enterpriseBranch, communityBranch] }` and the options graph resolves the correct branch at chart-creation time via `ModuleRegistry.isEnterprise()`. This replaces an earlier attempt that did post-theme pruning inside `sanitizeThemeModules`.
- **Default palette via `$enterprise` at the `colorScale` root** — bubble/scatter theme templates now set `colorScale: { $enterprise: [{ fills: [orange, yellow, green] }, undefined] }`. Community resolved options contain no `colorScale` sub-tree at all; enterprise gets the default palette. Shared fragments extracted into `BUBBLE_SCATTER_COLOR_SCALE_THEME` / `BUBBLE_SCATTER_GRADIENT_LEGEND_THEME` constants to avoid duplication.
- **`colorScale` moved to themeable contract** — `colorScale?: AgColorScale` relocated from `AgBubbleSeriesOptions` / `AgScatterSeriesOptions` into the respective themeable interfaces so theme templates can assign it without `@ts-expect-error`.
- **Tooltip symbol fill** — `legendItemSymbol` on `BubbleSeries` now accepts an optional `styleOverride`; `getTooltipContent` passes the per-datum `activeStyle` so the tooltip's marker reflects the resolved colour, matching heatmap.
- **Docs** — short "Colour Scale" sections on the scatter and bubble docs pages flag the enterprise requirement and link to the Heatmap page for the full `colorKey` / `colorName` / `colorScale` reference.
- **Minor refactor** — extracted registry-mode tracking into `packages/ag-charts-core/src/modules/registryMode.ts` to avoid a circular import introduced by consulting `isEnterprise()` from `state/validation.ts`.
- **Type system** — broadened `ExtendLiteralLeavesInner` in `operationOptions.ts` so operators (`$enterprise`, etc.) can replace nested object types, not just leaves and arrays. Matches the runtime capability.

## Manual verification Plunkers

Each Plunker isolates one fix with an identical 12-point dataset so only the gated behaviour varies. All three load the bundle from `https://localhost:4600/charts/dev/...` — **run `yarn nx dev` locally and accept the self-signed cert** (visit the UMD URL directly) before opening:

- **Community enterprise gating** — https://plnkr.co/edit/S8Ua0PBE5R79Dxo1?preview — community bundle, bubble with `colorKey` set. Expect series-blue markers and a console warning: *Option `series[0].colorKey` is an AG Charts Enterprise feature; ignoring.*
- **Default divergingColors palette** — https://plnkr.co/edit/ZF3lpCZDDU4IC5NG?preview — enterprise bundle, bubble with `colorKey` but no `colorScale.fills`. Expect orange → yellow → green markers and an auto-enabled gradient legend.
- **Tooltip swatch** — https://plnkr.co/edit/JnXCgz4WuaxfjQp8?preview — enterprise bundle, bubble with `colorKey` and explicit `colorScale.fills` (blue→red). Hover any point: tooltip swatch should match the on-canvas marker colour, not series blue.

## Verification

- `ag-charts-core` validation tests: 57 pass (incl. 4 new `enterprise` wrapper cases)
- `ag-charts-community`: 3281 pass, 1606 snapshots pass
- `ag-charts-enterprise`: 1859 pass, 1640 snapshots pass (no image-snapshot drift)
- `yarn nx format`, `yarn nx lint ag-charts-community ag-charts-core` (incl. depcruise), `yarn nx build:types` all green
- Full CI green (36 passing / 1 expected-skip Framework Package Tests)

Fix #AG-16048


packages/ag-charts-community/src/chart/factory/processModuleOptions.ts
alantreadway, line 1: Can we encode this in the theme via a `$enterprise` operator which allow specify community + enterprise cases distinctly?

iMoses (approved):
`enterprise` validator look good, though worth noting that if ever used with `required` it'll break. The two aren't compatible, and `enterprise` MUST remain optional.

packages/ag-charts-community/src/module/optionsGraphOperations.ts
```diff
@@ -127,17 +127,34 @@ function cacheMaxOperation(graph: OptionsGraphInterface, vertex: VertexInterface
...
+    // Attach neighbours from the chosen branch onto the vertex (mirrors $if for path propagation).
+    const neighbours = graph.neighboursWithEdgeValue(valueVertex, PATH_EDGE);
+    if (neighbours) {
+        for (const neighbour of neighbours) {
+            graph.addEdge(vertex, neighbour, PATH_EDGE);
+        }
+    }
+
+    return graph.resolveVertexValue(vertex, valueVertex);
+}
```
lsjroberts: This is fine, I'd just suggest we change it to `$isEnterprise` and have it return a boolean, to match the other chart operations. This would make it more flexible and save duplicating the `$if` behaviour.

lsjroberts: You'd then use it like `{ $if: [{ $isEnterprise: true }, then, else] }`

lsjroberts: Or you could do `{ $isPackageType: 'enterprise' }` / `{ $isPackageType: 'community' }`

alantreadway: Swapped to `$isPackageType` per your second alternative — keeps the predicate consistent with the other parameterised `$is*` operators (`$isChartType`, `$isSeriesType`) and composes through `$if` rather than reimplementing branch resolution. Theme templates now use `{ $if: [{ $isPackageType: 'enterprise' }, enterpriseBranch, communityBranch] }`. Pushed as ee009e01d4.

alantreadway:
Both review items addressed in ee009e01d4:

- **@lsjroberts** — `$enterprise` operator removed; replaced with boolean `$isPackageType` predicate composed through existing `$if`. Theme templates read as `{ $if: [{ $isPackageType: 'enterprise' }, enterpriseBranch, communityBranch] }`.
- **@iMoses** — `enterprise()` and `required()` now throw at definition time in either composition order, with an explicit JSDoc note on the incompatibility. Added a test covering both directions.

Verification: `build:types`, `test ag-charts-core` (+1 new case), `test ag-charts-community` (3281 pass / 1606 snapshots), `test ag-charts-enterprise` (1859 pass / 1640 snapshots — no drift), `lint` all green.

lsjroberts (approved):
LGTM

PR #6619, opened 2026-04-17
AG-17064 Address QA feedback on tooltip offset

alantreadway:
https://ag-grid.atlassian.net/browse/AG-17064

Follow-up on [#6593](https://github.com/ag-grid/ag-charts/pull/6593) addressing QA feedback from [comment 111104](https://ag-grid.atlassian.net/browse/AG-17064?focusedCommentId=111104):

- **`anchorTo: 'chart'` bug fix.** The directional `offset` is now applied inward from the anchored edge (e.g. `placement: 'top'` with default `offset: 8` now places the tooltip 8px below the chart's top edge). Previously the offset was silently ignored for chart-anchored tooltips, leaving them flush against the edge.
- **API reference rendering.** `AgTooltipPositionOptions.offset` JSDoc reformatted so the `Default: ` + `8` marker sits on its own paragraph and renders as a labelled default on the docs site.
- **Rule added.** `.rulesync/rules/defaults.md` now requires the `Default:` marker to be its own JSDoc paragraph, so we avoid the same miss in future API additions.
- **Refactor + tests.** Positioning math extracted from the private `Tooltip.getTooltipBounds` into a pure `tooltipBounds.ts` helper with 13 unit tests covering all 8 chart-anchor placements, default offset, corner composition, `xOffset`/`yOffset` interaction, and a pointer-anchor regression guard.

Verified against the repro plunk (https://plnkr.co/edit/2K6BPZ0AYIyu2OXM); full community + enterprise test suites pass.

Fix #AG-17064


lsjroberts (approved):
LGTM

PR #6392, opened 2026-03-16
Ag 9758/color scale

manapeirov:
PR for https://ag-grid.atlassian.net/browse/AG-9758

PR #6625, opened 2026-04-20
AG-9758 Decouple gradient legend display domain from stops

alantreadway:
https://ag-grid.atlassian.net/browse/AG-9758

Addresses the four issues raised by QA on 2026-04-20 against the colour
scale work completed in pt1 (#6392), all rooted in the same cause: the
colour scale conflated its interpolation-pivot domain (derived from fill
stops) with the user-visible axis domain shown on the gradient legend.

Introduces a separate `displayDomain` on `ColorScale` / `ColorScaleState`,
populated in `configureColorScale` from `props.domain ?? [dataMin, dataMax]`.
The gradient legend datum now uses it for the axis range and clips
continuous gradient stops to `[0, 1]`, sampling boundary colours via
`colorScale.convert()`. Named labels outside the visible range drop out,
and the hover arrow is hidden when the highlighted value is out of range.

Completes the two outstanding TODO ACs on the ticket (`stops do not
affect the colour domain` and explicit `domain` override).

Fix #AG-9758


iMoses (approved):
LGTM

PR #6634, opened 2026-04-21
AG-17082(3) Use the same `fill`/`fillOpacity` as zoom-selection

olegat:


alantreadway (approved):
These should probably be part of the theme rather than hardcoded, but this is step forward so LGTM

PR #6637, opened 2026-04-22
AG-17082(5) Bugfix: Call `data.selections.clear()` only once per data-set

olegat:


alantreadway (approved):
LGTM

PR #6640, opened 2026-04-22
AG-17085 Sync ag-shared and follow canary track for ag-dev-prompts

alantreadway:
https://ag-grid.atlassian.net/browse/AG-17085

Follow-up to the AG-17085 plugin rework. Two things:

1. **Subrepo-pull ag-shared** to pick up the post-rollout state: `.claude-settings.json` → `.claude-settings.template.json` (per-repo render introduced in the ag-studio migration), consumer-side `plugin-assignments.json` dropped (`stage.py` now reads from the fetched cache), and the retired `scripts/sync-to-plugin/` helpers removed.
2. **Opt ag-charts into the canary `ag-dev-prompts` track.** Consumer-side of the canary/stable release model documented in [`ag-dev-prompts/docs/release.md`](https://github.com/ag-grid/ag-dev-prompts/blob/latest/docs/release.md):
   - Extends the shared `.claude-settings.template.json` with a `${AG_DEV_PROMPTS_REF_SUFFIX}` placeholder.
   - `setup-prompts.sh` derives the suffix from the `AG_DEV_PROMPTS_REF` env var: unset or `latest` → empty suffix (stable track); any other value → `#<ref>` pinning the marketplace to that branch/tag.
   - ag-charts' `postinstall:setup-prompts` script sets `AG_DEV_PROMPTS_REF=canary`, so both the rendered `.claude/settings.json` (marketplace source: `ag-grid/ag-dev-prompts#canary`) and `rulesync-fetch/fetch.sh` now track canary. Any commit merged to `canary` reaches ag-charts immediately.

ag-grid and ag-studio leave the env unset → stay on `latest`. They pick up the shared template change on their next `subrepo pull ag-shared` — parallel PRs [ag-grid#13584](https://github.com/ag-grid/ag-grid/pull/13584) and [ag-studio#1318](https://github.com/ag-grid/ag-studio/pull/1318) sync the same ag-shared state.

## Verification

- `AG_DEV_PROMPTS_REF=canary ./external/ag-shared/scripts/setup-prompts/setup-prompts.sh` runs clean locally.
- Rendered `.claude/settings.json` contains `"repo": "ag-grid/ag-dev-prompts#canary"`.
- `~/.cache/ag-dev-prompts/repo` checks out the `canary` branch on first fetch.
- All canary skills (`example`, `dev-server`, `debug-trace`, `git-conventions`, `jira`) still resolve via the `/ag-prodeng:*` namespace.

## Test plan

[...]

PR #6642, opened 2026-04-22
AG-17085 Fix setup-prompts.sh aborting yarn install on CI agents

alantreadway:
https://ag-grid.atlassian.net/browse/AG-17085

Follow-up fix for the canary opt-in landed in [#6640](https://github.com/ag-grid/ag-charts/pull/6640). TeamCity's macOS agent was exiting `yarn install` with `postinstall:setup-prompts exited with 127` because two external-tool invocations in `setup-prompts.sh` were unguarded:

1. **Product detection at the `.claude/settings.json` render step** called `jq` unconditionally. When `jq` isn't installed, the product came back empty and the hard-fail `return 1` propagated via `set -e` — killing the whole script before rulesync or any other postinstall step could run.
2. **The plugin-delivered cleanup block** invoked `python3 "$cleanup_script"` without `|| true`, so a cleanup failure took down the whole postinstall.

Both paths now degrade to warnings and `return 0`. The existing `.claude/settings.json` stays in place if the render can't be completed, and the rest of postinstall finishes. Symptom was only "✓ Detected: Claude Code" then `ERROR: "postinstall:setup-prompts" exited with 127.` — matches the product-detect path (no `Warning:` lines visible because the error happened before the `Cannot render` echo on `set -e` kill paths vary across bash versions).

## What changes

- `command -v jq` guard before the `jq` invocation — missing jq no longer empties product via silent failure, it just skips the render.
- Both "no product detected" and "unknown product" branches now emit `!` (yellow warning) to stderr and `return 0` instead of `return 1`. The hard-fail was overly strict for a per-repo config render.
- `command -v python3` guard added on the cleanup block, plus `|| true` on the final `python3 "$cleanup_script"` so a cleanup hiccup doesn't abort postinstall.

## Verification

Tested locally with `env -i PATH=/bin:/usr/bin ... ./setup-prompts.sh --postinstall` (no jq, no python3, no node in PATH) — script now exits 0 with appropriate warnings instead of aborting.

## Subrepo note

The change touches `external/ag-shared/scripts/setup-prompts/setup-prompts.sh`, which is a subrepo. **Before merging this PR**, run `git subrepo push external/ag-shared` to publish the fix to the ag-shared mirror so ag-grid and ag-studio pick it up on their next subrepo pull. (I didn't do the push yet — wanted to keep it behind review.)

## Test plan

- [ ] CI green on this branch
- [ ] TeamCity `yarn install` completes on macOS agent
- [ ] Local `yarn` still renders `.claude/settings.json` correctly with canary pin (no regression on the happy path)
- [ ] `git subrepo push external/ag-shared` after merge

### DIFF
--- a/external/ag-shared/scripts/setup-prompts/setup-prompts.sh
+++ b/external/ag-shared/scripts/setup-prompts/setup-prompts.sh
@@ -255,24 +255,27 @@ copy_extra_configs() {
     local claude_settings_dest="$REPO_ROOT/.claude/settings.json"
 
     if [[ -f "$REPO_ROOT/$claude_settings_template" ]] && [[ "$targets" == *"claudecode"* || "$targets" == "*" ]]; then
+        # Detect product via AG_PRODUCT env var or by reading package.json .name.
+        # Fall back gracefully when neither source yields a match: skipping the
+        # render leaves any existing .claude/settings.json in place and lets
+        # the rest of postinstall complete. Failing here would otherwise abort
+        # `yarn install` on environments missing jq or on unknown checkouts.
         local product="${AG_PRODUCT:-}"
-        if [[ -z "$product" && -f "$REPO_ROOT/package.json" ]]; then
+        if [[ -z "$product" && -f "$REPO_ROOT/package.json" ]] && command -v jq >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


taktran (approved):
LGTM

PR #6648, opened 2026-04-22
Fix ag-dev-prompts marketplace canary ref rendering

alantreadway:
## Summary

Claude Code's plugin marketplace schema for `github` sources does not support a `repo#ref` suffix — `ref` is a sibling field on the source object ([docs](https://code.claude.com/docs/en/plugin-marketplaces)). The existing template rendered `ag-grid/ag-dev-prompts#canary` as the repo name, producing the malformed clone URL `https://github.com/ag-grid/ag-dev-prompts#canary.git/` and a "repository not found" error on install.

The fix moves the placeholder out of the repo string and emits a sibling `ref` field when `AG_DEV_PROMPTS_REF` is set:

```json
"source": {
  "source": "github",
  "repo": "ag-grid/ag-dev-prompts",
  "ref": "canary"
}
```

When `AG_DEV_PROMPTS_REF` is unset or `latest`, the source is emitted without a `ref` field (Claude picks the default branch), preserving the existing behaviour for ag-grid and ag-studio.

Note: the template file (`.claude-settings.template.json`) is no longer strictly valid JSON because the `${AG_DEV_PROMPTS_REF_FIELD}` placeholder now sits outside a string literal. The file name already conveys it is a template, and the rendered output remains valid JSON in both branches.

## Test plan

- [x] Render with `AG_DEV_PROMPTS_REF=canary`: `extraKnownMarketplaces.ag-dev.source.ref == "canary"`.
- [x] Render with `AG_DEV_PROMPTS_REF` unset: no `ref` field.
- [x] Claude Code successfully clones the canary branch on session start with the new settings.

### DIFF
--- a/external/ag-shared/.claude-settings.template.json
+++ b/external/ag-shared/.claude-settings.template.json
@@ -150,7 +150,7 @@
     "ag-dev": {
       "source": {
         "source": "github",
-        "repo": "ag-grid/ag-dev-prompts${AG_DEV_PROMPTS_REF_SUFFIX}"
+        "repo": "ag-grid/ag-dev-prompts"${AG_DEV_PROMPTS_REF_FIELD}
       }
     },
     "openai-codex": {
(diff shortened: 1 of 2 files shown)


AG-DavidG (approved):
LGTM

PR #6651, opened 2026-04-23
AG-16048 Bubble/scatter colour scale QA fixes (pt4)

alantreadway:
## Summary

Addresses two items from David's QA feedback on [AG-16048 comment 111360](https://ag-grid.atlassian.net/browse/AG-16048?focusedCommentId=111360):

### 1. `TypeError` when enterprise-only `colorScale` supplied in community

> Correctly warns. However chart does not load and there is an exception thrown `TypeError: currentValue.clear is not a function at ColorScaleProperties.clear`

Root cause: `BaseProperties.clear()` recurses into every decorated field where `isProperties(value)` is `true` — which includes `PropertiesArray`, but the array class had no `clear()` method. The community bundle's `enterprise()` validator strips a user-supplied `colorScale` to `null`, which drives `ColorScaleProperties.set(null) → clear()` → crash on the `fills` `PropertiesArray`.

Fix: add `clear()` to `PropertiesArray` matching the existing `set()`/`reset()`/`toJson()` API symmetry. Regression covered in `properties.test.ts`.

### 2. Marker colours fall back to red→blue instead of the heatmap palette

> If no `fills` is provided, use the default palette used by heatmap. Actual: markers are blue. Expected: markers range from orange to green.

Root cause: the options graph treats any user-supplied partial object as "user-supplied" for every descendant path (via `CHILDREN_SOURCE_EDGE` propagation in `OptionsGraph.hasUserOption`), which wipes the theme's `colorScale.fills` default wholesale. `$isUserOption` inherits the same propagation so it can't discriminate an explicitly-set leaf from an inherited one, and moving the default down to the `fills` leaf triggers element-wise array merging against user-supplied `fills` entries.

Fix: add an undocumented `colorRange` option (palette-bound via the theme template's `{ $palette: 'divergingColors' }`) that lives outside the `colorScale` sub-tree so a user-supplied partial can't wipe it. `BubbleSeries.processData` threads it through as the `fallbackRange` arg to `configureColorScale`, so an empty `colorScale.fills` resolves to the palette-appropriate gradient and tracks light/dark/financial palette switches. Mirrors heatmap's existing `colorRange` mechanism.

## Test plan

- [x] `yarn nx test ag-charts-community` — 3318 passed (new `properties.test.ts` regression)
- [x] `yarn nx test ag-charts-enterprise` — 1880 passed (new `gradientLegend.test.ts` regression for point 2)
- [x] `yarn nx format` / `build:types` / `lint` clean for core + community + enterprise
- [ ] Re-verify David's reproducer plunkers (6RcHTXkxqfOngpJa community-colorScale, yNVXgZ43XcrEEd1d empty colorScale heatmap fallback)

Fix #AG-16048

### DIFF
--- a/packages/ag-charts-community/src/chart/series/cartesian/bubbleSeries.ts
+++ b/packages/ag-charts-community/src/chart/series/cartesian/bubbleSeries.ts
@@ -318,7 +318,10 @@ export class BubbleSeries extends CartesianSeries<BubbleSeriesTypes> {
             const domain = extent(rawDomain);
 
             if (domain != null) {
-                configureColorScale(this.colorScale, this.properties.colorScale, domain, []);
+                // `colorRange` is the palette-bound fallback used when `colorScale.fills` is
+                // empty — see BubbleSeriesProperties.colorRange for why this can't live inside
+                // the `colorScale` sub-tree.
+                configureColorScale(this.colorScale, this.properties.colorScale, domain, this.properties.colorRange);
                 this.colorScaleValid = true;
             }
         }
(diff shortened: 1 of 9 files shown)


iMoses (approved):
LGTM

PR #6657, opened 2026-04-23
Retry Playwright image pull to reduce E2E shard flakes

alantreadway:
Split the `mcr.microsoft.com/playwright` pull out of `docker run` in `packages/ag-charts-website/playwright.sh` and wrap it in a 3-attempt loop with linear backoff (10s, 20s).

MCR occasionally returns a transient block (seen as `The request is blocked` / HTTP error in the docker daemon response) which causes `docker run` to exit 125 before Playwright even starts, sinking the whole shard. With 16 shards per run, every CI invocation rolls the dice 16× against MCR, so flakes have been creeping up.

On the happy path the behaviour is unchanged — `docker run` hits the local image cache after the pull. On a transient MCR hiccup, we now retry twice before failing the shard with a clear "Failed to pull ... after N attempts" message rather than an opaque docker exit-125.

This is a stopgap; the longer-term fix is to mirror the image to GHCR so CI doesn't depend on MCR at all.

### DIFF
--- a/packages/ag-charts-website/playwright.sh
+++ b/packages/ag-charts-website/playwright.sh
@@ -100,19 +100,21 @@ if [ "$1" == "--host" ] ; then
   cd $(git rev-parse --show-toplevel)
 
   playwright_image=mcr.microsoft.com/playwright:v1.57.0-jammy
-  pull_attempts=3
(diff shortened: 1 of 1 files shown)


taktran (approved):
LGTM

PR #6658, opened 2026-04-23
Introduce DynamicContext for module-scoped services

iMoses:
- Introduces `DynamicContext<TRegistry>`, a lazy service registry that replaces the static `ChartContext` class. Services are defined via `constant`/`ref`/`service`/`factory` and resolved on first access, with an idempotent destroy cascade (respecting `ref`-registered externals) and cycle detection.
- Renames `ModuleContext` → `ChartRegistry` as the typed surface, with `ChartAxisRegistry` and `ChartSeriesRegistry` for axis/series child registries.
- Moves owner-bound services out of `createChartContext` and into their plugin modules' `register(ctx)` hooks — so `legendManager`, `annotationManager`, `zoomManager`, and `contextMenuRegistry` only instantiate when their module is actually loaded. Fields on `ChartRegistry` are now optional; consumers use `?.` chaining.
- Follows the `sharedToolbar` precedent for multi-owner services (Zoom/Navigator/Scrollbar all guard-and-register `zoomManager`).


packages/ag-charts-community/src/main.ts
```diff
@@ -10,6 +10,17 @@ export { ModuleRegistry } from 'ag-charts-core';
...
+// (rather than via the `_ModuleSupport` barrel) so consumers that don't use annotations
+// can tree-shake it out — it has no community-internal runtime dependencies.
+export { AnnotationManager } from './chart/annotation/annotationManager';
+
+// Undocumented types used by Enterprise Modules. Exposed as top-level type-only exports
+// so enterprise can import them directly (without going through the _ModuleSupport barrel)
+// while keeping type identity consistent with the module resolution used everywhere else.
+export type { ChartRegistry, ChartAxisRegistry, ChartSeriesRegistry } from './module/moduleContext';
+export type { AxisContext } from './module/axisContext';
+
```
alantreadway: I'd rather we exported via `_ModuleSupport` to keep these out of the default import for users.

alantreadway (approved):
LGTM

PR #6660, opened 2026-04-23
AG-17130 Skip invisible nodes in leaf pickNode/pickNodes

alantreadway:
https://ag-grid.atlassian.net/browse/AG-17130

Leaf `Node.pickNode` / `Node.pickNodes` now respect `visible` and
`pointerEvents`, matching the guard already applied by `Group`.

Treemap hides tiles with no valid layout (`bbox == null`) by setting
`rect.visible = false` but leaves `x/y/width/height` untouched so the
rect can be reused on subsequent updates. Without this guard the
picker still hit-tested those invisible rects at their previous
positions. After an options-replacement update that rearranged the
layout, the tooltip could resolve a stale `datumIndex` against the new
root node and show the wrong datum.

Regressed in 12.0.0 with the AG-7126 refactor that moved tooltip
resolution onto `datumIndex`; before then the picked datum object was
used directly, which happened to mask the bug.

Fix #AG-17130

### DIFF
--- a/packages/ag-charts-community/src/scene/node.ts
+++ b/packages/ag-charts-community/src/scene/node.ts
@@ -357,12 +357,18 @@ export abstract class Node<TDatum = unknown> {
     }
 
     pickNode(x: number, y: number): Node | undefined {
+        if (!this.visible || this.pointerEvents === PointerEvents.None) {
+            return;
+        }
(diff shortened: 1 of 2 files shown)


iMoses (approved):
LGTM

PR #6662, opened 2026-04-24
Add entry-point hygiene rule for default exports

alantreadway:
Adds a new `.rulesync/rules/` file documenting the convention that default package entry points (`main.ts` for community and enterprise) expose only the documented public API.

Motivated by the discussion on #6658, where `AnnotationManager` (a runtime class) and `ChartRegistry` / `ChartAxisRegistry` / `ChartSeriesRegistry` / `AxisContext` (internal types) were added directly to `packages/ag-charts-community/src/main.ts`. The justification was tree-shakability, but that only addresses runtime bundle size — the type surface, IDE autocomplete, and `.d.ts` leakage costs are paid regardless, and once shipped the symbols become implicit public API.

The rule auto-loads whenever `packages/ag-charts-*/src/main.ts` is in the editing context, and cross-references the existing `module-support.md` rule which handles the symmetric barrel-contents discipline.

### DIFF
--- a/.rulesync/.gitignore
+++ b/.rulesync/.gitignore
@@ -24,6 +24,7 @@ mcp.json
 !/rules/data-model.md
 !/rules/defaults.md
 !/rules/dom-performance.md
+!/rules/entry-point-hygiene.md
 !/rules/module-support.md
 !/rules/series.md
 !/rules/server-side-rendering.md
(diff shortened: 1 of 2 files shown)


iMoses (approved):
LGTM

PR #6666, opened 2026-04-24
Stabilise flaky range-buttons visible-window-function e2e snapshot

alantreadway:
## Summary

- The \`range buttons › actions\` e2e test captures a canvas screenshot immediately after clicking **Visible Window Function**. Since it was added on 2026-03-03, its snapshot has been auto-regenerated **23 times** — the highest churn of any screenshot in \`ranges.spec.ts\`.
- Root cause: Playwright's \`animations: 'disabled'\` stops CSS animations but not canvas scene-graph animations, and after the click the cursor can be left hovering toolbar controls. The forward-nav (►) button in the zoom-pan toolbar rendered in varying hover/border states between runs — a consistent 24×24 diff region at (396, 454).
- Fix: move the mouse to a neutral position and \`waitForChartUpdate(wrapper)\` before the screenshot, to drain \`data-update-pending\` / \`data-animating\` before capture.

## Test plan

- [ ] E2E shard 15 passes locally on repeated runs (no snapshot drift).
- [ ] Other \`range-buttons-actions-*.png\` assertions are unaffected.
- [ ] CI \`E2E Tests (15/16)\` green on this PR.

After merge, the existing committed baseline is now effectively unreachable (different pixels from the new deterministic capture). A single follow-up snapshot-update commit will be needed to establish the new stable baseline.


taktran (approved):
LGTM

PR #6671, opened 2026-04-24
Remove jsonApply and decorators from the axis path

iMoses:
- **Commit 1 (`7b53df1c07`)**: Replace `Chart.createAxes` / `Chart.applyAxes` `jsonApply`+`jsonDiff` calls with an explicit `axis.applyOptions(options, skip)` method. Sub-objects (`AxisLine`, `AxisGridLine`, `AxisTick`, `AxisTitle`, `TimeAxisParentLevel`) migrate to plain classes with their own `applyOptions`; `AxisLabel` / `AxisInterval` / `CartesianAxisLabel` keep `BaseProperties` due to enterprise-series inheritance and expose an `applyOptions` bridge. Axis-plugin modules (Crosshair) configure via `applyOptions` too, eliminating jsonApply's recursion through the side-channel field on the axis.
- **Commit 2 (`50f7fb98f6`)**: Remove `@ObserveChanges`, `@ProxyPropertyOnWrite`, and `@ActionOnSet` from the axis files. `gridLength` now lives in a per-axis `ReactiveState`; callers invoke `axis.setGridLength()` and the constructor-registered observer handles cross-line re-init + grid-visibility change. `TimeAxis` emits its "use unit-time" warning from `applyOptions` instead of a setter proxy. `CategoryAxis` deletes its `bandAlignment` backing field (`layoutConstraints.align` is the single source of truth) and replaces the `requiredRange` `@ActionOnSet` with an `applyRequiredRange` method called explicitly by the chart.

### Out of scope
- Chart-level `jsonApply` (chart.ts:1610) and series-level `properties.set(...)` paths.
- `LabelStyle` / `LabelBorder` / `Caption` remain on `BaseProperties` — they are shared with enterprise series and migrate alongside them in a follow-up.

### DIFF
--- a/packages/ag-charts-community/src/chart/axis/axis.ts
+++ b/packages/ag-charts-community/src/chart/axis/axis.ts
@@ -13,7 +13,7 @@ import {
     ChartAxisDirection,
     ChartUpdateType,
     CleanupRegistry,
-    ObserveChanges,
+    ReactiveState,
     WeakCache,
     ZIndexMap,
     callWithContext,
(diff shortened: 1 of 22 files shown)


iMoses:
Closed in favour of #6683

PR #6672, opened 2026-04-27
AG-17059 Wire up waterfall item-level tooltip.renderer

alantreadway:
https://ag-grid.atlassian.net/browse/AG-17059

The waterfall series exposed `item.{positive,negative,total}.tooltip.renderer` on its public API surface but the runtime path only consulted the series-level renderer, so item-level renderers were silently ignored.

`getTooltipContent` now resolves the per-item config and overrides the series-level `SeriesTooltip` with the item-level renderer when set. Item-level wins over series-level; subtotal datums map to `item.total` (matching existing per-item style/label routing).

Tests added drive end-to-end through hover and assert on rendered tooltip HTML.

Fix #AG-17059


iMoses (approved):
LGTM

PR #6677, opened 2026-04-28
AG-10316 / AG-11689 Consolidate renderer signatures onto shared Renderer<P, R>

alantreadway:
https://ag-grid.atlassian.net/browse/AG-10316
https://ag-grid.atlassian.net/browse/AG-11689

Bundles the AG-10316 legend item tooltip follow-up with the AG-11689 caption tooltip fall-through (originally proposed in #6678) plus a global consolidation of renderer signatures onto the shared `Renderer<P, R>` type. Addresses [iMoses's review comment](https://github.com/ag-grid/ag-charts/pull/6677#issuecomment-4333946986) — _"Shouldn't we have a consistent behaviour on all renderers?"_ — by making `Renderer<P, R>` the single canonical renderer type across `ag-charts-types`, with a uniform "empty string suppresses, undefined falls through to default" contract.

#6678 is closed in favour of this PR.

### Type-level

- Widen `Renderer<P, R>` from `(params: P) => TextValue | R` to `(params: P) => TextValue | R | undefined`.
- Document the return-value contract on the shared type:
  - Return `TextValue` (string/number/Date) or `R` to provide content.
  - Return `undefined` (or omit) to fall through to the default content.
  - Return `''` to render empty content (tooltip-style consumers treat `''` as "suppress").
- Migrate the two bespoke renderer signatures to use the shared type:
  - `AgCaptionTooltipOptions.renderer` → `Renderer<…, string>`
  - `AgChartLegendItemTooltipOptions.renderer` → `Renderer<…, string>`
- Alias the community-internal `TooltipRenderer<P>` to `Renderer<P, AgTooltipRendererResult>`.
- Add `aglint/require-shared-renderer` lint rule preventing future bespoke renderer signatures.

### Runtime

- Widen runtime validators (caption / overlay / crosshair) so number + Date returns are accepted alongside the existing string / R / HTMLElement options.
- Coerce TextValue (number/Date) via `toTextString` at every Renderer<P, R> consumer call site (legend, caption, overlay, crosshair, waterfall, sparkline) so number/Date returns work consistently.
- Fix call sites that didn't previously honour `undefined`:
  - overlay → fall through to default text branch
  - crosshair label → fall through to `defaults`
  - waterfall + sparkline already handled it correctly via existing structured-tooltip logic; regression tests added.
- Caption tooltip: ported from #6678 — empty string suppresses, undefined falls through to `text ?? captionText`.
- Legend item tooltip:
  - Rename renderer param `enabled` → `visible` to match `AgInitialStateLegendOptions.visible` and the rest of the legend API surface.
  - Undefined falls through to `text` or default legend item label; empty string suppresses.

### Tests / docs / examples

- Regression coverage for undefined fall-through: legend, caption, overlay, crosshair, waterfall, sparkline.
- TextValue (number/Date) regression tests: legend tooltip, overlay.
- Aglint rule snapshot test (catches the 3 bespoke-signature patterns in test data; passes on union/intersection forms).
- Caption e2e: ported #6678's "Undefined Renderer" button + 2 specs.
- Legend example/docs: `enabled` → `visible` rename + new fall-through documentation.
- Titles docs page: fall-through / suppress contract documented.

## Test plan

- [x] `yarn nx format`
- [x] `yarn nx run-many -t build:types -p ag-charts-types,ag-charts-community,ag-charts-enterprise`
- [x] `yarn nx run-many -t lint -p ag-charts-types,ag-charts-community,ag-charts-enterprise`
- [x] `yarn nx test ag-charts-community --testPathPattern="legend|caption|overlay"` (3328 passed)
- [x] `yarn nx test ag-charts-enterprise --testPathPattern="crosshair"` / `--testPathPattern="waterfall"`
- [x] `yarn nx test ag-charts-eslint-rules` (new `require-shared-renderer` snapshot)
- [ ] Playwright e2e on CI (caption + legend item tooltip specs)

Fix #AG-10316
Fix #AG-11689

### DIFF
--- a/libraries/ag-charts-eslint-rules/rules/require-shared-renderer.mjs
+++ b/libraries/ag-charts-eslint-rules/rules/require-shared-renderer.mjs
@@ -1,0 +1,65 @@
+// -*- Mode: js2; -*-
+/**
+ * @fileoverview Enforce that all `renderer?:` declarations in ag-charts-types use the shared
+ * `Renderer<P, R>` type from `chart/callbackOptions`. This keeps the renderer return-value
+ * contract (TextValue | R | undefined; empty-string suppresses, undefined falls through to the
+ * default) uniform across the public API surface.
+ */
+
+/** @type {import('eslint').Rule.RuleModule} */
+export default {
+    meta: {
+        type: 'problem',
+        docs: {
+            description:
+                'Require that `renderer` properties in ag-charts-types use the shared `Renderer<P, R>` type from callbackOptions.',
+        },
+        messages: {
+            useSharedRenderer:
+                '`renderer` must use the shared `Renderer<P, R>` type from `chart/callbackOptions` (directly, or as part of a union/intersection) so the return-value contract stays uniform across the public API. Replace this bespoke signature with `Renderer<…, R>`. If you have a genuinely exceptional case, you can disable this rule with `// eslint-disable-next-line aglint/require-shared-renderer` and document why.',
+        },
+        schema: [],
+    },
+
+    create(context) {
+        function getTypeName(typeAnnotation) {
+            if (!typeAnnotation || typeAnnotation.type !== 'TSTypeReference') return undefined;
+            const { typeName } = typeAnnotation;
+            if (typeName.type === 'Identifier') return typeName.name;
+            if (typeName.type === 'TSQualifiedName') return typeName.right?.name;
+            return undefined;
+        }
+
+        // Walk the annotation tree so unions/intersections containing `Renderer<…>` still pass.
+        // Examples allowed: `Renderer<…>`, `Renderer<…> | undefined`, `Renderer<…> & Branded`.
+        // Examples flagged: `(params) => string`, bespoke aliases like `MyRenderer<P>`.
+        function containsRendererReference(annotation) {
+            if (!annotation) return false;
+            if (getTypeName(annotation) === 'Renderer') return true;
+            if (annotation.type === 'TSUnionType' || annotation.type === 'TSIntersectionType') {
+                return annotation.types.some(containsRendererReference);
+            }
+            if (annotation.type === 'TSParenthesizedType') {
+                return containsRendererReference(annotation.typeAnnotation);
+            }
+            return false;
+        }
+
+        return {
+            TSPropertySignature(node) {
+                const key = node.key;
+                if (!key || key.type !== 'Identifier' || key.name !== 'renderer') return;
+
+                const annotation = node.typeAnnotation?.typeAnnotation;
+                if (!annotation) return;
+
+                if (containsRendererReference(annotation)) return;
+
+                context.report({
+                    node: annotation,
+                    messageId: 'useSharedRenderer',
+                });
+            },
+        };
+    },
+};
--- a/packages/ag-charts-community/src/chart/caption.ts
+++ b/packages/ag-charts-community/src/chart/caption.ts
@@ -52,7 +52,7 @@ class CaptionTooltipProperties extends BaseProperties {
     text?: string;
 
     @Property
-    renderer?: Renderer<AgCaptionTooltipRendererParams, string>;
+    renderer?: Renderer<AgCaptionTooltipRendererParams, never>;
 }
 
 export class Caption extends BaseProperties implements CaptionLike {
--- a/packages/ag-charts-community/src/chart/caption.ts
+++ b/packages/ag-charts-community/src/chart/caption.ts
@@ -18,6 +18,7 @@ import type {
     AgCaptionTooltipRendererParams,
     FontStyle,
     FontWeight,
+    Renderer,
     TextAlign,
     TextOrSegments,
     TextWrap,
@@ -51,7 +52,7 @@ class CaptionTooltipProperties extends BaseProperties {
     text?: string;
 
     @Property
-    renderer?: (params: AgCaptionTooltipRendererParams) => string;
+    renderer?: Renderer<AgCaptionTooltipRendererParams, string>;
 }
 
 export class Caption extends BaseProperties implements CaptionLike {
@@ -209,7 +210,7 @@ export class Caption extends BaseProperties implements CaptionLike {
             const params: AgCaptionTooltipRendererParams = { text: captionText };
             const result = callWithContext(moduleCtx.chartService, renderer, params);
             if (result === '') return undefined;
-            return { type: 'raw', rawHtmlString: result };
+            if (result != null) return { type: 'raw', rawHtmlString: toTextString(result) };
         }
 
         const displayText = text ?? captionText;
--- a/packages/ag-charts-community/src/chart/legend/legend.ts
+++ b/packages/ag-charts-community/src/chart/legend/legend.ts
@@ -24,6 +24,7 @@ import {
     isTextTruncated,
     objectsEqual,
     toPlainText,
+    toTextString,
     truncateLine,
 } from 'ag-charts-core';
 import type { AgChartLegendContextMenuEvent, AgMarkerShapeFn } from 'ag-charts-types';
@@ -1069,7 +1070,7 @@ export class Legend {
             };
             const result = this.cachedCallWithContext(tooltipOpts.renderer, params);
             if (result === '') return undefined;
-            if (result != null) return [{ type: 'raw', rawHtmlString: result }];
+            if (result != null) return [{ type: 'raw', rawHtmlString: toTextString(result) }];
         }
 
         if (tooltipOpts?.text != null) {
--- a/packages/ag-charts-community/src/chart/legend/legend.ts
+++ b/packages/ag-charts-community/src/chart/legend/legend.ts
@@ -1058,17 +1058,18 @@ export class Legend {
         if (visible === 'never') return undefined;
         if (visible === 'auto' && !isTruncated) return undefined;
 
-        // Content precedence: renderer > text > default label
+        // Content precedence: renderer > text > default label.
+        // A renderer returning '' suppresses the tooltip; returning undefined falls through to text/label.
         if (tooltipOpts?.renderer) {
             const params = {
                 seriesId: datum.seriesId,
                 itemId: datum.itemId ?? datum.id,
                 text: toPlainText(datum.label.text),
-                enabled: datum.enabled,
+                visible: datum.enabled,
             };
             const result = this.cachedCallWithContext(tooltipOpts.renderer, params);
-            if (result == null || result === '') return undefined;
-            return [{ type: 'raw', rawHtmlString: result }];
+            if (result === '') return undefined;
+            if (result != null) return [{ type: 'raw', rawHtmlString: result }];
         }
 
         if (tooltipOpts?.text != null) {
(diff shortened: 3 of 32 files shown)


iMoses:
Similar concern as in #6678 about the renderer signature

packages/ag-charts-types/src/chart/callbackOptions.ts
```diff
@@ -71,5 +71,17 @@ export interface DatumItemCallbackParams<
 export type Formatter<P> = (params: P) => TextValue | undefined;
 export type RichFormatter<P> = (params: P) => TextOrSegments | undefined;
 export type Styler<P, S> = (params: P) => S | undefined;
-export type Renderer<P, R> = (params: P) => TextValue | R;
+/**
```
alantreadway: Drop this JSDoc, its unecessary

packages/ag-charts-types/src/chart/legendOptions.ts
```diff
@@ -112,8 +112,8 @@ export interface AgChartLegendItemTooltipRendererParams<TContext = ContextDefaul
     itemId: string | number;
     /** The full, untruncated legend item label text. */
     text: string;
-    /** Whether the legend item is currently enabled (series visible). */
-    enabled: boolean;
+    /** Whether the series for this legend item is currently visible. */
+    visible: boolean;
     /** Callback context for this renderer. */
     context?: TContext;
```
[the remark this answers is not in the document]
alantreadway: Ack - not previously released.

libraries/ag-charts-eslint-rules/rules/require-shared-renderer.mjs
```diff
@@ -0,0 +1,65 @@
...
+        function getTypeName(typeAnnotation) {
+            if (!typeAnnotation || typeAnnotation.type !== 'TSTypeReference') return undefined;
+            const { typeName } = typeAnnotation;
+            if (typeName.type === 'Identifier') return typeName.name;
+            if (typeName.type === 'TSQualifiedName') return typeName.right?.name;
+            return undefined;
+        }
+
+        // Walk the annotation tree so unions/intersections containing `Renderer<…>` still pass.
+        // Examples allowed: `Renderer<…>`, `Renderer<…> | undefined`, `Renderer<…> & Branded`.
```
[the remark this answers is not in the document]
alantreadway: Ack

packages/ag-charts-types/src/chart/legendOptions.ts
```diff
@@ -112,8 +112,8 @@ export interface AgChartLegendItemTooltipRendererParams<TContext = ContextDefaul
     itemId: string | number;
     /** The full, untruncated legend item label text. */
     text: string;
-    /** Whether the legend item is currently enabled (series visible). */
-    enabled: boolean;
+    /** Whether the series for this legend item is currently visible. */
+    visible: boolean;
```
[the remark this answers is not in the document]
alantreadway: Ack - `enabled` was unreleased

packages/ag-charts-community/src/chart/legend/legend.ts
```diff
@@ -1058,17 +1059,18 @@ export class Legend {
...
+        // Content precedence: renderer > text > default label.
+        // A renderer returning '' suppresses the tooltip; returning undefined falls through to text/label.
         if (tooltipOpts?.renderer) {
             const params = {
                 seriesId: datum.seriesId,
                 itemId: datum.itemId ?? datum.id,
                 text: toPlainText(datum.label.text),
-                enabled: datum.enabled,
+                visible: datum.enabled,
             };
```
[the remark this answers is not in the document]
alantreadway: Ditto

packages/ag-charts-community/src/chart/overlay/overlay.ts
```diff
@@ -58,10 +60,15 @@ export class Overlay extends BaseProperties {
...
-        if (this.renderer) {
-            const params: AgChartOverlayRendererParams<DatumDefault> = {};
-            const htmlContent = callWithContext(callers, this.renderer, params);
+        // The renderer is optional, and per the documented Renderer<P, R> contract may
+        // return `undefined` to fall through to the default overlay text. An empty string
+        // still renders (as an empty overlay). Number/Date returns (TextValue) are coerced
+        // via toTextString to keep behaviour consistent with other Renderer<P, R> consumers.
+        const params: AgChartOverlayRendererParams<DatumDefault> = {};
+        const rendered = this.renderer ? callWithContext(callers, this.renderer, params) : undefined;
+        const htmlContent = isNumber(rendered) || isDate(rendered) ? toTextString(rendered) : rendered;
```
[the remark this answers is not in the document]
alantreadway: Addressed in aa2a240dce — extracted `coerceTextValue` in `ag-charts-core` and updated both call sites.

packages/ag-charts-enterprise/src/features/crosshair/crosshair.ts
```diff
@@ -396,10 +399,16 @@ export class Crosshair extends AbstractModuleInstance {
...
         const defaults: AgCrosshairLabelRendererResult = { text: this.formatValue(value) };
-        if (this.label.renderer) {
-            return label.toLabelHtml(this.label.renderer({ value, fractionDigits }), defaults);
+        // Returning `undefined` from the renderer falls through to the default formatted value,
+        // matching the documented Renderer<P, R> contract. Empty strings still render an empty label.
+        // Number/Date returns (TextValue) are coerced via toTextString to keep behaviour consistent
+        // with other Renderer<P, R> consumers.
+        const rendered = this.label.renderer?.({ value, fractionDigits });
+        if (rendered === undefined) {
+            return label.toLabelHtml(defaults);
```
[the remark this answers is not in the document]
alantreadway: Addressed in aa2a240dce — see overlay reply.

alantreadway:
@iMoses This PR is the global fix you flagged on #6678 — see commits c594f8a83c (consolidate all renderer signatures onto the shared `Renderer<P, R>` from `ag-charts-types/src/chart/callbackOptions.ts`) and c371b22fb0 (allow `undefined` fall-through across all consumers). The new `require-shared-renderer` lint rule guards against future drift in `ag-charts-types`.

packages/ag-charts-community/src/chart/caption.ts
```diff
@@ -51,7 +52,7 @@ class CaptionTooltipProperties extends BaseProperties {
     text?: string;
 
     @Property
-    renderer?: (params: AgCaptionTooltipRendererParams) => string;
+    renderer?: Renderer<AgCaptionTooltipRendererParams, string>;
```
iMoses: nit: second parameter here is meaningless, as `TextValue` already includes `string`.
A minor optimisation could be to make the parameter optional by setting its default to `never`.
```
type Renderer<P, R = never> = (params: P) => TextValue | R;
```

packages/ag-charts-core/src/config/chartDefaults.ts
```diff
@@ -176,7 +176,7 @@ const chartCaptionOptionsDefs: OptionsDefs<AgChartCaptionOptions> = {
     tooltip: {
         visible: union('auto', 'always', 'never'),
         text: string,
-        renderer: callbackOf(string),
+        renderer: callbackOf(or(string, number, date)),
```
iMoses, line 179: Maybe out-of-scope for this PR, but shouldn't we allow passing HTML as an element as well?
Perhaps worth opening a ticket to add to the backlog.

packages/ag-charts-enterprise/src/features/crosshair/crosshair.ts
```diff
@@ -396,10 +397,13 @@ export class Crosshair extends AbstractModuleInstance {
     private getLabelHtml(value: any, label: CrosshairLabel) {
         const fractionDigits = this.axisLayout?.label?.fractionDigits ?? 0;
         const defaults: AgCrosshairLabelRendererResult = { text: this.formatValue(value) };
-        if (this.label.renderer) {
-            return label.toLabelHtml(this.label.renderer({ value, fractionDigits }), defaults);
+        // Returning `undefined` from the renderer falls through to the default formatted value,
+        // matching the documented Renderer<P, R> contract. Empty strings still render an empty label.
+        const rendered = this.label.renderer?.({ value, fractionDigits });
+        if (rendered === undefined) {
```
iMoses: What's the expected behaviour when the user returns `null`?
Shouldn't it be the same as `undefined`?

packages/ag-charts-types/src/chart/chartOptions.ts
```diff
@@ -102,8 +102,12 @@ export interface AgCaptionTooltipOptions<TContext = ContextDefault> {
...
     /** Static text to display in the tooltip. Overrides the default caption text. */
     text?: string;
-    /** Function to produce tooltip content. Return a plain string or an HTML string. Takes precedence over `text`. */
-    renderer?: (params: AgCaptionTooltipRendererParams<TContext>) => string;
+    /**
+     * Function to produce tooltip content. Return a plain string or an HTML string. Takes precedence over `text`.
+     *
+     * Returning `undefined` falls back to `text` (or the caption's own text). Returning an empty string suppresses the tooltip.
+     */
+    renderer?: Renderer<AgCaptionTooltipRendererParams<TContext>, string>;
```
iMoses: nit: Ditto about the second parameter being `string`

iMoses (approved):
LGTM. Some nit comments, only the null vs undefined is a real concern.

packages/ag-charts-community/src/chart/caption.ts
alantreadway: Good catch — addressed in 85de716bc7. `Renderer` now defaults R to `never` (`type Renderer<P, R = never> = (params: P) => TextValue | R | undefined`), and the three sites that were using `Renderer<P, string>` (caption, legend tooltip, chart caption tooltip) now use `Renderer<P, never>`. The `aglint/require-explicit-generic` rule still requires explicit args in `ag-charts-types`, so the default isn't visible at the API surface — but the placeholder is now meaningful.

packages/ag-charts-types/src/chart/chartOptions.ts
alantreadway: Same fix — see reply on caption.ts:55.

packages/ag-charts-enterprise/src/features/crosshair/crosshair.ts
alantreadway: Good catch — fixed in 85de716bc7. The crosshair now treats `null` the same as `undefined` (fall through to the default formatted value). The type contract still doesn't include `null`, but JS callers ignoring the contract no longer crash on the destructure in `toLabelHtml`. Added a regression test (`AG-10316 label.renderer returning null falls through (treated as undefined)`).

packages/ag-charts-core/src/config/chartDefaults.ts
alantreadway: Agreed it's out of scope here. Caption/legend tooltips would need DOM-overlay plumbing similar to `Overlay`, plus a sanitisation story for the `raw` tooltip path (currently the legend coerces with `toTextString` and feeds it into the raw HTML pipeline). I'll file a follow-up ticket to the backlog.

PR #6678, opened 2026-04-28
AG-11689 Fall back when caption tooltip renderer returns undefined

alantreadway:
## Summary

Addresses David's [AC feedback](https://ag-grid.atlassian.net/browse/AG-11689?focusedCommentId=112198) on AG-11689.

Previously a `tooltip.renderer` that returned `undefined` (or did not return) produced an empty raw HTML tooltip, hiding it entirely. The expected behaviour is to fall back to `tooltip.text` (or the caption's own text). Empty string still suppresses the tooltip per AC 9.2.6.

- `caption.ts` — only consume the renderer result when it is a non-empty string; otherwise fall through to `text ?? captionText`.
- `chartOptions.ts` — widen the renderer return type to `string | undefined` and document the fallback / suppress contract.
- E2E example + tests — adds an Undefined Renderer button and two regression tests covering the title (falls back to caption text) and subtitle (falls back to `tooltip.text`) paths.
- Titles docs page — note the fallback / suppress contract for the renderer.

## Test plan

- [x] `yarn nx build:types ag-charts-community`
- [x] `yarn nx lint ag-charts-community`
- [x] `yarn nx lint ag-charts-types`
- [x] `yarn nx test ag-charts-community --testPathPattern=caption`
- [ ] CI Playwright run for `caption-tooltip.spec.ts`

Fix #AG-11689

### DIFF
--- a/packages/ag-charts-community/src/chart/caption.ts
+++ b/packages/ag-charts-community/src/chart/caption.ts
@@ -51,7 +51,7 @@ class CaptionTooltipProperties extends BaseProperties {
     text?: string;
 
     @Property
-    renderer?: (params: AgCaptionTooltipRendererParams) => string;
+    renderer?: (params: AgCaptionTooltipRendererParams) => string | undefined;
 }
 
 export class Caption extends BaseProperties implements CaptionLike {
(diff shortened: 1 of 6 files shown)


iMoses:
Shouldn't we have a consistent behaviour on all renderers?
Because we currently have a `Renderer` type in `ag-charts-types/src/chart/callbackOptions.ts` that should ideally be used for all renderers (not sure if it's actually the case, it isn't the case here) and it doesn't support a return value of `undefined` OOTB.
Shouldn't we address this issue globally instead of one use-case at a time?

alantreadway:
Superseded by #6677, which now bundles the caption tooltip fall-through (this PR's content) plus a global consolidation of renderer signatures onto the shared `Renderer<P, R>` type. The caption tooltip changes (caption.ts + chartOptions.ts type widening + e2e example/spec/docs) were ported across verbatim. See https://github.com/ag-grid/ag-charts/pull/6677 for the merged scope.

PR #6686, opened 2026-04-29
AG-17015(6) Rewrite the `selectionChange` event logic

olegat:
https://ag-grid.atlassian.net/browse/AG-17015

This new behaviour avoids the need for copying the dataSetSelection.ts internal buffer, and it will make it easier to implement `selectionState:'none'` by making it easier to track the selection count.


alantreadway (approved):
LGTM

PR #6687, opened 2026-04-29
AG-17015(7) Keep count of the number items selected

olegat:
This allows us to provide a `selectionState: 'none'` property when no items are selected.


alantreadway (approved):
LGTM

PR #6692, opened 2026-04-30
AG-17005 Redraw using PERFORM_LAYOUT instead of FULL

olegat:

The data does not need to be pre-processed when updating the selection state.


alantreadway (approved):
LGTM

PR #6697, opened 2026-04-30
AG-17186 Remove dead highlightState from link.itemStyler params

alantreadway:
## Summary

- Stacked on #6696 (AG-17185).
- `link.itemStyler` callback params surfaced `highlightState`, but the runtime always emitted `'none'` regardless of which node was highlighted, providing no signal. Drop the field from the public type (widen the `Omit` on the inherited `DatumCallbackParams`) and from the runtime params builder.
- Re-introducing the field with `'highlighted-branch'` / `'unhighlighted-branch'` values is **deferred to a post-MVP follow-up sub-task** paired with AG-17187.
- Seeds `upgrade-to-ag-charts-13-3/index.mdoc` with a Breaking Changes entry for the removal; release-prep can amend with the rest of the breaking changes for the version.

JIRA: [AG-17186](https://ag-grid.atlassian.net/browse/AG-17186) — title + description updated to V1-only scope (manual edit pending).

## Test plan

- [x] `yarn nx test ag-charts-enterprise` (full suite green; no link rendering regression).
- [x] `yarn nx build:types` clean — type-level break is intentional and documented.
- [x] `yarn nx lint ag-charts-types`, `yarn nx lint ag-charts-enterprise` — 0 errors.

## Breaking change

`AgOrganizationSeriesLinkItemStylerParams` no longer carries a `highlightState` field. TypeScript consumers reading the field will hit a compile error; runtime semantics for any guard like `params.highlightState === 'none'` are preserved (`undefined === 'none'` is also `false`).

### DIFF
--- a/packages/ag-charts-website/src/content/docs/upgrade-to-ag-charts-13-3/index.mdoc
+++ b/packages/ag-charts-website/src/content/docs/upgrade-to-ag-charts-13-3/index.mdoc
@@ -1,0 +1,95 @@
+---
+title: 'Upgrade to AG Charts 13.3'
+description: "See what's new in AG Charts v13.3. View a full list of breaking changes, behaviour changes and deprecations. Read the release post for feature highlights."
+migrationVersion: '13.3.0'
+---
+
+## What's New
+
+<!-- Description with !BLOG_LINK! -->
+
+See the [release post](https://blog.ag-grid.com/whats-new-in-ag-charts-13-3/) for feature highlights of what's new in this version.
+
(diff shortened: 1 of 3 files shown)


packages/ag-charts-website/src/content/docs/upgrade-to-ag-charts-13-3/index.mdoc
alantreadway, line 1: Drop this - org chart isn't release yet so this isn't a breaking change

PR #6706, opened 2026-05-01
AG-17005 Add `Bitfield` class

olegat:


olegat:
Shelving this improvement for now – it's a lot of work for very small gain

PR #6691, opened 2026-04-30
AG-16046 Bugfix: Dismiss tooltip on missing-data hover for colour-scale series

alantreadway:
## Summary

Pt2 follow-up to [AG-16046](https://ag-grid.atlassian.net/browse/AG-16046) addressing David Glickman's review feedback: hovering a datum with a missing \`colorValue\` (or a series with \`tooltip.enabled = false\`) left the previous neighbour's tooltip stuck on screen for the colour-scale series (heatmap, map-shape, map-marker, line, treemap, sunburst).

## Fix

- **\`seriesAreaManager.showTooltip\`**: empty \`tooltipContent\` now triggers \`removeTooltip\` rather than calling \`updateTooltip\` with an empty array. **This broadens the dismissal semantics**: \`tooltip.enabled = false\` on a series now also dismisses any prior tooltip, because \`chart.getTooltipContent\` returns \`[]\` in that case. Flagging here so it isn't merged silently.
- **Per-series \`getTooltipContent\` early-returns** on missing \`colorValue\`: heatmap, map-marker, treemap (gated on \`isLeaf\`), sunburst.
- **Bubble / scatter parity**: \`missingDataFill\` is now applied in \`updateDatumStyles\` (the type already exposed it; the runtime ignored it). \`getTooltipContent\` early-returns on missing colorValue. Scatter inherits via class extension.

## Tests

- Hover-driven regression tests in \`heatmap.test.ts\` assert tooltip dismissal on (1) a missing-colorValue hover after a valid hover and (2) a hover after the series tooltip is disabled at runtime — exercising the \`seriesAreaManager.showTooltip\` empty-array path end-to-end.
- New enterprise test (\`colorScaleMissingData.test.ts\`) covers \`BubbleSeries\` / \`ScatterSeries\` tooltip suppression. \`colorKey\` on those series is enterprise-gated, so the test must run under enterprise registry mode.
- Tooltip-suppression assertions added / improved on heatmap, map-marker, sunburst, treemap unit tests; brittle index lookups and \`as any\` casts replaced with predicate-derived \`findIndex\` lookups and proper series types.
- A shared-tooltip-mode P2 case was investigated but dropped — overlapping heatmaps surfaced an unrelated JSDOM picker quirk; the same dismissal code path is exercised by the included tests.

## Test plan

- [x] \`yarn nx format\`
- [x] \`yarn nx run-many -t build:types -p ag-charts-community,ag-charts-enterprise\`
[...]

PR #6707, opened 2026-05-01
AG-16046 PR review: tier-2 colour-scale series keep tooltip on missing colour

alantreadway:
https://ag-grid.atlassian.net/browse/AG-16046

Addresses David's QA classification (breaking-change/regression) on the pt2 PR (#6691). Pt2 added per-series `getTooltipContent` early-returns gated on `colorScale.missingDataFill != null`, suppressing the tooltip whenever a missing-colour datum was rendered with the override fill. For series where the mark remains visible with shape/label/geometry intact, that suppression hides queryable data (size, label) and is more restrictive than `tooltip.renderer` can recover from.

Splits the rule by series category, per [David's tier-2 ruling](https://ag-grid.atlassian.net/browse/AG-16046?focusedCommentId=112958):

- **Tier 1 — colour is the only value (heatmap, map-shape):** keep the early-return. An empty cell has nothing to report.
- **Tier 2 — colour is supplementary (treemap leaves, sunburst, map-marker, map-line, bubble/scatter):** drop the early-return entirely. Tooltip now shows with the colour row omitted; the swatch follows the rendered fill (`missingDataFill` or `series.fill` fallback). For map-line specifically, also drop the long-standing `createNodeData` filter that previously omitted lines with missing colour values when no `missingDataFill` was set — those lines now render with the default series stroke, matching David's tier-2 "fall back to series.fill" expectation.

The `seriesAreaManager.showTooltip` empty-content → `removeTooltip` dismissal change from pt2 is retained.

## Test plan

- [x] `yarn nx format`
- [x] `yarn nx run-many -t build:types -p ag-charts-community ag-charts-enterprise`
- [x] `yarn nx run-many -t lint -p ag-charts-community ag-charts-enterprise` (only pre-existing warnings)
- [x] `yarn nx test ag-charts-community` — 3331 passed
- [x] `yarn nx test ag-charts-enterprise` — 1929 passed
- [x] New `assertTooltipPresentForAll` helper companion to `assertTooltipSuppressedForMissing`
- [x] Tier-2 tests flipped to expect tooltips for missing-colour datums (sunburst, treemap, map-marker, bubble/scatter, map-line in both with and without `missingDataFill` cases)
- [x] Tier-1 (heatmap) assertions unchanged
- [x] Existing "omits lines without color values" map-line test rewritten to assert the new tier-2 behaviour, snapshot refreshed

Fix #AG-16046


iMoses (approved):
LGTM

PR #6710, opened 2026-05-02
AG-17179 Org chart zoom support

alantreadway:
JIRA: [AG-17179](https://ag-grid.atlassian.net/browse/AG-17179)


packages/ag-charts-enterprise/src/series/network/networkLayout.ts
```diff
@@ -24,6 +24,13 @@ export interface NetworkLayoutUpdateOptions<TVertex, TEdge> {
 }
 
 export abstract class NetworkLayout<TVertex, TEdge> {
+    /**
+     * Bounding box of all laid-out nodes at native (scale-1) sizes, in layout-space coordinates.
+     * Set by concrete subclasses after each `update()` call. Used by `applyViewportTransform()`
+     * to compute the fit-to-viewport scale without requiring a layout re-run.
+     */
+    contentBBox?: _ModuleSupport.BBox;
```
alantreadway: Needs to be public?

packages/ag-charts-enterprise/src/series/organization/organizationSeries.test.ts
```diff
@@ -911,14 +942,135 @@ describe('OrganizationSeries', () => {
...
+            await waitForChartStability(chart);
+            setZoom(chart, 0.8, 1, 0, 1);
+            await waitForChartStability(chart);
+
+            // Spy after the initial zoom so only hover-induced pans count.
+            const zoomManager = deproxy(chart).ctx.zoomManager;
+            const panSpy = zoomManager ? jest.spyOn(zoomManager, 'panToBBox') : undefined;
+            const ratiosBefore = getZoomRatios(chart);
+
+            // `activeManager.update` is the canonical hover simulation — JSDOM canvas
```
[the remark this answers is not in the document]
alantreadway: Fixed

packages/ag-charts-enterprise/src/series/organization/__image_snapshots__/organization-series-test-ts-organization-series-series-area-clipping-should-clip-dragged-content-to-the-series-area-so-nodes-do-not-bleed-into-the-title-1-snap.png
iMoses, line 1: Snapshot looks fine, buy test description no longer matches the snapshot content.
Should pan the chart to see it's being clipped vertically under the titles.

alantreadway: Good catch — switched the clipping test over to a new TALL_ORG_CHART fixture (narrow + tall, fits horizontally, overflows vertically), applied y-only zoom, and dragged content up far enough that an upper card now clearly straddles the title boundary with its top half cut off. Snapshot updated.

packages/ag-charts-enterprise/src/series/network/networkSeries.ts
```diff
@@ -91,26 +97,24 @@ export abstract class AbstractNetworkSeries<
...
-            // bounds. Clipping prevents nodes/links overflowing into title/subtitle/footnote.
+            // Clip stops viewportGroup overflow into title/footnote during zoom/pan.
             alwaysClip: true,
+            supportsStandaloneZoom: true,
         });
 
         this.graph = this.createNetworkGraph();
         this.layout = this.createNetworkLayout();
 
         ctx.eventsHub.on('layout:complete', (event) => {
```
iMoses: Shouldn't all of the constructor's `ctx.eventsHub.on` and `ctx.chartState.observe` calls be part of the `this.cleanup.register` to be cleaned up when destroyed?

packages/ag-charts-enterprise/src/series/network/networkSeries.ts
```diff
@@ -120,13 +124,31 @@ export abstract class AbstractNetworkSeries<
...
             const activeItem = get('activeItem');
             if (activeItem?.seriesId === this.id) {
                 this.expandNetworkToItem(activeItem.itemId);
             }
         });
 
+        // Zoom is a transform-only update — no re-layout.
+        this.cleanup.register(
+            ctx.chartState.observe((get) => {
+                get('zoom');
```
iMoses: if we're not really using the state, shouldn't we use the `zoom:change-complete` event instead?

packages/ag-charts-enterprise/src/series/network/networkSeries.ts
alantreadway: This is pre-existing, but you're right I'll cleanup whilst we're here

alantreadway: Good call — wrapped the four un-registered subscriptions (`layout:complete`, `collapsed:restore`, the activeItem observer, and `series-area:click`) in `this.cleanup.register` to match the two that already were.

packages/ag-charts-enterprise/src/series/network/networkSeries.ts
alantreadway: Agreed — switched to `eventsHub.on('zoom:change-complete', ...)`. `applyViewportTransform` reads the zoom from chartState directly, so the observer's value tracking was unused and its initial-fire was a no-op (seriesRect/contentBBox aren't set yet at constructor time).

iMoses (approved):
LGTM

PR #6717, opened 2026-05-05
AG-17167: itemStyler cannot change marker size on scatter/bubble series

iMoses:
https://ag-grid.atlassian.net/browse/AG-17167


alantreadway (approved):
LGTM

PR #6720, opened 2026-05-05
AG-17005 Faster aggregation selection based on `Uint8Array.fill`

olegat:


alantreadway (approved):
Nice, LGTM

PR #6721, opened 2026-05-05
AG-16511: Incorrect layout when subtitle is blank string on initial load

iMoses:
https://ag-grid.atlassian.net/browse/AG-16511


alantreadway (approved):
LGTM

PR #6727, opened 2026-05-06
AG-17247: Global formatter option ignored on number axis in financial…

iMoses:
… chart preset

https://ag-grid.atlassian.net/browse/AG-17247


alantreadway (approved):
LGTM

PR #6728, opened 2026-05-06
AG-17207 TC3 Do not select the "nearest" datum

olegat:
https://ag-grid.atlassian.net/browse/AG-17207

`SeriesAreaClickEvent` now includes `distance` in the payload so that `dataSelection.ts` can distinguish between nearest & exact clicks.


packages/ag-charts-enterprise/src/features/data-selection/dataSelection.ts
olegat, line 1: This is the key change. The rest is all just plumbing to include `distance: number` in the event payload.

alantreadway (approved):
LGTM

PR #6740, opened 2026-05-06
AG-17004 TC2 Bugfix: only record nodes that are actually removed

olegat:


alantreadway (approved):
LGTM

PR #6741, opened 2026-05-06
AG-17004 TC3 Turn off selection when chart.selection.enabled is off

olegat:


alantreadway (approved):
LGTM

PR #6744, opened 2026-05-07
AG-17005 Support data selection on bubble and scatter series

alantreadway:

## Summary

- Add `DatumIndexSetReader` API to `Series` base class, with a concrete override in `BubbleSeries` that maps each rendered aggregated marker back to its underlying data indices
- Expose `indices` on `GroupedAggregation` so `BubbleSeries` can build the aggregate-to-source index mapping
- Wire `getAggregateIndexSetReader()` into `DataSelection.onSeriesAreaDragEnd` so drag-select fans out aggregated markers to all constituent data items
- Fix programmatic `setSelection()` not triggering node re-render — add `hasChangesOnSelection` flag (parallel to `hasChangesOnHighlight`) so `updateDatumNodes` runs after selection changes
- Add community tests (snapshot + unit) and enterprise tests (snapshot, programmatic selection, drag-select fan-out)

## Test plan

- [x] Community bubble series tests pass (`yarn nx test ag-charts-community --testPathPattern=bubbleSeries --testNamePattern=AG-17005`)
- [x] Enterprise data-selection tests pass (`yarn nx test ag-charts-enterprise --testPathPattern=dataSelection`)
- [x] Full community suite (3339 tests, 1612 snapshots) — no regressions
- [x] Full enterprise suite (1988 tests, 1703 snapshots) — no regressions
- [x] `build:types` and lint pass for both packages

Fix #AG-17005


packages/ag-charts-enterprise/src/features/data-selection/dataSelection.ts
```diff
@@ -283,6 +284,17 @@ export class DataSelection extends AbstractModuleInstance implements _ModuleSupp
                                 intervalSet.add(start, end);
```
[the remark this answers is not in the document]
alantreadway: False positive — the `else if (getIndexSetForAggregate)` on line 287 is **inside** the `asNumericDatumIndex` guard (line 277), not after it. The chain is: numeric index → try range reader (bar series) → try index-set reader (bubble series) → fallback. The drag-select test confirms fan-out works: it asserts `selectedCount > renderedMarkerCount`.

packages/ag-charts-community/src/chart/series/cartesian/bubbleSeries.test.ts
```diff
@@ -1074,4 +1074,105 @@ describe('BubbleSeries', () => {
...
+            const options: AgCartesianChartOptions = {
+                data: Array.from({ length: 20 }, (_, i) => ({ x: i, y: i % 5, s: 1 + (i % 3) })),
+                series: [{ type: 'bubble', xKey: 'x', yKey: 'y', sizeKey: 's', maxRenderedItems: 10 }],
+                legend: { enabled: false },
+            };
+            prepareTestOptions(options);
+            chart = AgCharts.create(options);
+            await waitForChartStability(chart);
+
+            const series = deproxy(chart).series[0] as any;
```
olegat: Optional: consider using the `Caster` utility instead of `as any`.

packages/ag-charts-enterprise/src/features/data-selection/dataSelection.test.ts
```diff
@@ -0,0 +1,388 @@
...
+                    x: { type: 'number', position: 'bottom' },
+                    y: { type: 'number', position: 'left' },
+                },
+            };
+
+            prepareEnterpriseTestOptions(options);
+            chart = AgCharts.create(options);
+            await waitForChartStability(chart);
+
+            const series = deproxy(chart).series[0] as any;
```
olegat: Optional: This is `as any` followed by `dataAggregation` is a repeated pattern that is time-consuming to maintain. If we move/rename `Series.dataAggregation` then all the tests that use this pattern would need to be repaired one-by-one. The utility getter in packages/ag-charts-community/src/chart/test/utils.ts might be better suited.

packages/ag-charts-enterprise/src/features/data-selection/dataSelection.ts
```diff
@@ -283,6 +284,17 @@ export class DataSelection extends AbstractModuleInstance implements _ModuleSupp
...
                             }
                         }
+                    } else if (getIndexSetForAggregate) {
+                        const indices = getIndexSetForAggregate(datum.datumIndex);
```
olegat: Feels unnecessary to allow the `getIndexSetForAggregate()` to return `undefined`.

Understandably this decision seems to follow the pattern of `type DatumRangeReader = (sampledDatumIndex: number) => [number, number] | undefined;` in `seriesTypes.ts` but I think that's also a mistake; that looks like a missing parentheses typo:
```
type DatumRangeReader = 
  | ((sampledDatumIndex: number) => [number, number])
  | undefined;
```
Though that type should simply just be `type DatumRangeReader = (sampledDatumIndex: number) => [number, number]`, because the implemention never return `undefined`
* https://github.com/ag-grid/ag-charts/blob/b31437dd8f405d1500ec113f75c297e62059501c/packages/ag-charts-community/src/chart/series/aggregationRangeReader.ts#L93

This would avoid the need for the nested `if (indices === undefined)` which duplicates the code about 10 lines down.

packages/ag-charts-enterprise/src/features/data-selection/dataSelection.ts
olegat: This check is obsolete.

`seriesTypes.ts`:
```
type DatumRangeReader = (sampledDatumIndex: number) => [number, number] | undefined;
```

should be:
```
type DatumRangeReader = (sampledDatumIndex: number) => [number, number] ;
```

Initially `DatumRangeReader` was declared with a `DatumIndexType` generic parameter which is what these `asNumericDatumIndex` were guarding against.

```
type DatumRangeReader<I extends DatumIndexType>  = (sampledDatumIndex: I) => [I, I];
// which roughly expands to:
type DatumRangeReader  = (sampledDatumIndex: number | object | undefined) => [number | object | undefined, number | object | undefined];
```
But this was changed to just a `number` for simplicity.

olegat (approved):
I'm approving to unblock, although we should consider cleaning up the unnecessary if-else nested logic complexity in `onSeriesAreaDragEnd` at some point. Happy to do this myself if necessary.

packages/ag-charts-community/src/chart/series/cartesian/bubbleSeries.test.ts
alantreadway: Addressed in `6b53d11c82` — extracted `getSeriesAggregationInternals` helper in `chart/test/utils.ts`, used here and in `dataSelection.test.ts`. Caster did not fit the method-call sites cleanly; a typed accessor was the lower-friction route.

packages/ag-charts-enterprise/src/features/data-selection/dataSelection.test.ts
alantreadway: Done in `6b53d11c82` — added `getSeriesAggregationInternals` and migrated all `(series as any).dataAggregation` sites to it.

packages/ag-charts-enterprise/src/features/data-selection/dataSelection.ts
alantreadway: Inner guard removed in `6b53d11c82`. Kept the outer one — `DatumIndexType` is still `number | object | undefined` (gauges/maps carry non-numeric indices).

packages/ag-charts-enterprise/src/features/data-selection/dataSelection.ts
alantreadway: Tightened `DatumIndexSetReader` in `6b53d11c82`; bubble yields `[]` for unknown indices, collapsing the duplicate fallback. Left `DatumRangeReader` as-is — bar legitimately returns undefined (`barSeries.ts:520-536`); tightening it needs different "no match" semantics for bar, out of scope here.

PR #6746, opened 2026-05-07
AG-17207 Default nodeClickRange to 10 when selection is enabled

olegat:
https://ag-grid.atlassian.net/browse/AG-17207


alantreadway (approved):
LGTM

PR #6747, opened 2026-05-07
AG-17257 Set click 'pointer' cursor when selection is enabled

olegat:
https://ag-grid.atlassian.net/browse/AG-17257


alantreadway (approved):
LGTM

PR #6737, opened 2026-05-06
Org chart roundup pt 2 — AG-17249, AG-17250, AG-17252, AG-17253

alantreadway:
## Summary

Second batch of high-priority org chart fixes for 35.3.0, building on the Pt 1 branch (AG-17243, AG-17246, AG-17192, AG-17205, AG-17010).

**Pt 2 tickets:**

- **AG-17249** — Always invalidate cached bbox on `Path` property changes. The `onChangeDetection` guard previously skipped `super` when `_dirtyPath` was already set, leaving the bbox cache stale after late position updates (e.g. `realign()`). This caused image-fill pattern transforms to anchor to the wrong coordinates, making images invisible at top/bottom positions.
- **AG-17250** — Stop org chart card height growing on every layout pass. The expander pill hangs below the card by half its height; `getDatumNodeBBox` was measuring the full group bbox including the pill overhang, feeding it back into `regularBBox` sizing each cycle. Adds `measureDatumNode` override that measures the card rect only.
- **AG-17252** — Honour `enabled: false` on org chart label entries. A disabled label entry with no `key` crashed `dataModel` with "no properties specified". Disabled tiers are now skipped in `processData` with slot alignment preserved downstream.
- **AG-17253** — Wrap and truncate org chart labels at narrow card widths. Adds `computeTextMaxWidth` / `wrapTextTier` helpers so long titles wrap when `maxWidth` or `width` is set. Nodes auto-size by default; users opt into wrapping by setting `node.maxWidth`.

**Pt 1 tickets (carried on this branch):**

- AG-17243, AG-17246, AG-17192, AG-17205, AG-17010

## Test plan

- [x] `yarn nx test ag-charts-community` — passes
- [x] `yarn nx test ag-charts-enterprise` — passes (94 org chart tests, 66 zoom tests)
- [x] `yarn nx build:types ag-charts-community` — passes
- [x] `yarn nx build:types ag-charts-enterprise` — passes
- [x] `yarn nx lint ag-charts-community` — passes
- [x] `yarn nx lint ag-charts-enterprise` — passes
- [ ] Visual snapshots reviewed (53 org chart + 1 zoom snapshot updated, 4 new)
- [ ] CI green

PR #6751, opened 2026-05-07
AG-17253 clip org chart card content when clamped under intrinsic size

alantreadway:
## Summary

Follow-up to AG-17253 / PR #6737 addressing the three TODOs raised by David Glickman during pt2 QA.

`node.maxWidth` / `node.maxHeight` clamp the card rect via the layout grid's `regularBBox`, but image and text tier scene nodes were unaware of the clamp and continued drawing at their intrinsic positions:

- 50px theme-default avatar visibly bulged out of a 30-wide card.
- Trailing label tiers spilled below the card border into the link / child rows.

Wraps the image and text tier scene nodes in a new inner `contentGroup` and `setClipRectCanvasSpace` to the clamped card bounds when intrinsic content size exceeds the regularBBox. Shape rect (with stroke) and the expander pill stay outside this group, so the card stroke and the pill that hangs below the card render unaffected.

The clip is no-op on cards that fit (zero per-frame `ctx.save/clip/restore` cost), so common usage without a narrow `maxWidth` / `maxHeight` is unchanged — verified by the full org chart suite passing without any other snapshot moves.

## Test plan

- [x] `nx build:types ag-charts-enterprise` clean
- [x] `nx lint ag-charts-enterprise` 0 errors (151 pre-existing warnings, none on changed files)
- [x] `nx test ag-charts-enterprise` org chart suite — 98 tests pass, only the four new `AG-17253 pt2` snapshots moved from buggy baseline → clipped render
- [x] Verified locally against David's four QA plunkers (`0maSWVkHxeSLQ5Fb`, `oq0FP2LDZkapxUO4`, `jxPaPCmKEyWoTi3b`, `qA05jRd478qdnhDe`)

Commits split for reviewable visual diff:

1. Regression tests + buggy-baseline snapshots
2. The fix + regenerated snapshots (`git show <sha> -- '*snap.png'` shows the four PNG diffs)
3. Comment trim

Fix #AG-17253


packages/ag-charts-enterprise/src/series/organization/organizationNode.ts
```diff
@@ -218,14 +242,21 @@ export class OrganizationNode extends _ModuleSupport.TranslatableGroup<Organizat
         applyStrokeStyles(this.shapeNode, styles);
     }
 
+    // Sits between `shapeNode` and `expanderNode` in z-order: children render on top of
+    // the shape rect, and the expander pill is appended later so it stays visually last.
+    private ensureContentGroup(): _ModuleSupport.Group {
```
alantreadway: Just remove this indirection, its unecessary and adds noise.

packages/ag-charts-enterprise/src/series/organization/organizationNode.ts
```diff
@@ -151,6 +164,17 @@ export class OrganizationNode extends _ModuleSupport.TranslatableGroup<Organizat
...
             this.expanderNode.translationY = bbox.height - expanderBBox.height / 2;
         }
+
+        // Conditional clip: only when `regularBBox` clamped the card under its intrinsic
+        // size (i.e. `maxWidth`/`maxHeight` kicked in). With no overflow we leave the
+        // contentGroup unclipped so the per-frame `ctx.save/clip/restore` cost is zero.
+        const intrinsic = this.intrinsicCardSize;
+        const overflows =
+            intrinsic != null &&
+            (intrinsic.width > bbox.width + CLIP_EPSILON || intrinsic.height > bbox.height + CLIP_EPSILON);
```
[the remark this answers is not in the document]
alantreadway: No-op — render path applies parent transforms to ctx before clipping (`Group.applyClip` does `ctx.rect(0, 0, w, h)` after parent translate/scale), so a local-space `(0, 0, cardW, cardH)` lands at the card's screen position regardless of layout offset, pan, or zoom. The `setClipRectCanvasSpace` name is misleading — it skips the canvas→local conversion `setClipRect` performs — and `miniChart.ts:263` follows the same local-rooted pattern. Local-space is in fact more robust under pan/zoom than canvas-space, which would go stale between update and the next render. Adding a clarifying comment in a follow-up to spare future readers the same trip.

packages/ag-charts-enterprise/src/series/organization/organizationNode.ts
alantreadway: Done in 673bc1a34f — helper deleted, lazy-init lifted to the top of `update()` (after `updateShapeNode` so z-order is preserved). Each tier helper now references `this.contentGroup` directly.

lsjroberts (approved):
LGTM

iMoses (approved):
LGTM

PR #6753, opened 2026-05-07
AG-17215 Call `preventDefault()` to avoid default touch-scrolling

olegat:
https://ag-grid.atlassian.net/browse/AG-17215


alantreadway (approved):
LGTM

PR #6757, opened 2026-05-07
AG-17005 Roll up datum selection into aggregation buckets

alantreadway:
## Summary

- Selection state for aggregating series now rolls up through the aggregation pyramid: any rendered bucket marker reflects "any underlying datum is selected", regardless of which datum the bucket happens to use as its representative. Fixes the zoom-out-loses-marker bug David flagged in [JIRA](https://ag-grid.atlassian.net/browse/AG-17005).
- Wires the roll-up across line, area, bar (split positive/negative arms), OHLC/candlestick, range-bar, and range-area through a single `BucketLookupFeature` companion class on `Series`. Per-series footprint: one `createBucketLookupFeature()` factory + the `aggregate({ onChange })` arg.
- The companion subsumes the previous `Series.getAggregateRangeReader()` virtual — both bucket-selected lookup and bucket→datum-range mapping share one cached prepared context per `(processedData, filter)` pair, so `getDataSelectionState` doesn't recompute the aggregation domain on every marker.

Fix #AG-17005


packages/ag-charts-community/src/chart/series/seriesTypes.ts
```diff
@@ -36,7 +36,39 @@ export type SeriesNodeEventTypes =
     | 'seriesNodeDoubleClick';
 
 export type DatumRangeReader = (sampledDatumIndex: number) => [number, number] | undefined;
```
olegat, line 38: Looks like you can remove `DatumRangeReader` too. It's exported in `module-support.ts` but it's probably no longer needed there either.

packages/ag-charts-enterprise/src/features/data-selection/dataSelection.ts
```diff
@@ -286,14 +301,17 @@ export class DataSelection extends AbstractModuleInstance implements _ModuleSupp
                             changed = true;
                             intervalSet.add(start, end);
                         }
-                    } else if (getIndexSetForAggregate) {
```
olegat, line 297: nit: The old `if {} else if {} else {}` code-style from before felt easier to follow compared to this new `if {} else { if {} else {} }` code-style.

packages/ag-charts-enterprise/src/features/data-selection/dataSelection.ts
```diff
@@ -304,12 +322,14 @@ export class DataSelection extends AbstractModuleInstance implements _ModuleSupp
...
 
             if (changed) {
-                changedSeries.push(series);
+                changedSeries.add(series);
             }
         }
 
-        this.dispatchInternalSelectionChange(changedSeries, changes);
+        // External listener fires first so any rollback completes before we
+        // refresh per-series bucket caches against the final bitset state.
```
olegat, line 329: Please add a test with selection styling for error-bars to verify this. There's one on the `selection-test` page that you may base the test one. Please test both cases when `preventDefault()` is and isn't called.

packages/ag-charts-enterprise/src/features/data-selection/dataSelection.ts
```diff
@@ -196,14 +197,18 @@ export class DataSelection extends AbstractModuleInstance implements _ModuleSupp
 
             if (clickMode === 'multiple' || modifierPressed) {
                 toggleSelection(changes, series, data, datumIndex);
+                internalRefreshTargets = [series];
             } else {
+                // Single-click clears every series — all of them need their
```
olegat: Well spotted. This is indeed a bug. If you click a series with error-bars, then the styling of error-bars on other series is not cleared

https://github.com/user-attachments/assets/bac2d507-b92e-4eb2-ae1e-1ae0469f7673

Please remove this comment and replace it with a test with 2 error-bars series. It should:
1. Click a node in series A node
    - node and error-bar in series A styling is drawn.
2. Click a node in series B node
    - node and error-bar in series A styling is cleared.
    - node and error-bar in series B styling is drawn.

olegat (commented):
I do not have a capacity to review the changes in `aggregation.ts` and `bucketLookupFeature.ts` – which seems to be doing most of the heavy lifting for this PR.

There are some minor code-styling improvements that could be made.

Claude has also found some genuine (but unrelated) bugs in `dataSelection.ts` involving error-bars. We haven't done much testing on selection+error-bar styling, so those changes are useful and definitely should be merged (either here or somewhere else).

packages/ag-charts-enterprise/src/features/data-selection/dataSelection.ts
```diff
@@ -257,12 +262,22 @@ export class DataSelection extends AbstractModuleInstance implements _ModuleSupp
 
         const changes = this.allocSelectionChanges();
         const shouldClearSelections: boolean = !hasAddToSelectionModifier(dragEndEvent);
+        const changedSeries = new Set<Series>();
         if (shouldClearSelections) {
+            // Every series with prior selections has its bitset cleared; their
```
olegat: The comment is misleading, this has nothing to do with aggregation/buckets. You need `changedSeries` to be population so that `dispatchInternalSelectionChange` can inform error-bars to update their styling. 

You can probably simplify all this to just "add everything from `this.ctx.chartService.series` into `changedSeries`".

packages/ag-charts-community/src/chart/series/seriesTypes.ts
alantreadway: Done in 9063ebdadb — dropped from the `module-support.ts` exports. Type itself stays in `seriesTypes.ts` since it's still used internally by the lookup managers.

packages/ag-charts-enterprise/src/features/data-selection/dataSelection.ts
alantreadway: Restored the flat ladder in 9063ebdadb. Reordered the branches so the negated-condition lint check stays happy without nesting.

packages/ag-charts-enterprise/src/features/data-selection/dataSelection.ts
alantreadway: Added two tests in 9063ebdadb covering both cases: `renders error-bar selection styling on the selected datum` (no preventDefault) and `rolls back error-bar selection styling when selectionChange listener calls preventDefault`.

packages/ag-charts-enterprise/src/features/data-selection/dataSelection.ts
alantreadway: Removed the comment and added `clears error-bar styling on series A when single-selecting on series B` in 9063ebdadb. The test selects on series A, then on series B, and snapshots confirm A's error-bar styling is cleared while B's is applied.

packages/ag-charts-community/src/chart/series/aggregationManager.ts
```diff
@@ -38,11 +38,23 @@ export class AggregationManager<TFilter extends AggregationFilterBase> {
...
+    /**
+     * Subscribe a single callback to every filter-set mutation: aggregation
+     * rebuilds (immediate + deferred coarser levels), staleness-driven
+     * discards (`markStale`), and on-demand level merges
+     * (`ensureLevelForRange`). Used by `BucketLookupFeature` to keep the
+     * per-bucket SELECTED roll-up cache in sync with the active filter list.
+     */
+    setOnFiltersChanged(cb: (() => void) | undefined): void {
+        this.onFiltersChanged = cb;
+    }
```
alantreadway: Don't we have a better mechanism for event subscription?

alantreadway: Switched to the codebase's standard `EventEmitter` pattern in 8efa200b01 — `aggregationManager.events.on('filtersChanged', cb)` returns an unsubscribe handle, consistent with `Series.events`, `Scene`, etc.

packages/ag-charts-enterprise/src/features/data-selection/dataSelection.ts
alantreadway: Good catch — the comment was misleading. Replaced the prior-selection filter with the simpler `new Set(this.ctx.chartService.series)` on clear-and-replace, with a comment that names the actual reason (error-bar / per-marker styling needs notification on Item/OtherItem state flips). 1a7ae99369.

iMoses:
**`packages/ag-charts-core/src/utils/aggregation.ts` — `populateBucketSelectedFromSparse` / `populateBucketSelectedFromSparseSplit`**

The bucket-index computation here (`scaleFactor`, `scaledX`, `Math.floor`, clamp to `bucketCount - 1`) is inlined for hot-loop performance, but it duplicates the canonical math in `aggregationXRatioForXValue` / `aggregationXRatioForDatumIndex` / `aggregationIndexForXRatio`. I verified the two are equivalent for in-domain datums, and out-of-range writes silently no-op on TypedArrays, so correctness is fine today.

The risk is drift: if the bucketing formula ever changes (e.g. rounding strategy, domain clamping, handling of edge values), it needs to be mirrored here too. Worth adding a cross-reference comment on the inlined math pointing back to the canonical helpers, something like:

```ts
// Bucket assignment mirrors aggregationXRatioForXValue / aggregationIndexForXRatio
// (inlined here to avoid per-datum function-call overhead in the sparse loop).
```

That way a future editor grepping for `aggregationXRatioForXValue` will find these sites.

alantreadway:
@iMoses Good catch — added the cross-reference comment to both `populateBucketSelectedFromSparse` and `populateBucketSelectedFromSparseSplit` in e90b527afd. Used your suggested wording, naming `aggregationXRatioForXValue` / `aggregationXRatioForDatumIndex` / `aggregationIndexForXRatio` as the canonical sites.

iMoses (approved):
One comment, other than that LGTM

PR #6758, opened 2026-05-07
AG-17065: Fix large dataset hang on zoom for ordinal-time axes

alantreadway:
https://ag-grid.atlassian.net/browse/AG-17065

## Summary

- Fixes chart hang when deeply zoomed on a small chart with 1M ordinal-time data points
- Root cause: the continuous-scale tick estimator returned `minTickCount ≈ 500` for large ordinal-time domains, but ordinal tick steps are discrete (time-interval-based) and don't change when `tickCount` decreases — so the overlap loop iterated ~1000 times redundantly
- Fix: override `minTickCount` to 0 for ordinal-time scales, letting the binary search in `buildTickData` exit immediately when ticks stop changing

## Test plan

- [x] All 3349 community tests pass (0 snapshot changes)
- [x] All 1993 enterprise tests pass (0 snapshot changes)
- [x] 60 ordinal-time axis tests pass (including parent-level/dual-label scenarios)
- [x] Manually verified: zoom on 1M-point ordinal-time chart at iPhone SE dimensions no longer hangs
- [x] Manually verified: high-frequency data update performance is not regressed

### DIFF
--- a/packages/ag-charts-community/src/chart/axis/generateTicks.test.ts
+++ b/packages/ag-charts-community/src/chart/axis/generateTicks.test.ts
@@ -1,0 +1,43 @@
+import { describe, expect, it } from '@jest/globals';
+
+import { OrdinalTimeScale } from '../../scale/ordinalTimeScale';
+import { estimateScaleTickCount } from './generateTicks';
+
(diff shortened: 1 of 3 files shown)


iMoses (approved):
LGTM

PR #6760, opened 2026-05-08
AG-17004 Do nothing in API calls when selection is disabled

olegat:


alantreadway (approved):
LGTM

PR #6764, opened 2026-05-08
AG-17288 Respect discrete mode in colour scale fallback path

alantreadway:
https://ag-grid.atlassian.net/browse/AG-17288

## Summary

- Fix `configureColorScale()` ignoring `colorScale.discrete: true` when `colorScale.fills` is not explicitly provided
- Unify the fills and discrete-fallback branches so both route through `computeColorBins` with the correct mode
- Add regression tests for discrete fallback with default palette and explicit domain override

## Test plan

- [x] Unit tests pass (`yarn nx test ag-charts-community` — 3350 tests)
- [x] Enterprise tests pass (`yarn nx test ag-charts-enterprise` — 2000 tests)
- [x] New test: discrete mode with fallback palette colours
- [x] New test: discrete fallback with explicit domain override
- [ ] Manual: verify heatmap/treemap renders discrete bins without explicit fills

Fix #AG-17288

### DIFF
--- a/packages/ag-charts-community/src/scale/colorScaleUtil.test.ts
+++ b/packages/ag-charts-community/src/scale/colorScaleUtil.test.ts
@@ -339,6 +339,36 @@ describe('configureColorScale', () => {
         ]);
     });
 
+    test('fallback path respects discrete mode (AG-17288)', () => {
+        const scale = new ColorScale();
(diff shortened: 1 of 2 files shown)


iMoses (approved):
LGTM

PR #6771, opened 2026-05-08
AG-17089 Do not zoom-pan and data-selection-drag at the same time

olegat:
https://ag-grid.atlassian.net/browse/AG-17089


alantreadway (approved):
LGTM

PR #6781, opened 2026-05-08
AG-17196 / AG-17195 / AG-17095 Map series & range-area missing-data fixes

alantreadway:
## Summary
- **AG-17196** map-marker hover redrew the marker with the series default fill, dropping the colour-scale-derived fill and leaving the tooltip swatch out of sync. Apply the colorScale fill in both highlight and base render paths so the highlight is conveyed via opacity (matching map-shape).
- **AG-17195** map-shape, map-marker, and map-line raised "invalid value" warnings for any datum whose colorKey resolved to null/undefined. Set \`invalidValue: undefined\` on the colorValue \`valueProperty\` so missing values fall through to \`missingDataFill\` / the series default silently — matching how missing xKey/yKey are handled.
- **AG-17095** range-area logged "invalid value" warnings whenever yLowKey/yHighKey resolved to null/undefined/NaN/Infinity (regression in v11.0.0 from AG-13253). Restore \`invalidValue: undefined\` on both \`valueProperty\` calls.

## Test plan
- [x] Added \`should preserve colorScale fill on highlighted marker\` snapshot test in mapMarkerSeries.test.ts.
- [x] Added \`AG-17195 should not warn when colorKey resolves to null/undefined/missing\` to all three map-series test files (asserts \`expectWarningsCalls().toEqual([])\`).
- [x] Updated three \`expectWarningsCalls\` inline snapshots in rangeArea.test.ts from a non-empty warning list to \`[]\`, asserting the regression's absence.
- [x] \`yarn nx test ag-charts-enterprise\` — full suite green (2001 pass, 1714 snapshots).
- [x] \`yarn nx lint\`, \`yarn nx build:types\`, \`yarn nx format\` clean.

Fix #AG-17196
Fix #AG-17195
Fix #AG-17095


iMoses (approved):
LGTM

PR #6788, opened 2026-05-08
Add runtime deprecation warning for `colorRange`

alantreadway:
## Summary

`colorRange` has been JSDoc-deprecated since v13.3.0 on `heatmap`, `mapMarker`, `mapShape`, `sunburst`, and `treemap`, but no runtime warning was emitted — users had no signal at runtime that they were on a deprecated path. This PR adds one, scoped strictly to user-supplied values.

- New `deprecated()` validator wrapper alongside `enterprise()` / `undocumented()` in `validation.ts`. It emits `Option \`<path>\` is deprecated. <message>` once per session via `warnOnce`, then falls through to the inner validator so the option keeps working.
- New `silentAdvisories` flag threaded through `validate()` (and through the per-module `validate` overrides on cartesian/polar chart modules and the mapMarker module). The post-theme validation passes in `optionsModule.slowSetup()` set it to `true`, so theme-injected `colorRange` values do not trip the warning — only genuine user input does.
- `colorRange` wrapped in `deprecated()` for the five public option defs: heatmap (series-level def) and the four themeable defs (mapMarker, mapShape, sunburst, treemap).
- Test/example inputs that exercised the deprecated path migrated to `colorScale.fills` so they stop tripping `setupMockConsole`. These are user-style call sites and the migration mirrors what end users will do (`heatmap.test.ts`, `highlight.test.ts`, `gradientLegend.test.ts`, `chart/series/test/examples.ts`, `agChartsServerSide.test.ts`, `simple-sunburst/main.ts`, plus the bundle-render and server-side-package test scenarios).

## Out of scope (deferred)

Internal usages of `colorRange` in theme templates and series fallback logic are intentionally left untouched here — too risky for the current release. A follow-up PR (new JIRA, fixVersion 36.0.0) will migrate:
- Theme templates: heatmap / sunburst / treemap / mapMarker / mapShape modules.
- Series-level fallback paths in `heatmapSeries.ts` and `hierarchySeries.ts` to derive fallback colours from `colorScale.fills` instead of `properties.colorRange`.
- Gallery / docs examples and the `treemap-series` mdoc references.

## Test plan

- [x] `yarn nx format`
- [x] `yarn nx lint ag-charts-core ag-charts-community ag-charts-enterprise`
- [x] `yarn nx test ag-charts-core` — 494 passed
- [x] `yarn nx test ag-charts-community` — 3349 passed
- [x] `yarn nx test ag-charts-enterprise` — 2000 passed (no failures)
- [x] `yarn nx test ag-charts-server-side` — 40 passed
- [ ] Verify on docs site that user-supplied `colorRange` triggers a single deprecation warning per session and the chart still renders identically.

### DIFF
--- a/packages/ag-charts-community/src/chart/cartesianChartModule.ts
+++ b/packages/ag-charts-community/src/chart/cartesianChartModule.ts
@@ -20,7 +20,7 @@ export const CartesianChartModule: ChartModuleDefinition<AgCartesianChartOptions
     create(options: ChartOptions, resources?: TransferableResources) {
         return new CartesianChart(options, resources);
     },
-    validate(options: any, optionsDefs, path) {
+    validate(options: any, optionsDefs, path, opts) {
         const additionalErrors: ValidationError[] = [];
         if (options?.series?.[0]?.type === 'histogram') {
             if (Object.values(options?.axes ?? {}).some(invalidHistogramAxis)) {
@@ -37,7 +37,7 @@ export const CartesianChartModule: ChartModuleDefinition<AgCartesianChartOptions
             }
         }
 
-        const result = validate(options, optionsDefs, path);
+        const result = validate(options, optionsDefs, path, opts);
         result.invalid.push(...additionalErrors);
         return result;
     },
--- a/packages/ag-charts-core/src/state/validation.ts
+++ b/packages/ag-charts-core/src/state/validation.ts
@@ -22,6 +22,7 @@ const requiredSymbol = Symbol('required');
 const markedSymbol = Symbol('marked');
 const undocumentedSymbol = Symbol('undocumented');
 const enterpriseSymbol = Symbol('enterprise');
+const deprecatedSymbol = Symbol('deprecated');
 export const unionSymbol = Symbol('union');
 
 const similarOptionsMap = [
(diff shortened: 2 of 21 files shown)


packages/ag-charts-community/src/chart/cartesianChartModule.ts
```diff
@@ -20,7 +20,7 @@ export const CartesianChartModule: ChartModuleDefinition<AgCartesianChartOptions
     create(options: ChartOptions, resources?: TransferableResources) {
         return new CartesianChart(options, resources);
     },
-    validate(options: any, optionsDefs, path) {
+    validate(options: any, optionsDefs, path, opts) {
```
iMoses, line 23: I'll comment once but it applies to all similar cases.
How about we name it `params` instead? Since there's already `options` and `optionsDefs` which means something different?

packages/ag-charts-core/src/state/validation.ts
```diff
@@ -77,6 +79,17 @@ export interface ValidatorResult extends ValidationResult<any> {
...
+    /**
+     * When true, validators that emit advisory warnings (e.g. `deprecated()`) should stay silent.
+     * Used during post-theme passes where the values originate from theme expansion rather than
+     * the user — emitting a deprecation warning for a theme-injected value would be a false
+     * positive.
+     */
+    silentAdvisories?: boolean;
+}
+
+export interface ValidateOptions {
```
iMoses, line 91: Same here. `ValidateParams`

iMoses (approved):
LGTM

PR #6790, opened 2026-05-08
AG-17296 Rename `ValidateOptions` to `ValidateParams`

alantreadway:
https://ag-grid.atlassian.net/browse/AG-17296

## Summary

Address reviewer feedback from #6788 (which landed before the feedback could be applied).

In the validation code path `options` already refers to the chart options being validated and `optionsDefs` to the validator definitions, so a third `Options`-named symbol was confusing. Rename the interface introduced in #6788 to `ValidateParams` and the corresponding `opts` parameter to `params` everywhere it was added.

Pure rename — no behaviour change.

> **Note:** Tracked under AG-17296 (the broader "migrate internal `colorRange` usages" follow-up) since this is a small ancillary fix and not its own ticket. Not using `Fix #AG-17296` because this PR does not complete that ticket's scope.

## Test plan

- [x] `yarn nx format:check`
- [x] `yarn nx run-many -t build:types -p ag-charts-community,ag-charts-enterprise,ag-charts-core`
- [x] `yarn nx run-many -t lint -p ag-charts-community,ag-charts-enterprise,ag-charts-core`
- [ ] CI green


iMoses (approved):
LGTM

PR #6801, opened 2026-05-11
Fix selection examples breaking in Angular and React [CRT-1105]

AG-DavidG:
https://ag-grid.atlassian.net/browse/CRT-1105

## Summary
- **selection-basic**: `updateStatus()` called from the `selectionChange` listener was placed as a class method in Angular but referenced as a bare function — `ReferenceError: updateStatus is not defined` on drag/click. Inlined the logic into the listener.
- **click-modes**: `updateSubtitle()` helper was placed at module scope by the generator (not HTML-referenced), but it accessed `options`/`chart` which are component-scoped. Broke both Angular and React when clicking any button. Inlined into each handler.

## Test plan
- [ ] Verify selection-basic example works in Angular (drag to select, status text updates)
- [ ] Verify click-modes example works in Angular (click mode/click-away buttons update subtitle)
- [ ] Verify click-modes example works in React (same)
- [ ] Verify vanilla TypeScript versions still work
- [ ] `yarn nx validate-examples` passes (confirmed locally — 838 examples, no issues)

### DIFF
--- a/packages/ag-charts-website/src/content/docs/selection/_examples/click-modes/main.ts
+++ b/packages/ag-charts-website/src/content/docs/selection/_examples/click-modes/main.ts
@@ -40,19 +40,18 @@ const options: AgCartesianChartOptions = {
 
 const chart = AgCharts.create(options);
 
-function updateSubtitle() {
+function setClickMode(value: AgSelectionClickMode) {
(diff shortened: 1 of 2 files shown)


alantreadway (approved):
LGTM

PR #6802, opened 2026-05-11
Fix gradient legend sliders producing NaN in framework examples [CRT-1109]

AG-DavidG:
https://ag-grid.atlassian.net/browse/CRT-1109

## Summary
- Gradient legend example sliders (`setThickness`, `setLength`, `setPadding`) used `oninput="handler(this.value)"` in HTML
- In vanilla JS, `this` is the input element — works fine
- In React, the generator produces `onInput={() => handler(this.value)}` where `this` in an arrow function is `undefined` → `Number(undefined)` = `NaN`
- In Angular, the generator produces `(input)="handler(this.value)"` where `this` is the component instance → `this.value` is `undefined` → `NaN`
- Fix: switch to `oninput="handler(event)"` and extract `(event.target as HTMLInputElement).value` in the function — the standard pattern used by all other range input examples

## Test plan
- [ ] Verify gradient legend sliders work in Angular (thickness/length/padding update without NaN)
- [ ] Verify gradient legend sliders work in React (same)
- [ ] Verify vanilla JS version still works
- [ ] `yarn nx validate-examples` passes (confirmed locally)

### DIFF
--- a/packages/ag-charts-website/src/content/docs/colour-scale/_examples/gradient-legend/index.html
+++ b/packages/ag-charts-website/src/content/docs/colour-scale/_examples/gradient-legend/index.html
@@ -7,11 +7,11 @@
         <button onclick="setPosition('top')">Top</button>
     </div>
     <div class="controls-row">
-        Thickness: <input type="range" min="5" max="60" value="16" step="1" oninput="setThickness(this.value)" />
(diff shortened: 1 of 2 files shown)


alantreadway (approved):
LGTM

PR #6803, opened 2026-05-11
Fix tooltip symbol colour in horizontal range bar gallery example

AG-DavidG:
[CRT-1106](https://ag-grid.atlassian.net/browse/CRT-1106)

## Summary
- Set `symbol.marker.fill` in the tooltip renderer return to match the `itemStyler` colour logic (green for positive growth, red for negative), so the tooltip symbol reflects the bar colour instead of the default series blue.

## Test plan
- [ ] Open the horizontal-range-bar-with-labels gallery example
- [ ] Hover bars with positive growth — tooltip symbol should be green (#10b981)
- [ ] Hover bars with negative growth — tooltip symbol should be red (#ef4444)


alantreadway (approved):
LGTM

PR #6804, opened 2026-05-11
CRT-1087 Fix pie/donut selection stroke default

olegat:
https://ag-grid.atlassian.net/browse/CRT-1087
https://ag-grid.atlassian.net/browse/CRT-1089 (duplicate; maps)

This stroke-palette is already set elsewhere in the options-graph. Pie/donut also has existing customisation to handle it's special-case.

Just setting the strokeWidth default to 2 is sufficient to show the stroke with the default palette color.

Tested on donut, bar, line.


alantreadway (approved):
LGTM

PR #6809, opened 2026-05-11
CRT-1091 Fix org inner and outer spacing

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-1091


packages/ag-charts-enterprise/src/series/organization/__image_snapshots__/organization-series-test-ts-organization-series-layout-should-use-outer-spacing-for-cousin-gaps-and-inner-spacing-for-sibling-gaps-1-snap.png
lsjroberts, line 1: Change here verifies that the spacing between Child F and Child H does not increase with each child of Child H, I, J etc.

alantreadway (approved):
LGTM

PR #6810, opened 2026-05-11
CRT-1107: Navigator - Handles retain keyboard focus style even when n…

iMoses:
…ot focused

https://ag-grid.atlassian.net/browse/CRT-1107


alantreadway (approved):
LGTM

PR #6818, opened 2026-05-11
CRT-1117 Drop saved zoom when axis type changes on chart update

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-1117

`chart.update()` that swaps an axis to a different scale type (e.g. category → number) used to leave the chart blank: ZoomManager preserved the saved `{min,max}` ratio against the new axis, and the downstream `preserveDomain` strategy then reinterpreted the captured visible range against the populated domain, collapsing the x zoom to `{0,0}`.

`ZoomManager.setAxes` now snapshots each axis's `type` before the swap and drops any saved entry whose type changed across the swap, so the new axis starts fully zoomed-out and `preserveDomain` no longer captures stale state.

## Notes for review

- Behaviour change for the `preserveRatios` strategy on a type swap: previously the original ratio survived (geometrically meaningless across a unit change); it now resets to `{0,1}` like every other strategy. `preserveDomain` (the default) is unaffected on axis-id reuse without a type change.
- Regression test added to the existing `data changes during zoom` suite in `zoom.test.ts`; full zoom suite (66 tests, 67 snapshots) passes.

Fix #CRT-1117

### DIFF
--- a/packages/ag-charts-community/src/chart/interaction/zoomManager.ts
+++ b/packages/ag-charts-community/src/chart/interaction/zoomManager.ts
@@ -53,6 +53,7 @@ export type SimpleAxis = {
 };
 
 export type CartesianAxisLike = SimpleAxis & {
+    type: string;
     visibleRange: [number, number];
     scale: Scale<any, any>;
     range: [number, number];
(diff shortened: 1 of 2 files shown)


lsjroberts (approved):
LGTM

PR #6827, opened 2026-05-12
CRT-1118 Do not open expandable annotation-buttons with Arrow Keys

olegat:
https://ag-grid.atlassian.net/browse/CRT-1118

This is a regression due to AG-15837

### DIFF
--- a/packages/ag-charts-community/src/components/toolbar/toolbarButtonWidget.ts
+++ b/packages/ag-charts-community/src/components/toolbar/toolbarButtonWidget.ts
@@ -101,9 +101,14 @@ export class ToolbarButtonWidget extends ButtonWidget {
     private lastTooltip?: string;
     private arrowKeyPredicate: KeyboardClickBindingPredicate = falsePredicate;
 
-    constructor(private readonly localeManager: LocaleManager) {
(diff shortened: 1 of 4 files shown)


packages/ag-charts-community/src/components/toolbar/toolbarButtonWidget.ts
```diff
@@ -101,9 +101,14 @@ export class ToolbarButtonWidget extends ButtonWidget {
...
     private arrowKeyPredicate: KeyboardClickBindingPredicate = falsePredicate;
 
-    constructor(private readonly localeManager: LocaleManager) {
+    constructor(
+        private readonly localeManager: LocaleManager,
+        opensWithArrowKey: boolean
+    ) {
         super();
-        this.addKeyboardClickBinding((ev) => this.arrowKeyPredicate(ev));
+        if (opensWithArrowKey) {
```
olegat, line 109: key-change:
* Defaults to `true` on all annotation buttons (to avoid risk of breakage).
* Set to `false` for buttons in the annotation-options toolbar.

alantreadway (approved):
LGTM

PR #6832, opened 2026-05-12
CRT-1117 Skip preserveDomain snapshot for fully zoomed-out axes

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-1117


olegat (approved):
LGTM.

The expectation is to use always fall back to `'preserveRatios'` when x zoom is `[0,1]`, regardless of the `strategy`. See https://ag-grid.atlassian.net/browse/AG-8627?focusedCommentId=94189

Might make more sense to update the `performUpdateStrategy()` to set `strategy` to `'preserveRatios'` instead of tweaking `performPreserveDomain()` so that the we don't call `resetZoom()` when `strategy: 'reset'`, but that's not so important right now.

PR #6837, opened 2026-05-13
AG-17296 Remove deprecated `colorRange` option for 14.0.0

alantreadway:
## Summary

- Migrate internal theme templates (heatmap, treemap, sunburst, mapMarker, mapShape) from `colorRange` to `colorScale.fills` via `$map` of the `divergingColors` palette
- Migrate series fallback paths to derive colours from `colorScale.fills` instead of the removed `colorRange` property
- Remove `colorRange` from `ag-charts-types` public option interfaces (5 series types + navigator ignored properties)
- Remove `deprecated()` validator, `silentAdvisories` machinery, and post-theme suppression calls — no remaining consumers
- Update treemap-series and sunburst-series docs to reference `colorScale.fills`

bubble/scatter/map-line `colorRange` (undocumented) is unaffected.

## Test plan

- [x] `build:types` passes across ag-charts-types, ag-charts-core, ag-charts-community, ag-charts-enterprise
- [x] `lint` passes with 0 errors across all packages
- [x] 3353 community tests pass (1617 visual snapshots unchanged)
- [x] 2001 enterprise tests pass (1712 visual snapshots unchanged)
- [ ] CI passes


packages/ag-charts-core/src/state/validation.ts
```diff
@@ -385,29 +374,6 @@ export function enterprise<T extends Validator | OptionsDefs<any>>(validatorOrDe
...
 
-/**
- * Marks an option as deprecated. Supplied values still pass through to the inner validator (so the
- * option remains functional during the deprecation window), but a one-shot warning is emitted via
- * `warnOnce` to nudge consumers toward the replacement API. The provided `message` should describe
- * the recommended migration (e.g. "Use `colorScale.fills` instead.").
- */
-export function deprecated(validator: Validator, message: string): Validator;
-export function deprecated<T extends OptionsDefs<any>>(validatorOrDefs: T, message: string): T;
-export function deprecated<T extends Validator | OptionsDefs<any>>(validatorOrDefs: T, message: string) {
```
lsjroberts, line 396: Do we want to keep this for future deprecations?

packages/ag-charts-community/src/chart/series/hierarchy/hierarchySeries.ts
```diff
@@ -254,7 +254,7 @@ export abstract class HierarchySeries<
         const colorDomain = [minColor, maxColor];
 
         const dataDomain: [number, number] = minColor < maxColor ? [minColor, maxColor] : [0, 1];
-        configureColorScale(this.colorScale, this.properties.colorScale, dataDomain, colorRange ?? ['black']);
+        configureColorScale(this.colorScale, this.properties.colorScale, dataDomain, []);
```
lsjroberts: Do we need this 4th param anymore? Every call to it here just passes `[]`, and the other instances in `bubbleSeries.ts` and `mapLineSeries.ts` probably need to be updated too?

packages/ag-charts-core/src/state/validation.ts
alantreadway: Removed intentionally — it's ~20 lines and easy to re-create when needed. Keeping it would also mean keeping the `silentAdvisories` machinery (which `deprecated()` depends on to suppress warnings for theme-injected values). No other deprecations are pending, so YAGNI for now.

packages/ag-charts-community/src/chart/series/hierarchy/hierarchySeries.ts
alantreadway: Good catch — went ahead and aligned all series. Second commit removes `colorRange` from bubble/scatter/mapLine too and drops the `fallbackRange` parameter from `configureColorScale` entirely. The `$map` theme expression survives partial `colorScale` overrides (tested in new cases added to `colorScaleMissingData.test.ts`), so the dual-path workaround documented in `bubbleSeriesModule.ts` was unnecessary. All 5367 tests pass.

lsjroberts (approved):
LGTM

PR #6838, opened 2026-05-13
AG-3390 - Add react to dev tools client:only

taktran:
To align with https://github.com/ag-grid/ag-studio/pull/1424


markdurrant (approved):
LGTM 👍

PR #6844, opened 2026-05-14
AG-13954 Update zoom initial state options for string categories

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13954


alantreadway (approved):
LGTM

PR #6857, opened 2026-05-15
Capture frame-cadence and GC stats per benchmark test case

alantreadway:
https://ag-grid.atlassian.net/browse/AG-16452


lsjroberts (approved):
LGTM

PR #6861, opened 2026-05-16
ci: run e2e shards inside Playwright container

alantreadway:
Stacked on #6860. Goal: minimise e2e shard spin-up overhead without increasing E2E execution time.

## Summary

- Replaces the dedicated `playwright_image` job (pull image → save tarball → upload as `actions/cache`) and the per-shard `Restore Playwright Image Cache` + `docker load` steps with a job-level `container: mcr.microsoft.com/playwright:v1.60.0-jammy` on the e2e job. GHA pulls the image at runner setup, so each of the 32 shards saves the docker load step on the critical path and the workflow saves a whole job.
- Adds an `--in-runner` mode to `playwright.sh` that runs astro + playwright inside the same container; the existing `--host` path is preserved for local dev.
- Drops `playwright_image` references from the `report` job (status, slack, needs list) and frees ~1.5 GB of cache budget (768 MB tarball × 2 refs).

## Test plan

- [ ] PR-level CI run completes under 10m wall-clock.
- [ ] All 32 e2e shards pass.
- [ ] Shard-level startup time drops vs the parent branch (visible in `Setup` → `nx test:e2e` step timing).
- [ ] Local `./playwright.sh --host test ...` still works (legacy path untouched).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -482,6 +482,12 @@ jobs:
               with:
                   name: e2e-init-outputs
 
+            - name: Allow Git in Container
+              # The workspace is owned by the host runner user but the container
+              # runs as root, which trips git's dubious-ownership check on every
+              # subsequent git invocation. Whitelist the workspace globally.
+              run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
+
             - name: Setup
               id: setup
               uses: ./.github/actions/setup-nx
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -456,47 +456,15 @@ jobs:
                       reports/
                       packages/**/__diff_output__/*
 
-    playwright_image:
-        runs-on: ubuntu-latest
-        name: Playwright Image
-        needs: [detect-changes]
-        if: needs.detect-changes.outputs.run_full_ci == 'true'
-        env:
-            PLAYWRIGHT_IMAGE: mcr.microsoft.com/playwright:v1.60.0-jammy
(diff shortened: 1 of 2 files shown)


alantreadway:
Closing — the container-runner approach is blocked by a fundamental incompatibility.

**Diagnosis:** `actions/cache` caches written by host-runner jobs are not readable by container-runner jobs (and vice versa). The cache *key* matches, but the cache *version* (an internal hash derived from the action's runtime environment) differs between host and in-container execution. Concrete evidence from this run's logs:

- Init job (host): `Cache hit for: node_modules-Linux-25819d152...`
- E2E shard (in container, same workflow run, same ref, same key): `Cache not found for input keys: node_modules-Linux-25819d152..., node_modules-Linux-`

With no `node_modules/` available, `npx astro dev` pulled astro afresh, failed to find the project's `@astrojs/markdoc` plugin, and the e2e step timed out at 15m. All 32 shards failed the same way.

**Secondary finding worth keeping:** image pull from MCR took ~25s, which is on par with the prior `actions/cache restore` + `docker load`. The image-cache dance wasn't actually saving meaningful critical-path time — it was just costing us ~1.5 GB of the 10 GB cache budget.

**Follow-up option to consider in a separate PR (not this one):** delete the `playwright_image` job entirely and replace per-shard `docker save/load + actions/cache` with an inline `docker pull` at the start of each e2e shard. Same wall-clock cost, one fewer job, ~1.5 GB cache freed, ~80 lines of YAML deleted. Does not require touching node_modules caching.

PR #6870, opened 2026-05-19
AG-17310 Add public-facing grid compatible color ref operation

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-17310


alantreadway (approved):
LGTM

PR #6873, opened 2026-05-20
AG-17390 Align org-chart formatter types with public RichFormatter

iMoses:
https://ag-grid.atlassian.net/browse/AG-17390

The public option `series[].node.title.formatter` (and analogous `subtitle` / `labels`) is declared in `ag-charts-types` as `RichFormatter<...>` — returning `TextOrSegments | undefined`. The internal `OrganizationSeriesNodeTextProperties` and the `formatText` helper in `organizationSeries.ts` declared the same field as `Formatter<...>`, returning only `TextValue | undefined`.

Runtime already supports `TextOrSegments` (segments are passed through `node.update` and rendered correctly); the internal types were just under-reporting the supported behaviour. Audited every other `RichFormatter` site in `ag-charts-types` (`AgChartLabelOptions`, axis label/caption, navigator label) — they consume via classes that already use `RichFormatter`, so no other drift to fix.

Fix #AG-17390


alantreadway (approved):
LGTM

PR #6903, opened 2026-05-26
AG-17394 Updated dark mode colors for docs example

stgrgic:
Updated dark mode colors for the Org chart example


AG-DavidG (approved):
LGTM

PR #6931, opened 2026-05-28
AG-17227 Fix sparkline context aliasing across cache hits

alantreadway:
https://ag-grid.atlassian.net/browse/AG-17227

Unblocks the optimised creation path for Grid sparkline cells, and fixes a latent context/container aliasing bug exposed by it.

- `optionsStructuralCache.ts` / `optionsModule.ts` — strip `container` and `context` from cached `processedOptions` on write and re-attach from `userOptions` on hit, matching the existing pattern for `data` (`VOLATILE_KEYS`).
- `optionsModule.ts` — add `context` to `FAST_PATH_OPTIONS` so per-row context deltas don't force slow setup on every cell refresh.
- `sparkline.ts` preset — stop capturing `context` in the memoised tooltip wrapper (read from `params.context` instead), emit user-supplied `context` onto `chartOpts`, and include `xValue` in the default tooltip content when `datumKey !== 'y'` so consumers don't need to install a renderer purely for that.

Companion Grid PR: ag-grid/ag-grid#13874. Land this first.

Fix #AG-17227

### DIFF
--- a/packages/ag-charts-community/src/api/agCharts.test.ts
+++ b/packages/ag-charts-community/src/api/agCharts.test.ts
@@ -87,9 +87,8 @@ describe('AgCharts', () => {
             height: sparklineOptions.height! + 50,
             data: sparklineOptions.data!.toReversed(),
             container: () => getDocument().createElement('div'),
-            // `context` change must take the fast path — the Grid sparkline cell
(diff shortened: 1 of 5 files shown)


iMoses (approved):
LGTM

PR #6938, opened 2026-05-29
AG-17306: fix polar-crosslines, restore theme defaults in TS/ESM mode

iMoses:
## Problem

Polar cross-lines worked with the JavaScript (UMD) bundle but were invisible in TypeScript/ESM and all framework builds. The chart loaded without errors and `PolarCrossLinesModule` was registered, but no crosslines appeared.

## Root cause

`sanitizeThemeModulesUncached` in `processModuleOptions.ts` prunes theme defaults for unregistered modules. It tracked missing axis-plugin modules by **name** — so when `CrossLinesModule` (name: `'crossLines'`) was absent, `'crossLines'` was added to the prune set and deleted from every polar axis theme config. That deleted the `$apply` block that `PolarCrossLinesModule` had written there, which is what injects `stroke`, `fill`, `strokeWidth`, and `enabled: true`.

The user's `crossLines: [...]` options survived (user options are not affected), so instances were created — but with `stroke: undefined`, `fill: undefined` they rendered zero visible pixels.

In UMD, `CrossLinesModule` is bundled and auto-registered, so its name never entered the prune set.

### `processModuleOptions.ts`
Before adding a missing axis-plugin module's optionsKey to the prune set, check whether any *registered* module already covers that same key. `PolarCrossLinesModule` (registered) shares `optionsKey: 'crossLines'` with `CrossLinesModule` (absent), so `'crossLines'` is now skipped and the theme defaults are preserved.

### `module-mappings.mjs` (ESLint rule)
- `PolarCrossLinesModule` was missing from `enterpriseModules`, causing auto-fix to emit `import { PolarCrossLinesModule } from 'ag-charts-community'` — a module that doesn't exist there, breaking the TS import.
- Added `polarAxisPluginToModule` map so the polar-specific lookup is data-driven rather than hardcoded in the rule.

### `modules.json`
Moved `CrossLinesModule` and `PolarCrossLinesModule` from the "Interactivity" section to "Data Elements" in the module selector, matching the docs nav.

### DIFF
--- a/libraries/ag-charts-eslint-rules/rules/module-mappings.mjs
+++ b/libraries/ag-charts-eslint-rules/rules/module-mappings.mjs
@@ -94,6 +94,9 @@ export const axisPluginToModule = new Map([
     ['bandHighlight', 'BandHighlightModule'],
 ]);
 
+// Polar-specific overrides for axisPluginToModule
+export const polarAxisPluginToModule = new Map([['crossLines', 'PolarCrossLinesModule']]);
+
 // Nested series plugin option → Module ID
 export const seriesPluginToModule = new Map([['errorBar', 'ErrorBarsModule']]);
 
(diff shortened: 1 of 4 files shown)


alantreadway (approved):
LGTM

PR #6945, opened 2026-05-29
AG-16608 Let BigInt keys flow through the data pipeline

alantreadway:

Next chunk of the BigInt (AG-16608) stacked PR train. Stacked on PR3b (#6943).

PR3b retained bigint *values* and rendered them at full precision; this PR lets bigint *keys* flow through the data pipeline safely:

- **`SORT_DOMAIN_GROUPS` comparator is now BigInt-safe.** Subtracting two bigint keys yields a bigint that `Array.sort` ToNumber-coerces and throws on. Bigint keys are now compared directly; number/Date keep their existing subtraction-based ordering.
- **Removed the two remaining PR1 key-side drops** (`dataExtractor.extractKeys` and the incremental processor) so bigint keys are retained rather than discarded — matching the existing bigint value support.

Full bigint-key numeric-axis prediction (AC #4) remains deferred to a later PR; ordering/uniqueness tracking simply skips bigint keys for now (no throw).

Histogram bigint bins (AC #13) and stacking rejection on `'date'`-tagged columns (AC #14) follow as separate PRs (PR3d/PR3e) — each is an independent feature that would exceed the per-PR size cap.

### Test plan
- New/updated unit tests in `dataModel.test.ts`: bigint keys retained in the key column; grouped bigint keys sort correctly without throwing.
- Full suites green: community 3375 passed, enterprise 2136 passed.
- `build:types`, lint (0 errors), and format all pass.

Fix #AG-16608

### DIFF
--- a/packages/ag-charts-community/src/chart/data/data-model/extraction/dataExtractor.ts
+++ b/packages/ag-charts-community/src/chart/data/data-model/extraction/dataExtractor.ts
@@ -266,9 +266,7 @@ export class DataExtractor<D extends object, K extends keyof D & string> {
 
                     const result = processKeyValue(data[datumIndex], datumIndex, scope);
 
-                    // bigint *values* are supported now, but bigint *keys* stay dropped until a later PR:
-                    // they break SORT_DOMAIN_GROUPS' comparator and need numeric-axis prediction (AC #4).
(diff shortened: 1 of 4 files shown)


lsjroberts (approved):
LGTM

PR #6946, opened 2026-05-29
AG-16608 Compute histogram bins in BigInt for BigInt columns

alantreadway:

Next chunk of the BigInt (AG-16608) stacked PR train. Stacked on PR3c (#6945). Implements AC #13.

A histogram with a `bigint` x-column previously rendered **empty**: bin boundaries were computed with `Math.floor`/`Math.log10`/`Math.pow` (NaN on bigint) and bigint x-values failed the `isNumber()` bucket gate, so every datum was dropped.

- **New `createBigIntBins(start, stop, count)` in `ag-charts-core`** — reuses the existing `bigIntTickStep`/`ceilToStep`/`floorToStep` helpers (added in PR3b) to produce contiguous, nice-aligned BigInt bins covering the domain.
- **Histogram branches onto it when the x-extent is bigint** (`computeBins`), preserving the existing number/Date paths exactly and the explicit-`bins`/`binCount` precedence.
- Bin boundaries are retained at **full precision** and converted to pixels by the already bigint-safe scale (PR2/PR3b), so values beyond `Number.MAX_SAFE_INTEGER` bucket exactly.
- Widened `CalculatedBin`/`HistogramNodeDatum` `domain` and `createDatumId`'s `KeyType` to accept bigint.

### Test plan
- New core unit tests for `createBigIntBins` (contiguity, nice extension, degenerate domain, spans beyond `Number.MAX_SAFE_INTEGER`).
- New histogram integration test: a `2^53 + 1`-based bigint column produces bigint bin boundaries at full precision, contiguous, with every datum bucketed (not dropped).
- Full suites green: core ticks 22, community 3376, enterprise 2136. `build:types`, lint (0 errors), format all pass.

Fix #AG-16608

### DIFF
--- a/packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
+++ b/packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
@@ -157,9 +157,8 @@ export class HistogramSeries extends CartesianSeries<HistogramSeriesTypes> {
         return this.calculatedBins.length > 0;
     }
 
-    // Resolves the bin boundaries from the user's configuration and the x-domain. A BigInt x-column
-    // (AG-16608) computes its boundaries in BigInt arithmetic for full precision; everything else keeps
-    // the existing number paths. Explicit `bins` win unless an explicit `binCount` is also set.
+    // A BigInt x-column (AG-16608) computes boundaries in BigInt for full precision; everything else
+    // keeps the number paths. Explicit `bins` win unless an explicit `binCount` is also set.
     private computeBins(xExtent: [number | bigint, number | bigint]): BinDomain[] {
         const bigIntExtent = typeof xExtent[0] === 'bigint' || typeof xExtent[1] === 'bigint';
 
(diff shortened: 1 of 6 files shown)


packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
```diff
@@ -88,8 +89,11 @@ const defaultBinCount = 10;
 
 type HistogramAnimationData = CartesianAnimationDataOf<HistogramSeriesTypes>;
 
+/** Bin boundaries are `bigint` for a BigInt x-column (AG-16608) and `number` otherwise. */
+type BinDomain = [number | bigint, number | bigint];
```
lsjroberts, line 93: I guess ideally this would be `[number, number] | [bigint, bigint]`, but probably tricky? Presumably the domains can't be mixed and that is handled by the data processor still?

lsjroberts (approved):
LGTM

alantreadway:

PR #6947, opened 2026-05-29
AG-16608 Reject stacking on a date-tagged column

alantreadway:

Final chunk of the original PR3 group of the BigInt (AG-16608) train. Stacked on PR3d (#6946). Implements AC #14.

Stacking / accumulation is not meaningful for date values. When a group accumulation or normalisation processor targets a column tagged `'date'`, the series is now rejected:

- Warn once: *"... column "X" is date-typed; stacking is not meaningful for date data. The series renders empty; other series in this chart are unaffected."*
- Blank the column and clear its domain so the series renders empty; sibling columns and series are unaffected.

Implemented at the `postProcessGroups` chokepoint, which only runs for stacked/accumulated columns — so no separate "is this stacking?" detection is needed (the prior concern about inferring stacking from prop definitions is avoided). Covers all stacking/normalising series: bar, line, area, radial-column, radial-bar, error-bar.

**Scope note:** single-series accumulated-value processors (`accumulativeValueProperty`, used by waterfall/donut) flow through the generic property-processor path and are out of scope for this PR — date-as-value there is already degenerate. Can be a follow-up if required.

### Test plan
- New dataModel tests: a stacked `'date'` column is tagged `'date'`, blanked to empty domains, and emits the warning once per column (inline-snapshot asserted); a stacked `number` column still stacks correctly (5 + 7 = 12).
- Full suites green: community 3378, enterprise 2136. `build:types`, lint (0 errors), format all pass.

Fix #AG-16608

### DIFF
--- a/packages/ag-charts-community/src/chart/data/data-model/aggregation/aggregator.ts
+++ b/packages/ag-charts-community/src/chart/data/data-model/aggregation/aggregator.ts
@@ -1,4 +1,4 @@
-import { first } from 'ag-charts-core';
+import { Logger, first } from 'ag-charts-core';
 
 import { ContinuousDomain, DiscreteDomain } from '../../dataDomain';
 import type { GroupedData, PropertySelectors, UngroupedData } from '../../dataModelTypes';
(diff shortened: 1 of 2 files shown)


lsjroberts (approved):
LGTM

alantreadway:

PR #6949, opened 2026-05-29
AG-16608 Accept bigint gauge values at runtime

alantreadway:
## Summary

Completes the **runtime** half of AG-16608 AC #11: gauges (radial, linear) accept `bigint` `value`, `scale.min`, `scale.max`, `target.value`, `scale.interval`, colour stops, and segmentation interval at runtime, not just statically.

PR4 widened the gauge public types to `AgNumericValue = number | bigint`, but the runtime still rejected or crashed on `bigint`, so those widened types were lying. This PR makes the runtime honour them.

## Runtime changes

- **`numericValue` validator** (`ag-charts-core`): a dedicated `number | bigint` validator, used only for the gauge bigint-eligible fields in `gaugePreset.ts` (`value`, `scale.min`/`max`, `scale.interval.step`/`values`, `target.value`, `AgGaugeColorStop.stop`, and `segmentation.interval.step`/`values`). The global `number` validator is deliberately left untouched so the rest of the library does not silently start accepting `bigint`.
- **`isComparable` is now bigint-aware**, so the `lessThan`/`greaterThan` comparators correctly validate mixed `number`/`bigint` scale bounds (e.g. `scale.min` < `scale.max`).
- **`stringifyValue` no longer throws on `bigint`** — `JSON.stringify` throws `TypeError: Do not know how to serialize a BigInt`, which would have turned a clean validation rejection into a crash on the error path.
- **`formatLabel` (gauge label) is bigint-safe**: a `bigint` value renders full-precision via `toLocaleString()`; only the decimal-place estimation narrows the bounds to `Number` (it never touches the value text).
- **Label count-up animation**: the tween endpoints narrow to `Number` (intermediate frames are visual only), while the resting frame re-formats from the full-precision value, so `bigint` labels animate without crashing the interpolator and settle at full precision.
- **`scale.interval`**: a bigint `step` narrows to `Number` for the magnitude arithmetic at every tick entry point (`LinearScale.ticks`/`getTickStep` and the radial gauge's `LinearAngleScale.ticks`). Explicit bigint `values[]` flow through full-precision — positioned by the bigint-safe `convert()` and labelled via a new bigint `toLocaleString` branch in `tickFormat`.
- **Colour stops**: a bigint `stop` narrows to `Number` at its single read point in `resolveStopPositions`, so the gauge gradient path (`getColorStops`) no longer throws on bigint/Number mixing. Stops are fractional thresholds where Number is correct.
- **Segmentation interval**: `getSegments` narrows a bigint `step` for boundary arithmetic and sorts explicit `values[]` with a Number comparator, while letting bigint boundaries flow to the bigint-safe scale.

## Precision discipline

- Full-precision (`bigint` preserved): the gauge **value label text** (`toLocaleString`) and explicit **`scale.interval.values`** tick values/labels.
- Narrowed to `Number` (safe — pixels/threshold/decimal estimation only): scale→pixel conversion (bigint-safe in the scale via PR2), the `fractionDigits` estimate, the count-up animation tween, colour-stop positions, and any **custom interval `step`** (scale or segmentation).

## Documented limitation (AC #17)

A **custom interval `step`** — for both `scale.interval` and `segmentation.interval` — narrows to `Number` for the tick/segment magnitude arithmetic. Full precision applies to the auto-stepping path and to explicit `values[]`. This matches the cartesian number-axis behaviour established in the earlier PRs in this train.

## Tests

- Radial + linear gauge bigint tests covering `value`/`scale.min`/`max`/`target.value`, `scale.interval.step`, `scale.interval.values`, colour `stop` values, and `segmentation.interval.step`/`values` — each asserting no-throw and a full-precision value label. The radial value/scale/target case also has an image snapshot.
- `numericValue` accepts `number` and `bigint`, rejects strings/NaN; mixed `number`/`bigint` comparators validate; an out-of-range `bigint` is rejected cleanly (no throw, exercising the `stringifyValue` bigint path).
- Core `stops.test.ts` unit tests: bigint colour stops produce the same normalised gradient as Number stops.

## Links

- JIRA: https://ag-grid.atlassian.net/browse/AG-16608
- Fix #AG-16608
- Stacked on PR4 (`AG-16608/pr4-bigint-public-types`); this completes AC #11's runtime half (the type half is PR4).


packages/ag-charts-core/src/config/gaugePreset.ts
```diff
@@ -157,8 +158,8 @@ export const linearGaugeSeriesOptionsDef: OptionsDefs<AgLinearGaugePreset> = {
             ...fontOptionsDef,
         },
         interval: {
-            values: arrayOf(number),
-            step: number,
+            values: arrayOf(numericValue),
```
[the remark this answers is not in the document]
alantreadway: Resolved by full runtime support (not a revert) in 97ef4cac5e.

`scale.interval.step` now narrows a bigint step to Number for the magnitude arithmetic at every entry point: `LinearScale.ticks` (`Math.abs(Number(interval))`), `LinearScale.getTickStep`, and the radial gauge's `LinearAngleScale.ticks`. A custom interval step is a Number concept (full precision applies to the auto-step path and to explicit `values[]`), per AC #17.

Explicit bigint `scale.interval.values` instead flow through full-precision: positioned by the bigint-safe `convert()` and labelled via a new bigint `toLocaleString` branch in `tickFormat`.

Coverage added: radial + linear gauge tests for bigint `scale.interval.step` and bigint `scale.interval.values`, each asserting no-throw and a full-precision value label.

packages/ag-charts-core/src/config/gaugePreset.ts
```diff
@@ -51,14 +52,14 @@ import {
 const fillsOptionsDef: OptionsDefs<FillsOptions> = {
     fills: and(
         arrayLength(2),
-        arrayOf(optionsDefs<AgGaugeColorStop>({ color: color, stop: number }, '')),
+        arrayOf(optionsDefs<AgGaugeColorStop>({ color: color, stop: numericValue }, '')),
```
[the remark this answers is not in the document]
alantreadway: Resolved by full runtime support (not a revert) in 97ef4cac5e.

A colour stop is a fractional threshold where Number is correct (bigint offers no precision benefit), but a bigint stop mixed with the Number domain previously threw a TypeError in the shared interpolation. The fix narrows the stop to Number at its single read point in `resolveStopPositions` (core `colorScaleUtil.ts`), so both the gauge gradient path (`createConicGradient`/`createLinearGradient` → `getColorStops`) and any other consumer are bigint-tolerant. `colorStopsOrderValidator` only uses relational comparison (no arithmetic), which is already bigint-safe.

Coverage added: radial + linear gauge tests with bigint colour `stop` values, plus core unit tests in `stops.test.ts` confirming bigint stops produce the same normalised gradient as Number stops.

packages/ag-charts-community/src/scale/linearScale.ts
```diff
@@ -21,7 +21,8 @@ export class LinearScale extends ContinuousScale<number> {
 
     protected static getTickStep(start: number, stop: number, ticks: ScaleTickParams<number>) {
         const { interval, tickCount = ContinuousScale.defaultTickCount, minTickCount, maxTickCount } = ticks;
-        return interval ?? tickStep(start, stop, tickCount, minTickCount, maxTickCount);
+        // A custom interval narrows to Number here; the magnitude is a Number concept (AG-16608 AC #17).
```
olegat, line 24: I cannot find `AC #17` in that ticket. Is this a hallucination?

packages/ag-charts-core/src/state/validation.ts
```diff
@@ -479,6 +480,12 @@ export const color = attachDescription(isColor, 'a color string');
 export const date = attachDescription(isValidDateValue, 'a date');
 export const defined = attachDescription(isDefined, 'a defined value');
 export const number = attachDescription(isFiniteNumber, 'a number');
+// Accepts a finite number or a bigint. Reserved for the gauge fields that opt into bigint at runtime
+// (AG-16608 AC #11) — do NOT widen the global `number` validator, or every numeric option accepts bigint.
+export const numericValue = attachDescription(
+    (value): value is number | bigint => isFiniteNumber(value) || typeof value === 'bigint',
+    'a number'
```
olegat, line 487: nit: `'a number or bigint'`

packages/ag-charts-enterprise/src/axes/angle-number/linearAngleScale.ts
```diff
@@ -30,7 +30,8 @@ export class LinearAngleScale extends LinearScale {
         const [d0, d1] = domain;
 
         if (interval) {
-            const step = Math.abs(interval);
+            // A custom interval step is a Number concept (AG-16608 AC #17); narrow a bigint step.
+            const step = Math.abs(Number(interval));
```
olegat, line 34: The spread of assertive bigint -> number conversions is a bit concerning, even if the numerical limits assumptions hold. In the interest of maintainability, as more defensive conversion utility might be a good idea:

```ts
function safeNumber(x: number | bigint): number {
  if (typeof x === 'bigint' && x > BigInt(Number.MAX_SAFE_INTEGER)) {
    /* log/throw error */
  }
  return Number(x);
}
```

packages/ag-charts-community/src/scale/linearScale.ts
alantreadway: https://docs.ag-grid.com/design-decisions/charts/AG-16608-bigint-and-iso8601-data-values#ag-16608--bigint

I'll update the JIRA to make sure the design doc is cross-referenced

olegat: Thanks

packages/ag-charts-enterprise/src/axes/angle-number/linearAngleScale.ts
olegat: Sorry. I meant `MAX_VALUE` instead of `MAX_SAFE_INTEGER`. We actually might want to different utility depending on the use case:

```ts
function safeLosslyNumber(x: number | bigint): number {
  if (typeof x === 'bigint' && x > BigInt(Number.MAX_VALUE)) {
    /* log/throw error */
  }
  return Number(x);
}

function safeLosslessNumber(x: number | bigint): number {
  if (typeof x === 'bigint' && x > BigInt(Number.MAX_SAFE_INTEGER)) {
    /* log/throw error */
  }
  return Number(x);
}
```

alantreadway: Done (PR10, bb0a7ce): added `toFiniteNumber` in ag-charts-core and routed the assertive conversions here through it. Implemented the lossy variant (guards `MAX_VALUE`/`Infinity`); the lossless (`MAX_SAFE_INTEGER`) variant has no consumer since the exact-value paths — tooltips, tick labels, aggregated totals — already retain bigint.

packages/ag-charts-core/src/state/validation.ts
alantreadway: Done in ce43a9f — now reads `'a number or bigint'`.

alantreadway:

PR #6950, opened 2026-05-29
AG-16654 ISO 8601 strings and bigint epochs on time axes

alantreadway:
JIRA: https://ag-grid.atlassian.net/browse/AG-16654

Fix #AG-16654

Stacks on the AG-16608 BigInt train (PR1 -> ... -> PR4b); base is `AG-16608/pr4b-gauge-bigint-runtime`. Opens the second ticket (AG-16654, ISO 8601 on time axes).

## Summary

- **Strict ISO 8601 detection** (`isISO8601`, new internal util in `chart/data/iso8601.ts`): regex + canonical round-trip that catches V8's silent calendar rollover (`2024-02-30` -> `2024-03-01`) and correctly handles timezone-offset forms.
- **Time-axis validation** (`basicTimeCheckDatumValidation`): accepts existing valid time values OR ISO strings; non-ISO strings on a time axis warn once (naming value, row index, accepted-format set). ISO on numeric/color axes stays rejected; ISO on category axes is used as-is.
- **Scale coercion**: `timeScale`, `unitTimeScale`, `discreteTimeScale` and `timeAxis` coerce Date / epoch number / bigint / ISO string to Date at `convert()`. Out-of-range bigint epochs resolve to `NaN` instead of an Invalid Date. Shared `timeValueToNumber` helper added to `ag-charts-core`.
- **Heterogeneous `'date'` tag**: a column mixing Date, ISO string and epoch number/bigint now promotes to `'date'` (resolves the PR1 `TODO(AG-16654, PR5)`); `number<->bigint` mixing without dates is still rejected. Original values are preserved in the column, so `params.xValue` returns the raw ISO string.

## Accepted / rejected ISO 8601 forms

Strict ISO 8601 **extended** format only. The date portion is always required; the time portion is optional, with **optional seconds** and **colon offsets only**.

- **Accepted**: `2024-01-15`, `2024-01-15T10:30Z`, `2024-01-15T10:30`, `2024-01-15T10:30+05:30`, `2024-01-15T10:30:00`, `2024-01-15T10:30:00Z`, `2024-01-15T10:30:00.123Z`, `2024-01-15T10:30:00+05:30`. (Fractional seconds are allowed only when seconds are present.)
- **Rejected**: non-ISO strings (`Dec 2024`, `01/02/2024`, `January 15, 2024`, `2024-01-15 10:30:00`), impossible/normalised dates (`2024-13-01`, `2024-02-30`), non-extended forms (`20240115`, `2024-W03`, `2024-015`), and **basic/short offsets** (`+0530`, `+05`) — rejected by design for consistency with the strict extended-format policy (which already rejects basic-format dates).

## Tests

- `isISO8601` unit test (accepted/rejected sets per AC #3, including minute-precision and basic-offset rejection, plus offset round-trip).
- Time-scale coercion tests across all three scales (ISO + bigint parity with Date, out-of-range bigint -> NaN, non-ISO -> NaN).
- DataModel integration: ISO accepted + preserved on time axis, non-ISO warns, `'date'` promotion, ISO rejected on numeric, ISO used as-is on category.

## Out of scope

Axis-type prediction (#6) and mixed-timezone diagnostics (#9) are PR6.

### DIFF
--- a/packages/ag-charts-community/src/chart/axis/timeAxis.ts
+++ b/packages/ag-charts-community/src/chart/axis/timeAxis.ts
@@ -101,14 +101,14 @@ export class TimeAxis<TOptions extends NormalisedTimeAxisOptions = NormalisedTim
     }
 
     override datumFormatParams(
-        value: number | Date,
+        value: number | bigint | string | Date,
         params: FormatDatumParams,
         _fractionDigits: number | undefined,
         timeInterval: AgTimeInterval | AgTimeIntervalUnit | undefined,
         style: DateFormatterStyle
     ): FormatterParams<any> {
-        if (typeof value === 'number') {
-            value = new Date(value);
+        if (!(value instanceof Date)) {
+            value = new Date(typeof value === 'bigint' ? Number(value) : value);
         }
 
         if (timeInterval == null) {
--- a/packages/ag-charts-community/src/chart/data/iso8601.ts
+++ b/packages/ag-charts-community/src/chart/data/iso8601.ts
@@ -3,15 +3,12 @@
 const ISO_8601 = /^\d{4}-\d{2}-\d{2}(T\d{2}:\d{2}(:\d{2}(\.\d+)?)?(Z|[+-]\d{2}:\d{2})?)?$/;
 
 /**
- * Returns true only for strings that are valid, canonical ISO 8601 date / date-time values.
+ * Returns true only for valid, canonical ISO 8601 date / date-time strings.
  *
- * The regex alone is insufficient: V8 silently rolls impossible calendar dates such as `"2024-02-30"`
- * forward to `"2024-03-01"`. We round-trip the leading `YYYY-MM-DD` (which the regex guarantees is
- * present) through `Date`: parsing the date portion alone fixes it at UTC midnight, so re-deriving the
- * calendar components in UTC and requiring them to match catches any silent rollover. Parsing the date
- * portion in isolation deliberately ignores any time-zone offset — comparing UTC components against the
- * full instant would wrongly reject valid offset forms such as `"2024-01-15T02:00:00+05:30"`, which land
- * on the previous UTC day.
+ * The regex alone is insufficient — V8 silently rolls impossible dates (`"2024-02-30"` → `"2024-03-01"`),
+ * so we round-trip the leading `YYYY-MM-DD` through `Date` (UTC midnight) and require the calendar
+ * components to match. The date portion is parsed in isolation so an offset form landing on the previous
+ * UTC day (e.g. `"2024-01-15T02:00:00+05:30"`) is not wrongly rejected.
  */
 export function isISO8601(value: unknown): value is string {
     if (typeof value !== 'string' || !ISO_8601.test(value)) return false;
--- a/packages/ag-charts-community/src/chart/data/iso8601.ts
+++ b/packages/ag-charts-community/src/chart/data/iso8601.ts
@@ -1,0 +1,27 @@
+// Strict ISO 8601 detection. String parsing is a time-axis concern, not a general one, so this lives
+// alongside the data pipeline rather than in the shared value utilities.
+const ISO_8601 = /^\d{4}-\d{2}-\d{2}(T\d{2}:\d{2}:\d{2}(\.\d+)?(Z|[+-]\d{2}:\d{2})?)?$/;
+
+/**
+ * Returns true only for strings that are valid, canonical ISO 8601 date / date-time values.
+ *
+ * The regex alone is insufficient: V8 silently rolls impossible calendar dates such as `"2024-02-30"`
+ * forward to `"2024-03-01"`. We round-trip the leading `YYYY-MM-DD` (which the regex guarantees is
+ * present) through `Date`: parsing the date portion alone fixes it at UTC midnight, so re-deriving the
+ * calendar components in UTC and requiring them to match catches any silent rollover. Parsing the date
+ * portion in isolation deliberately ignores any time-zone offset — comparing UTC components against the
+ * full instant would wrongly reject valid offset forms such as `"2024-01-15T02:00:00+05:30"`, which land
+ * on the previous UTC day.
+ */
+export function isISO8601(value: unknown): value is string {
+    if (typeof value !== 'string' || !ISO_8601.test(value)) return false;
+    if (Number.isNaN(new Date(value).getTime())) return false;
+    const datePart = value.slice(0, 10);
+    const d = new Date(datePart);
+    if (Number.isNaN(d.getTime())) return false;
+    const y = d.getUTCFullYear();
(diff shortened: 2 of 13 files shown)


packages/ag-charts-community/src/chart/data/iso8601.ts
```diff
@@ -0,0 +1,27 @@
+// Strict ISO 8601 detection. String parsing is a time-axis concern, not a general one, so this lives
+// alongside the data pipeline rather than in the shared value utilities.
```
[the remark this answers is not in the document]
alantreadway: Fixed in 08dce53144. The seconds component is now optional in the regex (`T\d{2}:\d{2}(:\d{2}(\.\d+)?)?...`), so minute-precision forms like `2024-01-15T10:30Z`, `2024-01-15T10:30`, and `2024-01-15T10:30+05:30` are accepted; fractional seconds remain gated behind seconds. The leading `YYYY-MM-DD` round-trip canonical check is unchanged and still holds for the no-seconds forms. Added those forms to the accepted test set.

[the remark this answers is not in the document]
alantreadway: By design — basic/short offsets (`+0530`, `+05`) stay rejected. This is deliberate and consistent with the strict extended-format policy already in effect (the detector also rejects basic-format dates such as `20240115`): we accept only the ISO 8601 *extended* format with colon offsets (`+HH:MM`). Accepting a mix of basic and extended offset forms while rejecting basic dates would be inconsistent. Locked the policy in with `2024-01-15T10:30:00+0530` and `2024-01-15T10:30+05` added to the rejected test set (08dce53144).

packages/ag-charts-core/src/utils/time/timeFormatDefaults.ts
```diff
@@ -7,6 +7,19 @@ export function dateToNumber(value: any) {
     return value instanceof Date ? value.getTime() : value;
 }
 
+/**
+ * Coerces a time-axis value (Date, epoch number/bigint, or ISO 8601 string) to epoch milliseconds.
+ * Returns `NaN` for out-of-range bigint epochs and any value `Date` cannot parse, so callers can treat
+ * the result like any other off-domain time value rather than letting an Invalid Date propagate.
+ */
+export function timeValueToNumber(value: Date | number | bigint | string): number {
+    if (typeof value === 'number') return value;
```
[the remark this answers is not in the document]
alantreadway: By design — `timeValueToNumber` stays lenient. Strict ISO 8601 enforcement lives at the validation layer (`basicTimeCheckDatumValidation`), which rejects non-ISO strings before they ever reach a scale. By the time a value hits `convert()` it has already been validated, so the scale assumes validated input. Adding a per-`convert()` regex would be a hot-path cost (one regex per data point per render) and risks regressing legitimate internal Date-string uses that don't flow through the validation layer. Keeping the strictness in one place (validation) and the scale coercion lenient is the intended separation of concerns.

packages/ag-charts-community/src/chart/data/iso8601.dataModel.test.ts
alantreadway, line 1: Nit: Lets nest this in the dataModel.test.ts test suite

packages/ag-charts-community/src/chart/axis/timeAxis.ts
```diff
@@ -101,14 +101,14 @@ export class TimeAxis<TOptions extends NormalisedTimeAxisOptions = NormalisedTim
...
+        value: number | bigint | string | Date,
         params: FormatDatumParams,
         _fractionDigits: number | undefined,
         timeInterval: AgTimeInterval | AgTimeIntervalUnit | undefined,
         style: DateFormatterStyle
     ): FormatterParams<any> {
-        if (typeof value === 'number') {
-            value = new Date(value);
+        if (!(value instanceof Date)) {
+            value = new Date(typeof value === 'bigint' ? Number(value) : value);
```
olegat, line 111: nit: This `new Date(typeof value === 'bigint' ? Number(value) : value);` pattern is repeated throughout the PR. Consider adding an ag-charts-core wrapper function for that.

packages/ag-charts-community/src/chart/data/iso8601.ts
```diff
@@ -0,0 +1,24 @@
...
+const ISO_8601 = /^\d{4}-\d{2}-\d{2}(T\d{2}:\d{2}(:\d{2}(\.\d+)?)?(Z|[+-]\d{2}:\d{2})?)?$/;
```
olegat, line 3: This regex is sufficient for this release 👍🏻

However, be aware that `Temporal.PlainDate` also allows 6-digits years when prefixed with a plus (+) or minus (-), which this regex won't match. I believe that ISO-8601 also formally allows years with >4 digits but I cannot confirm this because the spec seems to be paywalled.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/PlainDate#rfc_9557_format

packages/ag-charts-community/src/scale/ordinalTimeScale.test.ts
```diff
@@ -554,6 +554,27 @@ describe('OrdinalTimeScale', () => {
...
+            ];
+            scale.range = [0, 300];
+            return scale;
+        }
+
+        it('coerces ISO string and bigint epoch to the same band as the equivalent Date', () => {
+            const scale = bandedScale();
+            const date = new Date(Date.UTC(2024, 1, 1));
+            const expected = scale.convert(date);
+            expect(scale.convert('2024-02-01T00:00:00Z')).toBeCloseTo(expected);
```
olegat, line 573: The `.toBeCloseTo()` assertion feels odd. I would expect the timestamps to be integers and match exactly. Does this test fail with `.toBe()`.

packages/ag-charts-community/src/chart/axis/timeAxis.ts
olegat: Actually, I just noticed that `timeValueToNumber` is added in this PR. That could be used here and in other places where `new Date(typeof value === 'bigint' ? Number(value) : value);` is written.

packages/ag-charts-community/src/scale/ordinalTimeScale.test.ts
alantreadway: You're right — switched to `toBe` (ce43a9f); the bands match exactly.

packages/ag-charts-community/src/chart/axis/timeAxis.ts
alantreadway: Done in ce43a9f — reused `timeValueToNumber` at all four `new Date(bigint ? Number : value)` sites (discrete/unit time scales + time axis).

alantreadway:

PR #6952, opened 2026-05-29
AG-16654 Predict ISO/bigint time axes; warn on mixed timezones

alantreadway:
JIRA: https://ag-grid.atlassian.net/browse/AG-16654

Fix #AG-16654

Stacks on the AG-16608 / AG-16654 bigint + ISO 8601 train (base: `AG-16654/pr5-iso-time-scale`).

## Summary

PR6 of the train. Two features:

1. **Axis-type prediction** (`chart/series/cartesian/util.ts`) — AG-16654 #6, AG-16608 #4
   - Extracted `isTimeNumeric(value) = isNumber(value) || typeof value === 'bigint'`, so a **bigint** under a time-named key (`time`/`timestamp`/`date`/`datetime`) now predicts a time axis (previously bigint keys never predicted).
   - Added an ISO branch: a strict ISO 8601 string under a time-named key predicts a time axis. Reuses the strict `isISO8601` detector from PR5 (no reimplementation). The strict regex rejects `'Jan'`, `'Q1'`, etc., so false-positive risk is near-zero, and there is no behaviour change for non-time-named keys or non-ISO strings.

2. **Mixed-timezone diagnostic** (`chart/data/data-model/extraction/dataExtractor.ts`) — AG-16654 #9
   - When a single date-tagged time column contains both timezone-explicit ISO strings (suffix `Z` or `±HH:MM`) and timezone-implicit ones (no offset), emits one `Logger.warnOnce` per column naming an explicit example + row and an implicit example + row, explaining offset-less points are interpreted in local time. The chart still renders — all values parse.
   - Detection lives in the existing per-column extraction loop (key and value paths), so it runs once per column per pass with zero impact on validation-function identity / request merging.
   - **Asymmetry (intentional):** mixed `number`/`bigint` is rejected (undefined arithmetic); mixed-timezone only warns (defined-but-surprising). Mixed Date + UTC-explicit ISO + epoch numbers do **not** warn — all unambiguous UTC instants.

## Release notes callout

This **reverses the original AG-16654 AC 2** (which stipulated no string-driven prediction). String-driven prediction is now enabled by design review. Please flag for release notes.

## Tests

- Prediction matrix in `util.test.ts` for both `predictCartesianAxis` and `predictCartesianFinancialAxis`: under a time-named key — Date/number/bigint/strict-ISO → time(/ordinal-time); non-ISO string → category; ISO under a non-time key → category.
- Mixed-timezone cases in `iso8601.dataModel.test.ts`: mixed explicit+implicit warns once and still renders; all-explicit / all-implicit / Date+UTC+epoch do not warn.

### DIFF
--- a/packages/ag-charts-community/src/chart/data/data-model/extraction/dataExtractor.ts
+++ b/packages/ag-charts-community/src/chart/data/data-model/extraction/dataExtractor.ts
@@ -114,6 +114,25 @@ function isDateColumnType(type: ColumnValueType): boolean {
     return type === 'date';
 }
 
+// An ISO 8601 string carries an explicit timezone iff it ends with `Z` or a trailing `±HH:MM` offset.
+// Date-only (`2024-01-15`) and offset-less date-times (`2024-01-15T10:30:00`) are timezone-implicit.
+const ISO_8601_EXPLICIT_TZ = /(Z|[+-]\d{2}:\d{2})$/;
+
+/** First-seen examples of explicit- and implicit-timezone ISO strings within a single column. */
+interface TimezoneTracker {
+    explicit: { value: string; index: number } | undefined;
+    implicit: { value: string; index: number } | undefined;
+}
+
+function updateTimezoneTracker(tracker: TimezoneTracker, value: unknown, datumIndex: number): void {
+    if (typeof value !== 'string' || !isISO8601(value)) return;
+    if (ISO_8601_EXPLICIT_TZ.test(value)) {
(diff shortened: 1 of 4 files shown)


packages/ag-charts-community/src/chart/data/iso8601.dataModel.test.ts
alantreadway, line 1: Move into dataModel.test.ts as per PR5 feedback

alantreadway:

PR #6955, opened 2026-05-30
AG-16608 Fix bigint render crashes in area, heatmap and waterfall

alantreadway:
## What

Stacked on #6954 (PR8). A per-family **render smoke test** — added here for community (`bar`, `line`, `area`, `scatter`, `bubble`, `histogram`) and enterprise (`range-bar`, `range-area`, `box-plot`, `heatmap`, `waterfall`) cartesian families — flushed out three `isNegative`-class crashes: number-only math performed on a `bigint` data value, reachable now that the public types (PR4) widened to accept `bigint`.

Only `bar` was render-tested before (via the docs example that caught the original `isNegative` crash). This closes the open question from the AG-16608 plan: *do the other families render large bigint values?* — three did not.

## Crashes fixed

| Family | Site | Cause |
|---|---|---|
| **area** | `areaSeries.ts` `handleDatum` | unary `+` on the cumulative bigint threw; `Number.isFinite(bigint)` also dropped the point |
| **waterfall** | `waterfallSeries.ts` `getSeriesDomain` | `Math.min/max` on the bigint cumulative extent |
| **heatmap** | `colorScaleUtil.ts` `computeColorBins` + `deriveNormalizedStops`, `colorScale.ts` `convert` | bigint colour domain/value mixed with the Number colour maths |

## Approach

Each site narrows `bigint` → `Number` for the **position / colour** arithmetic (finite precision, consistent with how bar positions already work). Raw `bigint` values are untouched on the datum, so tooltips and labels keep full precision. Area validity switches from `Number.isFinite` to the existing `isContinuous` predicate, which accepts `bigint`.

## Verification

- New smoke tests: community 6/6, enterprise 5/5 green.
- No regressions: existing `colorScale`, `colorScaleUtil`, `areaSeries` (328 tests), `heatmap` + `waterfall` (81 tests) suites pass.
- `build:types`, `lint`, `format` clean across core/community/enterprise.

### DIFF
--- a/packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
+++ b/packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
@@ -1065,8 +1065,10 @@ export class AreaSeries extends CartesianSeries<AreaSeriesTypes> {
 
         scratch.datum = ctx.rawData[datumIndex];
         scratch.yDatum = ctx.yRawValues[datumIndex];
-        scratch.yCumulative = +ctx.yCumulativeValues[datumIndex];
-        scratch.validPoint = Number.isFinite(scratch.yDatum) && ctx.invalidData?.[datumIndex] !== true;
+        // Coerce to Number for positioning (bigint values plot at finite precision); the raw value
+        // is retained on yDatum for tooltips. isContinuous accepts bigint where Number.isFinite would not.
+        scratch.yCumulative = Number(ctx.yCumulativeValues[datumIndex]);
+        scratch.validPoint = isContinuous(scratch.yDatum) && ctx.invalidData?.[datumIndex] !== true;
 
         // Compute marker coordinates
         this.computeMarkerCoordinate(ctx, scratch);
--- a/packages/ag-charts-community/src/scale/colorScale.ts
+++ b/packages/ag-charts-community/src/scale/colorScale.ts
@@ -106,6 +106,11 @@ export class ColorScale extends AbstractScale<number, string> {
     convert(x: number) {
         this.refresh();
 
+        // A bigint colour value (AG-16608 heatmap) narrows to Number; the output colour is derived from
+        // the value's position in the domain, so finite precision is sufficient and arithmetic on a
+        // bigint mixed with the Number domain would throw.
+        x = Number(x);
+
         const { domain, range, parsedRange } = this;
         const d0 = domain[0];
         const d1 = domain.at(-1)!;
--- a/packages/ag-charts-community/src/scale/colorScale.ts
+++ b/packages/ag-charts-community/src/scale/colorScale.ts
@@ -106,9 +106,8 @@ export class ColorScale extends AbstractScale<number, string> {
     convert(x: number) {
         this.refresh();
 
-        // A bigint colour value (AG-16608 heatmap) narrows to Number; the output colour is derived from
-        // the value's position in the domain, so finite precision is sufficient and arithmetic on a
-        // bigint mixed with the Number domain would throw.
+        // A bigint colour value (AG-16608 heatmap) narrows to Number: the colour derives from the value's
+        // position in the domain, so finite precision suffices and bigint/Number mixing would throw.
         x = Number(x);
 
         const { domain, range, parsedRange } = this;
--- a/packages/ag-charts-community/src/scale/colorScale.ts
+++ b/packages/ag-charts-community/src/scale/colorScale.ts
@@ -58,7 +58,7 @@ export class ColorScale extends AbstractScale<number, string> {
      * `domain` (which carries interpolation pivots) so that colour-stop
      * positions do not distort the legend axis range.
      */
-    displayDomain?: [number, number];
+    displayDomain?: [number | bigint, number | bigint];
 
     private parsedRange = this.range.map(convertColorStringToOklcha);
 
@@ -103,12 +103,12 @@ export class ColorScale extends AbstractScale<number, string> {
         return;
     }
 
-    convert(x: number) {
+    convert(x: number | bigint) {
         this.refresh();
 
         // A bigint colour value (AG-16608 heatmap) narrows to Number: the colour derives from the value's
-        // position in the domain, so finite precision suffices and bigint/Number mixing would throw.
-        x = Number(x);
+        // position in the (Number) domain, so finite precision suffices and bigint/Number mixing would throw.
+        const xn = Number(x);
 
         const { domain, range, parsedRange } = this;
         const d0 = domain[0];
@@ -116,11 +116,11 @@ export class ColorScale extends AbstractScale<number, string> {
         const r0 = range[0];
         const r1 = range.at(-1)!;
 
-        if (x <= d0) {
+        if (xn <= d0) {
             return r0;
         }
 
-        if (x >= d1) {
+        if (xn >= d1) {
             return r1;
         }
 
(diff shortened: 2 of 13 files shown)


packages/ag-charts-community/src/scale/colorScale.ts
```diff
@@ -106,6 +106,11 @@ export class ColorScale extends AbstractScale<number, string> {
     convert(x: number) {
         this.refresh();
 
+        // A bigint colour value (AG-16608 heatmap) narrows to Number; the output colour is derived from
+        // the value's position in the domain, so finite precision is sufficient and arithmetic on a
```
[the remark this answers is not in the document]
alantreadway: False positive: `colorScale.domain` never holds bigint at runtime. It is only ever set via `configureColorScale → computeColorBins`, and the PR9 change in `colorScaleUtil.ts` narrows the domain endpoints to Number (`const d0 = Number(domain[0])`) before building the stops/range. So by the time `convert()` reads `d0`/`d1` from `this.domain`, both they and the narrowed input `x` are Number — no bigint/Number mixing. The PR9 heatmap smoke test renders a bigint heatmap beyond `Number.MAX_SAFE_INTEGER` and passes, confirming the crash is fixed.

(`convert()` does rely on the caller having narrowed the domain rather than self-protecting; a defensive `Number(d0)`/`Number(d1)` would be belt-and-braces but isn't required for correctness, and convert() is a hot path.)

packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
```diff
@@ -1065,8 +1065,10 @@ export class AreaSeries extends CartesianSeries<AreaSeriesTypes> {
 
         scratch.datum = ctx.rawData[datumIndex];
         scratch.yDatum = ctx.yRawValues[datumIndex];
-        scratch.yCumulative = +ctx.yCumulativeValues[datumIndex];
-        scratch.validPoint = Number.isFinite(scratch.yDatum) && ctx.invalidData?.[datumIndex] !== true;
+        // Coerce to Number for positioning (bigint values plot at finite precision); the raw value
+        // is retained on yDatum for tooltips. isContinuous accepts bigint where Number.isFinite would not.
+        scratch.yCumulative = Number(ctx.yCumulativeValues[datumIndex]);
```
olegat, line 1070: This doesn't seem right. It's true that bigint values use finite precision, but that's not really relevant. 

AFAICT `yCumulativeValues` is retrieved/computed from the input `data`, which obviously may not fit in a `number` primitive. This simple example confirms this:
https://plnkr.co/edit/1UXr5RV4KqXtuNT2?open=main.js&preview

Image

packages/ag-charts-community/src/scale/colorScale.ts
```diff
@@ -106,6 +106,10 @@ export class ColorScale extends AbstractScale<number, string> {
     convert(x: number) {
         this.refresh();
 
+        // A bigint colour value (AG-16608 heatmap) narrows to Number: the colour derives from the value's
+        // position in the domain, so finite precision suffices and bigint/Number mixing would throw.
+        x = Number(x);
```
olegat: At face value, this looks unnecessary because `x` is already of type `number`. Perhaps it would make more sense to change the signature to `convert(x: number | bigint)` and handle the `number` and `bigint` cases separately rather than coercing both types.

packages/ag-charts-core/src/scale/colorScaleUtil.ts
```diff
@@ -119,7 +119,10 @@ export function computeColorBins(
         return { domain: [], range: [], bins: [] };
     }
 
-    const [d0, d1] = domain;
```
olegat, line 122: If `domain` is unexpectedly of type `[bigint, bigint]` at this point, then the function signature should be declared as `computeColorBins(fills, domain: [number, number] | [bigint, bigint], mode)`

packages/ag-charts-core/src/scale/colorScaleUtil.ts
```diff
@@ -173,16 +176,19 @@ export function deriveNormalizedStops(colorScale: ColorScaleState): GradientColo
     const { domain, range, mode, displayDomain } = colorScale;
     if (range.length === 0) return [];
 
-    const [d0, d1] = displayDomain ?? [domain[0], domain.at(-1)!];
```
olegat, line 176: likewise, `displayDomain` such be redeclared from `[number, number]` to `[number, number] | [bigint, bigint]`

packages/ag-charts-enterprise/src/series/waterfall/waterfallSeries.ts
```diff
@@ -277,7 +277,9 @@ export class WaterfallSeries extends _ModuleSupport.AbstractBarSeries<WaterfallS
         } else {
             const yCurrIndex = dataModel.resolveProcessedDataIndexById(this, 'yCurrent');
             const yExtent = values[yCurrIndex];
-            const fixedYExtent = [Math.min(0, yExtent[0]), Math.max(0, yExtent[1])];
+            // The cumulative extent can be bigint; narrow to Number for the domain (finite precision,
+            // consistent with bar) since Math.min/max throw on bigint.
+            const fixedYExtent = [Math.min(0, Number(yExtent[0])), Math.max(0, Number(yExtent[1]))];
```
olegat, line 282: same as area-series. I think that `Number(yExtent[i])` will cause precision loss.

packages/ag-charts-community/src/chart/series/cartesian/areaSeries.ts
alantreadway: This is a simplification to avoid needing to cater for `bigint` and `number` cases explicitly - stacked/cumulative cases do not require `bigint` precision for rendering purposes, so we are coercing to `number` for these cases to avoid needing to handle edge cases with `bigint` downstream.

olegat: This is not a problem with precision, it's a problem with upper-limits. `number` is a fixed-width type with an upper limit `Number.MAX_VALUE`, whereas `bigint` a variable-width type with no upper limit.

If `ctx.yCumulativeValue[datumIndex]` fits in a number (albeit with precision-loss), then that's okay for rendering purposes.

If `ctx.yCumulativeValue[datumIndex]` does not fit a number (example, `BigInt(Number.MAX_VALUE) * BigInt(Number.MAX_VALUE)`), then `Number(ctx.yCumulativeValues[datumIndex])` will expand to `Infinity`

Screenshot 2026-06-02 at 11 46 30

alantreadway: Agreed — addressed in PR10 (bb0a7ce). Added a shared `toFiniteNumber(number | bigint)` in ag-charts-core that warns once and clamps to `±Number.MAX_VALUE` when a bigint would otherwise coerce to `Infinity`, and adopted it here. Finite-precision positioning stays intentional for cumulative rendering; the helper just guarantees a finite result. Unit-tested with `10n ** 400n`.

packages/ag-charts-enterprise/src/series/waterfall/waterfallSeries.ts
alantreadway: Same fix as area — now via `toFiniteNumber` (PR10, bb0a7ce), so a beyond-`MAX_VALUE` extent clamps to a finite value instead of `Infinity`.

packages/ag-charts-community/src/scale/colorScale.ts
alantreadway: The signature is already `convert(x: number | bigint)`; the narrow is now `toFiniteNumber(x)` (PR10, bb0a7ce). Colour derives from the value's position in the Number domain, so a single narrow at the read point is correct rather than branching.

packages/ag-charts-core/src/scale/colorScaleUtil.ts
alantreadway: `computeColorBins` already declares `domain: [number | bigint, number | bigint]`, and the endpoints now narrow via `toFiniteNumber` (PR10, bb0a7ce) — signature is honest about the bigint input.

packages/ag-charts-core/src/scale/colorScaleUtil.ts
alantreadway: `displayDomain` is already typed `[number | bigint, number | bigint]` on both `ColorScaleState` and `ColorScale`; the narrow now routes through `toFiniteNumber` (PR10, bb0a7ce).

alantreadway:

PR #6962, opened 2026-06-01
AG-17310 Fix public operations over default private operations

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-17310


alantreadway (approved):
LGTM

PR #6965, opened 2026-06-01
fix: remove duplicate OrganizationSeriesModule entry from module mappings

AG-DavidG:
## Summary

- Removes a duplicate `OrganizationSeriesModule` entry from `packages/ag-charts-website/src/content/module-mappings/modules.json`

## Test plan

- [ ] Verify the org chart docs page still renders correctly with the module mapping intact (the first entry remains)


alantreadway (approved):
LGTM

PR #6967, opened 2026-06-01
AG-17377 Fix docs: highlight.enabled references and series/axes update warning

AG-DavidG:
https://ag-grid.atlassian.net/browse/AG-17377

- Remove incorrect `highlight.enabled` references from Treemap and Sunburst series docs (this option does not exist on these series types).
- Add a warning on the Create/Update page clarifying that when updating `series` or `axes` options, the complete array must be supplied with all properties for each item.

Fix #AG-17377


alantreadway (approved):
LGTM

PR #6972, opened 2026-06-02
AG-16608 Make scale/axis bigint types honest; sum aggregates exactly

alantreadway:

Stacked on #6955 (PR9). Follow-up to a reviewer type-confusion flag: the runtime threaded `bigint` through `number`-typed scale/axis signatures, so the load-bearing coercions looked like dead defensive code. This makes `D` itself bigint-capable through the continuous-scale and number-axis surface, so the compiler now flags every site that narrows a bigint domain to Number.

## Summary
- `ContinuousScale<D extends number|bigint|Date>`, `LinearScale extends ContinuousScale<number|bigint>`; `ticks()`/`niceDomain()` honestly take and return `(number|bigint)[]` — the `as unknown as number[]` casts are removed.
- `NumberAxis`/`AngleNumberAxis`/`RadiusNumberAxis` `D` widened to `number|bigint`. Zero cascade into the ~50 generic `Scale<>` consumers (the `Axis<S=Scale<any>>` defaults absorb it; `convert()` returns Number pixels; ticks flow to the bigint-aware formatter).
- New `Scale.domainMin`/`domainMax` accessors: `AbstractScale` derives them from `getDomainMinMax()`; `ContinuousScale` returns the true (order-independent) min/max preserving the exact bigint endpoint — replacing the `Math.min/max(...domain.map(Number))` ceremony at the gauge sites. `getDomainMinMax()`'s narrowed contract is left intact for its zoom consumers.
- Gauge/segmentation/gradient-legend narrow only at genuine Number-space maths; the per-tick label keeps the exact value. `linear-gauge` `labelDatum` stops narrowing bigint (its datum field and `formatLabel` were already bigint-capable), preserving label precision.
- `LinearAngleScale` documents its Number-precision nice-stepping limitation (AC #17); value positioning stays full-precision via `convert()`.

Precision note: forcing the bigint through surfaced no latent precision bug — `convert()` already positions exactly via `d0Big`/`d1Big`/`convertBigInt` even where narrowed endpoints would collapse. The remaining narrows are genuinely Number-space quantities.

## Follow-up: cast tightening + bigint aggregation precision

Two further commits, from a full audit of bigint→number coercion across the PR1–PR10 train:

- **Type honesty (no behaviour change).** Consolidated the three scattered `as D[]` domain casts in `ContinuousScale` into one documented narrowing helper, replaced `d0/d1 as unknown as number` with `Number()`, and turned `linearScale`'s `domain as number[]` into an honest `domain.map(Number)`. Widened `NumberAxis.getVisibleDomain`/`tickFormatParams`/`normaliseDataDomain` params from `number` to `number|bigint`, matching the bigint-capable axis `D`. Full bigint retention in `_domain` was investigated and **rejected** — it is semantically wrong for `LogScale` (`D=number`) and `TimeScale` (`D=Date`), where Number-narrowing is correct, and would cascade crashes into `logScale`/`radialBar`. Exact bigint stays exposed (cast-free) via `domainMin`/`domainMax`.
- **Bigint aggregation precision.** `sumValues` silently dropped bigint y-values, so a histogram summing a bigint `yKey` column produced zero-height bins. It now sums in bigint via `addAccumulated`, widening the aggregation result surface (`[number, number]` → `[number | bigint, number | bigint]`) through `sum`/`groupSum`/`groupAverage`/`area`. Narrowing happens only at genuine Number points: the mean and area-density divisions, histogram `getSeriesRange`'s pixel `Math.max`, and the y-scale position. The exact total reaches the bin tooltip via `aggregatedValue` (widened to `AgNumericValue`). Also fixes a latent throw for an areaPlot histogram over bigint bins (`keyWidth`).

## Test plan
- New `domainMin`/`domainMax` unit block (bigint flow-through, true min/max ordering, precision beyond `Number.MAX_SAFE_INTEGER`).
- New `sumValues` bigint unit tests (incl. a >2^53 case) and an end-to-end histogram test asserting an exact bigint bin total (3 × (2^53 + 1)).
- `build:types` + `lint` + `format` green on types/community/enterprise.
- `bigintRenderSmoke` (community + enterprise), colour-scale, `numberAxisBigInt`, `aggregateFunctions`, core `extent`/`ticks` suites green. Pre-existing `dataModel`/histogram image-snapshot failures in the worktree are unrelated (baseline-confirmed; CI re-validates).

Fix #AG-16608

### DIFF
--- a/packages/ag-charts-community/src/chart/axis/numberAxis.ts
+++ b/packages/ag-charts-community/src/chart/axis/numberAxis.ts
@@ -11,7 +11,7 @@ import { CartesianAxis } from './cartesianAxis';
 
 export class NumberAxis<
     TOptions extends NormalisedNumberAxisOptions = NormalisedNumberAxisOptions,
-> extends CartesianAxis<LinearScale | LogScale, number, TOptions> {
+> extends CartesianAxis<LinearScale | LogScale, number | bigint, TOptions> {
     static readonly className: string = 'NumberAxis';
     static readonly type: string = 'number';
 
--- a/packages/ag-charts-community/src/chart/axis/numberAxis.ts
+++ b/packages/ag-charts-community/src/chart/axis/numberAxis.ts
@@ -33,7 +33,7 @@ export class NumberAxis<
         return this.options.label.format;
     }
 
-    override normaliseDataDomain(d: DomainWithMetadata<number>) {
+    override normaliseDataDomain(d: DomainWithMetadata<number | bigint>) {
         const { min, max, preferredMin, preferredMax } = this.options;
         const { extent, clipped } = normalisedExtentWithMetadata(
             d.domain,
@@ -52,7 +52,7 @@ export class NumberAxis<
         return [this.options.min == null && this.nice, this.options.max == null && this.nice];
     }
 
-    protected getVisibleDomain(domain: number[]): [number, number] {
+    protected getVisibleDomain(domain: (number | bigint)[]): [number, number] {
         // Narrow to Number: a bigint domain reaches here as formatter-context metadata only — the tick
         // label itself carries the exact bigint value, so this range context can narrow safely.
         const d0 = Number(domain[0]);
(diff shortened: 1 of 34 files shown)


alantreadway:

PR #6974, opened 2026-06-02
AG-NONE Fix flaky e2e test: range buttons > dropdown

olegat:
We have a `waitForAllChartUpdates(page)` call, however the zoom toolbar buttons (zoom in/out, pan left/right, reset buttons) are just plain HTML `<button>` elements. These can take a bit longer to render. This means that the `.toHaveScreenshot()` assertion can sometimes grab an outdated screenshot where the disabled buttons have not been grayed out yet.

To solve this, just reset the zoom by clicking somewhere else where the zoom toolbar remains hidden.


alantreadway (approved):
Nice, LGTM

PR #6979, opened 2026-06-02
jira-agent-pipeline: collapse stub to single jira-resume entry + 4 jobs

alantreadway:
## Why

Now that the resume preflight routes `→ In Progress` mechanically (and triage joins it as a route — see ag-grid/ag-dev-prompts#299), the per-stage JIRA dispatch types are vestigial. This collapses the stub to a single JIRA-facing entry point and removes the near-identical job duplication.

## What

**One JIRA entry point** (`repository_dispatch: [jira-resume]`) and **4 jobs** (was ~10, −288 lines):

- **`preflight`** — one cheap gate, branches by event: resume router (dispatch / `workflow_dispatch`) emits `stage`; ci-failure eligibility walk (`workflow_run · failure`) emits `ci_failure_eligible`.
- **`agent-run`** — one job for **every** `/fr` stage (resume routes + ci-failure-iterate), sharing the whole token→setup-nx→install→jira-pipeline bootstrap. Stage precedence: `ci-failure-iterate` (eligible) → resume route → forced `workflow_dispatch` stage.
- **`ci-success-handler`** + **`pr-plnkr`** — kept separate (terminal REST work / a different agent action gating on its outputs).

The legacy `jira-triage` / `jira-build` / `jira-pr-iterate` dispatch types + their standalone jobs are deleted. The route→stage map (`iterate`→`pr-iterate`) lives in the preflight on the ag-dev-prompts side, so the stub passes `stage` straight through — no shell mapping.

## Notes
- Distribution unchanged (npm package + composite actions; no reusable workflow).
- actionlint clean (only the pre-existing unrelated `setup-nx` metadata warning).
- **Depends on ag-grid/ag-dev-prompts#299** (the `triage` route + `stage` output) reaching the pinned `canary` dist-tag. Merge/publish that first.
- Manual JIRA follow-up: the single `→ In Progress → jira-resume` rule is live; the legacy per-status rules are disabled.

### DIFF
--- a/.github/workflows/jira-agent-pipeline.yml
+++ b/.github/workflows/jira-agent-pipeline.yml
@@ -10,15 +10,20 @@ name: Jira Agent Pipeline
 #   recycle             — /fr --ci --recycle --prior <merged|abandoned> <KEY>      (new -rN PR after prior PR merged/closed)
 #
 # A single `→ In Progress` JIRA rule dispatches `jira-resume`; the cheap
-# `resume-preflight` job reads AI pr-url + the PR's GitHub state and routes to
-# one of resume-build / resume-iterate / resume-recycle (no AI-status guard in
-# the rule — the preflight does the disambiguation JIRA can't, since it can see
-# whether the PR is open / merged / closed). See ag-dev-prompts
+# `resume-preflight` job reads AI pr-url + the PR's GitHub state + whether
+# AI plan is populated, and emits a route (triage / build / iterate / recycle)
+# that the single `resume-run` job runs as a parameterised stage (no AI-status
+# guard in the rule — the preflight does the disambiguation JIRA can't: it sees
+# whether the PR is open / merged / closed, and whether a build-ready plan
+# exists yet).
+# First-pass triage hands back to Needs Review for plan review; the human's
+# next drag (now finding a plan) routes to build. See ag-dev-prompts
 # docs/jira-agent-pipeline.md §"Re-engagement after Needs Review".
 #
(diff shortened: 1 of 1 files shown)


alantreadway:
Thanks Codex — both P1s are false positives; leaving the stub as-is:

**P1 "preflight omits required prompts setup".** The `preflight` job is a faithful merge of the two original preflight jobs (`resume-preflight` + `ci-failure-preflight`), neither of which ran `setup-nx` or minted an App token — they only run the REST-only gate actions (`jira-resume-preflight` / `jira-ci-failure-preflight`), which don't invoke the Claude agent, so there's no `.claude/settings.json` to render. `install-ag-dev-prompts` defaults `github-token` to `${{ github.token }}` and the job grants `packages: read`, so the package resolves without the mint. `setup-nx` + the App-token mint live only in `agent-run` (and `pr-plnkr`), exactly as before. Nothing was dropped.

**P1 "forced manual path skipped via needs".** `agent-run`'s `if` begins with `!cancelled()`, which is the documented idiom for running a job when a `needs` dependency is **skipped** (skipped ≠ cancelled ≠ failure). On a forced `workflow_dispatch` (stage ≠ resume) `preflight` is skipped, `!cancelled()` is true, and the `(github.event_name == 'workflow_dispatch' && inputs.stage != 'resume')` clause runs `agent-run` with `inputs.stage`. Without `!cancelled()` GHA would skip-propagate; with it, it runs. This is the proven pattern the review asks for.

Stephanemw (approved):

PR #6982, opened 2026-06-03
AG-16608, AG-16654 Support bigint values and ISO/bigint time axes

alantreadway:

Consolidated bigint-value and ISO/bigint time-axis support, replacing blind
`as number` casts with a typed, runtime-checked contract through the data
pipeline. Squashed from the prior AG-16608 + AG-16654 PR stack onto `latest`.

**AG-16608 — bigint numeric values**
- Public/gauge types widened to accept `bigint`; bigint keys and values flow
  through the data pipeline and are accepted at runtime.
- `resolveColumnById` gains an optional expected-type discriminator that both
  types the column as `number|bigint` and asserts it against the runtime
  `ColumnValueType` tag, warning once when a value column holds non-continuous
  (string/category) data. Axis- or category-capable columns (xValue, scatter
  yValue, organization labels, selectedRaw) are deliberately left unchecked.
- Pure bigint-promoting `addValues`/`subtractValues` added to `ag-charts-core`;
  `addAccumulated` delegates to `addValues`. Aggregation result/domain types
  widened to `number|bigint`; Number-only aggregates (groupCount, area density)
  opt back into `[number, number]`.
- Histogram bins computed in BigInt; continuous-scale domain selection exact
  beyond `Number.MAX_VALUE`; coercion warns rather than clamps.
- An auditing sweep of every `Number()`/`toNumber()`/`Number.isFinite`/`Math.*`
  coercion on possibly-bigint data hardened the render/domain paths across
  series: raw bigint domain/datum values are preserved exact (not narrowed),
  `Number.isFinite` gates that would drop bigints switched to `isContinuous`,
  and `Math.min|max|abs`/`+`/`-` that throw or coerce on bigint switched to the
  bigint-safe `minValue`/`maxValue`/`absValue`/`addValues`/`subtractValues`
  helpers — covering cartesian domains, gauge segmentation, OHLC/range/box-plot
  node data, donut/radar, and the measurer-annotation volume total. Coercion is
  confined to genuine screen-space (`scale.convert()`) and `[number, number]`
  contract boundaries.

**AG-16654 — time axes**
- ISO 8601 strings and bigint epochs rendered on time axes; ISO/bigint axis
  prediction with mixed-timezone warnings; `AgTimeValue` alias and docs.
- Fixed ISO 8601 datetime *string* data rendering blank on time/unit-time axes:
  the scales parsed the strings but the domain classes dropped them, leaving an
  empty extent. `ContinuousDomain`/`DiscreteDomain` now coerce ISO strings via
  `coerceIso8601Date` (the discrete path gated on a new `timeDomain` flag so
  category axes keep ISO-shaped labels as strings).

**Testing**
- Folded the ad-hoc standalone bigint smoke tests into uniform per-series
  sub-suites with a visual snapshot per stacking/grouping combo, an ISO-datetime
  snapshot where the series is time-capable, and a pixel-identical
  magnitude-invariance test (true `toMatchImage` of small- vs large-magnitude
  renders) proving values below and above `Number.MAX_SAFE_INTEGER` render
  identically.
- Added an interactive bigint/ISO test page (axes-types-test) switching across
  series × data-mode × stacking, plus two ~1M-point benchmarks
  (`high-volume-bigint`, `high-volume-iso-datetime`).

Docs and gallery examples added for both features.

**Review follow-up (post-review hardening)**
- ISO 8601 string keys now compute the same smallest/largest key interval as Date/epoch keys (`finiteKey` parsed them to `NaN` before, so an ISO-keyed time axis got no key interval).
- `histogram` `computeBins` falls back to the Number path when only one extent endpoint is bigint and the other is non-integral, instead of `BigInt()` throwing.
- `area`/mean density width subtracts the bigint bin edges exactly before narrowing, so adjacent edges beyond 2^53 no longer collapse to a zero width.
- `epochColumnForTimeScale` now returns `{ values, needsValueOf }`, removing a 13x duplicated identity-check idiom across the aggregators.
- Corrected the mixed-numeric column warning (the series renders narrowed, not empty) and tightened the bigint-invariant comments.
- Added a v14 migration-guide entry for the `AgNumericValue`/`AgTimeValue` widening.

Fix #AG-16608
Fix #AG-16654


alantreadway:
## Carried-forward review items from the superseded PR stack

This PR consolidates the AG-16608 + AG-16654 stack (#6946, #6947, #6948, #6949, #6950, #6952, #6953, #6954, #6955, #6972), which are being closed in its favour. Their review threads were triaged against this branch's content. Most items are resolved-and-present or by-design; the following remain genuinely applicable here and were **not** carried into the squash:

### Divergence — `toFiniteNumber` clamp not present
PR10 (`bb0a7ce`) introduced a shared `toFiniteNumber()` to clamp a bigint that would otherwise coerce to `Infinity` (olegat's P1 on #6955: `Number(hugeBigint) → Infinity` beyond `Number.MAX_VALUE`). This branch instead adopts a **warn-not-clamp** design: `toNumber()` warns on non-finite (used in `waterfallSeries`), while `areaSeries` still narrows `cumulativeValue` via a bare `Number(scratch.yCumulative)`. Decision needed: confirm warn-not-clamp is the intended final design and apply it consistently (areaSeries currently has neither clamp nor warn at that site).

### Unanswered reviewer questions (code present on this branch)
- **memento decode** (`memento.ts`, #6948, @lsjroberts): "Could this check be left to the JS engine via try/catch?" — re the `/^-?\d+$/` guard on bigint restore.
- **chord example** (`gallery/_examples/chord/main.ts`, #6948, @lsjroberts): "Users won't need this conversion unless using bigints? Just extra caution?"
- **histogram domain typing** (`histogramSeries.ts`, #6946, @lsjroberts): ideal `[number, number] | [bigint, bigint]`. Partly addressed — this branch types it as `BinDomain = [number | bigint, number | bigint]`; worth a confirming reply.

### Housekeeping
- #6949 (AC#17 thread): cross-reference the AG-16608 bigint/ISO design doc from the JIRA ticket.

### Not carried (moot here)
PR7's per-series `xKey`/`yKey` bigint/ISO JSDoc additions (#6953) are absent from this branch, so the Codex "overstates time-axis support" comments no longer apply. `AgTimeValue` itself did land.

packages/ag-charts-community/src/chart/data/data-model/extractor/__snapshots__/dataExtractor.test.ts.snap
alantreadway, line 1: Is this (and similar) snapshot changes correct?

ag-alantreadway: This is due to empty-array being treated as `number` before.

PR #6984, opened 2026-06-03
AG-14484 revert enabled-false validation bypass in validate()

iMoses:
https://ag-grid.atlassian.net/browse/AG-14484

Reverts the `enabled: false` shortcut introduced in #6977. That change caused `validate()` to skip required-field enforcement and discriminated-union type checks whenever a node had `enabled: false`.

Removed:
- `optionsDisabled` local variable and its two uses in `validate()`
- The "Disabled nodes (enabled: false)" describe block in `validation.test.ts`

Required-field and type-union checks now run unconditionally again, matching the pre-#6977 behaviour.

Fix #AG-14484

### DIFF
--- a/packages/ag-charts-core/src/state/validation.test.ts
+++ b/packages/ag-charts-core/src/state/validation.test.ts
@@ -341,39 +341,6 @@ describe('Validation utils', () => {
         });
     });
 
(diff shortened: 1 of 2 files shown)


alantreadway (approved):
LGTM

PR #6986, opened 2026-06-03
AG-16926 add getDataId callback for non-1:1 series (histogram, sankey/chord)

iMoses:
## Summary

Adds a `getDataId` callback to the **histogram**, **sankey** and **chord** series so aggregated nodes (which don't map 1:1 to datums) carry a stable, user-facing identifier, surfaced as `itemId` in events (`seriesNodeClick`, `activeChange`, `selectionChange`) and active state. Builds on the existing `dataIdKey` datum-id infrastructure (AG-14484 / AG-16076).

Jira: [AG-16926](https://ag-grid.atlassian.net/browse/AG-16926)

## What changed

- **Histogram**: bins now default to `bin:<start>,<end>` (previously the numeric group index).
- **Sankey/Chord**: nodes now default to the **node name** from `fromKey`/`toKey` (previously index-based `node-N`), so ids stay stable when nodes are added/reordered.
- **`getDataId`** overrides the default. Invoked via `cachedCallWithContext`, so the chart `context` is passed to the callback and a throwing callback is contained (falls back to the default id).
- Relaxed `DataModelSeriesNodeDatum.itemId` from `never` to `string` so the aggregated histogram bin can carry an explicit id (required by the `CartesianSeriesTypes` generic bound; runtime-identical for other cartesian series, which never set it).
- Public types added to `ag-charts-types`; docs updated on all three series pages.

## ⚠️ Behavioural change

The emitted `itemId` default changes for these series:
- Histogram bins: numeric group index → `bin:<start>,<end>` string.
- Sankey/chord nodes: `node-<index>` → node name.

Consumers reading these ids from events/active state, or passing them to `setActiveItem`, are affected. The `callbacks.test.ts` assertions/snapshots were updated accordingly.

## Acceptance criteria

- [x] Histogram bins have auto-generated ids exposed in events and active state
- [x] Sankey/chord nodes have auto-generated ids (from node names) exposed in events and active state
- [x] `getDataId` overrides the auto-generated id when provided
- [x] Ids remain stable when bin boundaries / node names don't change
- [x] API docs updated; types added to `ag-charts-types`

## Testing

- New `getDataId` tests (default, override, context passthrough, stability across update, negative/fractional bins, throwing-callback fallback) for all three series.
- `yarn nx test ag-charts-community` (3374 passed) and `yarn nx test ag-charts-enterprise` (2184 passed).
- `build:types`, `build:test`, and `lint` clean across types/community/enterprise.

### DIFF
--- a/packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.test.ts
+++ b/packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.test.ts
@@ -1,4 +1,4 @@
-import { afterEach, describe, expect, it } from 'vitest';
+import { afterEach, describe, expect, it, vi } from 'vitest';
 
 import { mapValues } from 'ag-charts-core';
 import type { AgCartesianChartOptions, AgChartOptions } from 'ag-charts-types';
@@ -11,6 +11,7 @@ import {
     IMAGE_SNAPSHOT_DEFAULTS,
     cartesianChartAssertions,
     deproxy,
+    expectWarningsCalls,
     extractImageData,
     hoverAction,
     prepareTestOptions,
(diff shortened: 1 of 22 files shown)


alantreadway (approved):
LGTM

PR #6990, opened 2026-06-03
AG-17153 trailing slash logo link

skewis6:


packages/ag-charts-website/src/components/SiteLogo.tsx
```diff
@@ -4,12 +4,14 @@ import LogoMark from '@components/logo/LogoMark';
...
 
+import { ensureTrailingSlash } from '../utils/ensureUrl';
+
 export const SiteLogo: FunctionComponent = () => {
     const [isLogoHover, setIsLogoHover] = useState(false);
 
     return (
         <a
-            href={SITE_BASE_URL.replace(/\/?$/, '/')}
+            href={ensureTrailingSlash(SITE_BASE_URL)}
```
taktran: Actually, I do have an existing function for this

```
import { addTrailingSlash } from '@ag-website-shared/utils/addTrailingSlash';
```

Can you use this instead

taktran (changes requested):

taktran (approved):

PR #6996, opened 2026-06-03
fix(website): flatten API reference search to avoid stack overflow

iMoses:
The `/themes-api` documentation page failed to load with `RangeError: Maximum call stack size exceeded`.

### Root cause
The page builds its search index by rooting `extractSearchData` at `AgChartTheme`. That tree is finite but large — it expands to ~150k entries (the `overrides` subtree alone is ~149.5k). The previous implementation assembled the index with `results.push(...extractSearchData(...))` and `[leaf, ...extractSearchData(...)]`. Spreading a large array as call arguments overflows V8's argument limit (throws at ~125k in Node, lower in browsers), which surfaces as `Maximum call stack size exceeded` — not an infinite recursion (real recursion depth here is only ~14).

This regressed when generics threading landed in the API-reference helpers: resolving generic member type-params to concrete interfaces grew the `AgChartTheme` index from ~63k entries (under the browser limit, so it still loaded) to ~150k (over the limit → crash). The options page (~1.7k entries) was never close, so it was unaffected.

### Fix
Flatten the four search-extraction helpers (`extractSearchData`, and the interface/union collectors) to push into a single shared accumulator one entry at a time, so no large array is ever spread as arguments. `extractSearchData` keeps its public signature as a thin wrapper.

The produced index is **byte-for-byte identical** to before — verified by comparing serialised output (label + searchable + navPath) for `AgChartTheme`, `AgChartOptions`, and the axis option roots: same counts, matching SHA hashes.

### Test plan
- New unit regression test (`apiReferenceHelpers.test.ts`): a self-contained large finite reference reproduces the argument-spread overflow on the previous code and asserts the flattened version returns the expected entry count without throwing.
- New e2e test (`api-ref-page.spec.ts`): loads `/themes-api` and asserts the navigation tree renders; the suite's intrinsic `pageerror` assertion fails on any uncaught overflow.
- `yarn nx lint ag-charts-website` passes (0 errors); output equivalence verified against the pre-fix implementation.

### DIFF
--- a/packages/ag-charts-website/e2e/api-ref-page.spec.ts
+++ b/packages/ag-charts-website/e2e/api-ref-page.spec.ts
@@ -88,6 +88,15 @@ test.describe('api-ref-page', () => {
         ]);
     });
 
+    // The themes API page roots its search index at AgChartTheme, whose tree is far larger than the
+    // options page. Regression: building that index overflowed V8's argument limit and the page
+    // failed to load with "Maximum call stack size exceeded". setupIntrinsicAssertions captures any
+    // such pageerror and fails the test; the visible nav tree guards against a silently blank page.
+    test('themes API page loads without overflowing', async ({ page }) => {
+        await gotoUrl(page, toPageUrl('themes-api/'));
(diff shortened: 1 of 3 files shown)


alantreadway (approved):
LGTM

PR #6998, opened 2026-06-04
AG-17445 Add guard for error-throwing in selectionChange callbacks

olegat:
https://ag-grid.atlassian.net/browse/AG-17445


alantreadway (approved):
LGTM

PR #7003, opened 2026-06-04
AG-17107 QA Feedback: Clear candidacy when pressing Escape

olegat:
https://ag-grid.atlassian.net/browse/AG-17107


alantreadway (approved):
LGTM

PR #7017, opened 2026-06-05
AG-17444 Fix blank polar chart on detached create-then-update

alantreadway:
https://ag-grid.atlassian.net/browse/AG-17444

A polar chart created on a **detached** container and then `update()`-d with its series renders blank — the AG Charts Studio lifecycle (downstream incident AS-774).

**Root cause:** adding the series changes the chart type (default `CartesianChart` → `PolarChart`), which re-creates the chart and transfers the scene to the replacement. The previous chart's teardown is deferred behind its update mutex, so `DOMManager.destroy()` runs *after* the new chart has adopted the transferred `<canvas>` — and it removed every tracked child, ripping the live canvas back out of the DOM. The chart then renders to an orphaned, off-DOM canvas → blank. Cartesian charts were unaffected because adding a cartesian series does not change the chart type.

**Fix:** `DOMManager.destroy()` now only removes child nodes still parented under this manager, leaving a re-homed (transferred) canvas intact.

**Tests:** the JSDOM regression coverage now lives in two homes — a `DOMManager.destroy()` unit test (`domManager.test.ts`) pinning the ownership check directly (a transferred/re-homed child survives teardown; a still-owned child is removed), and an `AgCharts.create()` → `update(polar)` integration test (`agChart.test.ts`) asserting the canvas stays attached after a type-changing update. Both fail on the old code for donut/pie (and the unit test for the transferred child) and pass with the fix; the line cartesian control case is unaffected. donut/pie/radar-area verified rendering in a real browser via the dev server.

Fix #AG-17444


iMoses (approved):
LGTM

PR #7020, opened 2026-06-05
AG-14484 Tidy up crossline options and validations

iMoses:
## Summary

Follow-up QA fixes for the crossline discriminated-union rework. Addresses items 1, 3, 5 and 6 from the QA comment; item 2 needs no code change (see below).

- **`enabled: false` cross-lines no longer warn (item 1).** Options are validated twice: once on the raw user options, then again after enrichment. `removeDisabledOptions` runs between the two passes and strips a disabled node down to `{ enabled: false }`, so the second pass would warn about the just-removed required/discriminant properties. Re-introduced the disabled-node bypass behind a new opt-in `skipDisabledNodeValidation` flag, set only by the second pass; the first pass over raw user options is unchanged. The flag is routed through `ValidatorContext` so the array validators forward it to nested cross-lines.
- **Unknown `fill`/`fillOpacity` on `line` cross-lines now warn (item 3).** `fill`/`fillOpacity` are valid on the `range` variant only. Moved them off the shared `AgCommonCrossLineOptions` into `AgRangeCrossLineOptions`, split the validator so only the `range` branch accepts them, and made the theme template apply them conditionally (`$if type == 'range'`) so default `line` cross-lines don't trip the new warning. The theme-override path keeps both (via `AgCrossLineThemeOptions`).
- **Radius-category `line` cross-line renders as a single band-centred stroke (item 5).** Previously drew a filled annulus straddling the band edge. It now collapses to one ring and is offset to the band centre (`+ bandwidth / 2`, matching the cartesian cross-line), with fill applied to `range` only.
- **Cross-lines docs API Reference no longer blank (item 6).** The page embedded the aliased union `AgCartesianCrossLineOptions`, which renders empty. Replaced with two tabs referencing the concrete `AgLineCrossLineOptions` / `AgRangeCrossLineOptions` interfaces.

### Item 2 (ordinal-time cross-lines not rendering) — no code change
The reporter's example uses `ag-charts-community`, which has no `ordinal-time` axis (it is enterprise-only), so the axis and its cross-lines never render. The fix is to build the example against `ag-charts-enterprise`.

## Notes for reviewers
- **Breaking type change:** `fill`/`fillOpacity` are removed from the `line` variant type, so `{ type: 'line', fill }` no longer type-checks. Intended.
- **New runtime warning:** existing configs that set `fill`/`fillOpacity` on a `line` cross-line now log an "unknown option" warning and drop the value. Visually unchanged — the runtime already ignored line fill.

## Test plan
- `yarn nx test ag-charts-community` — full suite green (3392); crossline suite incl. new `#unknown options` (line renders + warns) and `#disabled options` (disabled cross-line emits no warning) groups.
- `yarn nx test ag-charts-enterprise` — full suite green (2195); polar crossline suite incl. regenerated radial-bar-circle snapshot (band-centred stroke, verified by pixel measurement).
- `ag-charts-core` validation unit tests — new disabled-node block, asserting the bypass is off by default and only applies with the opt-in flag.
- `yarn nx build:types` + `lint` across types/core/community/enterprise — green.

Fix #AG-14484


alantreadway (approved):
LGTM

PR #7023, opened 2026-06-05
AG-16608, AG-16654 Bigint/ISO render hardening — core (1/3)

alantreadway:

**Stacked PR 1 of 3** — splits the bigint/ISO render-hardening work into reviewable slices. This is the base of the stack:

1. **core (this PR)** → `latest`
2. tests → this branch
3. docs/examples → tests branch

## Summary

Core implementation for full-precision bigint and ISO-8601 datetime values across the data model, scales, axes and every series render path.

- Flow `number | bigint` (`AgNumericValue`) and ISO datetime values through aggregation, domains, ticks and min/max paths, coercing to Number only at the final screen-space mapping so precision is preserved beyond ±2^53.
- Add `ag-charts-core` `iso8601`/numeric helpers and `ag-charts-types` `AgNumericValue`/`AgTimeValue` aliases; widen formatter, axis and aggregation option/types accordingly.
- Harden the hierarchy series (treemap + sunburst) so out-of-safe-range bigint `sizeKey`/`colorKey` values drive tile area and the colour domain, instead of collapsing to equal tiles / a single colour.
- Update the affected gallery examples for the widened `number | bigint` formatter value type, and teach the doc-reference generator about the BigInt keyword.
- Regenerate the `dataModel`/`dataExtractor` snapshots and adjust the four unit tests whose existing assertions are forced by the new column-type, scale-domain and warning behaviour.

## Review follow-ups

- Order `ContinuousScale` `domainMin`/`domainMax` by the exact (possibly bigint) endpoints rather than the Number-narrowed caches, so a descending domain whose two bigint endpoints collapse to the same double (differ below one ULP near 2^53) still returns the correct min/max. Regression test added in the tests PR (#7024).

## Test plan

- `yarn nx build:types` / `build:test` (community + enterprise) ✓
- `yarn nx lint` (community, enterprise, core, types) — 0 errors ✓
- `yarn nx test ag-charts-community` ✓ / `yarn nx test ag-charts-enterprise` ✓
- `yarn nx validate-examples ag-charts-website` ✓ (gallery fixups type-check against the widened types)

Fix #AG-16608
Fix #AG-16654

### DIFF
--- a/packages/ag-charts-community/src/chart/data/__snapshots__/dataModel.test.ts.snap
+++ b/packages/ag-charts-community/src/chart/data/__snapshots__/dataModel.test.ts.snap
@@ -667,7 +667,7 @@ exports[`DataModel > grouped processing - calculated grouping > should generated
     },
   ],
   "columnValueType": [
-    "number",
+    undefined,
   ],
   "columns": [
     [
--- a/packages/ag-charts-community/src/chart/data/data-model/extraction/dataExtractor.ts
+++ b/packages/ag-charts-community/src/chart/data/data-model/extraction/dataExtractor.ts
@@ -1,4 +1,4 @@
-import { Logger, first, isNumberObject, iterate } from 'ag-charts-core';
+import { Logger, first, isISO8601, isNumberObject, iterate } from 'ag-charts-core';
 
 import { ContinuousDomain } from '../../dataDomain';
 import {
@@ -76,33 +76,63 @@ function valueColumnType(value: unknown): ColumnValueType | undefined {
             return 'number';
         case 'bigint':
             return 'bigint';
+        case 'boolean':
+            return 'boolean';
         case 'string':
-            return 'string';
+            return isISO8601(value) ? 'date' : 'string';
         case 'object':
             if (value == null) return undefined;
-            // A NumberObject ({ valueOf(): number }) is numeric, not a date.
-            return isNumberObject(value) ? 'number' : 'date';
+            if (value instanceof Date) return 'date';
+            return isNumberObject(value) ? 'number' : 'object';
         default:
             return undefined;
     }
 }
 
-function updateColumnTypeTracker(tracker: ColumnTypeTracker, value: unknown, datumIndex: number): void {
+function updateColumnTypeTracker(
+    tracker: ColumnTypeTracker,
+    value: unknown,
+    datumIndex: number
+): ColumnValueType | undefined {
     const observed = valueColumnType(value);
-    if (observed == null) return;
+    if (observed == null) return undefined;
 
-    if (tracker.type == null) {
+    if (tracker.type == null || tracker.type === observed) {
         tracker.type = observed;
-    } else if (tracker.type !== observed && isNumericColumnType(tracker.type) && isNumericColumnType(observed)) {
-        // TODO(AG-16654, PR5): number<->date coexistence (epoch number + Date in one column) should also
-        // promote to 'date'; only number<->bigint mixing is handled today.
+    } else if (isDateColumnType(tracker.type) || isDateColumnType(observed)) {
+        tracker.type = 'date';
+    } else if (isNumericColumnType(tracker.type) && isNumericColumnType(observed)) {
+        // Mixed number/bigint narrows bigints to Number (lossy beyond ±2^53); warned once at series level.
         tracker.mixedAtIndex ??= datumIndex;
         tracker.type = 'mixed-numeric';
     }
+
+    return observed;
 }
 
 function isNumericColumnType(type: ColumnValueType): boolean {
     return type === 'number' || type === 'bigint' || type === 'mixed-numeric';
+}
+
+function isDateColumnType(type: ColumnValueType): boolean {
+    return type === 'date';
+}
+
+// Explicit timezone iff the ISO string ends with `Z` or a trailing `±HH:MM` offset.
+const ISO_8601_EXPLICIT_TZ = /(Z|[+-]\d{2}:\d{2})$/;
+
+interface TimezoneTracker {
+    explicit: { value: string; index: number } | undefined;
+    implicit: { value: string; index: number } | undefined;
+}
+
+function updateTimezoneTracker(tracker: TimezoneTracker, value: unknown, datumIndex: number): void {
+    if (typeof value !== 'string') return;
+    if (ISO_8601_EXPLICIT_TZ.test(value)) {
+        tracker.explicit ??= { value, index: datumIndex };
+    } else {
+        tracker.implicit ??= { value, index: datumIndex };
+    }
 }
 
 /**
@@ -236,6 +266,8 @@ export class DataExtractor<D extends object, K extends keyof D & string> {
 
             // Track ordering/uniqueness for this key definition
             const tracker = createKeyTracker();
+            const typeTracker: ColumnTypeTracker = { type: undefined, mixedAtIndex: undefined };
+            const tzTracker: TimezoneTracker = { explicit: undefined, implicit: undefined };
 
             for (const scope of keyScopes ?? []) {
                 const data = sources.get(scope)?.data ?? [];
@@ -270,6 +302,9 @@ export class DataExtractor<D extends object, K extends keyof D & string> {
                         keys.push(result.value);
                         // Track ordering/uniqueness for valid keys
                         updateKeyTracker(tracker, result.value);
+                        if (updateColumnTypeTracker(typeTracker, result.value, datumIndex) === 'date') {
+                            updateTimezoneTracker(tzTracker, result.value, datumIndex);
+                        }
                         continue;
                     }
 
@@ -301,6 +336,7 @@ export class DataExtractor<D extends object, K extends keyof D & string> {
 
             // Store the computed sort order entry for this key definition
             keySortOrders.set(keyDefIndex, trackerToSortOrderEntry(tracker));
+            this.warnMixedTimezoneColumn(keyDef.property, typeTracker.type, tzTracker);
         }
         return {
             invalidData,
@@ -364,7 +400,7 @@ export class DataExtractor<D extends object, K extends keyof D & string> {
         const columns: unknown[][] = [];
         const allColumnScopes: Set<ScopeId>[] = [];
         const columnNeedValueOf: boolean[] = [];
-        const columnValueType: ColumnValueType[] = [];
+        const columnValueType: (ColumnValueType | undefined)[] = [];
         let maxDataLength = 0;
         const valueProcessors = valueDefs.map((def) => getProcessValue(def));
 
@@ -383,6 +419,7 @@ export class DataExtractor<D extends object, K extends keyof D & string> {
             const invalidKeys = scopeInvalidKeys.get(columnScope);
             let needsValueOf = false;
             const typeTracker: ColumnTypeTracker = { type: undefined, mixedAtIndex: undefined };
+            const tzTracker: TimezoneTracker = { explicit: undefined, implicit: undefined };
             for (let datumIndex = 0; datumIndex < columnSource.length; datumIndex++) {
                 if (columnSource[datumIndex] == null || typeof columnSource[datumIndex] !== 'object') {
                     // Count non-object items as invalid data
@@ -401,8 +438,8 @@ export class DataExtractor<D extends object, K extends keyof D & string> {
                     this.markScopeDatumInvalid(def.scopes, columnSource, datumIndex, invalidData, invalidDataCount);
                 } else if (result.missing) {
                     this.markScopeDatumMissing(def.scopes, columnSource, datumIndex, missingData);
-                } else {
-                    updateColumnTypeTracker(typeTracker, result.value, datumIndex);
+                } else if (updateColumnTypeTracker(typeTracker, result.value, datumIndex) === 'date') {
+                    updateTimezoneTracker(tzTracker, result.value, datumIndex);
                 }
 
                 if (invalidKey) {
@@ -424,11 +461,13 @@ export class DataExtractor<D extends object, K extends keyof D & string> {
             if (typeTracker.type === 'mixed-numeric' && typeTracker.mixedAtIndex != null) {
                 this.warnMixedNumericColumn(columnScope, def.property, typeTracker.mixedAtIndex);
             }
+            this.warnMixedTimezoneColumn(def.property, typeTracker.type, tzTracker);
 
             columns.push(column);
             allColumnScopes.push(columnScopes);
             columnNeedValueOf.push(needsValueOf);
-            columnValueType.push(typeTracker.type ?? 'number');
+            // Leave unobserved columns undefined rather than guessing 'number', so type assertions don't false-positive.
+            columnValueType.push(typeTracker.type);
             maxDataLength = Math.max(maxDataLength, column.length);
         }
 
@@ -445,8 +484,17 @@ export class DataExtractor<D extends object, K extends keyof D & string> {
     private warnMixedNumericColumn(seriesId: ScopeId, key: K, atIndex: number) {
         Logger.warnOnce(
             `Series "${seriesId}": column "${String(key)}" mixes 'number' and 'bigint' values ` +
-                `(first detected at row ${atIndex}). Each column must be uniformly typed. ` +
-                `The series renders empty; other series in this chart are unaffected.`
+                `(first detected at row ${atIndex}); the bigints are narrowed to Number and may lose precision ` +
+                `beyond ±2^53. Use one numeric type per column.`
+        );
+    }
+
+    private warnMixedTimezoneColumn(key: K, columnType: ColumnValueType | undefined, tracker: TimezoneTracker) {
+        const { explicit, implicit } = tracker;
+        // Only date-tagged columns interpret ISO strings as instants; mixed offsets in a category column are not ambiguous.
+        if (columnType !== 'date' || explicit == null || implicit == null) return;
+        Logger.warnOnce(
+            `Time axis: column "${String(key)}" contains both timezone-explicit values (e.g. "${explicit.value}", row ${explicit.index}) and timezone-implicit values (e.g. "${implicit.value}", row ${implicit.index}). Ambiguous timezone semantics may produce unexpected positions — points without an explicit offset are interpreted as local time. Use explicit offsets (Z or ±HH:MM) for cross-environment determinism.`
         );
     }
 
--- a/packages/ag-charts-community/src/chart/data/data-model/extraction/dataExtractor.ts
+++ b/packages/ag-charts-community/src/chart/data/data-model/extraction/dataExtractor.ts
@@ -1,4 +1,5 @@
-import { Logger, first, isISO8601, isNumberObject, iterate } from 'ag-charts-core';
+import { Logger, first, isISO8601, isNumberObject, iterate, timeValueToNumber } from 'ag-charts-core';
+import type { AgNumericValue } from 'ag-charts-types';
 
 import { ContinuousDomain } from '../../dataDomain';
 import {
@@ -21,7 +22,7 @@ import { createArray } from '../utils/helpers';
 
 /** Tracks ordering/uniqueness during key extraction */
 interface KeyExtractionTracker {
-    lastValue: number | undefined;
+    lastValue: AgNumericValue | undefined;
     sortOrder: 1 | -1 | 0; // 0 = undetermined, 1 = ascending, -1 = descending
     isUnique: boolean;
     isOrdered: boolean;
@@ -32,27 +33,41 @@ function createKeyTracker(): KeyExtractionTracker {
 }
 
 function updateKeyTracker(tracker: KeyExtractionTracker, value: unknown): void {
-    // Only track numeric values (including Date.valueOf())
-    const numericValue = typeof value === 'number' ? value : (value as Date)?.valueOf?.();
-    if (typeof numericValue !== 'number' || !Number.isFinite(numericValue)) return;
+    // Resolve to a comparable value: bigint stays exact, ISO 8601 strings parse to epoch ms, Date narrows
+    // via valueOf(). Without this, bigint/ISO keys are skipped and the series is wrongly flagged unordered.
+    let current: AgNumericValue | undefined;
+    if (typeof value === 'number') {
+        current = Number.isFinite(value) ? value : undefined;
+    } else if (typeof value === 'bigint') {
+        current = value;
+    } else if (isISO8601(value)) {
+        const epoch = timeValueToNumber(value);
+        current = Number.isFinite(epoch) ? epoch : undefined;
+    } else {
+        const viaValueOf = (value as Date)?.valueOf?.();
+        current = typeof viaValueOf === 'number' && Number.isFinite(viaValueOf) ? viaValueOf : undefined;
+    }
+    if (current === undefined) return;
 
-    if (tracker.lastValue === undefined) {
-        tracker.lastValue = numericValue;
+    const { lastValue } = tracker;
+    if (lastValue === undefined) {
+        tracker.lastValue = current;
         return;
     }
 
-    const diff = numericValue - tracker.lastValue;
-    if (diff === 0) {
+    // A column is uniformly typed here, so `===` is exact; the relational `>` for direction stays type-safe
+    // (and never throws as `bigint - number` subtraction would) if a malformed mixed column slips through.
+    if (current === lastValue) {
         tracker.isUnique = false;
     } else if (tracker.isOrdered) {
-        const direction = diff > 0 ? 1 : -1;
+        const direction = current > lastValue ? 1 : -1;
         if (tracker.sortOrder === 0) {
             tracker.sortOrder = direction;
         } else if (tracker.sortOrder !== direction) {
             tracker.isOrdered = false;
         }
     }
-    tracker.lastValue = numericValue;
+    tracker.lastValue = current;
 }
 
 function trackerToSortOrderEntry(tracker: KeyExtractionTracker): SortOrderEntry {
--- a/packages/ag-charts-community/src/chart/data/data-model/extraction/dataExtractor.ts
+++ b/packages/ag-charts-community/src/chart/data/data-model/extraction/dataExtractor.ts
@@ -348,7 +348,7 @@ export class DataExtractor<D extends object, K extends keyof D & string> {
 
             // Store the computed sort order entry for this key definition
             keySortOrders.set(keyDefIndex, trackerToSortOrderEntry(tracker));
-            this.warnMixedTimezoneColumn(keyDef.property, typeTracker.type, tzTracker);
+            this.warnMixedTimezoneColumn(keyDef.property, typeTracker.type, tzTracker, keyDef.timeDomain === true);
         }
         return {
             invalidData,
@@ -480,7 +480,7 @@ export class DataExtractor<D extends object, K extends keyof D & string> {
             if (typeTracker.mixedDateAtIndex != null) {
                 this.warnMixedDateColumn(columnScope, def.property, typeTracker.mixedDateAtIndex);
             }
-            this.warnMixedTimezoneColumn(def.property, typeTracker.type, tzTracker);
+            this.warnMixedTimezoneColumn(def.property, typeTracker.type, tzTracker, def.timeDomain === true);
 
             columns.push(column);
             allColumnScopes.push(columnScopes);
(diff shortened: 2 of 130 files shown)


packages/ag-charts-community/src/chart/data/data-model/extraction/dataExtractor.ts
```diff
@@ -76,35 +76,65 @@ function valueColumnType(value: unknown): ColumnValueType | undefined {
...
         default:
             return undefined;
     }
 }
 
-function updateColumnTypeTracker(tracker: ColumnTypeTracker, value: unknown, datumIndex: number): void {
+function updateColumnTypeTracker(
+    tracker: ColumnTypeTracker,
+    value: unknown,
+    datumIndex: number
```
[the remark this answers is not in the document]
alantreadway: Fixed in 598139d469. `updateColumnTypeTracker` now only promotes to `date` when the other observed type is numeric (epoch values); a date column mixed with a string/boolean/object records the row and emits a one-shot warning, mirroring the mixed number/bigint path. Added a regression test in the tests PR (#7024). Rendering of the incompatible values is unchanged (still best-effort), but the user is now told.

packages/ag-charts-community/src/scale/linearScale.ts
```diff
@@ -36,32 +37,33 @@ export class LinearScale extends ContinuousScale<number> {
 
     override ticks(
         { interval, tickCount = ContinuousScale.defaultTickCount, minTickCount, maxTickCount }: ScaleTickParams<number>,
-        domain: number[] = this.domain,
+        domain: AgNumericValue[] = this.domain,
```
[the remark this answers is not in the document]
alantreadway: Triaged as a false positive — the default axis path doesn't rely on the Number-narrowed `this.domain`. Tick generation in `generateTicks.ts` computes `niceDomain` from the axis' `dataDomain.domain` (which retains bigint) and passes it explicitly: `scale.ticks(tickParams, niceDomain, visibleRange)` (generateTicks.ts:290/327/385). That's the same shape the passing bigint tick tests exercise (`scale.ticks(params, [0n, 100n])`), so the full-precision BigInt branch does run for the full unzoomed domain. `this.domain` is only the narrowed public accessor; the exact endpoints also remain on `domainMin`/`domainMax`. (The custom-interval / zoomed sub-range cases intentionally fall back to Number — documented at linearScale.ts:49.)

packages/ag-charts-core/src/utils/aggregation.ts
```diff
@@ -18,6 +21,99 @@ const SMALLEST_INTERVAL_MIN_RECURSE = 3;
...
+const relativeNumericColumnCache = new WeakMap<readonly unknown[], unknown[]>();
+
+/**
+ * Narrow a bigint column to Number after subtracting its bigint minimum, so a high-magnitude narrow-range
+ * column keeps full resolution instead of collapsing onto one double. The offset destroys sign and absolute
+ * magnitude, so use ONLY for the comparison-based extrema envelope (line/area/ohlc/range); bar and bubble
+ * must use {@link narrowBigIntColumn}.
+ * @returns the input `values` unchanged when the column holds no bigint.
+ */
+export function narrowBigIntColumnRelative(values: any[]): any[] {
```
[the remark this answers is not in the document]
alantreadway: Working as intended for this path. `narrowBigIntColumnRelative` is scoped (per its JSDoc) to the comparison-based min/max envelope only, and only runs on columns already flagged as mixed-invalid (which now also emit a warning). The same offset is subtracted from every entry — exactly as bigint, and as `Number(minBig)` for numbers — so relative ordering is preserved for the extrema computation even though absolute magnitude/sign is deliberately destroyed. Bar/bubble, which need absolute magnitude, use `narrowBigIntColumn` instead. Leaving as-is.

packages/ag-charts-enterprise/src/series/gauge-util/segmentation.ts
```diff
@@ -1,27 +1,32 @@
...
             ticks = [d0, ...segments, d1];
         } else if (step != null) {
-            const segments: number[] = [];
-            for (let i = d0; i < d1; i += step) {
+            const numericStep = Number(step);
+            const segments: AgNumericValue[] = [];
+            for (let i = d0n; i < d1n; i += numericStep) {
                 segments.push(i);
             }
             segments.push(d1);
```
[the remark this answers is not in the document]
alantreadway: Real hang, but pre-existing and out of scope for this PR: the loop `for (i = d0; i < d1; i += step)` already hung for `step: 0` (and negative) before this change — the PR only widened `step` from `number` to `number | bigint`, so `0n` is newly *accepted* but `0` always hung. Tracking a finite-&-positive guard for segmentation `step`/`count` as a separate hardening task rather than expanding this PR's scope.

packages/ag-charts-types/src/api/initialStateOptions.ts
```diff
@@ -47,9 +47,9 @@ export interface AgInitialStateZoomOptions {
 
 export interface AgInitialStateZoomRange {
     /** The start value of the zoom range. */
-    start?: AgStateSerializableDate | string | number;
+    start?: AgStateSerializableDate | AgStateSerializableBigInt | string | number | bigint;
```
lsjroberts: `start` and `end` should probably not allow non-serialised bigint?

packages/ag-charts-types/src/chart/annotationsOptions.ts
```diff
@@ -516,7 +516,7 @@ interface Extendable {
     extendEnd?: boolean;
 }
 
-export type ValueType = string | number | AgStateSerializableDate;
+export type ValueType = string | number | bigint | AgStateSerializableDate | AgStateSerializableBigInt;
```
lsjroberts: Same here, it should probably not allow non-serialised bigint

packages/ag-charts-core/src/state/memento.ts
```diff
@@ -93,12 +93,25 @@ export class MementoCaretaker {
...
 
     private static decodeTypes(this: any, key: any, value: any) {
-        if (isObject(this[key]) && '__type' in this[key] && this[key].__type === 'date') {
-            return new Date(this[key].value);
+        if (isObject(this[key]) && '__type' in this[key]) {
+            if (this[key].__type === 'date') {
+                return new Date(this[key].value);
+            }
+            if (this[key].__type === 'bigint') {
+                // Untrusted input: only decode a valid integer string, else leave it for guardMemento to reject.
```
lsjroberts, line 109: We may want to check the various `guardMemento` calls are doing enough to reject this (separate from this PR).

packages/ag-charts-core/src/utils/time/ticks.ts
```diff
@@ -220,7 +223,9 @@ export function tickFormat(ticks: any[], format?: string): ((n: number | { value
         }
     }
     const formatter = createNumberFormatter(options);
-    return (n) => formatter(Number(n));
+    // A bigint tick renders full-precision rather than narrowing to Number; pin en-US grouping to match
+    // createNumberFormatter and the rest of the chart's numeric ticks (toLocaleString() defaults to the runtime locale).
+    return (n) => (typeof n === 'bigint' ? n.toLocaleString('en-US') : formatter(Number(n)));
```
lsjroberts: I would assume we want to apply the user's tick formatter to bigints as well?

packages/ag-charts-enterprise/src/features/status-bar/statusBar.ts
```diff
@@ -392,7 +392,9 @@ export class StatusBar extends AbstractModuleInstance implements _ModuleSupport.
 
             value.setFont(opts[labelStyle]);
             value.fill = opts[labelStyle].color;
-            value.text = typeof datumValue === 'number' ? formatter.format(datumValue) : '';
+            // Intl.NumberFormat.format accepts bigint natively, so a bigint OHLC/volume renders full-precision.
+            value.text =
+                typeof datumValue === 'number' || typeof datumValue === 'bigint' ? formatter.format(datumValue) : '';
```
lsjroberts: Could use `isNumericValue()` here?

packages/ag-charts-enterprise/src/gradient-legend/axisTicks.ts
```diff
@@ -271,10 +271,14 @@ export class AxisTicks {
         const ticks: TickDatum[] = [];
         const domain = tickParams.nice ? this.scale.niceDomain(tickParams) : this.scale.domain;
         const rawTicks = this.scale.ticks(tickParams, domain)?.ticks ?? [];
-        const fractionDigits = rawTicks.reduce((max, tick) => Math.max(max, countFractionDigits(tick)), 0);
+        const fractionDigits = rawTicks.reduce<number>(
```
lsjroberts: Given we do `rawTicks.map(Number)` below, in case of bigints, should we be casting `rawTicks` to `number[]` here? Does it need to handle `bigint` as well?

packages/ag-charts-community/src/chart/data/__snapshots__/dataModel.test.ts.snap
```diff
@@ -667,7 +667,7 @@ exports[`DataModel > grouped processing - calculated grouping > should generated
     },
   ],
   "columnValueType": [
-    "number",
+    undefined,
```
lsjroberts, line 670: Is this expected?

packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
```diff
@@ -272,7 +305,7 @@ export class HistogramSeries extends CartesianSeries<HistogramSeriesTypes> {
                 if (isDate(xValue)) {
                     xValue = xValue.getTime();
                 }
-                if (!isNumber(xValue)) return [];
+                if (!isNumber(xValue) && typeof xValue !== 'bigint') return [];
```
lsjroberts: `!isNumericValue(xValue)`?

lsjroberts (changes requested):

packages/ag-charts-core/src/utils/time/ticks.ts
alantreadway: Agreed

packages/ag-charts-types/src/api/initialStateOptions.ts
alantreadway: Fixed in 60a8b030d3 — dropped the raw `| bigint` from `start`/`end`, leaving `AgStateSerializableDate | AgStateSerializableBigInt | string | number`.

These are serialised-state inputs restored via `MementoCaretaker.decode`, whose inner `JSON.stringify` throws on a raw bigint. The `AgStateSerializableBigInt` wrapper (`{ __type: 'bigint', value }`) is the JSON-safe form — mirroring how raw `Date` is already excluded here in favour of `AgStateSerializableDate`. The runtime `AgZoomRange` was never widened to bigint, so there's no save-path change, and the `{ __type: 'bigint' }` round-trip is covered at `memento.test.ts:120`. `build:types` passes across types/community/enterprise.

packages/ag-charts-types/src/chart/annotationsOptions.ts
alantreadway: Same fix in 60a8b030d3 — `ValueType` is now `string | number | AgStateSerializableDate | AgStateSerializableBigInt` (raw `bigint` removed).

The runtime annotation point type is `number | string | Date | undefined` and never sees bigint, and the existing round-trip test (`annotations.test.ts:738`) already exercises the `{ __type: 'bigint' }` wrapper form. No runtime change.

packages/ag-charts-core/src/state/memento.ts
alantreadway: Agreed it's worth tightening — and as you note it's separate from this PR. It's also pre-existing: the same gap exists for the `__type: 'date'` Invalid-Date path (a malformed wrapper flows through `guardMemento` rather than being rejected — see `memento.test.ts:153`). The encode path here is already JSON-safe.

Logged as a follow-up: **AG-17522** — drop unrecognised/invalid `{ __type }` wrappers in `decodeTypes` (bigint *and* date), and tighten the value-level guards in `zoomManager`/`annotationManager`/`legendManager`. Left out of this PR.

packages/ag-charts-core/src/utils/time/ticks.ts
alantreadway: Done in 19e39b01ca — bigint ticks now route through the same `createNumberFormatter` result as number ticks (`return (n) => formatter(typeof n === 'bigint' ? n : Number(n))`), so they honour the user's format string (prefix/suffix) instead of emitting a bare `toLocaleString`. en-US grouping is preserved (the formatter's bigint branch pins it).

Red→green test added in #7024: `tickFormat([0, 1], '$#{,} USD')!(9_007_199_254_740_993n) === '$9,007,199,254,740,993 USD'`.

packages/ag-charts-enterprise/src/features/status-bar/statusBar.ts
alantreadway: Done in b62f7bb738 — replaced the inline `typeof datumValue === 'number' || typeof datumValue === 'bigint'` with `isNumericValue(datumValue)`. Byte-identical predicate, just DRYer.

packages/ag-charts-enterprise/src/gradient-legend/axisTicks.ts
alantreadway: Good catch — though the real issue is the double Number-coercion rather than the cast: `rawTicks` is coerced to Number twice (once in the `fractionDigits` reduce, once in `rawTicks.map(Number)`, which feeds a formatter parameter that's actually unused).

`rawTicks` is already correctly typed `AgNumericValue[]` (from `LinearScale.ticks()`), so it should *not* be cast to `number[]` — that'd be a dishonest narrowing. Fixed in 693d4f43e7 by hoisting a single `const numericTicks = rawTicks.map(Number)` and reusing it. The per-tick `convert(tick)`/label deliberately keep the exact (raw) value.

packages/ag-charts-community/src/chart/data/__snapshots__/dataModel.test.ts.snap
alantreadway: Expected and intentional. The `count` column comes from `rowCountProperty` (`missingValue: 1`); the source rows have no `count` field, so every datum resolves as `result.missing` and `updateColumnTypeTracker` is never called for it.

This PR deliberately stopped defaulting unobserved columns to `'number'` (`dataExtractor.ts`: `columnValueType.push(typeTracker.type)` instead of `?? 'number'`) so that `assertColumnValueType` no longer false-positives on a column whose type was never observed. Both consumers null-guard: `aggregator.ts` treats `undefined !== 'date'` as non-date, and `resolvers.ts` early-returns on `actualType == null`.

packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
alantreadway: Done in b62f7bb738 — `!isNumericValue(xValue)` replaces `!isNumber(xValue) && typeof xValue !== 'bigint'` (De Morgan-identical). `isNumber`/`isBigInt` stay imported — both are still used for the bin-count and bin-edge checks.

packages/ag-charts-community/src/chart/data/data-model/extraction/dataExtractor.ts
```diff
@@ -445,8 +503,31 @@ export class DataExtractor<D extends object, K extends keyof D & string> {
     private warnMixedNumericColumn(seriesId: ScopeId, key: K, atIndex: number) {
         Logger.warnOnce(
             `Series "${seriesId}": column "${String(key)}" mixes 'number' and 'bigint' values ` +
-                `(first detected at row ${atIndex}). Each column must be uniformly typed. ` +
-                `The series renders empty; other series in this chart are unaffected.`
+                `(first detected at row ${atIndex}); the bigints are narrowed to Number and may lose precision ` +
+                `beyond ±2^53. Use one numeric type per column.`
```
olegat: nit: MIN_SAFE_INTEGER / MAX_SAFE_INTEGER are ±(2⁵³ - 1)
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MIN_SAFE_INTEGER
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

packages/ag-charts-community/src/chart/data/processors.ts
```diff
@@ -459,8 +500,8 @@ function animationValidationProcessValue(def: DatumPropertyDefinition<unknown>,
     let lastValue = column[0]?.valueOf();
     for (let d = 1; validation !== 0 && d < column.length; d++) {
         const keyValue = column[d]?.valueOf();
-        if (!Number.isFinite(keyValue) || lastValue > keyValue) validation &= ~ANIMATION_VALIDATION_ORDERED_KEYS;
-        if (Number.isFinite(keyValue) && lastValue === keyValue) validation &= ~ANIMATION_VALIDATION_UNIQUE_KEYS;
+        if (!isContinuous(keyValue) || lastValue > keyValue) validation &= ~ANIMATION_VALIDATION_ORDERED_KEYS;
```
olegat: Can `keyValue` be an ISO-8601 string like `'2026-06-09'`? If yes, then this may be incorrect, because `isContinuous()` will always return `false` on string types.

packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
olegat: `total` is of type `bigint | number` and `0n === 0` expands to false.

packages/ag-charts-enterprise/src/series/linear-gauge/linearGaugeSeries.ts
olegat: This can expand to `false` if the gauge value type change. For example you can end up with `5n === 5` which expands to `false`. The same issue is present in `radialGaugeSeries.ts` too.

packages/ag-charts-core/src/utils/data/numbers.ts
```diff
@@ -1,7 +1,71 @@
...
+    if (typeof a === 'bigint' || typeof b === 'bigint') {
+        return bothIntegral(a, b) ? BigInt(a) - BigInt(b) : Number(a) - Number(b);
+    }
+    return a - b;
+}
+
+/** Smaller of two operands via comparison, preserving an exact `bigint` (`Math.min` throws on bigint). */
+export function minValue(a: AgNumericValue, b: AgNumericValue): AgNumericValue {
+    if (typeof a === 'number' && typeof b === 'number') return Math.min(a, b);
+    return a < b ? a : b;
```
olegat, line 52: nit: `Math.min(NaN, x)` expands to `NaN` but `NaN < x ? NaN : x` expands to `x` because `NaN < x` is always false.

If the intent of `minValue` to be a type-cohesive version of `Math.min` then adding a check for `NaN` might be worth considering.

Same issue with the mirrored `maxValue` function.

packages/ag-charts-community/src/chart/data/data-model/utils/helpers.ts
```diff
@@ -39,13 +40,26 @@ export function toKeyString(keys: any[]): string {
...
-    const allFinite = mapped.every((v) => typeof v === 'bigint' || Number.isFinite(v));
-    return allFinite ? (mapped as unknown as [number, number]) : [];
+    return mapped.every(isFiniteNumericValue) ? mapped : [];
+}
+
+/**
+ * Extends a numeric `[min, max]` extent to include the zero baseline, so bars and areas anchor at zero.
+ * A bigint baseline is emitted as `0n` to keep both endpoints bigint; a numeric `0` would narrow them.
+ */
+export function extendDomainToZero(extent: ReadonlyArray<AgNumericValue>): AgNumericValue[] {
```
olegat: Do we need to worry about mixed number/bigint domains like `[number, bigint]`? ContinuousScale seems to make assumptions that the min/max are either both numbers or both bigints, so the return type `AgNumericValue[]` might be too permissive. A stricter return types like `[] | [number, number] | [bigint, bigint]` might be more suitable.

Same with `fixNumericExtent`

packages/ag-charts-community/src/chart/series/cartesian/barSeries.ts
```diff
@@ -820,23 +832,24 @@ export class BarSeries extends AbstractBarSeries<BarSeriesTypes> {
...
+        // Metadata only (tooltips/error-bars); geometry below uses the exact bigint, so narrowing here is fine.
+        const cumulativeValue = phantom ? prepared.yFilterValue! : (prepared.yFilterValue ?? Number(params.yEnd));
         const nodeLabelText = phantom ? undefined : prepared.labelText;
 
-        let currY: number;
-        if (phantom) {
-            currY = params.yEnd;
-        } else if (prepared.yFilterValue == null) {
+        // Stay in domain space (possibly bigint) so yScale.convert() positions proportionally; the cross-filter
+        // branch recombines in Number space, so a bigint baseline beyond MAX_VALUE degrades there rather than throws.
```
olegat: This reasoning is misleading. `yFilterValue` is already narrowed to a number at line 691, which is why the `Number(params.yStart)` is needed in the addition.

In fact, it's unclear whether the number-narrowing at line 691 is actually safe, because `createNodeDatumContext` processes that input data as `number | bigint`. I don't know what `yStart` and `yFilterValue` are meant for; but if they can go above `Number.MAX_VALUE` then there's a good chance of ending up with Infinity in the addition.

packages/ag-charts-core/src/state/memento.ts
alantreadway: https://ag-grid.atlassian.net/browse/AG-17522

packages/ag-charts-community/src/chart/series/cartesian/barSeries.ts
```diff
@@ -621,7 +631,8 @@ export class BarSeries extends AbstractBarSeries<BarSeriesTypes> {
             barWidth,
             range,
             yReversed: yAxis.isReversed(),
-            bboxBottom: yScale.convert(0),
+            // 0n so a bigint y-domain takes the full-precision convert() path; a numeric 0 against it yields NaN.
```
olegat: This claim is false. `ContinuousScale.convert()` accepts both `number` and `bigint` input types.

If we're seeing `ContinuousScale.convert(0) => NaN` then that's probably a bug, I would expect `ContinuousScale.convert(0)` to return a finite number.

The same false claim is made at lines 1076 and 1131.

packages/ag-charts-enterprise/src/series/ohlc/ohlcSeriesBase.ts
```diff
@@ -115,10 +117,11 @@ class OhlcSeriesNodeEvent<
 interface PreparedOhlcNodeDatumState {
     datum: any;
     xValue: any;
-    openValue: number;
-    closeValue: number;
-    highValue: number;
-    lowValue: number;
+    // bigint-capable so yScale.convert() keeps full precision; narrowed to number for the stored datum.
```
olegat: The comment contradicts the code. bigints are stored as-is in the datum and not narrowed to a number.

packages/ag-charts-community/src/chart/data/data-model/extraction/dataExtractor.ts
alantreadway: Good catch — fixed in 587ae3c03b. The warning now reads `…may lose precision beyond ±(2^53 - 1) (Number.MAX_SAFE_INTEGER)…`, with the matching inline-snapshot assertion updated. (Also corrected the sibling `lossy beyond ±2^53` comment in `dataExtractor` and the `2^53` reference in `memento.encodeTypes`.)

packages/ag-charts-community/src/chart/data/processors.ts
alantreadway: Confirmed and fixed in 833e93140b. ISO strings (and bigints) were dropped for ordered/unique detection in **two** places: `updateKeyTracker` (the key path feeding `KEY_SORT_ORDERS`) skipped any non-`number` value, and `animationValidationProcessValue` (this value path) ran `isContinuous()` on the raw ISO string (→ `false`). Both now resolve bigint exactly / ISO via `timeValueToNumber` before comparing, so an ordered ISO/bigint series keeps its smooth update animations. Render output was already correct — only animation enablement was affected.

Regression tests added in #7024 (10d7ff243e): `KEY_SORT_ORDERS` is now asserted for ascending bigint and ISO 8601 keys.

packages/ag-charts-community/src/chart/series/cartesian/histogramSeries.ts
alantreadway: Fixed in 833e93140b — `Number(total) === 0`, so a bigint `0n` total (a non-empty bin summing to exactly zero) also skips its label. Regression test in #7024 asserts the bin's `aggregatedValue` is `0n` and its `label` is `undefined`.

packages/ag-charts-enterprise/src/series/linear-gauge/linearGaugeSeries.ts
alantreadway: Fixed in 833e93140b for both linear and radial gauges — the equality now compares via `Number(...)`, consistent with the count-up tween (which already narrows to Number), so a value type change (`5n` → `5`) doesn't force a no-op animation.

packages/ag-charts-core/src/utils/data/numbers.ts
alantreadway: Agreed — done in 833e93140b. `minValue`/`maxValue` now surface a NaN number operand like `Math.min`/`Math.max` instead of returning the bigint. (Inputs are NaN-free by contract upstream, but the type-cohesive behaviour now matches `Math.min`/`max`.)

packages/ag-charts-community/src/chart/data/data-model/utils/helpers.ts
alantreadway: Good call — tightened in 833e93140b, though not quite to the 2-tuple. `extendDomainToZero` genuinely is `[min, max]`, so it's now `[] | [number, number] | [bigint, bigint]`. But `fixNumericExtent` can't be a 2-tuple: `domainForClippedRange` flat-maps multiple key domains (e.g. OHLC's `[highMin, highMax, lowMin, lowMax]`), which it must preserve for the scale to span the full range — so it's `number[] | bigint[]` (length-general, still no mixed pair). I caught this when the 2-tuple version clipped the OHLC/box-plot/candlestick y-domains; the enterprise snapshot tests flagged it.

packages/ag-charts-community/src/chart/series/cartesian/barSeries.ts
alantreadway: You're right — the comment conflated the two branches. Reworded in 587ae3c03b: the non-filtered branch keeps `params.yEnd` in domain space (possibly bigint) for a full-precision `convert()`; the cross-filter branch is Number space precisely because `yFilterValue` is already Number-narrowed at line 691 (hence the `Number(params.yStart)` in the addition).

packages/ag-charts-community/src/chart/series/cartesian/barSeries.ts
alantreadway: Confirmed — `convert(0)` returns `range[0]` (via the `x === d0` short-circuit), not NaN; the claim was wrong. Reworded all three sites (634/1076/1131) in 587ae3c03b to state the real reason: a same-typed zero (`0n` / `zeroLike`) keeps a bigint baseline on the full-precision bigint `convert()` path, where a numeric `0` narrows to the Number path. No code change — the `0n`/`zeroLike` was correct, only the rationale wasn't.

packages/ag-charts-enterprise/src/series/ohlc/ohlcSeriesBase.ts
alantreadway: Right — the values are stored as `AgNumericValue` (bigint preserved), not narrowed to Number. Reworded in 587ae3c03b.

lsjroberts (approved):
Changes related to my feedback LGTM

packages/ag-charts-community/src/scale/continuousScale.ts
```diff
@@ -187,12 +192,32 @@ export abstract class ContinuousScale<D extends number | Date, I = number> exten
...
     }
 
+    // Returns the exact (possibly bigint) endpoint; d0Cache/d1Cache order even when bigints narrow to equal Numbers.
+    private exactEndpoint(index: 0 | 1): D {
+        return ((index === 0 ? this.d0Big : this.d1Big) ?? this._domain[index]) as D;
+    }
+
+    override get domainMin(): D | undefined {
+        if (this._domain.length < 2) return this._domain.at(0);
+        return this.exactEndpoint(this.d0Cache <= this.d1Cache ? 0 : 1);
```
olegat: My Claude Code session is flagging this (and also `domainMax()` mirror) a low-risk vulnerability, but I do not understand what it's saying:

> The min/max endpoint is chosen by comparing the Number-narrowed caches
> d0Cache/d1Cache. If two distinct bigint endpoints collapse to the same
> double (differ below one ULP at that magnitude) AND are stored
> descending, d0Cache === d1Cache so index 0 is always picked, returning
> the larger bigint as the minimum -- defeating the exact-endpoint purpose
> of these getters. Unaddressed since round 1. Near-unreachable in practice
> because normalizeContinuousDomains emits [min,max] ascending, so a
> descending bigint domain only occurs if one is set directly. Defensive
> fix: break the d0Cache===d1Cache tie on the exact bigints d0Big/d1Big.

alantreadway: Ok, I think this should be a quick fix

alantreadway: Practical example of what its trying to show - a result of `1` (end of domain) is not reachable:
```
> d0 = 1n*(2n**100n)
1267650600228229401496703205376n
> d1 = d0 + 2n
1267650600228229401496703205378n
> d0Cache = Number(d0)
1.2676506002282294e+30
> d1Cache = Number(d1)
1.2676506002282294e+30
> d0Cache <= d1Cache ? 0 : 1
0
```

alantreadway: Good catch — the flag was right, if narrow. `domainMin`/`domainMax` were ordering the endpoints on the Number-narrowed caches (`d0Cache`/`d1Cache`). Two bigints closer than one ULP near 2^53 narrow to the same double, so a **descending** bigint domain tied on the caches and returned the larger endpoint as the minimum — the exact `d0Big`/`d1Big` were retained but never consulted for ordering.

Fixed in f7f981e663: order on the exact endpoints instead — `startIsMin() => exactEndpoint(0) <= exactEndpoint(1)`. Relational `<=` on bigints is exact, and for number/Date domains it's identical to the old cache comparison, so there's no behaviour change outside the narrow-collapse case.

On why it slipped through: the existing tests covered descending-with-distinct-Numbers and ascending-narrow-collapse separately, but not the intersection (descending **and** narrow-collapse) — and the render path always normalises domains ascending, so it was unreachable end-to-end. Added that missing case to `linearScale.test.ts` in the tests PR (#7024); verified it fails before the fix and passes after.

PR #7024, opened 2026-06-05
AG-16608, AG-16654 Bigint/ISO enterprise render-test coverage

alantreadway:

Enterprise render-test coverage for bigint and ISO 8601 datetime support. This PR was restacked: it now contains only the enterprise-package tests, with the runtime fixes and shared test fixtures moved to #7052 (its base), and the core unit / community render tests split into sibling PRs (#7053, #7054) for parallel review.

## What's covered

- Bigint render snapshots across enterprise series (range bar/area, waterfall, radial bar/column, nightingale, radar, heatmap, treemap, sunburst, sankey, gauges) and axes (time, ordinal-time, polar, crosslines), including out-of-safe-range magnitudes and aggregation paths.
- ISO 8601 datetime string values on time/unit-time axes.
- Interaction probes in their 1:1 feature modules: zoom, chart sync, annotations, crosshair.
- Value-preserving tests for the API widening in #7052: the same value as `number` and as `bigint` must render pixel-identically (heatmap `colorScale` domain/stops, annotation `y`/channel heights, zoom `rangeX` endpoints, map-marker/map-line `sizeDomain`). Each was verified to FAIL against the un-widened code, using a single-chart `chart.update()`/`setState()` pattern (the mock canvas only tracks the first chart per test).

- **Product fix**: mixed number/bigint extents (e.g. a literal `0` baseline against bigint data) collapsed waterfall, radial-column, radial-bar and nightingale domains at high magnitudes — `getSeriesDomain` now uses `zeroLike` baselines. Exposed by the de-vacuoused pixel-identity helper (see #7052).

Six radial-column/nightingale bigint baselines were regenerated: the old baselines encoded the broken pre-fix render (collapsed domains). New renders visually verified. All other image snapshots are for new tests only.


packages/ag-charts-enterprise/src/features/sync/chartSync.test.ts
```diff
@@ -97,4 +97,38 @@ describe('ChartSync', () => {
...
+            // domain to match.
+            charts = [
+                AgCharts.create(bigintLineChart(bigintData)),
+                AgCharts.create(bigintLineChart([...bigintData, { x: 4, y: BIG * 8n }])),
+            ];
+            await waitForAllChartStability();
+
+            const yDomain = (c: AgChartInstance<AgCartesianChartOptions>) =>
+                deproxy(c).axes.find((axis) => axis.direction === ChartAxisDirection.Y)!.dataDomain.domain;
+            expect(yDomain(charts[0])).toEqual(yDomain(charts[1]));
```
olegat: That test description says that the expectation is for both domains to be equal to their union (`[BIG, BIG*8n]`), however a bug where both domains are too small (e.g. `[BIG, BIG*4n]`) will also satisfy this assertion. Explicitly comparing the domains to a constant like `[BIG,BIG*8n]` would guard against this.

alantreadway: Good point — now pinned to the explicit constant in 1ca0aba332. The converged domain is `[0, BIG * 8n]` (the nice number axis pulls the low end to 0), asserted on both charts.

olegat (approved):

PR #7034, opened 2026-06-08
AG-17484 Waterfall: undefined datum and original itemId for total/subtotal bars

iMoses:
https://ag-grid.atlassian.net/browse/AG-17484

Waterfall total/subtotal bars are synthetic — they come from the `totals` series option and have no row in the user's `data` array. Previously the synthetic bar's `datum` flowing into user callbacks was an internal object (`{ totalType, index, axisLabel }`) that lacked `yKey`, so any callback reaching into `params.datum[yKey]` got `undefined` and could throw.

## What changed
- Synthetic total/subtotal bars now expose `datum: undefined` across **every** callback surface: `tooltip.renderer`, per-item `tooltip.renderer`, `itemStyler`, label `formatter`, and the `seriesNodeClick` / `seriesNodeDoubleClick` / `activeChange` events. Real (positive/negative) bars keep their original user object.
- Real bars' node-event `itemId` is now the **original data-array index** rather than the augmented index that synthetic bars shifted. It is emitted as a string so the active-state round-trip (`getItemId` → `findNodeDatum`) stays consistent: real bars resolve via `node.itemId` (string branch), synthetic bars via their numeric `datumIndex`.
- New `totalLabel` callback param on `itemStyler`, `tooltip.renderer` and label `formatter`: the total/subtotal axis label (from the `totals` option), `undefined` for positive/negative bars.
- Fixed a highlight regression surfaced by the above: the item-highlight overlay (the layer that produces `highlightState: 'highlighted-item'`) was gated on the highlighted datum being truthy, so synthetic bars with `datum: undefined` never rendered it and their `itemStyler` only ever saw `'unhighlighted-item'`. `cartesianSeries` now discriminates item highlights by a concrete `datumIndex` (series-level highlights use `NaN`), matching `isItemHighlighted`.
- `itemType` (`'positive' | 'negative' | 'total' | 'subtotal'`) is unchanged and remains the way to identify synthetic bars.

## Breaking change
Callbacks/events for total and subtotal bars now receive `datum: undefined` (was an internal object). For real bars, `event.itemId` is now the original data index as a `string` (was a shifted number). Branch on `itemType === 'total' | 'subtotal'` to detect synthetic bars. The `totalLabel` addition is additive (new optional param).

## Test plan
- Unit tests in `waterfall.test.ts` assert the contract across `tooltip.renderer`, `itemStyler`, node `datum`, the `itemId` round-trip via `findNodeDatum`, `totalLabel`, and a regression test that highlighting a synthetic bar invokes its `itemStyler` with `'highlighted-item'` (TDD: confirmed failing before the fix).
- Updated `callbacks.test.ts` assertions + regenerated the affected `activeChange` and waterfall `formatter`/`itemStyler` snapshots (synthetic `datum`→undefined; real-bar `itemId` number→string; new `totalLabel` field).
- Full `ag-charts-community` and `ag-charts-enterprise` suites pass (pre-existing, unrelated `text.test.ts` / `treemapSeries.test.ts` sub-pixel snapshot failures excepted — tracked separately).
- Verified end-to-end in a headless browser against the built bundle: synthetic bars get `datum: undefined`; real bars emit string original-index `itemId`; the total's `itemStyler` observes `'highlighted-item'` on hover and keyboard focus.

Fix #AG-17484


alantreadway (approved):
LGTM

PR #7041, opened 2026-06-08
AG-17134 Enforce charts CSP (report-only → enforce)

taktran:
https://ag-grid.atlassian.net/browse/AG-17134

Flips the charts CSP from **report-only to enforced**. `getCspHtaccessBlock` switches to `enforce` mode, which unsets **both** inherited headers (the grid root CSP and the legacy wildcard) and sets the tightened policy as the enforced `Content-Security-Policy` on `/charts` pages. Applies to staging + production builds.

**Deploy ordering:** charts (and studio) must be enforced in production **before** grid root is enforced — `/charts` inherits the grid root policy, so this PR going live first is required.

**Pre-deploy check:** confirm the charts prod report-only window is clean (page-load crawl + a consent-accept / pricing-form / example pass). After this is the real enforcement — any unfixed violation will now block.

Fix #AG-17134


alantreadway (approved):
LGTM

PR #7048, opened 2026-06-09
AG-17134 Enforce charts staging CSP on latest (production stays report-only)

taktran:
## What

Enable the tightened CSP in **enforce** mode on **staging only** on `latest`. Production stays **report-only** on `latest` for now — it will be flipped to enforce after the release-branch enforce rollout (charts → studio → grid) is deployed and verified in production.

Makes the `.htaccess` CSP mode env-conditional in `getHtaccessContent()`:

```ts
const cspMode = env === 'staging' ? 'enforce' : 'report-only';
```

In enforce mode `getCspHtaccessBlock` unsets both inherited headers (the grid root policy and the legacy wildcard) and sets the tightened policy as the enforced `Content-Security-Policy`.

## Why staging can land independently

On production, `/charts` inherits the grid-root CSP, which is why the production rollout is ordered (charts/studio before grid). Staging is a separate host with no such inheritance, so charts staging can enforce without waiting on the others.

## Testing

Generated the CSP block for both envs:
- **staging:** `Header always set Content-Security-Policy "…"` (enforced)
- **production:** `Header always set Content-Security-Policy-Report-Only "…"` (unchanged)

Formatter applied; file lints clean.

Part of AG-17134.


iMoses (approved):
LGTM

PR #7052, opened 2026-06-10
AG-16608 Widen remaining data-space options to AgNumericValue

alantreadway:
Fix #AG-16608

## What

Follow-up to the bigint render-hardening stack: an inventory of `ag-charts-types` found data-space options still typed plain `number`, so `bigint` data could not be expressed. This PR widens them and hardens the consuming runtime paths.

- **Project rule**: adds `.rulesync/rules/numeric-values.md` documenting `number` vs `bigint` (`AgNumericValue`) conventions, the bigint-safe utilities in `ag-charts-core`, and the widening checklist.
- **Axis bounds & interval**: number/log/angle-number/radius-number axes now extend `AgContinuousAxisOptions<AgNumericValue, AgNumericValue>`, covering `min`/`max`/`preferredMin`/`preferredMax` and `interval.step`. Validators switch to `numericValue`/new `positiveNumericValueNonZero`. `LogAxis` narrows before `Math.*` and rejects a `0n` endpoint like `0`.
- **Cross-lines**: number/log/angle-number/radius-number cross-lines widen to `AgNumericValue`; time-axis cross-lines align with `AgTimeValue` (`Date | number | bigint | string`).
- **Colour scale**: `AgColorScale.domain` and `AgColorScaleColorStop.stop` widen to `AgNumericValue` (runtime was already hardened).
- **Annotations**: `AgAnnotationPoint.y`, channel `height`/`startHeight`/`endHeight` widen; `getGrouping` no longer misreads a raw bigint as a grouping object; channel/measurer maths use the bigint-safe helpers.
- **Zoom range**: `AgZoomRange`/`AgZoomEventRange` `start`/`end` accept `bigint`; `rangeAlignment` and `zoomOnDataChange` domain narrowing handle bigint domains.
- **Size domains**: bubble `domain` and map-marker/map-line `sizeDomain` widen to `AgNumericValue` (validators: `numericValue` / new `positiveNumericValue`); the consuming `LinearScale.domain` path already accepts bigint. Note this overlaps with #7047's rename of bubble `domain` to `sizeDomain` — whichever lands second takes the rebase.
- **Runtime fixes** (moved here from the tests PR so the parallel test PRs share one base): pie/donut angle normalisation coerces bigint values; label formatters may return bigint. Each carries its regression test.
- **Shared test fixtures**: `bigintExamples.ts` (exported via `ag-charts-community-test`) and the enterprise `renderEnterpriseChartImage` helper, used by the stacked test PRs.
- **Lint rule**: the in-house `aglint/change-detection` rule now recognises `bigint` and union element types (e.g. `readonly [number | bigint, ...]`) as allowed primitives for `@SceneArrayChangeDetection` — its allow-list predated bigint. Covered by new rule fixtures; the rule-test text snapshot churn is line-number shifts only.

A custom `interval.step` remains a Number concept at runtime (narrowed via `Number()`), consistent with the documented bigint limitation for custom intervals (AC #17).

Not in scope: ratio/pixel options, key-based options.

## Tests

`build:types` and lint pass for types/core/community/enterprise. Full community + enterprise suites pass except pre-existing image-segment snapshot failures (`text.test.ts`, `treemapSeries.test.ts`) which reproduce identically on `latest`. Cross-line warning expectations updated for the new validator description. Value-preserving regression tests follow on the stacked test PRs (#7053 core unit, #7054 community render, #7024 enterprise render).

## Review-pass follow-ups

A review pass over the stack landed further fixes here (the stack base):

- **Pixel-identity test helper rewritten**: the shared `expectPixelIdenticalAcrossUpdate` helper previously compared a destroyed chart's canvas with itself (the mock canvas only tracks the first chart per test), making the value-preserving tests vacuous. It now reuses a single chart via `chart.update()` with a non-blank-image guard. De-vacuousing it exposed three real rendering bugs, fixed below.
- **Exact Number conversion against bigint domains**: `ContinuousScale.convert` returned NaN for Number values (e.g. the zero baseline of an area series straddling zero) when the domain endpoints were bigint; Numbers are now decomposed and routed through the exact bigint conversion path.
- **Bigint histogram bins**: the bigint bin path produced different bin boundaries from the Number path; bins are now split evenly from the exact domain minimum, with a tick-aligned variant for the default-count path.
- **Log-axis narrowing**: bigint narrowing no longer emits a spurious out-of-range warning for invalid (undefined) domain endpoints.


packages/ag-charts-community/src/chart/axis/numberAxisBigInt.test.ts
```diff
@@ -67,4 +74,32 @@ describe('NumberAxis BigInt labels', () => {
...
+                },
+            },
+        };
+        prepareTestOptions(options);
+        chart = AgCharts.create(options);
+        await waitForChartStability(chart);
+
+        // The custom formatter returns the bigint verbatim, so labels are plain digit strings (no grouping).
+        expect(yAxisLabelText(chart)).toContain('1000000000000000000000');
+        expectWarningsCalls().toMatchInlineSnapshot(`[]`);
```
lsjroberts: This line shouldn't be needed? The mock console by default expects no calls.

packages/ag-charts-community/src/chart/data/processors.ts
```diff
@@ -422,18 +423,24 @@ function normalisePropertyFnBuilder({
         return result;
     };
 
+    // Normalisation maps to an angle/ratio with Number factors, so narrow bigint endpoints/values to
+    // Number — a bigint would otherwise throw when mixed. Missing values keep raw coercion semantics
+    // (null -> 0, undefined -> NaN) without warning via toNumber.
+    const toRenderValue = (value: unknown): number => (isFiniteNumericValue(value) ? toNumber(value) : Number(value));
```
lsjroberts: This mirrors `narrowDomainValue()` in `logAxis.ts`, could perhaps be a shared util?

packages/ag-charts-core/src/state/memento.ts
```diff
@@ -80,8 +80,12 @@ export class MementoCaretaker {
     private decode(originator: MementoOriginator, encoded: unknown) {
         if (encoded == null) return encoded;
         try {
+            // Encode on the way in too: a provided blob (e.g. initialState) may hold a raw bigint/Date,
+            // which JSON.stringify cannot serialise without the replacer.
             // eslint-disable-next-line @typescript-eslint/unbound-method
-            return JSON.parse(JSON.stringify(encoded), MementoCaretaker.decodeTypes) as unknown;
+            const json = JSON.stringify(encoded, MementoCaretaker.encodeTypes);
```
lsjroberts: I'm not convinced by this, the provided blob should already be encoded. It was intentional that we don't support incoming unencoded blobs.

packages/ag-charts-types/src/chart/zoomOptions.ts
```diff
@@ -13,9 +14,9 @@ export type AgZoomOnDataChangeStrategy = 'reset' | 'preserveDomain' | 'preserveR
 
 export interface AgZoomRange {
     /** The start of the axis zoom range. */
-    start?: Date | string | number;
+    start?: AgTimeValue;
```
lsjroberts: This isn't technically correct since `string` here also represents a category (pending the change to `AgGroupingValueType` https://ag-grid.atlassian.net/browse/AG-13954?focusedCommentId=115775)

packages/ag-charts-types/src/chart/eventOptions.ts
```diff
@@ -135,8 +136,8 @@ export interface AgZoomEvent<TContext = ContextDefault> {
 }
 
 export interface AgZoomEventRange {
-    start?: Date | string | number;
-    end?: Date | string | number;
+    start?: AgTimeValue;
```
lsjroberts: Also not correct since `string` is also for categories.

packages/ag-charts-core/src/state/memento.ts
lsjroberts: If there's a specific product reason for this then it's probably fine

packages/ag-charts-community/src/chart/axis/numberAxisBigInt.test.ts
alantreadway: Agreed — removed in 5c1fb6dc02 (the `setupMockConsole` afterEach already asserts no warnings).

packages/ag-charts-community/src/chart/data/processors.ts
alantreadway: Done in 5c1fb6dc02 — extracted a shared `narrowToNumber()` into `ag-charts-core` (`utils/data/numbers.ts`) and switched both `logAxis.ts` and the normalise processor to it.

packages/ag-charts-types/src/chart/zoomOptions.ts
alantreadway: Good point — reverted to the explicit `Date | string | number | bigint` union in 5c1fb6dc02, so the `AgTimeValue` JSDoc (which describes strings as strict ISO time values) doesn't misdocument category strings here. Happy to revisit once AG-13954's `AgGroupingValueType` lands.

packages/ag-charts-types/src/chart/eventOptions.ts
alantreadway: Same fix as zoomOptions — reverted to the explicit union in 5c1fb6dc02.

packages/ag-charts-core/src/state/memento.ts
alantreadway: Following your lead here — reverted the decode change (and its unit test) in f33bc5baf2, so decode keeps the strict pre-encoded contract. The bigint `{ __type }` encode/decode replacers are unchanged.

packages/ag-charts-core/src/utils/time/ticks.ts
```diff
@@ -298,22 +298,50 @@ export function createBigIntTicks(start: bigint, stop: bigint, count: number): b
...
+ * power of 10; that snap is scale-dependent so it is deliberately not replicated here — an even split
+ * keeps the bin count and boundary proportions identical for a BigInt domain that is a pure scaling
+ * of a Number domain.
+ */
 export function createBigIntBins(start: bigint, stop: bigint, count: number): [bigint, bigint][] {
     const lo = start < stop ? start : stop;
     const hi = start < stop ? stop : start;
+    if (lo === hi) return [[lo, hi]];
+
+    const segments = BigInt(Math.max(1, Math.floor(count) || 1));
```
olegat: The intent of the `|| 1` is unclear. `x || 1` is more or less equivalent to `x !== 0 ? x : 1`, which appears to be redundant because of the `Math.max(1, ..)`.

lsjroberts (approved):
LGTM

packages/ag-charts-core/src/utils/time/ticks.ts
alantreadway: Good spot — the `|| 1` was actually guarding NaN (`Math.max(1, NaN)` is NaN and `BigInt(NaN)` throws) rather than zero, but that was far from obvious. Rewritten with an explicit `Number.isFinite` check in eeb7b40846.

PR #7053, opened 2026-06-10
AG-16608, AG-16654 Bigint/ISO core unit-test coverage

alantreadway:

Core unit-test coverage for bigint and ISO 8601 datetime support, split out of the former combined tests PR so the three test PRs (core unit / community render / enterprise render) can be reviewed in parallel. Stacked on #7052, which carries the API widening, the runtime fixes and the shared test fixtures.

## What's covered

- `memento` encode/decode round-trips for bigint payloads (including the `{ __type: 'bigint' }` wire form and malformed payloads).
- `validation` — `numericValue` / `positiveNumericValueNonZero` validator acceptance and descriptions.
- `aggregation`, `ticks` — bigint domain arithmetic and full-precision tick generation.
- `iso8601`, `numbers`, `strings` — parsing/narrowing/stringification helpers used by the render paths.

No implementation changes in this PR (the fixes these tests exercise live in #7052).


packages/ag-charts-core/src/utils/time/ticks.test.ts
```diff
@@ -1,4 +1,29 @@
...
+    createBigIntTicks,
+    createTicks,
+    niceBigIntDomain,
+    tickFormat,
+} from './ticks';
+
+describe('tickFormat', () => {
+    test('formats a bigint tick with the chart-wide en-US grouping', () => {
+        // Bigint ticks must pin en-US grouping to match createNumberFormatter, regardless of the
+        // runtime's default locale.
```
olegat, line 13: Using a hard-coded 'en-US' locale in `format.util.ts` and other places feels incomplete.

Perhaps it would make more sense to add string to `packages/ag-charts-locale/src/en-US.ts` (e.g. `'iso-3166': 'en-US'`) and use that; only falling back to 'en-US' if the localisation string is not available.

alantreadway: Agreed the hard-coded locale is a smell, but it's not specific to the bigint paths — `format.util.ts` already pins `en-US` for its pre-existing `percentFormatter` and `formatNumber` caches, and the new bigint branches deliberately match that so a bigint renders identically to an equal `number` tick. Threading a locale through `ag-charts-locale` (with an `en-US` fallback) would need to cover all of those formatters together, which feels like a follow-up ticket rather than something to fold into this test-only PR — happy to raise one if you agree.

olegat: SGTM. It's a pre-existing issue so we can descope and follow-up later.

olegat (approved):

PR #7054, opened 2026-06-10
AG-16608, AG-16654 Bigint/ISO community render-test coverage

alantreadway:

Community render-test coverage for bigint and ISO 8601 datetime support, split out of the former combined tests PR so the three test PRs (core unit / community render / enterprise render) can be reviewed in parallel. Stacked on #7052, which carries the API widening, the runtime fixes and the shared test fixtures (`bigintExamples.ts`).

## What's covered

- Bigint render snapshots for the cartesian series (line/area/bar/scatter/bubble/histogram), pie/donut, and the number/log axes — including out-of-safe-range magnitudes that force the full-precision bigint path, and high-volume datasets that flow through aggregation.
- ISO 8601 datetime string x-values on time axes.
- Value-preserving tests for the API widening in #7052: the same value supplied as `number` and as `bigint` must render pixel-identically (axis `min`/`max`/`preferredMin`/`preferredMax`/`interval.step`, cross-line `value`/`range`, bubble size `domain`). Each was verified to FAIL (real pixel diffs) against the un-widened validators.
- The probes relocated into their 1:1 feature test modules (no `*BigInt.test.ts` modules), retiring `numberAxisBigInt.test.ts` into `numberAxis.test.ts`.

One histogram bigint baseline was regenerated: the old baseline encoded the pre-fix bin boundaries (see the bin-semantics fix on #7052). All other image snapshots are for new tests only. Also removes four stale committed `__diff_output__` artefacts.


packages/ag-charts-community/src/chart/data/aggregateFunctions.test.ts
```diff
@@ -1,8 +1,24 @@
...
+import { accumulatedValue, addAccumulated, area, sum, sumValues, trailingAccumulatedValue } from './aggregateFunctions';
 
 describe('aggregateFunctions bigint support (AG-16608)', () => {
+    describe('sumValues', () => {
+        it('splits numbers into negative and positive sums', () => {
+            expect(sumValues([1, -2, 3, -4])).toEqual([-6, 4]);
+        });
+
+        it('sums bigint values in bigint so a large-magnitude total stays exact', () => {
+            // The negative side keeps its untouched Number seed; the positive side promotes to bigint.
```
olegat: Misleading comment: The negative side is a bigint `-3n` not a number seed.

alantreadway: Right you are — both seeds promote (the assertion is `-3n`). Comment fixed in 3caecb0022.

olegat (approved):

PR #7059, opened 2026-06-10
AG-16654 ISO 8601 time-column performance: sort order + parse-once epochs

alantreadway:

**Stacked on #7025** (docs/benchmarks) — uses the aligned `high-volume-iso-datetime` benchmark example for verification.

## Summary

Two fixes for ISO 8601 datetime string columns at high data volumes (1M rows), found via the benchmark comparison against the epoch-number/`Date` equivalents:

**1. Sort-order detection (zoom path).** `valuesSortOrder` bailed out for string values, so sorted ISO columns were flagged unsorted and zooming fell back to the O(n) `domainForVisibleRange` scan — ~3.9s of `Date.parse` per wheel event — instead of the O(log n) binary-search path. ISO strings now resolve to epoch ms during detection.

**2. Parse-once epoch columns (initial-load path).** Initial load re-parsed the same 1M strings in four separate passes: domain extension allocated a `Date` per value via `coerceIso8601Date`, the extraction key tracker, aggregation bucketing and sort-order detection each ran their own `Date.parse` sweep. A shared parse-once cache in `ag-charts-core` (`ensureEpochColumn`/`getEpochColumn`/`invalidateEpochColumn`, keyed by column identity) replaces aggregation's private cache:

- `ContinuousDomain.extend` ignores strings; the extractor materialises the epoch column for ISO date columns in one post-pass and extends domains from it (ISO domains are now epoch ms, not `Date`).
- Whole-column domain recomputation substitutes the epoch column for continuous domains; incremental in-place mutation invalidates the cache.
- Sort-order detection reuses the cached epoch column.

## Results (high-volume-iso-datetime, 1M rows, vs axes-1M-time epoch baseline)

| Test case | Before | After | Epoch baseline |
| --- | --- | --- | --- |
| Zoom | 3937ms | 157ms | 148ms |
| Initial load | 629ms | 293ms | 123ms |

The remaining initial-load gap is the single irreducible parse pass plus ISO regex type-tracking during extraction.

Also adds `.rulesync/rules/test-harness-contracts.md`, a repo-local rule documenting the JSDOM mock-canvas contract (first-chart-only snapshots, `setupMockConsole`, fixture-extraction discipline, caching-test checklist), prompted by the stack-wide review pass.

## Test plan

- New unit tests: `sortOrder.test.ts` (number/bigint/ISO/Date columns), `epochColumns.test.ts` (cache/invalidations), DataModel ISO-column tests (raw strings retained, epoch domains, cache materialisation) ✓
- `yarn nx test ag-charts-core` ✓ / `ag-charts-community` ✓ / `ag-charts-enterprise` ✓ (only pre-existing local image-snapshot failures in `text.test.ts`/`treemapSeries.test.ts`, present without these changes)
- `yarn nx build:types` ✓ / `yarn nx lint` ✓ for core + community

Fix #AG-16654


packages/ag-charts-core/src/utils/data/epochColumns.ts
```diff
@@ -0,0 +1,37 @@
...
+ * later) must still parse, and the result is cached by identity so a wrong guess would never
+ * self-correct.
+ * @returns the input `values` unchanged when the column holds no strings.
+ */
+export function ensureEpochColumn(values: unknown[]): unknown[] {
+    const cached = epochColumnCache.get(values);
+    if (cached !== undefined) return cached;
+
+    const hasStrings = values.some((v) => typeof v === 'string');
+    const converted = hasStrings ? values.map(parseEpochValue) : values;
```
olegat, line 26: This funnels into `Date.parse` by calling `parseEpochValue -> timeValueToNumber`, which allows non-ISO date strings like "Jan 2024". Is that a concern?

alantreadway: Good question — it isn't reachable with non-ISO strings in practice: a string only ever enters a date-typed column through the strict `isISO8601` gates upstream (`dataExtractor.ts` value classification, `processors.ts` continuity check, `sortOrder.ts`), which reject things like "Jan 2024" before any column reaches `ensureEpochColumn`. So `Date.parse`'s leniency is dead defence rather than an accepted input path. I've added a comment documenting that invariant at the parse site in f1126e3bc1.

olegat (approved):

PR #7061, opened 2026-06-10
AG-15933 image labels: rename verticalAlign/cornerRadius, drop border, fix blank-url & same-origin example images

iMoses:
Addresses QA feedback on the Images in Labels feature (AG-15933, comment 118793). The feature is unreleased, so the option renames are not breaking.

## Option surface changes

- **`verticalAlign`: `'alphabetic'` → `'baseline'`**, and the surface trimmed to `'baseline' | 'top' | 'middle' | 'bottom'` (dropped `'hanging'` and `'ideographic'`). `'baseline'` is the public spelling of the canvas `'alphabetic'` baseline; it is translated back at the render boundary via a new `toCanvasTextBaseline` helper, so internal layout is unchanged. Defaults are per-type: inline images default to `'baseline'`, block images (`block: true`) to `'middle'`.
- **`borderRadius` → `cornerRadius`** on image segments.
- **Removed the per-image `border` option.** Text segments never supported a border, and it did not reserve room for a wider stroke.

## Behaviour changes

- **Empty image `url`** now paints the `backgroundFill` box and logs a single warning instead of silently rendering nothing.
- **Overflowing images wrap to their own line** rather than being dropped, keeping width-shrinking monotonic. The text line left behind is counted under `maxHeight` so a wrap is only taken when both lines fit.
- **Text-segment edge whitespace is preserved across wraps** — the gap between a segment and an adjacent image (e.g. a flag then `" Germany "`) is restored onto the first/last content line after a wrap break trims it.
- **Image segments are clipped to `cornerRadius`** when a corner radius is set.

## Examples now use same-origin assets

The doc examples no longer pull images from third-party CDNs (flagcdn.com): flags and brand icons are served as local `${baseWWWUrl}/example-assets/...` assets. Same-origin images don't taint the canvas, so chart export (`toDataURL`) works without any loader-level CORS handling.

## Out of scope (later chunks / separate tickets)

Germany label hidden and the validation-message ticket.

## Test plan

- `build:types` + `lint` pass across types, core, community, enterprise.
- Updated/added unit tests: `text`, `textWrapper` (wrapping/whitespace), `caption`, `overlay`, plus an empty-url warning test.
- Renamed/removed-option snapshots cleaned up (`border` snapshots removed, `cornerRadius`/`baseline` snapshots added).
- SVG icon glyphs in `text` and `treemap` snapshot tests are now drawn as vector `<path>`s instead of `<text>`, so they render identically on macOS and CI (system fonts differ); baselines regenerated accordingly.
- Docs (`fonts` page) and inline-image examples updated to the new option names and same-origin asset URLs.

Fix #AG-15933


alantreadway (approved):
LGTM

PR #7076, opened 2026-06-11
AG-17484 Waterfall callbacks: correct itemType/itemId and surface event metadata

iMoses:
## Summary

Follow-up to AG-17484 (PR #7034) fixing the waterfall callback/event contract and unifying how `itemId` is resolved across every surface.

1. **Label-formatter `itemType` was wrong.** It reported `total` for both total *and* subtotal bars; `itemStyler`/tooltip already distinguished them. The label-formatter param now receives the real `itemType` (the `subtotal → total` alias is removed). The bucket lookup (`item.total` covers both) is unchanged.

2. **`itemId` is now the single identifier across all callbacks and events**, resolved via `getItemId()` (`node.itemId → dataIdKey → datumIndex`):
   - Real (positive/negative) bars resolve via `dataIdKey`, then `datumIndex`.
   - Total/subtotal bars use the new public `totals[].itemId?: string` option as their stable identifier, falling back to the bar's `axisLabel` when omitted.
   - The item-level `itemStyler`, tooltip, label formatter, the label-level `itemStyler`, and `seriesNodeClick`/`seriesNodeDoubleClick`/`nodeContextMenuAction`/`activeChange` all now report the same `itemId`.

3. **Node/active events carry the bar type.** Synthetic bars correctly carry `datum: undefined`, so listeners get `itemType` (via a `WaterfallSeriesNodeEvent` subclass mirroring `HistogramSeriesNodeEvent`); `activeChange` carries it too, attached only when present so other series' events are unchanged.

## Removed

- The interim `totalLabel` callback/event field has been removed. It existed only to identify total/subtotal bars; `itemId` (with the `axisLabel` fallback) now does that job and is consistent with every other surface.

## Funnel & cone-funnel label `itemId`

Funnel and cone-funnel previously surfaced bespoke, inconsistent label identifiers: the label *formatter* received `valueKey`, while the label *itemStyler* received `stageKey` (funnel) / `valueKey` (cone-funnel) — neither matched the series' item-level/event identity. Both label surfaces now resolve `itemId` through the shared `getItemId()` (`dataIdKey → datumIndex`):

- With no `dataIdKey`, both report the positional `datumIndex`.
- With a `dataIdKey` set, both report the `dataIdKey` value.

This makes the formatter and styler consistent with each other and with the events/`activeItem` identity — the funnel node datum carries no `itemId`, so events already resolve to `datumIndex`. A shared `BaseFunnelSeries.resolveItemId()` helper guarantees funnel and cone-funnel use identical logic, and `itemId` is no longer stored on the funnel label datum (it is resolved on read).

## Cross-series behaviour change

Unifying the label-level styler on `getItemId` (in the shared `labelUtil`) means series that set a string `node.itemId` — treemap, sunburst, histogram — now receive that id as `itemId` in their **label** `itemStyler` instead of the positional `datumIndex`, aligning the label styler with those series' events. Per-series label *formatters* are unchanged. (Funnel/cone-funnel are handled explicitly above; sankey and chord pass `undefined` to the label styler and are unaffected.)

## Notes

- `itemType` is a base `SeriesNodeDatum` field and an `AgChartCallbackParams` concept, so `activeChange.itemType` is populated for any series that sets it (waterfall, OHLC, candlestick, range area). `AgItemType` is exported from the types package for the optional event field.

## Test plan

- `waterfall.test.ts`: label-formatter `itemType` distinction; real-bar `itemId` via `dataIdKey`/`datumIndex`; totals `itemId` + `axisLabel` fallback; node-click and `activeChange` carrying `itemType`/`itemId`.
- `callbacks.test.ts`: funnel label formatter/styler snapshots updated (both now report `datumIndex`, previously `valueKey`/`stageKey`); added cone-funnel label formatter/styler coverage mirroring funnel (new snapshots confirm parity with funnel).
- Regenerated label-styler snapshots (waterfall formatter + sunburst/treemap/histogram) — diffs are `itemId`-only.
- `build:types`, `lint` (0 errors), and the callbacks + waterfall suites pass. Pre-existing image-segment snapshot failures (`text.test.ts`, treemap block-image; 1–14 px) confirmed unrelated — they fail identically on clean `HEAD`.

Fix #AG-17484


alantreadway (approved):
LGTM

PR #7078, opened 2026-06-11
AG-16452 Switch CI benchmarks to browser-based runs on hosted runners

alantreadway:
https://ag-grid.atlassian.net/browse/AG-16452

Completes the switch to the browser-based benchmarks and migrates all GHA benchmark runs off the self-hosted `macos-benchmark` runner onto standard GitHub-hosted runners.

## What changed

- **Legacy Jest/jsdom benchmarks removed entirely**: `packages/ag-charts-{community,enterprise}/benchmarks/`, their nx targets (`benchmark`, `benchmark:report`, `benchmark:rebuild`), the ci.yml benchmark step, comparison/Slack tooling, and lint-config carve-outs.
- **Static-serve tooling**: benchmark pages are now served from a static `nx build ag-charts-website` output via a zero-dep server (`tools/benchmark/serve-static.js`, `serve-and-run.sh`) instead of the Astro dev server. Sites are built with `PUBLIC_SITE_URL` pinned to the port they are served from (example pages bake absolute library-script URLs at build time).
- **`benchmark.yml` reworked** (browser-only, `ubuntu-24.04`): resolve → parallel head/base builds (pruned ~210MB artifacts) → 4-shard example matrix → fan-in report. Each shard runs its examples against **both** head and base on the same runner VM, so the relative comparison stays valid on shared hardware. Shards are computed from `tools/benchmark/benchmark-examples.js` (shared with `browser-benchmark.ts`, so new examples are picked up automatically). Same path for nightly schedule, `workflow_dispatch` and `/benchmarks` PR comments.
- **Noise handling**: comparisons use per-test **median** timings; results with sample CV > 10% are flagged `~`; "notable regressions" require >10% **and** ≥2ms absolute delta (sub-10ms timings are timer quantisation).

## Noise validation (Phase 0 spike)

A/A runs (same commit, 5 repeats, 6 examples) on `ubuntu-24.04` via the temporary `benchmark-noise-spike.yml`:

| Run | median A/A error | p95 | Notes |
|---|---|---|---|
| [27367150228](https://github.com/ag-grid/ag-charts/actions/runs/27367150228) | 1.7% | 9.0% | p95 driven by one hit-testing case |
| [27368509989](https://github.com/ag-grid/ag-charts/actions/runs/27368509989) | 3.0% | 11.2% | noisier VM allocation |

Conclusion: typical single-comparison error is ~2–3% with tails to ~10% on specific test cases depending on VM allocation. Changes below ~10% on a single run should be treated as indicative only — consistent with the 10% notable threshold, soft-fail and human review. Full pipeline validation run: [27368508418](https://github.com/ag-grid/ag-charts/actions/runs/27368508418) — all green, **29 min wall-clock** (vs ~60 min on the macOS runner).

## Before merge

- [x] Remove the temporary `push` triggers from `benchmark.yml` and delete `benchmark-noise-spike.yml` + `tools/benchmark/noise-stats.js` (done in db79c9b98e).
- [ ] Companion change in **ag-dev-prompts**: `benchmarks.md`, `testing.md`, `playbook-bug-fix.md` rules and the optimize-series/benchmark-profile/sonar-fix skills still reference `yarn nx benchmark`; benchmarks rule should also document the `npm:<version>` base mode.
- [ ] After merge + one good production run: decommission the `macos-benchmark` self-hosted runner.

## Follow-up: build-time and long-pole optimisation (post-validation)

- **Benchmark-only website build** (`build:benchmarks` target): an `AG_BENCHMARK_ONLY_BUILD` flag filters `getStaticPaths` to the `vanilla/benchmarks` example pages (gallery/docs/options/themes-api routes emit nothing) and drops the thumbnail-generation dependency. CI head build: **3m54s vs 8m12s** (validated in run 27407539033). Old refs without the target fall back to the full build (both in `benchmark.yml` and `compare-browser-latest.sh`).
  - Nx loads `.env.<targetName>` per task, so a minimal `.env.build:benchmarks` provides `PUBLIC_BASE_URL=/charts` plus the Algolia keys (the benchmark-only chunk graph evaluates the search client at module load during prerender).
- **`data-selection-zoom` split by series family** into 4 examples (`-line-area`, `-bar`, `-financial`, `-xy`) — same 24 test cases, each family in a fresh page session. The monolith accumulated memory pressure over 24 cases x 1M points and dominated wall-clock (~18 min shard). All four validated locally: 4/4 success in 182s total. Sharding is back to plain round-robin at `AG_SHARD_COUNT: 5`.
- **ag-shared precommit gate fix**: `precommit-check.sh` now skips symlinks (staged blob is the target path, which prettier mangles). SYNC-LOG entry added.
- **Published-npm base mode** (`456c64960a`): base can be `npm:<version>` as well as a git ref — the head-built site is reused with its `dev/`-served library bundles replaced by the published UMD bundles (`tools/benchmark/swap-published-lib.sh`, validated locally against `npm:13.3.1`). This restores easy "current benchmarks vs any published version" comparisons (the old Jest `collect-version-benchmark.sh` capability), with no base site build. The **nightly schedule now compares `latest` against the most recently published npm version** (`npm view ag-charts-community version`) instead of the newest release branch: head's example set is measured, and the base build leg is skipped entirely. `workflow_dispatch` `compare-base` accepts either form.

Expected post-merge wall-clock: ~4 min builds + ~5-7 min balanced shards + 1 min report = **~12 min** (vs 29 min validated earlier, vs ~60 min on the old macOS runner). First comparison run after merge will report the old `data-selection-zoom` keys as removed and the new family keys as added (one-off).

## Follow-ups (out of scope)

- The `summary/` historical dashboard's writer (`collate-reports.js`) is gone; ticket needed to feed it from browser results or remove it.
- Shard balance: the shard containing `data-selection-zoom` dominates wall-clock (19 min vs 7 min) — could use weighted sharding.
- If sub-10% detection is ever needed: interleaved repeat runs (A/B/A/B) or higher harness sample counts.

Fix #AG-16452


iMoses (approved):
LGTM

PR #7080, opened 2026-06-12
AG-17386 Rename 'select' -> 'crossFilterSelect'

olegat:
https://ag-grid.atlassian.net/browse/AG-17386

In the past, this cross-filtering internals has confused coding agents into thinking that it's related to data-selection, which it isn't.

This sign-posting to hopefully help to avoid this confusion.

The `selectedKey` cannot be renamed because it's a (undocumented) key that is read from the user option; renaming this will certainly break things.


alantreadway (approved):
LGTM

PR #7083, opened 2026-06-12
AG-15505 Padding

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-15505

### DIFF
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -468,7 +468,7 @@ export abstract class Chart extends Observable implements ModuleInstance, ChartS
         // Publish processed options to chartState immediately so option-derived reads
         // (mode, padding, etc.) work for the rest of construction. `applyOptions` will
         // refresh this on each subsequent update.
-        ctx.chartState.setValue('options', options.processedOptions as ChartState['options']);
+        ctx.chartState.setValue('options', options.processedOptions as unknown as ChartState['options']);
 
         this.chartCaptions = new ChartCaptions(ctx);
         this.titleGroup.append(this.title.node);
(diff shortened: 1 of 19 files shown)


packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -468,7 +468,7 @@ export abstract class Chart extends Observable implements ModuleInstance, ChartS
         // Publish processed options to chartState immediately so option-derived reads
         // (mode, padding, etc.) work for the rest of construction. `applyOptions` will
         // refresh this on each subsequent update.
-        ctx.chartState.setValue('options', options.processedOptions as ChartState['options']);
+        ctx.chartState.setValue('options', options.processedOptions as unknown as ChartState['options']);
```
lsjroberts, line 471: Do we think this is ok? Without it these changes throw the following seemingly unrelated error:

```
Conversion of type 'AgChartOptions' to type 'ResolvedChartOptions' may be a mistake because neither type sufficiently overlaps with the other. Ifa this was intentional, convert the expression to 'unknown' first.
   Type 'AgStandaloneChartOptions<any, unknown>' is not comparable to type 'ResolvedChartOptions'.
   Property 'mode' is missing in type 'AgStandaloneChartOptions<any, unknown>' but required in type '{ animation?: NormalisedAnimationOptions | undefined; background: NormalisedBackgroundOptions; dataSource?: NormalisedDataSourceOptions | undefined; ... 21 more ...; annotations?: NormalisedAnnotationsOptions | undefined; }'.
```

packages/ag-charts-enterprise/src/features/navigator/miniChart.ts
```diff
@@ -243,25 +243,22 @@ export class MiniChart extends AbstractModuleInstance {
...
         const resized = this.seriesRect?.width !== width || this.seriesRect?.height !== height;
 
         this.seriesRect = seriesRect;
-        this.seriesRoot.translationY = paddingTop;
-        this.seriesRoot.setClipRectCanvasSpace(new BBox(0, -paddingTop, width, height));
+        this.seriesRoot.translationY = padding.top;
+        this.seriesRoot.setClipRectCanvasSpace(new BBox(0, -padding.top, width, height));
 
         for (const axis of this.axes) {
             const { position = 'left' } = axis;
```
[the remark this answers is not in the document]
lsjroberts: It goes blank which is expected.

packages/ag-charts-community/src/chart/chart.ts
alantreadway: Even before this change the `as` cast is a forcing function for types - lets roll with it for the moment.

alantreadway (approved):
LGTM

PR #7088, opened 2026-06-12
AG-17481 Index size-bounds warning and enforce 2-element sizeDomain

iMoses:
## Summary

Addresses QA follow-up feedback (comment 119421) on the merged size-scaling unification PR (#7047) for Bubble, Map Marker and Map Line series.

1. **Inverted-bounds warning now identifies the series.** The `minSize > maxSize` (and `minStrokeWidth > maxStrokeWidth`) warning previously read `series[].minSize …`; it now includes the series index (`series[0].minSize …`) and the wording changed from `reverting both to theme defaults.` to `ignoring both.`. Behaviour is unchanged — both bounds are still dropped so theme defaults re-apply.
2. **`sizeDomain` must be a 2-element array.** A malformed `sizeDomain` (e.g. `[5]`) previously flowed through to the size scale, producing a `NaN` size and an empty render. The option-def validators are tightened to `and(arrayOf(...), arrayLength(2, 2))`, so a malformed value is now rejected by the existing validation layer (which emits an indexed warning and drops the value) and the series falls back to the data-derived domain, rendering at default sizes.

The third item raised in the comment — distinguishing user `theme.overrides` from default theme values for the inverted-bounds warning — is intentionally **not** included. Default theme values already never trigger the warning (the check runs on raw user series options before theme merge); warning on `theme.overrides` was flagged as optional/complicated and is left as a possible follow-up.

## Changes

- `optionsModule.ts` — thread the series index into `validateSizeBounds` and update the warning text.
- `bubbleSeriesOptionsDef.ts`, `enterpriseThemeableOptionsDef.ts` — tighten `sizeDomain` validators to require exactly two elements.

## Test plan

- Updated the three existing inverted-bounds snapshots to the new indexed/`ignoring both.` text.
- Added a malformed-`sizeDomain` test per series asserting the warning and non-`NaN` rendering at the data domain.
- `yarn nx format`, `build:types`, `lint` (community + enterprise) — clean.
- Full suites: community 3638 passed, enterprise 2378 passed.

Fix #AG-17481

### DIFF
--- a/packages/ag-charts-community/src/chart/series/cartesian/bubbleSeries.test.ts
+++ b/packages/ag-charts-community/src/chart/series/cartesian/bubbleSeries.test.ts
@@ -1275,7 +1275,21 @@ describe('BubbleSeries', () => {
             expectWarningsCalls().toMatchInlineSnapshot(`
               [
                 [
-                  "AG Charts - series[].minSize (30) cannot be greater than maxSize (5), reverting both to theme defaults.",
+                  "AG Charts - series[0].minSize (30) cannot be greater than maxSize (5), ignoring both.",
+                ],
+              ]
+            `);
+        });
+
+        it('rejects a sizeDomain that is not a 2-element array and falls back to the data domain', async () => {
+            await createBubble({ sizeDomain: [5] });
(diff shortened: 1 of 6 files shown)


alantreadway (approved):
LGTM

PR #7089, opened 2026-06-12
AG-16720 TC7 Do not stop mousedown propagation on color palette

olegat:


alantreadway (approved):
LGTM

PR #7092, opened 2026-06-12
AG-17558 report the offending property in segment formatter warnings

iMoses:
https://ag-grid.atlassian.net/browse/AG-15933

When a label/caption formatter returns a rich-text/segment value with a bad sub-property — e.g. an image segment with `verticalAlign: 't0p'` — the runtime warned by dumping the whole returned array and saying only "expecting a string, a number, a date or text or image". The structured per-property error already existed but the array branch of `warnCallbackErrors` discarded it.

This drops that array short-circuit so array results use the same per-error path as objects. The warning now names the offending property, e.g.:

> Callback `axes[0].label.formatter` returned an invalid property `[0].verticalAlign`: `"t0p"`; expecting a keyword such as 'baseline', 'top', 'middle' or 'bottom', ignoring.

The valid part of the segment is still cleared and rendered (the bad property is dropped), so behaviour is unchanged aside from the message.

This is the "separate ticket for validation" called out in the AG-15933 review (raised against AG-15933 pending its own key).

Tests: rewrote the callback test that asserted the single generic dump to expect targeted per-element/property messages, and added segment cases (using the real `textOrSegments` validator) covering the `verticalAlign` typo and a valid sibling segment being preserved.

Fix #AG-15933

### DIFF
--- a/packages/ag-charts-core/src/state/validation.test.ts
+++ b/packages/ag-charts-core/src/state/validation.test.ts
@@ -1,6 +1,7 @@
 import type { Mock } from 'vitest';
 import { afterEach, beforeEach, describe, expect, vi } from 'vitest';
 
+import { textOrSegments } from '../config/chartDefaults';
 import { reset as resetLogger } from '../logging/logger';
 import { RegistryMode, reset as resetRegistry, setRegistryMode } from '../modules/moduleRegistry';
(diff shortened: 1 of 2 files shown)


alantreadway (approved):
LGTM

PR #7103, opened 2026-06-15
AG-17564 Expose Waterfall cumulative value to callbacks and events

iMoses:
https://ag-grid.atlassian.net/browse/AG-17564

Waterfall `total` and `subtotal` bars render a computed cumulative value, but callbacks and events had no clean way to read it: for `total` bars it is the displayed value, and for `subtotal` bars the displayed value is the segment delta — leaving the absolute running total unavailable.

This surfaces the existing cumulative as a new public `totalValue` field, populated only for `total` and `subtotal` items (`undefined` for `positive`/`negative` and all non-Waterfall series).

**Surfaces it reaches:**
- Label formatter params (`AgWaterfallSeriesLabelFormatterParams`)
- Tooltip renderer params (`AgWaterfallSeriesTooltipRendererParams`)
- Item-styler params (`AgWaterfallSeriesItemStylerParams`)
- Node events — `seriesNodeClick`, `seriesNodeDoubleClick`, `nodeContextMenuAction` (`AgNodeClickEvent`)
- `activeChange` event (`AgActiveChangeEvent`)

**Design:** the value is stored once on the node datum (mirroring the existing `datums`/`itemType` plumbing) and read by every consumer. It is sourced from the bigint-capable cumulative, so precision is preserved consistently across all callbacks.

**Tests:** new `waterfall.test.ts` cases assert `totalValue` on the label formatter, tooltip renderer, and `seriesNodeClick` event (150/200 cumulatives for subtotal/total, `undefined` otherwise). Existing AG-15850 callback snapshots updated to include the new field.

Fix #AG-17564


alantreadway (approved):
LGTM

PR #7112, opened 2026-06-16
AG-17459 re-render chart when external fonts finish loading

iMoses:
## Summary

Canvas text rendering never triggers a font download the way DOM text does. When a chart renders before a referenced web/icon font (FontAwesome, a user `@font-face`) has downloaded, it measures and draws with a fallback and never revisits — so an icon glyph in a title is silently dropped until the font happens to be cached on a later load.

We already solve this for Google fonts (`loadGoogleFonts`): options are scanned for referenced families, the font is loaded, and the chart re-lays-out on a `font:load` event. This generalises that re-render-on-load behaviour to **any** externally loaded font, out of the box (no new option), using the `FontFaceSet` (`document.fonts`) API.

## Changes

- **Extraction (`optionsModule.ts`)**: while walking options for Google fonts, also collect a weight/style-specific `FontFaceSet` shorthand for every concrete referenced family (e.g. `900 16px "Font Awesome 6 Free"`). Weight/style are part of the key because a family like FontAwesome ships a separate font file per weight. CSS generic keywords (`sans-serif`, etc.) are excluded via the shared `CSS_GENERIC_FAMILIES` set (now exported from `ag-charts-core`).
- **`FontManager.waitForFonts()`**: for each shorthand, `document.fonts.check()` skips already-available fonts (so system fonts cause no redundant re-render) and `document.fonts.load()` forces the download canvas wouldn't; once the loads settle it clears the text-measurer cache and emits the existing `font:load` event, which `chart.ts` already handles by re-laying-out. SSR-safe (`getDocument('fonts')` returns `undefined` with no document) and guarded against late emits after destroy.
- **Wiring**: `agCharts.ts` calls `waitForFonts(chartOptions.fonts)` alongside the existing `updateFonts(...)`; `fonts` is threaded through `ChartOptions` and the structural cache.
- **Docs/example**: the `inline-font-icons` example now creates the chart once the FontAwesome stylesheet's `@font-face` rules are registered (instead of pre-awaiting the font files), demonstrating the automatic re-render. Docs prose updated to match.

## Test plan

- New unit tests:
  - `fontManager.test.ts` — `waitForFonts` loads the exact weight-specific shorthand and emits `font:load` once loads settle; skips already-available fonts; no-op without a `FontFaceSet`; no emit after destroy.
  - `optionsModule.test.ts` — extraction collects weight-specific shorthands for nested title `text` segments and excludes generic families; Google fonts route to the CDN set.
- `yarn nx test ag-charts-community` (3656) and `yarn nx test ag-charts-enterprise` pass; `build:types` and `lint` clean.
- Manually verified in the dev server: with the pre-await removed, the FontAwesome icons in the example title/subtitle render after the chart's automatic re-render.

## Notes

- The general font path intentionally does not exclude Google families; `check()` skips them when already loaded and the Google CDN path still drives their re-render, so there is no double work or regression.

Fix #AG-17459

### DIFF
--- a/packages/ag-charts-community/src/api/agCharts.ts
+++ b/packages/ag-charts-community/src/api/agCharts.ts
@@ -263,6 +263,7 @@ class AgChartsInternal {
         }
 
         chart.ctx.fontManager.updateFonts(chartOptions.googleFonts);
+        chart.ctx.fontManager.waitForFonts(chartOptions.fonts);
 
         if (data != null) {
             chart.ctx.dataService.restoreData(data);
--- a/packages/ag-charts-community/src/chart/fonts/fontManager.test.ts
+++ b/packages/ag-charts-community/src/chart/fonts/fontManager.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it, vi } from 'vitest';
+import { afterEach, describe, expect, it, vi } from 'vitest';
 
 import { type DynamicContext, EventEmitter } from 'ag-charts-core';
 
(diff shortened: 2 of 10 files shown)


alantreadway (approved):
LGTM

PR #7114, opened 2026-06-16
AG-17516 warn when ordinal-time axis requires ag-charts-enterprise

iMoses:
https://ag-grid.atlassian.net/browse/AG-17516

In community-only builds, an axis configured with `type: 'ordinal-time'` was silently stripped and re-inferred as a plain `time` axis, with no warning that `ordinal-time` is an enterprise feature.

`validateAxesOptions` now mirrors `validateSeriesOptions`: a known-but-unregistered enterprise axis type is collected as a missing module and threaded into `processModuleOptions`, which emits the existing "required modules are not registered" warning naming `OrdinalTimeAxisModule`. Genuinely unknown axis types still produce the "Unknown type ... ignoring" warning.

Scope is the warning only — the axis still falls back to `time`, now accompanied by a clear warning, matching how unregistered enterprise series already behave.

Includes a regression test covering an unregistered enterprise axis in a community build.

Fix #AG-17516

### DIFF
--- a/packages/ag-charts-community/src/module/optionsModule.test.ts
+++ b/packages/ag-charts-community/src/module/optionsModule.test.ts
@@ -461,6 +461,20 @@ describe('ChartOptions', () => {
             expect(message).toContain("'number'");
             expect(message).toContain('ignoring.');
         });
+
+        it('warns when an enterprise axis type is not registered', () => {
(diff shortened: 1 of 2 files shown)


alantreadway (approved):
LGTM

PR #7121, opened 2026-06-16
Skip benchmark examples whose base version lacks the feature

alantreadway:
## Summary

The nightly browser benchmark compares the head example set against an older published base library. Examples exercising **new** features (bigint data values, ISO-8601 strings on a time axis) render no valid data points on a base that lacks the feature, so the base is meaninglessly fast — producing bogus regressions (e.g. +280%) that inflate the nightly counts.

This adds a per-example **`minVersion`** marker so the compare step excludes such cases against a base below that version, rather than reporting them as regressions.

## Changes

- **Examples** — stamp `minVersion: '14.0.0'` in the `metadata` of `high-volume-bigint` and `high-volume-iso-datetime` (flows into the report JSON; no harness/type changes).
- **`compare-browser-results.js`** — new `--base-version` flag, semver gating helpers, and a `skippedBelowMinVersion` bucket. Gated cases are diverted out of `matched`, so they no longer affect the rankings, the notable-regressions list, or the improved/regressed summary. Surfaced in both table and JSON output.
- **`compare-browser-latest.sh`** — passes `--base-version <npm version>` in published-npm mode; git-ref mode derives the version from the base report.
- **`format-browser-slack-message.js`** — adds a "N skipped (base below min version)" note so excluded cases are visible.

## Fail-open semantics

Only a **released** base strictly older than `minVersion` is excluded. A prerelease/dev base, an unparseable ref (e.g. `origin/b13.2.0`), or a missing version is treated as the `latest` reference and **not** excluded — so real comparisons are never silently hidden.

## Test plan

Verified against the real Jun-16 CI reports (`benchmark-results` artifact):

- base `13.3.1` (released, < min) → 4 new-feature cases skipped, ranked 116 → 112; counts and notable list exclude them.
- base `14.0.0` (== min) and `14.1.0` (> min) → not skipped (116).
- base `13.3.0-beta.x` (prerelease) → fail-open, not skipped.
- base `origin/b13.2.0` (unparseable) → fail-open, not skipped.
- no `--base-version` → version derived from base report (`13.3.1`) → 4 skipped.

`prettier --check` and `bash -n` pass on the changed files.

### DIFF
--- a/packages/ag-charts-website/src/content/docs/benchmarks/_examples/high-volume-bigint/main.ts
+++ b/packages/ag-charts-website/src/content/docs/benchmarks/_examples/high-volume-bigint/main.ts
@@ -81,6 +81,9 @@ function getBenchmarkConfig(): BenchmarkConfig {
             axisType: 'number',
             dataType: 'bigint',
             version: VERSION,
+            // bigint values are a new feature; an older published base renders no valid data
+            // points, so the compare step skips this example when the base is below minVersion.
+            minVersion: '14.0.0',
         },
     };
 }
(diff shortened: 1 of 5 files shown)


iMoses (approved):
LGTM

PR #7122, opened 2026-06-16
AG-17271 support dot-notation paths in DatumKey type

iMoses:
## Summary

`DatumKey<TDatum>` now resolves to dot-notation string paths for nested data (e.g. `"x" | "x.date"`) in addition to top-level `keyof TDatum & string`. AG Charts already supports dot-notation for nested field access at runtime, but the type previously rejected paths like `"x.date"`, forcing users with nested data to reach for `as any` / `@ts-ignore`. This makes those references type-safe across `xKey`, `yKey`, `sizeKey`, `colorKey`, `labelKey`, `idKey`, and the other key properties.

This is a **type-only change** — there is no runtime behaviour change.

## Details

- Recursion is bounded to depth 4 via a tuple depth-counter, so deeply nested or self-referential datum types cannot trigger unbounded type expansion.
- Array-, `Date`-, and function-valued properties are treated as leaves (no `items.name`, no `date.<method>` expansion), matching what the runtime dot-notation resolver supports.
- Optional nested properties contribute both the bare key and the nested path.
- Index signatures and untyped / non-object datums fall back to `string`, keeping the common `DatumDefault = any` path ergonomic.

Backwards-compatible: the type only widens, so every previously valid key value remains valid.

## Test plan

- Added `datumKey.typecheck.ts` — compile-time assertions covering bare/nested keys, deep recursion, optionals, array/Date leaves, index-signature and untyped fallbacks. The file is compiled by `build:types`, so a regression in `DatumKey` fails the build.
- `yarn nx build:types ag-charts-types` ✓
- `yarn nx build:types ag-charts-community` ✓ (verifies the wider type breaks none of the ~25 consuming option files)
- `yarn nx lint ag-charts-types` ✓

Fix #AG-17271


alantreadway (approved):
LGTM

PR #7127, opened 2026-06-16
AG-17564 fix waterfall totalValue for subtotals after the first

iMoses:
## Summary

Follow-up to #7103. QA found that the `totalValue` param exposed to Waterfall callbacks/events was wrong for any `subtotal` after the first: it reported the absolute grand running total instead of the segment the bar actually represents.

- **Reproducer:** 2nd `subtotal` bar — rendered label/`value` `-686` (correct), but `totalValue` reported `-133` (the absolute cumulative) instead of `-686`.
- **Root cause:** `totalValue` was sourced from the raw `cumulativeValue` (`ctx.yCurrTotalValues[i]`, the absolute running total across all bars). For the **first** subtotal this coincidentally equals the displayed segment value because the trailing baseline is `0`, which is why the original single-subtotal test passed vacuously. Every later subtotal diverges.
- **Fix:** base `totalValue` on the already-computed display `value`, which resolves the segment relative to the previous total/subtotal. For `total` items this is the absolute cumulative; for `subtotal` items the segment delta; `undefined` for `positive`/`negative`. Every consumer (label formatter, tooltip renderer, item styler, node-click events) reads `nodeDatum.totalValue`, so the two datum-construction sites are the single fix point.

## Test plan

- Added a regression test with **two** subtotals separated by data (the original fixture had only one, so `value === totalValue` everywhere and the bug was invisible). It asserts the 2nd subtotal's `totalValue` equals its segment (`50`, matching the rendered value) rather than the absolute cumulative (`200`). Confirmed failing before the fix, passing after.
- `yarn nx test ag-charts-enterprise` waterfall + callbacks suites (200 tests) pass; no snapshot changes.
- `yarn nx format`, `build:types`, `lint` all green.

Fix #AG-17564

### DIFF
--- a/packages/ag-charts-enterprise/src/series/waterfall/waterfall.test.ts
+++ b/packages/ag-charts-enterprise/src/series/waterfall/waterfall.test.ts
@@ -260,6 +260,32 @@ describe('WaterfallSeries', () => {
             expect(seriesNodeClick.mock.calls[0][0].totalValue).toBe(SUBTOTAL_VALUE);
             expect(seriesNodeClick.mock.calls[1][0].totalValue).toBeUndefined();
         });
+
+        it('computes each subtotal totalValue relative to the previous subtotal, not the absolute total', async () => {
+            // Two subtotals: 1st after A,B (segment 150, absolute 150 — coincide); 2nd after C,D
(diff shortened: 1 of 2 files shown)


alantreadway (approved):
LGTM

PR #377, opened 2023-10-18
AG-8557 Zoom fixes and enableDoubleClickToReset option

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-8557

This adds the `zoom.enableDoubleClickToReset` option and fixes and cleans various bits on zoom.

PR #7132, opened 2026-06-17
Re-enable l10n translations test and add advisory CI review

alantreadway:
## Summary

Brings AG Charts' locale package back under a maintenance contract, with the reusable rule + skill now sourced from the shared `ag-dev-prompts` marketplace rather than committed locally.

- **Re-enables the `translations` test** in `packages/ag-charts-locale/src/main.test.ts` (identical key sets, allowed formatters, matching `${variables}`).
- **Fixes the drift it surfaced** — all 31 locales were missing `ariaAnnounceSelectedItem` / `ariaAnnounceUnselectedItem`; adds translated values.
- **Adds an advisory CI translation review** (`tools/l10n/review-translations.mjs` + `.github/workflows/l10n-translation-review.yml`) — dependency-free, advisory only, never fails the build.

## Rule + skill now shared

The `l10n-maintenance` rule and `l10n-translate` skill originally added here have been promoted to the `ag-dev-prompts` marketplace (`ag-eng` plugin) so AG Grid and AG Studio get them too (ag-dev-prompts #377, now on `latest`). The repo-local copies and their `.rulesync/.gitignore` allowlist entries are removed in this PR — Charts consumes `/ag-eng:l10n-translate` and the glob-triggered `l10n-maintenance` rule from the marketplace like the other products.

## Test plan

- `yarn nx test ag-charts-locale` — `translations` suite green.
- `yarn nx format` / `lint` / `build:types ag-charts-locale` clean.
- The advisory workflow is paths-filtered to `packages/ag-charts-locale/src/**`.


AG-DavidG (approved):
LGTM

alantreadway:
## Addressing review feedback

Codex's automated review found no issues here. For consistency with the Grid/Studio PRs (which share this workflow), I applied the same security hardening to the advisory review workflow in two follow-up commits:

- **Run the review script from the base ref** — the script is now extracted from `$BASE_SHA` via `git show` and executed from `$RUNNER_TEMP`, never the PR head, so PR-controlled code can't run with `ANTHROPIC_API_KEY` in scope. Falls back to the checked-out copy only on the PR introducing the script (base ref doesn't have it yet).
- **Same-repo fork gate** — added a job-level `if:` mirroring `ci.yml`/`pr-review.yml` so the job is skipped entirely for fork PRs; fork PRs can be reviewed via maintainer `workflow_dispatch`.
- **`workflow_dispatch` PR-head resolution** — when a PR number is supplied, the dispatch path now fetches and diffs that PR's head/base instead of the checked-out ref.

CI green against the latest commit.

PR #7133, opened 2026-06-17
AG-17494 Update default theme params for grid parity

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-17494


packages/ag-charts-enterprise/src/features/error-bar/__image_snapshots__/error-bar-test-ts-error-bars-ag-16943-should-hide-error-bars-when-series-visible-is-set-to-false-via-update-2-snap.png
lsjroberts, line 1: Unrelated: I'm 99% sure this test is wrong. I did see it correctly generated at one point, with the blue line hidden and the bars visible on the orange line. But no longer.

packages/ag-charts-community/src/chart/__image_snapshots__/scene-test-ts-scene-on-translation-only-change-should-render-bar-series-correctly-after-update-1-snap.png
lsjroberts, line 1: Intentional change on these tests, since they now use `prepareTestOptions()`.

alantreadway (approved):
LGTM

PR #7135, opened 2026-06-17
AG-17548 ignore parentResize on a destroyed chart

iMoses:
## Summary

A series-type switch replaces the chart instance via `destroy({ keepTransferableResources })`: the outgoing chart is flagged destroyed and hands its scene to the replacement **synchronously**, but its teardown — which unsubscribes the `dom:resize` listener — is queued **asynchronously**. A container measurement arriving in that window was still applied by the dead chart to the now-shared scene, parking a pending size that the live chart then treated as a no-op, leaving the chart unrendered until the next interaction.

Guard `Chart.parentResize` on the `destroyed` flag so a late `dom:resize` delivered to a destroyed chart is ignored.

## Test plan

- Added a regression test in `detachedContainerAsyncData.test.ts` asserting that a `dom:resize` emitted after `destroy({ keepTransferableResources: true })` does not mutate the shared scene's dimensions. Confirmed failing before the fix, passing after. (jsdom can't model the real-browser timing of the full race, so the test asserts the invariant directly.)
- `npx vitest run` enterprise `detachedContainerAsyncData` suite (3 tests) passes.

Fix #AG-17548

### DIFF
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -1323,7 +1323,7 @@ export abstract class Chart extends Observable implements ModuleInstance, ChartS
     }
 
     private parentResize(size: { width: number; height: number; pixelRatio: number } | undefined) {
-        if (size == null || (this.width != null && this.height != null)) return;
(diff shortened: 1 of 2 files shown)


alantreadway (approved):
LGTM

PR #7137, opened 2026-06-17
AG-13954 Fix zoom event to use AgGroupingValueType shape

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-13954


alantreadway (approved):
LGTM

PR #7139, opened 2026-06-17
Add release-review CI jobs for docs and options reviews

alantreadway:
Adds a `release-review` workflow that runs the release docs and options reviews headlessly via Claude Code.

## What changed

- **`.github/workflows/release-review.yml`**: parallel `docs-review` and `options-review` jobs that invoke the shared `release-review` composite action from `@ag-grid/dev-prompts`.
  - **Triggers:** `workflow_dispatch` (pick docs / options / both, optional branch overrides) and `push` to a `bX.Y.Z` branch **guarded by `github.event.created`** — so it runs once when a release branch is cut, not on every subsequent commit.
  - Each job uploads its report as an artifact and posts a summary + the full report (as a file) to **#charts_teamcity_status**.
  - Reuses existing secrets/vars (`ANTHROPIC_API_KEY`, `SLACK_BOT_OAUTH_TOKEN`, the ag-dev-prompts read token).

## Why

So release-readiness docs and API/breaking-change reviews run automatically at release-branch cut (and on demand against `latest`), instead of only by hand.

## Dependency

Depends on ag-grid/ag-dev-prompts#381 (the `release-review` action + skill changes) being merged and published to the `canary` channel — this workflow installs `@ag-grid/dev-prompts@canary`.

### DIFF
--- a/.github/workflows/release-review.yml
+++ b/.github/workflows/release-review.yml
@@ -114,8 +114,9 @@ jobs:
               uses: ./.ag-dev-prompts/node_modules/@ag-grid/dev-prompts/.github/actions/release-review
               with:
                   review_type: docs
-                  previous_branch: ${{ inputs.previous_branch }}
-                  current_branch: ${{ inputs.current_branch }}
+                  # inputs.* only exist on workflow_dispatch; on push they auto-resolve in the action.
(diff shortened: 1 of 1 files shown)


AG-DavidG (approved):
LGTM

PR #7142, opened 2026-06-17
Make l10n translation review comment less verbose

alantreadway:
Tweaks the advisory l10n translation-review output so it no longer adds a `### <locale>` heading for every locale. Now:

- If nothing is concerning, it stays a single line ("translations look correct").
- Otherwise, only locales **with** findings get a section; all clean locales are summarised together on one line (e.g. "No issues found in: de-DE, fr-FR, ja-JP").

Previously every locale — including clean ones — got its own `## <locale>` / "No issues found." block, which made the comment very long on PRs that touch all 31 locales (see ag-studio#1752).

This is a one-line change to the shared review script's system prompt; pushed to the `ag-shared` source via `subrepo push`. Companion `subrepo pull` PRs follow in ag-grid and ag-studio.

### DIFF
--- a/external/ag-shared/.gitrepo
+++ b/external/ag-shared/.gitrepo
@@ -6,7 +6,7 @@
 [subrepo]
 	remote = https://github.com/ag-grid/ag-shared.git
 	branch = latest
-	commit = 06b051e5e1bc7da80598b1038a9a410958d3e41a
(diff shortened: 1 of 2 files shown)


AG-DavidG (approved):
LGTM

PR #7145, opened 2026-06-18
AG-16032 Align ordinal-time day labels with day boundaries

iMoses:
https://ag-grid.atlassian.net/browse/AG-16032

Ordinal-time axis day labels now land on the bar at the day boundary (the first datum at/after midnight) by default, without enabling `parentLevel`. Previously labels were evenly sampled and could sit on a mid-day bar (e.g. "Sep 8" not lining up with the 00:00 candle).

Implementation: in `calculateRawTicks()`, ordinal-time now infers a tick interval, and a gate boundary-aligns labels only when the data is intraday **and** the inferred interval is day-level (>= 1 day and < 1 week). Sub-day intervals (zoomed in), week-and-coarser intervals (zoomed out), and daily-or-coarser data keep the legacy even-sampling — so finer time-of-day labels and existing wide-view spacing are preserved.

Behaviour change: default ordinal-time day labelling shifts for wide intraday views (no opt-out) — worth a release note.

Test plan:
- Two behavioural tests added: continuous hourly data -> every labelled tick is at local midnight; gapped market-hours data -> each labelled tick is the first datum of its day.
- Full enterprise suite green (2457 passed), zero snapshot churn; community axis tests green.

Fix #AG-16032

### DIFF
--- a/packages/ag-charts-community/src/chart/axis/generateTicks.ts
+++ b/packages/ag-charts-community/src/chart/axis/generateTicks.ts
@@ -337,7 +337,8 @@ function calculateRawTicks<TScale extends Scale<TDatum, number, TickInterval<TSc
                     niceDomain.length > 0 &&
                     tickParams.interval == null &&
                     (UnitTimeScale.is(scale) ||
-                        (generatePrimaryTicks && (TimeScale.is(scale) || OrdinalTimeScale.is(scale))))
+                        OrdinalTimeScale.is(scale) ||
+                        (generatePrimaryTicks && TimeScale.is(scale)))
                 ) {
                     const dates = niceDomain as (Date | number)[];
                     const start = Math.min(dates[0].valueOf(), dates.at(-1)!.valueOf());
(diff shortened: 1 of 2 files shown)


packages/ag-charts-community/src/chart/axis/generateTicks.ts
```diff
@@ -363,6 +364,26 @@ function calculateRawTicks<TScale extends Scale<TDatum, number, TickInterval<TSc
...
+                    // default ticks:
+                    //  - sub-day intervals (zoomed in) preserve the finer-grained time-of-day labels;
+                    //  - week-and-coarser intervals (zoomed out) keep even spacing, where a few hours'
+                    //    misalignment of a weekly/monthly label is invisible anyway;
+                    //  - daily-and-coarser data has ~one datum per label, so alignment is moot.
+                    const intervalMs = intervalMilliseconds(timeInterval);
+                    const dataIsIntraday =
+                        minimumTimeGranularity != null && intervalMilliseconds(minimumTimeGranularity) < dayMs;
+                    const intervalIsDayLevel = intervalMs >= dayMs && intervalMs < weekMs;
+                    if (!(dataIsIntraday && intervalIsDayLevel)) {
```
jacobp100, line 382: Could some of these checks not be hoisted to the top-level if?

iMoses:
Due to it being too close to the release and having a performance regression, let's delay merging this.
The solution does work, but it is addressing a symptom and not the underlying problem.
We need to look at all parent-level related tickets together and try to come up with a global, generic solution.

PR #6491, opened 2026-03-23
CRT-1071 Disable custom tooltip renderers in waterfalls

olegat:
https://ag-grid.atlassian.net/browse/CRT-1071

PR #7147, opened 2026-06-18
AG-16997 Restore and refine waterfall tooltip renderers

AG-DavidG:
## Summary

Restores the custom tooltip renderers removed in CRT-1071 (PR #6491) for the `customised-waterfall` and `horizontal-waterfall` gallery examples, and fixes the underlying reason they were disabled.

The original renderers dereferenced the user `datum` unconditionally. For synthetic **total/subtotal** bars `datum` is `undefined`, so the renderers threw — which is why CRT-1071 disabled them. They now branch on `itemType` and read the cumulative value from `totalValue` (the detail recently added to the waterfall tooltip params), with the heading supplied by the series/axis label for totals.

The tooltip content is also trimmed for clarity (PREVis-reviewed):
- Dropped the redundant `title` line that echoed the series name and duplicated a data row.
- Normal bars show only the per-bar values; totals show a single cumulative row.
- Replaced generic labels with specific ones.

### Resulting tooltips
| Case | Tooltip |
|---|---|
| Customised – normal | `17 Oct` / Daily Change +0.58% / Close Price £7,675.21 |
| Customised – total | `Monthly Net` / Net Change -4.12% |
| Horizontal – normal | `Income Tax` / Amount £185 billion |
| Horizontal – total | `Total Budget` / Amount £553 billion |

## Test plan
- `tsc` against the examples tsconfig passes.
- E2E gallery snapshots for these two examples will change (the triggered-tooltip content differs). The snapshot baselines need regenerating — left to the CI snapshot job (per the CRT-1071 pattern, where snapshots landed via a separate automated commit).

Fix #AG-16997


alantreadway (approved):
LGTM

PR #7149, opened 2026-06-18
AG-17584 Show PixelSize branch for Padding-typed options in API reference

iMoses:
https://ag-grid.atlassian.net/browse/AG-17584

The API reference for `Padding`-typed options (e.g. axis `label.padding`) stopped
showing that a plain `number` is accepted, displaying only the `PaddingOptions`
object branch of `Padding = PixelSize | PaddingOptions`.

A direct union type alias was being intercepted by the cross-line aliased-union
expansion (added for discriminated cross-line options) and flattened into its
interface variants, which dropped both the `type Padding = …` alias line and the
hidden `PixelSize` (number) branch. The expansion is now gated to the cross-line
indirection only (an empty interface extending a union alias), so direct union
aliases render via `formatUnionTypeAlias` as before — emitting the alias line plus
each non-hidden interface body.

Verified the cross-line options page still lists its per-axis discriminated
variants; `ag-charts-website` lint and tests pass.

Fix #AG-17584

### DIFF
--- a/packages/ag-charts-website/src/components/api-documentation/components/ApiReference.tsx
+++ b/packages/ag-charts-website/src/components/api-documentation/components/ApiReference.tsx
@@ -391,12 +391,10 @@ function useMemberAdditionalDetails(member: MemberNode): NodeTypes | NodeTypes[]
 
     const resolvedDetails = resolve(memberType);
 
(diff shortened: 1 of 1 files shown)


alantreadway (approved):
LGTM

PR #7152, opened 2026-06-18
perf(data-model): skip redundant column rescans in extraction/updates

alantreadway:
## Summary

Pre-release performance work on the data-model column-extraction path, targeting regressions introduced by the new per-datum extraction machinery (column type-tracking, continuous/time validation, bigint narrowing) added since 13.3.1.

Changes:
- **Avoid full epoch-column rescan on incremental updates** — incremental processor carries forward epoch column state instead of rescanning every column on each update.
- **Seed epoch cache at extraction** — populates the epoch cache during `extractValues` so subsequent column scans are skipped.
- **Seed numeric column cache as identity** for pure-`number` columns at extraction (`seedNumericColumnIdentity`), skipping the O(n) bigint-narrowing scan per full extraction. Identity-only, not the offset cache.
- **Settled-type fast-path** in `updateColumnTypeTracker` for `number`/`bigint`/`boolean` (typeof maps 1:1), skipping the `valueColumnType` call; `string`/`object` keep the full path.
- **Correctness: invalidate identity epoch cache when an incremental insert introduces a string** — a string-free time column seeded as its own epoch identity must drop that identity when an `applyTransaction` insert writes an ISO string into it, or `ensureEpochColumn`'s lazy parse leaves the raw string in the continuous key domain. Tracks strings in the delta so the all-numeric/Date streaming case stays scan-free. (addresses PR review)
- **Benchmark tooling**: CPU-profile capture for benchmark examples (`profile-benchmark.ts`), with Chromium closed in a `finally` on failure.

## Why

`updateDelta({data})` recreates the DataSet and re-extracts all points each update, so the new machinery's per-datum cost dominates full re-extraction. These changes remove redundant column rescans without changing extraction semantics.

## Cross-release benchmark validation

HEAD vs the latest patch of each published minor line (curated 8-example set). `before` = published npm bundle (minified UMD), `after` = HEAD (local build); negative % = HEAD faster. HEAD's local build carries a ~3–4% parse penalty vs minified published bundles, measured via the Initial-Load/Chart-Create control:

| base | median Δ | chart-create offset | datum-highlight | reading |
|---|---|---|---|---|
| npm:13.3.1 | +4.3% | +4.0% | **−20.9%** | at parity (median ≈ offset); real highlight win |
| npm:13.2.1 | +8.5% | +11.9% | −29.1% | faster than raw (offset > median) |
| npm:13.1.0 | +0.9% | +2.9% | −29.6% | at parity; highlight win |
| npm:13.0.1 | +3.1% | +2.8% | −23.2% | at parity; highlight win |
| npm:12.3.1 | −2.2% | +14.6% | −69.6% | faster |
| npm:12.2.0 | −8.7% | +4.9% | −72.7% | faster |
| npm:12.1.2 | −14.8% | −42.2% | −83.6% | much faster |
| npm:12.0.2 | −23.1% | −40.4% | −94.5% | much faster |

- Decisively faster than every 12.x line; **at algorithmic parity with recent 13.x** once the unminified-build offset is removed, with a real ~20–30% win on the extraction/highlight path.
- Flagged "regressions" are Initial-Load/Chart-Create cases dominated by minified-vs-unminified bundle parse cost, not algorithmic regressions.
- `high-freq-line` update residual is now within offset+noise (+6.9%, flagged noisy; was +18% pre-branch). The deferred diff-aware `updateDelta` is the only remaining lever (post-release follow-up).

## Test plan

- `nx test ag-charts-community -- dataModel incrementalProcessor` (204 pass, incl. the new epoch-cache invalidation regression test; example-data tests must run through nx, not raw vitest).
- Browser benchmark compare + CPU profile diff vs npm:13.3.1 and the per-minor sweep above.

### DIFF
--- a/packages/ag-charts-community/src/chart/data/data-model/extraction/dataExtractor.ts
+++ b/packages/ag-charts-community/src/chart/data/data-model/extraction/dataExtractor.ts
@@ -124,6 +124,15 @@ function updateColumnTypeTracker(
     value: unknown,
     datumIndex: number
 ): ColumnValueType | undefined {
+    // Hot-loop fast path: for the primitive types whose `typeof` maps 1:1 to a column type
+    // (number/bigint/boolean), a value of the column's already-settled type cannot change it or
+    // introduce a string — skip the full classify and merge. `string` (→ string|date via isISO8601)
+    // and `object` (→ date|number|object) are not 1:1, so they fall through to the slow path.
+    const settled = tracker.type;
+    if ((settled === 'number' || settled === 'bigint' || settled === 'boolean') && typeof value === settled) {
+        return settled;
+    }
+
     const observed = valueColumnType(value);
     if (observed == null) return undefined;
 
--- a/packages/ag-charts-community/src/chart/data/data-model/extraction/dataExtractor.ts
+++ b/packages/ag-charts-community/src/chart/data/data-model/extraction/dataExtractor.ts
@@ -5,6 +5,7 @@ import {
     isISO8601,
     isNumberObject,
     iterate,
+    seedEpochColumnIdentity,
     timeValueToNumber,
 } from 'ag-charts-core';
 import type { AgNumericValue } from 'ag-charts-types';
@@ -93,6 +94,9 @@ interface ColumnTypeTracker {
     mixedAtIndex: number | undefined;
     /** Row index where a date column was first observed to mix with a non-date, non-numeric value. */
     mixedDateAtIndex: number | undefined;
+    /** Whether any column value was a string. A string-free column needs no epoch-ms parse, so its
+     *  epoch cache can be seeded as identity to spare downstream consumers a redundant string scan. */
+    sawString: boolean;
 }
 
 function valueColumnType(value: unknown): ColumnValueType | undefined {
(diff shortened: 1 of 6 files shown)


alantreadway:
Thanks — reviewed all three findings against the code.

**[P1] Identity epoch cache stays valid after inserts that introduce strings** — confirmed real, fixed in bf69cdb. An `applyTransaction` insert can write a raw ISO string into a previously all-`Date`/numeric-epoch time column, and `ensureEpochColumn`'s lazy parse never re-runs on an identity cache hit, so the string leaked unparsed into the continuous key domain. `applyChangeDescWithCache` now tracks whether the delta writes a string into an identity column and invalidates only then — the common all-numeric/Date streaming case stays scan-free. Added a regression test (`reparses a string-free time key column when an incremental insert introduces an ISO string`).

**[P2] Browser can be left running when profiling fails** — valid, fixed in b43658f. `browser.close()` now runs in a `finally`.

**[P1] Fast path can misclassify invalid numbers after the column type settles** — false positive, no change. The premise is "if `valueColumnType` rejects non-finite/`NaN` numeric values", but it does not: every `typeof === 'number'` returns `'number'` unconditionally (`dataExtractor.ts:103-120`), `NaN` included. The fast path returns the settled `'number'` only when `typeof value === 'number'`, which is exactly what the slow path would classify, and the slow path's `sawString`/merge branches are all no-ops for a settled-number column receiving another number. Invalid-numeric handling (NaN filtering) lives in the validation processors, not column type-tracking, so nothing is suppressed.

iMoses (approved):
LGTM

PR #7160, opened 2026-06-18
AG-14806: Expand union-typed API props into discrete types

iMoses:
## Summary

API-reference properties whose type is a **union of named interfaces** (e.g. `fill`, `format`, `crossLines`) previously dumped every variant's definition into one concatenated code block — hard to read and easy to confuse with object child-properties.

They now expose a **"See available types"** action that lists the discrete named interface variants as child-property-style rows. Each variant uses the standard **"See child properties"** affordance to expand into its own properties rendered as standard API-reference rows (reusing `NodeFactory` recursion) — not a code block.

- **TC1** — chevron on the prop name + a "See available types" action; either expands the list.
- **TC2** — variants render underneath the property, styled like child properties.
- **TC3** — each variant expands (via "See child properties") into its own properties as standard rows.
- **Deep-links** — variant anchor IDs follow the navigation hash convention (`…-{discriminator}-{child}`, matching `getNavigationDataFromPath`), so links into a variant's sub-property auto-expand the chain (prop → variant → child).

## How it works

- `useMemberAdditionalDetails` resolves a union into `{ kind: 'unionTypes', variants }`, where each variant carries its anchor segment (the `type` discriminator value, else the interface name), the interface node, and resolved generics. Special-type members (`series`, `axes`, mini-chart series, annotations) are excluded so they keep their existing page-navigation behaviour.
- `getCollapsibleType` orders the `nestedPath` guard ahead of `unionTypes`/`code` so nested-page members keep their "See property details" link.
- A new `UnionVariantNode` renders the variant header (a child-prop-style row) and recurses into the variant interface's members via `NodeFactory`.

## Test plan

- Updated the `fill` e2e test to assert variant sub-property **rows** appear (and that no `<pre>` code block renders), plus a new deep-link test.
- `yarn nx format`, `yarn nx lint ag-charts-website` (0 errors), and TS language-server checks pass.
- Logic verified against the generated interfaces data for `fill`, `crossLines`, `format`, `series`.

## Verification status

- ✅ Static gates (format / lint / type-check) and data-shape logic verified.
- ⚠️ **Runtime/visual not yet verified by me** — the Playwright e2e harness needs Docker (unavailable in my env) and I couldn't drive a browser. Please eyeball the dev server (or rely on CI e2e): `options/series/bar/#reference-AgBarSeriesOptions-fill` and `options/axes/time/#reference-AgTimeAxisOptions-label-format`, in light **and** dark mode.

## Known limitation

Union members that are themselves **named typeAliases** (not interfaces) are dropped from the expandable list when the union also contains interface variants (they remain in the signature text). A typeAlias has no members to recurse into, and rendering it as a code sub-block would reintroduce the pattern this PR removes. Unions of *only* typeAliases fall through to the unchanged code-block rendering. None of the cited props hit this.

Fix #AG-14806


iMoses:
image

markdurrant (approved):
Walked through design/UX: LGTM 👍

PR #7163, opened 2026-06-18
Make Google Fonts E2E test hermetic with recorded HAR

alantreadway:
Proper fix following the temporary skip in #7159. **Re-enables** the `fonts › google fonts` E2E test and removes its dependency on the live Google Fonts CDN.

## Root cause
The `google-fonts` example renders with `loadGoogleFonts: true`, so at render time `FontManager.loadFonts` injects `@import url('https://fonts.googleapis.com/css2?…')` and the browser fetches CSS + woff2 from `fonts.googleapis.com`/`fonts.gstatic.com`. The test waits via `gotoExample` → `waitForLoadState('networkidle')`, so every run (×6 frameworks) was hostage to the live CDN. When the CDN was slow/unreachable from the runners the page never reached `networkidle` and the test timed out — which is why it was skipped.

## Fix
Replay the CDN responses from a committed HAR (`e2e/fixtures/google-fonts.har.zip`, ~54 KB) via `page.routeFromHAR`, scoped to the two font hosts. The test no longer touches the live network: matched requests are served from the HAR and, with Playwright's default `notFound: 'abort'`, nothing falls through to the internet. `networkidle` now settles instantly.

The HAR was recorded by reproducing the runtime's exact `@import` URL and rendering the example's text in each family, so it contains exactly the googleapis CSS + the three `latin`/weight-400 woff2 the chart downloads. Because replay always serves the recorded CSS, the referenced gstatic URLs are fixed regardless of the runner's browser/UA.

Re-record with `UPDATE_FONT_HAR=1` if the example's fonts change.

## Verification
- Offline replay validated locally: with `update:false` all three families load purely from the HAR, with zero font-CDN network requests.
- The screenshot baseline (`google-fonts-chromium-linux.png`) is unchanged — only font *delivery* changes, not what's rendered — so CI validates the end-to-end render against the existing baseline. (Couldn't run the full website e2e locally; example generation doesn't run in a bare worktree.)


AG-DavidG (approved):
LGTM

PR #7164, opened 2026-06-18
AG-10468 Fix mini chart processing data before its request has completed

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-10648


alantreadway (approved):
LGTM

PR #7170, opened 2026-06-19
AG-17474 Document navigator mini chart source in async data page

AG-DavidG:
https://ag-grid.atlassian.net/browse/AG-17474

Updates the Asynchronous Data documentation page following QA feedback:

- The `async-data` example now enables a navigator mini chart, which issues its own full-range `getData` call with `source: 'mini-chart'` independent of the main chart's windowed fetches. Adds an `initialState.zoom` so the windowed behaviour is visible on load, plus a **Reload** control demonstrating `updateDelta({})`.
- Documents the `'mini-chart'` source value and reorders the `AgDataSourceCallbackParams` list; clarifies that `getData` runs on initial load and whenever the visible window changes; moves the first/last data point requirement into a note.

Also tidies wording in a few v14 upgrade-page breaking-change sections (these were staged alongside the doc edits).

Fix #AG-17474


alantreadway (approved):
LGTM

PR #7177, opened 2026-06-19
Clarify the simple option on mixed simple/complex union props in API reference

iMoses:
## Summary

Many `ag-charts-types` properties accept a union of a **simple** value and a **complex** interface/array (e.g. `fill` = a colour string *or* a gradient/pattern/image object; `padding` = a number *or* a per-side object). In the generated API reference the complex branch expands into a noisy nested tree, and the simple primitive form, which is often what users actually want, is easy to miss.

This change appends a terse, simple-form-first clause to the JSDoc of every such property so the simple option is spelled out alongside the complex one. The rendered description comes only from the consuming property's JSDoc (never the `export type` alias), so the clause is added per consumer.

~84 properties across 25 files, grouped by union type:

| Union | Clause |
|---|---|
| `AgColorType` / `[]` / `AgColorTypeStrict` | "A colour string, or an object for a gradient, pattern, or image fill." (array/strict variants worded accordingly) |
| `Padding` | "A number applies uniform padding; an object sets each side." |
| `boolean \| AgBorderThemeParam` | "`true` for the default border, `false` to disable, or an object to customise it." |
| `FontFamilyFull` | "A single family name, or an array of names used as fallbacks." |
| `TextOrSegments` | "Plain text, or an array of segments for rich content." |
| `AgCssColorOrRef` | "A colour string, or a theme-colour reference object." |
| `AgChartLegendPosition` | "A placement keyword, or an object for fine-grained positioning." |
| `AgTimeAxisFormattableLabelFormat` | "A format string, or an object specifying a format per time unit." |
| `AgTimeInterval \| AgTimeIntervalUnit` | "A unit keyword (or number), or an object describing the interval." |
| state-serializable (`number \| AgStateSerializableBigInt`, initial-state `start`/`end`) | "A number, or a serialised value object." |

Properties with no prior JSDoc gained a full one-liner. `Default:` paragraphs were preserved with the clause inserted ahead of them. Properties already naming both forms (e.g. `theme`) were left untouched. While editing affected lines, also corrected one pre-existing typo (`formating`) and two US spellings (`colors`).

## Notes / known limitations

- `FontFamilyFull` also admits a single `GoogleFontFamily` object; the clause frames the complex form only as an array of fallbacks for terseness. The google-font object still renders as its own variant.

## Test plan

- `yarn nx format` clean
- `yarn nx build:types ag-charts-types` passes (JSDoc-only edits)
- `yarn nx lint ag-charts-types` 0 errors
- Spot-check the options/api-reference rendering for a few touched props (legend `padding`, series `fill`, a `themeParamsOptions` colour, time-axis `format`) to confirm the clause appears and the simple form reads clearly above the expanded complex tree.

### DIFF
--- a/packages/ag-charts-types/src/chart/axisOptions.ts
+++ b/packages/ag-charts-types/src/chart/axisOptions.ts
@@ -46,9 +46,9 @@ export interface AgAxisCaptionOptions {
     fontWeight?: FontWeight;
     /** The font size in pixels to use for the title. */
     fontSize?: FontSize;
-    /** The font family to use for the title. */
+    /** The font family to use for the title. A single family name, or an array of names used as fallbacks. */
     fontFamily?: FontFamilyFull;
-    /** The colour to use for the title. */
+    /** The colour to use for the title. A colour string, or a theme-colour reference object. */
     color?: AgCssColorOrRef;
     /** Spacing between the axis labels and the axis title. */
     spacing?: PixelSize;
@@ -114,7 +114,7 @@ export interface AgContinuousAxisOptions<
      * __Note:__ This does not override the `min` or `max` options.
      */
     nice?: boolean;
-    /** Configuration for the axis ticks interval. */
+    /** Configuration for the axis ticks interval. A unit keyword (or number), or an object describing the interval. */
     interval?: AgAxisContinuousIntervalOptions<TInterval>;
 }
 
@@ -170,11 +170,11 @@ export interface AgBaseAxisLabelStyleOptions extends LabelBoxOptions {
     fontWeight?: FontWeight;
     /** The font size in pixels to use for the labels. */
     fontSize?: FontSize;
-    /** The font family to use for the labels */
+    /** The font family to use for the labels. A single family name, or an array of names used as fallbacks. */
     fontFamily?: FontFamilyFull;
     /** Spacing in pixels between the axis label and the tick. */
     spacing?: PixelSize;
-    /** The colour to use for the labels */
+    /** The colour to use for the labels. A colour string, or a theme-colour reference object. */
     color?: AgCssColorOrRef;
 }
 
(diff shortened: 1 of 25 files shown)


packages/ag-charts-types/src/chart/axisOptions.ts
```diff
@@ -236,7 +236,7 @@ export interface AgTimeAxisFormattableLabelUnitFormat {
 export type AgTimeAxisFormattableLabelFormat = string | AgTimeAxisFormattableLabelUnitFormat;
 
 export interface AgTimeAxisFormattableLabelOptions<TContext = ContextDefault> extends AgBaseAxisLabelOptions<TContext> {
-    /** Format string used when rendering labels. */
+    /** Format string used when rendering labels. A format string, or an object specifying a format per time unit. */
```
AG-DavidG, line 239: "an object specifying a format per time unit" - is this correct?

packages/ag-charts-types/src/chart/themeOptions.ts
```diff
@@ -66,7 +66,7 @@ export type AgChartThemeName =
     | 'ag-financial-dark';
 
 export interface AgPaletteColors {
-    /** The fill colour for the palette. */
+    /** The fill colour for the palette. A colour string, or a gradient object. */
```
AG-DavidG, line 69: Does this not take images and patterns?

AG-DavidG (approved):
LGTM - few queries

packages/ag-charts-types/src/chart/themeOptions.ts
iMoses: That's the `AgPaletteColors` `fill` property. It's either a string or an object with `colorStops`

packages/ag-charts-types/src/chart/axisOptions.ts
iMoses: I think so.
```
interface AgTimeAxisFormattableLabelUnitFormat {
    millisecond?: string;
    second?: string;
    hour?: string;
    day?: string;
    month?: string;
    year?: string;
}
```

PR #7187, opened 2026-06-19
Make browser-benchmark comparisons like-for-like

alantreadway:
## Summary

The `npm:<version>` browser-benchmark comparison was not apples-to-apples, which manufactured phantom regressions in the overnight report.

**Root cause:** head was served from a local `nx build` bundle while the base was the release-packed *published* tarball. Both are unminified UMD, but the published artifact is production-built (sourcemaps off, debug asserts dropped), so head read a few percent slower purely from the build pipeline. Combined with a measured 5–14% single-run noise floor, sub-15% deltas were unreliable.

Worked example — `high-perf-line` Initial Load, the same head build measured every way:
| Method | Result |
|---|---|
| npm 1-run (asymmetric) | **+14% / +112ms** (reported as a major regression) |
| symmetric pack-and-swap, 2 runs | +7.7%, +10.5% (one freak-fast base run) |
| symmetric pack-and-swap, **5 runs** | **−2.2%, not notable** (converged — not a regression) |

## Changes

- **`swap-local-lib.sh` (new):** production-packs the local head packages and installs their UMD into the site, mirroring `swap-published-lib.sh` for the base. Both sides become a packed-tarball bundle through the same pipeline → like-for-like.
- **`compare-browser-latest.sh`:** swap the local pack into head for `npm:` bases; add `-n N` to run N times per side; fix an unbound-variable crash in the reporter invocation. Kept bash 3.2 compatible (no namerefs).
- **`compare-browser-results.js`:** accept multiple report files per side, aggregate across runs (median of run medians), compute cross-run CV, and gate notable regressions on `>min-pct` AND `>=min-abs-ms` AND not noisy AND above 2× cross-run CV. Backward compatible with a single file.
- **`benchmark.yml`:** pack local head bundles for `npm:` bases in the build job, so the overnight and release-cut runs are like-for-like.

## Test plan

- Reporter verified on real reports: single-file path unchanged; multi-file aggregates and the cross-run-CV gate demotes a noisy regression out of `notable`.
- `swap-local-lib.sh` produces a production UMD tarball (sourcemap-stripped) and installs it.
- End-to-end `compare-browser-latest.sh -j -n 5 npm:13.3.1 -- --examples high-perf-line` runs head-swap + 5 runs/side and reports the converged, denoised result.
- Scripts syntax-checked under bash 3.2 (macOS `/bin/bash`); CI runs bash 5.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -219,6 +219,15 @@ jobs:
                     yarn nx build ag-charts-website
                   fi
 
+            # Like-for-like vs a published base: serve head from its own locally-packed UMD,
+            # so head and the published base are each a packed-tarball bundle through the same
+            # pipeline. Without this, a locally-built head bundle vs a release-packed published
+            # bundle differ by a build-asymmetry offset that inflates head's apparent cost.
+            # git-ref bases are built from source here too, so no swap is needed for them.
+            - name: Pack local head bundles (npm-base parity)
+              if: matrix.role == 'head' && startsWith(needs.resolve.outputs.base, 'npm:')
+              run: ./tools/benchmark/swap-local-lib.sh dist/packages/ag-charts-website
+
             # Keep only what the benchmark pages need: the example pages
             # themselves, the dev-served library bundles, example-runner
             # boilerplate, and hashed astro assets.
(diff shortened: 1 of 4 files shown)


iMoses (approved):
LGTM

PR #7195, opened 2026-06-19
AG-17592 Thread colour-ref fill/stroke types across all packages

alantreadway:

Extends @laurencerobertz's WIP so colour-reference objects (`{ ref }`, `{ ref, mix }`, `{ ref, mix, onto }`) are accepted by the public types on `fill` and `stroke` everywhere the runtime already resolves them, while the internal/resolved code paths stay strongly typed.

### What changed
- **`ag-charts-types`** — widen `stroke` to `AgCssColorOrRef` on the cross-line, crosshair, band-highlight, range-area, waterfall connector-line, map-line-background and sparkline-crosshair options. These are the fields whose option-defs already validate via the shared ref-accepting validators, so the type now matches the runtime. Other narrow `CssColor` fields are intentionally left alone (their validators are still string-only — widening them would be a type-lie and is out of scope).
- **`ag-charts-core`** — complete the normalised-options layer Laurence started: morph `stroke` on the marker / pagination / axis-label / border normalised types, add normalised bar/histogram/line series styles, and relocate the internal colour types (`InternalAg*`, `NormalisedGradientColor`, …) into `normalisedCommonOptions` to break the `optionsDefaults ↔ normalisedCommonOptions` import cycle introduced by the widening.
- **`ag-charts-community` & `ag-charts-enterprise`** — thread the normalised style types through every series/feature that reads a resolved style and pushes it into the scene graph (bar, line, area, bubble, histogram, box-plot, heatmap, candlestick/ohlc, funnel/cone-funnel, chord, flow-proportion, radar/radial/range-area/range-bar, gauges, map-*, sankey/sunburst/treemap/waterfall, plus crosshair, band-highlight, ranges, scrollbar, gradient-legend, error-bar and annotation scenes). Refs are accepted on input; resolved colours flow to render.

### Notes
- **Type-only change** — only type annotations, type aliases, `import type`s and narrowing `as` casts at already-resolved boundaries (no `as any`/`as unknown as` introduced). Emitted JavaScript and runtime behaviour are unchanged.

### Test plan
- `yarn nx build:types` — clean across types/core/locale/community/enterprise.
- `yarn nx lint` — 0 errors across all four packages (pre-existing warnings only).
- `yarn nx test ag-charts-community` — 3687 passed / 6 skipped.
- `yarn nx test ag-charts-enterprise` — 2478 passed; 7 pre-existing transition/update **snapshot** diffs (integrated-charts crossfilter chart-type transitions + error-bar update tests). The diff images show identical rendered content (same colours/structure) — anti-aliasing/animation-frame noise only — and a type-only change cannot alter rendering, so these are environment baseline-drift, not regressions. Snapshots were **not** regenerated.

Fix #AG-17592

### DIFF
--- a/packages/ag-charts-core/src/config/optionsDefaults.ts
+++ b/packages/ag-charts-core/src/config/optionsDefaults.ts
@@ -5,15 +5,11 @@ import type {
     AgColorScaleColorStop,
     AgColorType,
     AgCssColorOrRef,
-    AgGradientColorBounds,
     AgGradientColorStop,
     AgGradientColorStrict,
-    AgGradientType,
     AgHighlightStyleOptions,
-    AgImageFill,
     AgLineHighlightStyleOptions,
     AgNumericValue,
-    AgPatternColor,
     AgSelectionContainment,
     AgSelectionStyleOptions,
     AgSeriesLineSegmentOptions,
@@ -56,7 +52,11 @@ import {
     undocumented,
     union,
 } from '../state/validation';
-import type { NormalisedGradientColor } from '../types/normalised-options/normalisedCommonOptions';
+import type {
+    InternalAgGradientColor,
+    InternalAgImageFill,
+    InternalAgPatternColor,
+} from '../types/normalised-options/normalisedCommonOptions';
 import { isObject } from '../utils/types/typeGuards';
 
 // Validator for internal theme operators.
@@ -181,41 +181,6 @@ export const gradientStrict = optionsDefs<AgGradientColorStrict>(
     gradientStrictDefs,
     'a gradient object with colour stops'
 );
-
-export interface InternalAgGradientColor extends NormalisedGradientColor {
-    /** Format of the gradient */
-    gradient?: AgGradientType;
-    /** The domain of the colour gradient, defaults to item. */
-    bounds?: AgGradientColorBounds;
-    /** Reverse the order of colour stops. */
-    reverse?: boolean;
-    /** Colour space to use when interpolating colours in the gradient. */
-    colorSpace?: ColorSpace;
-}
-
-export type ColorSpace = 'rgb' | 'oklch';
-
-export interface InternalAgPatternColor extends AgPatternColor {
-    /** Padding for the shape in the pattern unit. */
-    padding?: number;
-}
-export interface InternalAgImageFill extends AgImageFill {}
-
-export type RequiredInternalAgImageFill = Required<Omit<InternalAgImageFill, 'url' | 'width' | 'height'>> &
-    Pick<Partial<InternalAgImageFill>, 'url'> &
-    Pick<InternalAgImageFill, 'width' | 'height'>;
-
-export type RequiredInternalAgPatternColor = Required<Omit<InternalAgPatternColor, 'path'>> &
-    Pick<InternalAgPatternColor, 'path'>;
-
-export type RequiredInternalAgGradientColor = Required<InternalAgGradientColor>;
-
-export type InternalAgColorType = CssColor | InternalAgGradientColor | InternalAgPatternColor | InternalAgImageFill;
-export type RequiredInternalAgColorType =
-    | CssColor
-    | RequiredInternalAgGradientColor
-    | RequiredInternalAgPatternColor
-    | (RequiredInternalAgImageFill & Pick<InternalAgImageFill, 'url'>);
 
 export const strokeOptionsDef: OptionsDefs<StrokeOptions> = {
     stroke: colorOrRef,
--- a/packages/ag-charts-core/src/config/optionsDefaults.ts
+++ b/packages/ag-charts-core/src/config/optionsDefaults.ts
@@ -5,7 +5,6 @@ import type {
     AgColorScaleColorStop,
     AgColorType,
     AgCssColorOrRef,
-    AgGradientColor,
     AgGradientColorBounds,
     AgGradientColorStop,
     AgGradientColorStrict,
@@ -57,6 +56,7 @@ import {
     undocumented,
     union,
 } from '../state/validation';
+import type { NormalisedGradientColor } from '../types/normalised-options/normalisedCommonOptions';
 import { isObject } from '../utils/types/typeGuards';
 
 // Validator for internal theme operators.
@@ -182,7 +182,7 @@ export const gradientStrict = optionsDefs<AgGradientColorStrict>(
     'a gradient object with colour stops'
 );
 
-export interface InternalAgGradientColor extends AgGradientColor {
+export interface InternalAgGradientColor extends NormalisedGradientColor {
     /** Format of the gradient */
     gradient?: AgGradientType;
     /** The domain of the colour gradient, defaults to item. */
--- a/packages/ag-charts-core/src/types/normalised-options/normalisedCartesianSeries.ts
+++ b/packages/ag-charts-core/src/types/normalised-options/normalisedCartesianSeries.ts
@@ -21,8 +21,8 @@ export type NormalisedLineSeriesStylerResult = Normalised<
     { stroke?: CssColor; marker?: NormalisedSeriesMarkerStyle }
 >;
 
-export type NormalisedSeriesSegmentation<SegmentOptions = NormalisedSeriesShapeSegmentOptions> = Normalised<
-    AgSeriesSegmentation<SegmentOptions>,
+export type NormalisedSeriesSegmentation = Normalised<
+    AgSeriesSegmentation,
     never,
     { segments: NormalisedSeriesShapeSegmentOptions[] }
 >;
--- a/packages/ag-charts-core/src/types/normalised-options/normalisedCartesianSeries.ts
+++ b/packages/ag-charts-core/src/types/normalised-options/normalisedCartesianSeries.ts
@@ -1,0 +1,12 @@
+import type { FillStrokeMorph } from 'ag-charts-core';
+import type { AgSeriesSegmentation, AgSeriesShapeSegmentOptions } from 'ag-charts-types';
+
+import type { Normalised } from './normalise';
+
+export type NormalisedSeriesSegmentation<SegmentOptions = NormalisedSeriesShapeSegmentOptions> = Normalised<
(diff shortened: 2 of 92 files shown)


packages/ag-charts-core/src/config/optionsDefaults.ts
alantreadway, line 1: FYI: Moved to avoid circular modular dependencies.

packages/ag-charts-core/src/types/normalised-options/normalisedCartesianSeries.ts
```diff
@@ -0,0 +1,30 @@
...
+export type NormalisedBarSeriesStyle = Normalised<AgBarSeriesStyle, never, FillStrokeMorph>;
+
+export type NormalisedHistogramSeriesStyle = Normalised<AgHistogramSeriesStyle, never, FillStrokeMorph>;
+
+export type NormalisedLineSeriesStylerResult = Normalised<
+    AgLineSeriesStylerResult,
+    never,
+    { stroke?: CssColor; marker?: NormalisedSeriesMarkerStyle }
+>;
+
```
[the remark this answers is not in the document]
lsjroberts: This can be solved on a case-by-case basis if needed.

packages/ag-charts-core/src/types/normalised-options/normalisedCommonOptions.ts
```diff
@@ -1,13 +1,81 @@
...
+
+export type RequiredInternalAgGradientColor = Required<InternalAgGradientColor>;
+
+export type InternalAgColorType = CssColor | InternalAgGradientColor | InternalAgPatternColor | InternalAgImageFill;
+export type RequiredInternalAgColorType =
+    | CssColor
+    | RequiredInternalAgGradientColor
+    | RequiredInternalAgPatternColor
+    | (RequiredInternalAgImageFill & Pick<InternalAgImageFill, 'url'>);
+
```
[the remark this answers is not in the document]
lsjroberts: There won't be any unresolved ref objects by that point

packages/ag-charts-core/src/types/normalised-options/normalisedLegendOptions.ts
```diff
@@ -36,7 +33,7 @@ export type NormalisedLegendLabelOptions = Normalised<
 export type NormalisedPaginationMarkerStyle = Normalised<
     AgPaginationMarkerStyle,
     'fill' | 'strokeWidth' | 'strokeOpacity',
-    { fill: string }
+    { fill: string; stroke?: CssColor }
```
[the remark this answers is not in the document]
lsjroberts: I expect in this case the fill is never allowed to be anything but a css color or ref, i.e. not a gradient etc. In which case this is probably fine.

packages/ag-charts-enterprise/src/features/annotations/scenes/capScene.ts
```diff
@@ -30,7 +30,8 @@ export class ArrowCapScene extends CapScene {
         const leftEnd = Vec2.rotate(Vec2.from(0, armLength), angle + offsetAngle, origin);
         const rightEnd = Vec2.rotate(Vec2.from(armLength, 0), angle - offsetAngle, origin);
 
-        path.setProperties(rest);
+        // Stroke refs are resolved before reaching the scene node.
+        path.setProperties(rest as StrokeOptions & { stroke?: CssColor });
```
lsjroberts: This `update()` function should use `& NormalisedStrokeOptions` instead I expect

packages/ag-charts-enterprise/src/features/annotations/scenes/startEndScene.ts
```diff
@@ -112,12 +115,13 @@ export abstract class StartEndScene<Datum extends StartEndProperties> extends Li
     }
 
     protected updateHandles(datum: Datum, coords: Bounds4, bbox?: _ModuleSupport.BBox) {
+        // Colour refs in the handle styles are resolved at runtime before reaching the scene node.
         this.start.update({
-            ...this.getHandleStyles(datum, 'start'),
+            ...(this.getHandleStyles(datum, 'start') as NormalisedAnnotationHandleStyles),
```
lsjroberts, line 120: This doesn't seem the right place to cast this. The `Datum` type itself should use `Normalised` types.

packages/ag-charts-enterprise/src/features/annotations/scenes/withBackgroundScene.ts
```diff
@@ -37,7 +37,8 @@ export class WithBackgroundScene {
         background.path.closePath();
         background.checkPathDirty();
 
-        const backgroundStyles = this.getBackgroundStyles?.(datum) ?? datum.background;
+        // Colour refs are resolved during theme-merge, so the fill is a concrete colour by render time.
+        const backgroundStyles = (this.getBackgroundStyles?.(datum) ?? datum.background) as NormalisedFillOptions;
```
lsjroberts, line 41: Same here, `Datum` should use `Normalised` types, rather than casting it here.

packages/ag-charts-enterprise/src/features/band-highlight/bandHighlight.ts
```diff
@@ -159,12 +161,18 @@ export class BandHighlight extends AbstractModuleInstance {
 
         const { stroke, strokeWidth, strokeOpacity, lineDash, fill, fillOpacity, lineDashOffset } = options;
 
-        node.stroke = stroke;
+        // Colour refs are resolved during theme-merge, so stroke/fill are concrete by render.
+        node.stroke = stroke as CssColor;
```
lsjroberts: `options.stroke` should come from `NormalisedStrokeOptions`, and not be cast directly here

packages/ag-charts-enterprise/src/features/band-highlight/bandHighlight.ts
```diff
@@ -159,12 +161,18 @@ export class BandHighlight extends AbstractModuleInstance {
...
         node.strokeWidth = strokeWidth;
         node.strokeOpacity = strokeOpacity;
         node.lineDash = lineDash;
         node.lineDashOffset = lineDashOffset;
-        node.fill = getShapeFill(fill, this.fillGradientDefaults, this.fillPatternDefaults, this.fillImageDefaults);
+        node.fill = getShapeFill(
+            fill as InternalAgColorType,
```
lsjroberts: `options.fill` should come from `NormalisedFillOptions`, and not be cast directly here

packages/ag-charts-enterprise/src/features/crosshair/crosshair.ts
```diff
@@ -242,7 +243,8 @@ export class Crosshair
         const isVertical = this.isVertical();
 
         lineGroupSelection.each((line) => {
-            line.stroke = stroke;
+            // Colour refs are resolved during theme-merge, so this is a concrete CSS colour by render.
+            line.stroke = stroke as CssColor;
```
lsjroberts: `options.stroke` should come from `NormalisedStrokeOptions`, and not be cast directly here

packages/ag-charts-enterprise/src/features/error-bar/errorBarNode.ts
```diff
@@ -193,7 +196,8 @@ export class ErrorBarNode extends _ModuleSupport.Group<ErrorBarNodeDatum> {
         const { xBar, yBar, capDefaults } = this.datum;
 
         const whisker = this.whiskerPath;
-        this.applyStyling(whisker, whiskerStyle);
+        // Refs resolved at runtime before render.
+        this.applyStyling(whisker, whiskerStyle as NormalisedErrorBarStylingOptions);
```
lsjroberts: `whiskerStyle` should not be cast here, instead `this.formatStyles()` should return the correct type

alantreadway:
Thanks @codex — useful pass. The review's core thesis ("normalised colour aliases still inherit public shapes that can contain refs, so the render-time casts are risky") rests on a piece of the pipeline the diff doesn't show, so let me close the loop on each finding.

**The invariant holds because refs are resolved upstream of every cast site.** Colour references (`{ref}`/`mix`/`onto`, including gradient stops) are resolved in `optionsGraphOperations.ts` during options-graph processing — before any value reaches a series' properties. Runtime `styler`/`itemStyler` callback results, which bypass the graph, are run back through the same resolver via `optionsGraphService.resolvePartial(...)` before being narrowed (see `radialUtil.ts:134` and its comment). So by the time a value is cast to `NormalisedColorType`/`InternalAgColorType`/`CssColor`, it is provably ref-free on both paths.

Against the individual findings:

- **P1 – image fills / P1 – pattern fills:** `AgImageFill` and `AgPatternColor` were *not* widened — their colour fields (`backgroundFill`, `fill`, `stroke`) remain `CssColor`. The only public widenings in this PR are gradient *stops* (`AgGradientColorStop.color`) and `StrokeOptions.stroke`. So no ref can sit in those nested fields even at the type level; nothing to normalise.
- **P1 – `InternalAgColorType` aliases the normalised gradient:** this is sound precisely because gradient-stop refs are resolved in the graph stage (and by `resolvePartial` for styler results). `InternalAgGradientColor.colorStops: CssColor` is therefore honest at the properties layer, which is the only place these aliases are populated.
- **P2 – segmentation generic ignored:** agreed, it was dead surface — the generic is removed in 0f8baa2.
- **P2 – pagination marker fill vs stroke:** in the *normalised* type both are plain ref-free strings (`fill: string`, `stroke?: CssColor`), so they're consistent. The public `fill: AgColorType` asymmetry is pre-existing legend shape, out of scope here.
- **P2 – fallback stroke cast:** tidied in 0f8baa2 to `(stroke ?? '...') as CssColor`, matching the radial-bar sibling.

Separately worth noting for reviewers: the bulk of the render-time casts are sound but not strictly necessary — they exist because `resolvePartial<T>` returns the public (ref-inclusive) shape `Partial<T>` rather than a normalised one. Tightening that return type would concentrate them into a single assertion at the resolver; tracking that as a follow-up rather than widening this PR.

lsjroberts (approved):
LGTM

PR #7198, opened 2026-06-19
AG-17592 Return Resolved<T> from resolvePartial to drop boundary casts

alantreadway:

**Stacked on #7195** (base branch `ag-17592/fix-fill-and-stroke-types`) — review/merge that first.

### Why
#7195 widened public colour options to accept theme colour references and threads normalised (ref-free) colour types through render code. Colour refs (`ref`/`mix`/`onto`, incl. gradient stops) are resolved by the options-graph resolver — for runtime styler callbacks, via `optionsGraphService.resolvePartial(...)` — before any value reaches render code. But `resolvePartial` was typed to return the public, ref-inclusive shape (`Partial<T>`), so every styler boundary had to re-narrow with an `as` cast.

### What
- Add `Resolved<T>` (`ag-charts-core`): a mapped type that strips the `AgColorRef`/`AgColorRefMixOnto` members from every colour union, recursing through nested objects and arrays. Functions (e.g. marker-shape callbacks) and primitives pass through unchanged — the function guard is load-bearing, otherwise callbacks collapse to `{}`.
- Retarget `resolvePartial` (`OptionsGraph`, the `OptionsGraphAccessor` interface, and `OptionsGraphService`) to return `Resolved<Partial<T>>`.
- Remove the now-redundant colour-ref casts at the styler boundaries: `labelUtil`, `areaSeries`, `barSeries`, `lineSeries`, `donutSeries`, series marker (`series.ts`), `boxPlotSeries`, `rangeBarSeries`, and `radialUtil` (whose local `stylerResult` is retyped so its `fill` cast drops).

This encodes the refs-resolved guarantee in the type system rather than asserting it by hand at each call site.

### Scope
Type-only; emitted JS is unchanged. Removes ~13 boundary casts (≈25–30% of the colour casts in #7195). Out of scope, left for a possible follow-up: downstream casts on Properties-stored / node-datum styles (candlestick, ohlc, pyramid, map-shape/line, scrollbar, radial-gauge, error-bar background), which would need those fields retyped; and orthogonal `Required<>`/`RequireOptional<>`/`ResolvedStyleMixin` requiredness casts.

### Test plan
- `build:types` passes for `ag-charts-core`, `ag-charts-community`, `ag-charts-enterprise`.
- `lint` clean for all three; `nx format` applied.

Fix #AG-17592

### DIFF
--- a/packages/ag-charts-community/src/chart/labelUtil.ts
+++ b/packages/ag-charts-community/src/chart/labelUtil.ts
@@ -102,11 +102,12 @@ export function getLabelStyles<TParams>(
             selectionState: series.getSelectionStateString(nodeDatum?.datumIndex),
             candidateState: series.getCandidateStateString(nodeDatum?.datumIndex),
         };
-        const stylerResult = (series.ctx.optionsGraphService.resolvePartial(
-            labelPath,
-            series.cachedCallWithContext(label.itemStyler, { ...params, ...styleParams }),
-            { pick: false }
-        ) ?? {}) as NormalisedChartLabelStyleOptions;
+        const stylerResult =
+            series.ctx.optionsGraphService.resolvePartial(
+                labelPath,
+                series.cachedCallWithContext(label.itemStyler, { ...params, ...styleParams }),
+                { pick: false }
+            ) ?? {};
 
         return mergeDefaults(stylerResult, styleParams);
     }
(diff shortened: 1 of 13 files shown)


lsjroberts (approved):
LGTM

PR #7199, opened 2026-06-19
AG-17592 Remove colour-ref `as any` casts from examples

AG-DavidG:

Removes the `as any` casts that two generated examples used to compile colour-ref objects against the public types (added in PR #7162). The types now accept colour-ref objects on series `fill` and theme params, so the casts are no longer needed.

- `colours/_examples/colour-references/main.ts` — `fill: { ref: 'accentColor' }` (series fill)
- `themes/_examples/theme-colour-references/main.ts` — `textColor: { ref, mix }` theme-param refs (two sites)

## ⚠️ Merge order
Depends on the AG-17592 type widening (**PR #7195**) landing first. Merging this before #7195 will break example type-checking.

## Test plan
- [ ] After #7195 lands: `yarn nx validate-examples`
- [ ] Confirm the colour-references and theme-colour-references examples render across frameworks

Fix #AG-17592


alantreadway (approved):
LGTM

PR #7202, opened 2026-06-22
CRT-1122 Only show pointer cursor when datum selection is enabled

iMoses:
https://ag-grid.atlassian.net/browse/CRT-1122

The hover pointer cursor was gated on `series.isDatumSelectable(datumIndex)` alone. The base `Series.isDatumSelectable()` returns `true` unconditionally — it is a per-datum refinement intended to run *after* selection is confirmed enabled — so the pointer cursor was applied on hover to every series and gauge regardless of interactivity, falsely implying clickability.

The fix requires `isSelectionEnabled()` as well, restoring the pre-regression intent while preserving the new per-datum selectability check. The same gate covers the treemap (whose `isDatumSelectable` checks `isLeaf` but not `selection.enabled`) and all gauges.

**Test plan**

Added `seriesAreaManager.test.ts` (bar series, whose Rect nodes are hit-testable in JSDOM):
- non-interactive series → default cursor (fails before the fix, passes after)
- `selection.enabled: true` → pointer cursor
- `seriesNodeClick` listener present → pointer cursor

Verified: `yarn nx test ag-charts-community` (3690 pass), `build:types`, `lint` clean.

Fix #CRT-1122

### DIFF
--- a/packages/ag-charts-community/src/chart/series/seriesAreaManager.test.ts
+++ b/packages/ag-charts-community/src/chart/series/seriesAreaManager.test.ts
@@ -1,0 +1,69 @@
+import { afterEach, describe, expect, it } from 'vitest';
+
+import type { AgBarSeriesOptions, AgCartesianChartOptions } from 'ag-charts-types';
+
+import {
+    createChart,
+    getCursor,
+    hoverAction,
+    setupMockCanvas,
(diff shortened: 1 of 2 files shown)


alantreadway (approved):
LGTM

PR #7205, opened 2026-06-22
CRT-1130 Interrupt highlight animation on legend keyboard focus

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-1130

## Summary

Tabbing through legend items with the keyboard while a series highlight/dim animation was in progress did not update the dim: non-focused series stayed opaque until the animation finished, then dimmed.

**Root cause** (`packages/ag-charts-community/src/chart/legend/legend.ts`): legend highlight updates funnel through `updateHighlight` → `highlightNodeDatum`. During `InteractionState.Animation` the only handling deferred the update via `animationManager.onBatchStop(...)`, for both keyboard focus and mouse hover. That deferral exists for CRT-1034 (a passing hover must not interrupt a show/hide animation) but over-applied to keyboard focus, so a focus-driven dim only landed after the animation completed.

A second, subtler defect surfaced once the first was addressed: keyboard navigation fires `blur` (outgoing item) then `focus` (incoming item). The blur's deferred `clearHighlight` survives on the animation batch's stopped-callbacks and, when the next batch starts (`startBatch` → `AnimationBatch.destroy` → `dispatchStopped`), fires and clobbers the freshly-applied focus highlight.

**Fix:** thread a `fromKeyboardFocus` flag (derived from `event instanceof FocusEvent`, an idiom already used in this file) through `onHover` / `onLeave` / `clearHighlight` / `updateHighlight`. During `Animation`, a keyboard focus/blur now applies/clears the highlight immediately (interrupting the animation); pointer hover keeps the existing `onBatchStop` deferral, so CRT-1034 is preserved. The blur path clearing immediately also removes the stale deferred clear that was clobbering the incoming focus.

## Changes

- `legend.ts` — `fromKeyboardFocus` flag through the highlight path; keyboard focus/blur interrupt during animation, pointer hover still defers.
- `legendDOMProxy.ts` — `onLeave` accepts the blur event; the `blur` listener forwards `ev.sourceEvent` (a `FocusEvent`). `mouseleave` is unchanged (still defers).
- `test/utils.ts` — typed `getLegendModule()` test accessor.
- `legend.test.ts` — `CRT-1130` regression test asserting the highlight reflects the focused item immediately during the animation and survives the batch stop.

## Test plan

- `nx build:types ag-charts-community` + `ag-charts-enterprise`: pass.
- `nx lint ag-charts-community`: pass.
- Legend/highlight unit suites: the new `CRT-1130` test passes and the existing `CRT-1034` test still passes.

## Reviewer notes

- A `blur` caused by clicking elsewhere mid-animation is also a `FocusEvent`, so it now clears the legend highlight immediately too. This is intended and consistent with "deliberate focus change takes effect immediately"; low risk, but worth a glance.
- No keyboard E2E added: jsdom does not implement `:focus-visible`, so the real DOM `focus`→`onHover` path cannot be exercised end-to-end in a unit test (the regression test drives `onHover`/`onLeave` with real `FocusEvent`s directly). A faithful Playwright test belongs in `keyboard-nav.spec.ts` but is screenshot-based and needs baselines generated in the Docker E2E host.
- Pre-existing, unrelated to this change: the 4 `Legend > Large series count chart > should render legend correctly at width [200/400/600/800]` image-snapshot tests fail on the clean `b14.0.0` baseline (confirmed via `git stash`); not regressions, snapshots not regenerated.

Fix #CRT-1130

### DIFF
--- a/packages/ag-charts-community/src/chart/legend/legend.ts
+++ b/packages/ag-charts-community/src/chart/legend/legend.ts
@@ -1093,25 +1093,26 @@ export class Legend {
             this.ctx.tooltipManager.removeTooltip(this.id, undefined, true);
         }
 
-        this.updateHighlight(datum?.enabled, datum, series);
+        this.updateHighlight(datum?.enabled, datum, series, undefined, event instanceof FocusEvent);
         this.ctx.eventsHub.emit('legend:item-hover', null);
     }
 
-    onLeave() {
+    onLeave(event?: FocusEvent | MouseEvent) {
         if (this.checkInteractionState()) return;
         this.ctx.tooltipManager.removeTooltip(this.id, undefined, true); // true = delayed
-        this.clearHighlight();
+        this.clearHighlight(event instanceof FocusEvent);
     }
 
-    private clearHighlight(): void {
-        this.updateHighlight(undefined, undefined, undefined);
+    private clearHighlight(fromKeyboardFocus = false): void {
+        this.updateHighlight(undefined, undefined, undefined, undefined, fromKeyboardFocus);
     }
 
     private updateHighlight(
         enabled: boolean | undefined,
         legendDatum: CategoryLegendDatum | undefined,
         series: SeriesType | undefined,
-        event?: ActiveLoadMementoEvent
+        event?: ActiveLoadMementoEvent,
+        fromKeyboardFocus = false
     ): void {
         if (this.checkInteractionState()) return;
         type InternalUpdateOpts = {
@@ -1140,11 +1141,17 @@ export class Legend {
             if (this.ctx.interactionManager.isState(InteractionState.Default) || event?.initialState) {
                 updateManagers(opts);
             } else if (this.ctx.interactionManager.isState(InteractionState.Animation)) {
-                // Updating the highlight can interrupt animations, so defer both setting and
-                // clearing highlights until the current animation batch completes.
-                this.ctx.animationManager.onBatchStop(() => {
+                // A keyboard focus change is a deliberate navigation that must take effect immediately,
+                // interrupting any in-progress animation. Pointer hover, by contrast, defers both setting
+                // and clearing highlights until the current animation batch completes so that a passing
+                // hover does not interrupt a show/hide animation.
+                if (fromKeyboardFocus) {
                     updateManagers(opts);
-                });
+                } else {
+                    this.ctx.animationManager.onBatchStop(() => {
+                        updateManagers(opts);
+                    });
+                }
             } else if (opts === undefined) {
                 updateManagers(opts);
             }
--- a/packages/ag-charts-community/src/chart/legend/legend.ts
+++ b/packages/ag-charts-community/src/chart/legend/legend.ts
@@ -101,6 +101,10 @@ export class Legend {
 
     private readonly truncatedItems: Set<string | number> = new Set();
 
+    /** Incremented on every highlight update so a deferred (animation-batch) update can detect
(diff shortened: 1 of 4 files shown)


packages/ag-charts-community/src/chart/legend/legend.ts
```diff
@@ -1140,11 +1141,17 @@ export class Legend {
             if (this.ctx.interactionManager.isState(InteractionState.Default) || event?.initialState) {
                 updateManagers(opts);
             } else if (this.ctx.interactionManager.isState(InteractionState.Animation)) {
-                // Updating the highlight can interrupt animations, so defer both setting and
-                // clearing highlights until the current animation batch completes.
-                this.ctx.animationManager.onBatchStop(() => {
+                // A keyboard focus change is a deliberate navigation that must take effect immediately,
```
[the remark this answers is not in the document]
alantreadway: Confirmed and fixed in 2c36703bc1.

Reproduced empirically: with a pointer hover queued during the animation followed by a keyboard focus on a different item, the highlight applied correctly on focus but reverted to the pointer's target on batch stop (the stale onBatchStop callback overwrote it).

Fix: `highlightNodeDatum` now bumps a per-update token (`highlightUpdateToken`) on every update; the deferred onBatchStop callback only runs if its captured token is still the latest, so any later update — immediate keyboard focus or a newer deferred hover — invalidates it. CRT-1034 is preserved: a lone deferred pointer hover is still the latest token and runs as before.

Added a regression test (`keyboard focus is not clobbered by a deferred pointer-hover update on batch stop`) covering exactly this sequence: deferred pointer hover -> immediate keyboard focus in the same batch -> assert the focused highlight survives batch stop.

packages/ag-charts-community/src/chart/legend/legendDOMProxy.ts
```diff
@@ -114,7 +114,7 @@ export class LegendDOMProxy {
             button.addListener('mouseenter', (ev) => itemListener.onHover(ev.sourceEvent, markerLabel));
             button.addListener('mouseleave', () => itemListener.onLeave());
             button.addListener('contextmenu', (ev) => itemListener.onContextClick(ev, markerLabel));
-            button.addListener('blur', () => itemListener.onLeave());
+            button.addListener('blur', (ev) => itemListener.onLeave(ev.sourceEvent));
```
olegat: nit: The `ev instanceof FocusEvent` is little bit weird. We normally use clearly named listeners like `onLeave`, `onFocus`, `onBlur`, and so on. Perhaps an `onBlur() -> onLeave(true)` wrapper might be more consistent with the code-style that's typically used.

olegat (approved):
LGTM, just an optional nit.

packages/ag-charts-community/src/chart/legend/legendDOMProxy.ts
alantreadway: Good call — refactored in 5894c6ec1c. The keyboard signal is now an explicit `fromKeyboardFocus` boolean threaded through `onHover`/`onLeave` rather than sniffing `ev instanceof FocusEvent`. The DOM proxy wires the `blur` listener to `onLeave(true)` and the keyboard `focus` path to `onHover(..., true)`; mouse listeners use the default `false`. Behaviour is unchanged (keyboard focus/blur still interrupt immediately, pointer hover still defers). The only remaining `instanceof FocusEvent` is in `toTooltipMeta`, where it legitimately picks tooltip positioning (node centre vs pointer offset), not the highlight behaviour.

PR #7208, opened 2026-06-22
CRT-1134 Expose Scrollbar.enabled so chart.update preserves zoom

iMoses:
https://ag-grid.atlassian.net/browse/CRT-1134

Switching `scrollbar.vertical.position` at runtime (the docs `scrollbar/#position` example) removed the scrollbar entirely instead of moving it.

Root cause is not in the scrollbar layout. `Chart.hasViewportSupport()` reads `.enabled` off the module *instance* (`getModule('scrollbar').enabled`), as `zoom.ts` and `navigator.ts` both expose via `get enabled() { return this.opts.enabled; }`. The migration off `jsonApply` (aa1bdbd913) dropped the scrollbar's old `public enabled` field without adding the getter, so `hasViewportSupport()` returned `false` for scrollbar-only charts. With no viewport support, every `chart.update()` hits the "reset zoom to initial state" branch and resets the zoom to full extent — and the auto-visible scrollbar hides because `span === 1`. Any `update()` triggered it, not just a position change.

The fix re-exposes `enabled` on the scrollbar instance, mirroring the sibling modules, so the zoom is preserved across updates.

**Test plan**

Added to `scrollbar.test.ts`:
- switching `vertical.position` from `right` to `left` keeps the zoom and the scrollbar laid out (fails before the fix, passes after)
- an unrelated update (title change) keeps the zoom
- a `same-side-single-axis` scenario for the existing placement matrix (covers single-axis same-side, horizontal/vertical × inner/outer)

Verified: `yarn nx test ag-charts-enterprise` scrollbar suite (33 pass), `build:types`, `lint` clean, spec `tsc --noEmit` clean.

Fix #CRT-1134


alantreadway (approved):
LGTM

PR #7209, opened 2026-06-22
CRT-1124 Fix gauge keyboard navigation stuck across layers

alantreadway:
## Ticket

[CRT-1124](https://ag-grid.atlassian.net/browse/CRT-1124) — [NR] Gauge keyboard navigation buggy with multiple layers

## Problem

Gauge keyboard navigation could get stuck. On a gauge with multiple layers (e.g. a segmented bar plus targets), navigating Down into the targets, Right to a non-first target, then Up should return focus to the main section — but the arrow keys stopped responding.

### Root cause

In `pickGaugeFocus`, an Up/Down move that changes layer carried the previous within-layer `datumIndex` across to the new layer and clamped it. When the new layer collapsed its data into a single focusable node (segmented bar `datumUnion`), the clamped index pointed at a datum that no longer existed as a focusable node, so the pick returned `undefined`. The focus write-back was then skipped and the stale focus state persisted, leaving navigation stuck until a later delta happened to re-land on a valid index. Non-segmented configurations masked the bug because their `nodeData.length === 1` forced the index to 0.

## Fix

`packages/ag-charts-enterprise/src/series/gauge-util/pick.ts`: when focus actually moves to a different layer, reset the within-layer datum index to 0; otherwise clamp-preserve it. This guarantees the pick always resolves to a valid datum (the union node on layer 0, target 0 on layer 1) so the focus write-back always runs. Covers both linear and radial gauges via the shared `pickFocus`.

## Tests

- `linearGaugeSeries.test.ts`: 7 keyboard-navigation unit tests — AC1 regression (direct `pickFocus` and full JSDOM keyboard pipeline), AC2 first-item path, single-layer/no-target no-op, and a non-first-target ArrowDown boundary no-op.
- `keyboard-nav.spec.ts` (E2E): extended the gauge case with the direct ArrowUp-from-target → bar assertion (reuses existing screenshots).

## Acceptance criteria

| AC | Criterion | Status |
|----|-----------|--------|
| 1 | Up from a non-first lower-layer item returns to the main section (not stuck) | PASS |
| 2 | No regression of first-item / single-layer cases | PASS |

## Notes

Pre-existing enterprise a11y defect (not a b14 regression). No public API or documentation changes. The shared `seriesAreaManager` focus structure (write-back-on-failure) was left untouched to keep blast radius minimal late in the release; a candidate follow-up tech-debt note.

### DIFF
--- a/packages/ag-charts-enterprise/src/series/gauge-util/pick.ts
+++ b/packages/ag-charts-enterprise/src/series/gauge-util/pick.ts
@@ -37,7 +37,17 @@ export function pickGaugeFocus(self: GaugeSeries, opts: PickFocusInputs): PickFo
     const { data, selection } = others[otherIndex];
     if (data == null || data.length === 0) return;
 
-    const datumIndex = clamp(0, opts.datumIndex, data.length - 1);
+    // When focus actually moves between layers (an Up/Down that lands on a different layer),
+    // the carried datumIndex belongs to the previous layer and has no meaning in the new one:
+    // layer 0 is a single merged node (`datumUnion`) and the target layer has no positional
+    // relationship to it. Reset to 0 so the pick always resolves to a datum present in the
+    // chosen layer's selection (the union datum for layer 0, target 0 for the targets layer);
+    // a stale non-zero index could otherwise reference a datum absent from the selection,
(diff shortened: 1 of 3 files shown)


olegat (approved):
LGTM

PR #7210, opened 2026-06-22
CRT-1140 - Fix arrow keys moving annotations by bandwidth instead of 1px

alantreadway:
## Summary

Fixes [CRT-1140](https://ag-grid.atlassian.net/browse/CRT-1140) — arrow keys move a selected annotation by ~10px instead of the documented 1px.

The keyboard move step in `onKeyDown` was clamped to a floor of the axis `bandwidth`:

```ts
const xStep = Math.max(context?.xAxis.scale.bandwidth ?? 0, ctrlShift ? 10 : 1);
const yStep = Math.max(context?.yAxis.scale.bandwidth ?? 0, ctrlShift ? 10 : 1);
```

The move is performed in **pixel space**, so the intended semantic is the literal pixel step. On band/ordinal x-axes (e.g. the financial preset's `ordinal-time` axis) the per-band width is several pixels, inflating an unmodified arrow press to roughly the band width rather than 1px. The fix removes the bandwidth floor:

```ts
const xStep = ctrlShift ? 10 : 1;
const yStep = ctrlShift ? 10 : 1;
```

## Changes

- `packages/ag-charts-enterprise/src/features/annotations/annotations.ts` — remove the `bandwidth` floor from the keyboard move step.
- `packages/ag-charts-enterprise/src/features/annotations/annotations.test.ts` — add a regression block asserting a plain arrow moves 1px and Ctrl/Shift + arrow moves 10px on a band/ordinal x-axis, independent of bandwidth.

## Testing

- `yarn nx build:types ag-charts-enterprise` — PASS
- `yarn nx lint ag-charts-enterprise` — PASS (0 errors)
- Annotation suite — 37/37 PASS (2 new regression tests). Verified the new tests fail against the old code (annotation moved ~128px) and pass against the new code.

## Notes

- No docs change required — docs already state the correct 1px / 10px behaviour; the code was the defect.
- Pre-existing regression (shipped in 13.x), low-risk single-file change. Category-snapping (`groupPercentage`) is a separate downstream concern and is unaffected by the step size.

### DIFF
--- a/packages/ag-charts-enterprise/src/features/annotations/annotations.test.ts
+++ b/packages/ag-charts-enterprise/src/features/annotations/annotations.test.ts
@@ -1,7 +1,13 @@
 import { afterEach, describe, expect, it } from 'vitest';
 
 import { type AgCartesianChartOptions, AgCharts } from 'ag-charts-community';
-import { extractImageData, setupMockCanvas, setupMockConsole, waitForChartStability } from 'ag-charts-community-test';
+import {
+    deproxy,
+    extractImageData,
+    setupMockCanvas,
+    setupMockConsole,
+    waitForChartStability,
+} from 'ag-charts-community-test';
 
 import { prepareEnterpriseTestOptions } from '../../test/utils';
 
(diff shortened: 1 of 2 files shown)


iMoses (approved):
LGTM

PR #7212, opened 2026-06-22
CRT-1131 Call chart.update() when interacting with the controls

olegat:
https://ag-grid.atlassian.net/browse/CRT-1131


alantreadway (approved):
LGTM

PR #7213, opened 2026-06-22
CRT-1142 Fix linear-gauge labels example for Angular generation

iMoses:
https://ag-grid.atlassian.net/browse/CRT-1142

The linear-gauge "labels" example wrapped the range-input value with the global `Number()` inside its inline event handler. Angular evaluates template expressions against the component instance, so the generated Angular example resolved `Number` to `ctx.Number` and threw `TypeError: ctx.Number is not a function` at runtime.

Move the string-to-number coercion into the `setValue` handler (where globals are available) and pass the raw DOM value from the template, matching the example's sibling handlers (`setLabelPlacement`, `setAvoidCollisions`). The generated output is now clean and idiomatic across all frameworks; no generator change required.

Validated with `yarn nx validate-examples ag-charts-website` (872 examples, no issues) and by inspecting the regenerated Angular/React/Vue/TS output.

Fix #CRT-1142

### DIFF
--- a/packages/ag-charts-website/src/content/docs/linear-gauge/_examples/labels/index.html
+++ b/packages/ag-charts-website/src/content/docs/linear-gauge/_examples/labels/index.html
@@ -24,8 +24,8 @@
             min="0"
             max="100"
             value="50"
-            oninput="setValue(Number(event.target.value))"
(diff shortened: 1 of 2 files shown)


alantreadway (approved):
LGTM

PR #7215, opened 2026-06-22
CRT-1145, CRT-1146 Honour theme-override padding in $applyPadding

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-1145
https://ag-grid.atlassian.net/browse/CRT-1146

## Summary

Legend padding supplied through `theme.overrides` was silently dropped, breaking layout — `legend.padding`/`legend.item.padding` (CRT-1145: chart bunched at top, no legend) and `legend.item.marker.padding` (CRT-1146: legend item text not shown). The same options supplied as direct chart options worked, confirming a theming-pipeline defect rather than a legend bug.

**Root cause** (regression from the AG-15505 `$applyPadding` rework): `applyPaddingOperation` resolved padding only from user options via `dangerouslyGetUserOption`, never from theme overrides. Compounding this, the `$applyPadding` edge-priority special-case in `resolveVertexInEdgePriority` let a *plain* override value short-circuit and win raw — skipping the operation entirely. The override value therefore reached the resolved options un-expanded (a bare number) or with default sides dropped (a partial-side object), so downstream padding arithmetic (`padding.top + padding.bottom`) produced `NaN`.

Both ticket repros flow through this one operation, so a single fix clears both (and their integrated-charts mirrors RTI-3385 / RTI-3386).

## Changes

- `applyPaddingOperation` resolves the theme override and folds it into the merge with precedence **user > override > default**, preserving number→object expansion and partial-side merge over defaults.
- `resolveVertexInEdgePriority` routes a plain override through the default `$applyPadding` operation (instead of grafting it raw) when only the default carries the operation; an override-supplied operation template still wins as before.
- New `dangerouslyGetThemeOverride` accessor mirrors `dangerouslyGetUserOption`, resolving the series-type / `common` / direct override namespaces (the constructor's unwrap order).

## Test plan

- `optionsGraph.test.ts` — new `$applyPadding` cases: number-override expansion, partial-side override merge over defaults, and user-option-beats-override precedence.
- `chartTheme.test.ts` — chart-level regression asserting override-supplied legend padding resolves identically to direct options, for both ticket scenarios.
- Full `ag-charts-community` suite green (3697 passed). Enterprise failures observed are pre-existing/flaky on `b14.0.0` (verified against the clean baseline) and unrelated to this change.

Fix #CRT-1145
Fix #CRT-1146


iMoses (approved):
LGTM

iMoses:
Worth testing with partial padding objects on 2 separate theme overrides paths.
E.g. 
```
series.common.legend.padding = { top: 10, bottom: 10 };
series.bar.legend.padding = { left: 20, right: 20 };
```

PR #7216, opened 2026-06-22
CRT-1148 Hide caption node when disabled via options update

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-1148

A v14 caption migration moved the `node.visible = enabled` write into a code path that `ChartCaptions.positionCaptions()` only reaches inside the `if (caption.enabled)` guard. Caption scene nodes are long-lived and reused across updates, so disabling a previously-enabled title/subtitle/footnote (via options or theme overrides) skipped the write entirely — the node stayed `visible: true` and un-repositioned (hence it also failed to re-centre on resize).

This restores the unconditional pre-migration behaviour: `applyToNode()` is now called for all three captions at the top of `positionCaptions()` on every layout, so `node.visible` is synced to `enabled` even for disabled captions. The now-redundant call inside `positionCaption()` is removed, leaving a single source of truth.

Adds a `test.each` over title/subtitle/footnote covering enable -> disable -> re-enable through the public `proxy.update(options)` API. Verified failing without the fix and passing with it.

Fix #CRT-1148

### DIFF
--- a/packages/ag-charts-community/src/chart/caption.test.ts
+++ b/packages/ag-charts-community/src/chart/caption.test.ts
@@ -18,6 +18,7 @@ import {
     setupMockConsole,
     waitForChartStability,
 } from './test/utils';
+import type { AgChartProxy } from './test/utils';
 
 describe('Caption', () => {
     setupMockConsole();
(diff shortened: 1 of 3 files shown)


iMoses (approved):
LGTM

PR #7217, opened 2026-06-22
CRT-1143 Make full update() replace options, dropping omitted subtrees

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-1143

Switching between chart configurations via `AgChartInstance.update()` — e.g. nightingale to radar-area in the homepage gallery switcher — left a stale subtitle on screen. The same latent problem affected any option subtree (legend, axes, zoom, captions) omitted from a subsequent `update()`.

`update()` is documented to take complete options — its JSDoc states "Options provided should be complete and not partial" — i.e. REPLACE semantics. The live-update path instead computed `jsonDiff(prev, new)` and merged the diff onto the previous user options. `jsonDiff` had no removal channel — an omitted key looked identical to "unchanged" — and `merge` ignores `undefined`, so any subtree present before and omitted now was retained.

This restores the intended replace semantics for `update()`, while `updateDelta()` remains the deep-merge API:

- `jsonDiff` gains an opt-in removal sentinel: keys present in the source but omitted from the target are expressed as removals rather than dropped. Default behaviour is unchanged for the existing caller.
- The full-update path (reached only when no explicit delta is supplied) uses the sentinel and strips the resulting symbols, reusing the existing `merge`/`removeLeftoverSymbols` removal pipeline.
- `updateDelta()` always supplies an explicit delta (`DeepPartial`), so it never enters this branch — its merge semantics are unchanged.

### Performance — fast path is preserved for no-removal updates

The fix deliberately does **not** disable `update()`'s fast-path cache for all calls. Symbols are stripped — and the slow path taken — **only when the diff actually contains a removal sentinel** (an omitted subtree). The decision is `stripSymbols ||= containsRemovalSentinel(diff)`, gated on a real removal rather than on "is this an `update()`".

- A normal no-removal `update()` (re-applying complete options where nothing was dropped) produces a diff with no sentinels, so it keeps the existing fast path — no behavioural or caching change.
- `containsRemovalSentinel` is an early-exit recursive scan over the (small) diff — not the full options — and skips the opaque `data` payload.
- The Grid sparkline cell-renderer hot path uses `updateDelta()`, which never enters this branch at all — zero overhead, zero behaviour change there. Normal fast-path `update()`/`updateDelta()` calls (incl. the sparkline `context`/`data` path) are untouched.
- All 16 sparkline fast-path optimisation assertions (fast setup for `width`/`height`/`data`/`container`/`context` via both `update()` and `updateDelta()`, pooling, instance reuse) pass.

Trade-off: an `update()` that genuinely removes a subtree now takes the slow path — correct, since a removal needs full reprocessing, and exactly the CRT-1143 scenario that was previously broken.

Test plan:
- `jsonDiff` unit tests for the removal sentinel, including a fast-path-safety case (same-shape value change emits no sentinel).
- Integration tests: a full `update()` drops an omitted subtitle; `updateDelta()` retains it.
- Full community suite green (3694). Enterprise pre-existing/flaky snapshot failures verified identical on the clean base branch (unrelated to this change).
- core + community lint/types/format green.
- Website e2e deferred to CI: the local Dockerised Playwright runner is infra-flaky here (mount-permission/timeout), and the CRT-1143 repro is already covered by the integration test above. CI runs it cleanly.

Fix #CRT-1143

### DIFF
--- a/packages/ag-charts-community/src/chart/agChartV2.test.ts
+++ b/packages/ag-charts-community/src/chart/agChartV2.test.ts
@@ -370,4 +370,55 @@ describe('AgChartV2', () => {
             await waitForChartStability(chart);
         });
     });
+
+    describe('CRT-1143 update() replace semantics', () => {
+        const withSubtitle = (): AgChartOptions => ({
+            title: { text: 'Title' },
+            subtitle: { text: 'Subtitle' },
+            data: [
+                { x: 'a', y: 1 },
+                { x: 'b', y: 2 },
+            ],
+            series: [{ type: 'bar', xKey: 'x', yKey: 'y' }],
+        });
+
+        const withoutSubtitle = (): AgChartOptions => ({
+            title: { text: 'Title' },
+            data: [
+                { x: 'a', y: 1 },
+                { x: 'b', y: 2 },
+            ],
+            series: [{ type: 'bar', xKey: 'x', yKey: 'y' }],
+        });
+
+        it('should drop a subtitle omitted from a full update()', async () => {
+            const initial = withSubtitle();
+            prepareTestOptions(initial, container);
+            chart = AgCharts.create(initial);
+            await waitForChartStability(chart);
+            expect(chart.getOptions().subtitle).toEqual({ text: 'Subtitle' });
+
+            const updated = withoutSubtitle();
+            prepareTestOptions(updated, container);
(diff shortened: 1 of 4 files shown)


iMoses (approved):
LGTM

PR #7218, opened 2026-06-22
CRT-1144 Hide box-plot shapes on legend toggle in combos

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-1144

In a box-plot + scatter combo, toggling the box-plot off via the legend left its
shapes drawn and stretched (a b14.0.0 regression).

Box-plot extends `AbstractBarSeries`, so it is treated as a grouped series and the
shared `cartesianSeries` hidden-series early-return does not fire — `populateNodeData()`
re-filled the node array while hidden. Box-plot has no datum hide animation, so those
stale nodes were re-projected through the changed y-scale and rendered stretched once
the scatter series kept the y-domain populated.

The fix produces no nodes when the series is hidden, both contained within
`boxPlotSeries.ts`:
- gate `canIncrementallyUpdate` on `this.visible` (do not alias the stale node array)
- early-return from `populateNodeData()` when `!this.visible`

The shared `cartesianSeries` path and the `processData` guard are deliberately left
untouched. Waterfall (solo-only) and OHLC/candlestick (resolve to empty node data on
hide) were verified not to share the defect, so no change was warranted there.

A regression test was added to `boxPlotSeries.test.ts` using a box-plot + scatter combo
fixture mirroring the repro: an image snapshot (no residual box shapes) plus a
node-data-cleared invariant. Both were confirmed failing pre-fix and passing post-fix.

Fix #CRT-1144


packages/ag-charts-enterprise/src/series/box-plot/boxPlotSeries.ts
```diff
@@ -426,6 +429,13 @@ export class BoxPlotSeries extends _ModuleSupport.AbstractBarSeries<BoxPlotSerie
      * Populate node data by iterating over raw data.
      */
     protected override populateNodeData(ctx: BoxPlotSeriesNodeDatumContext): void {
+        // A hidden box-plot must leave no shapes behind. Box-plot belongs to the bar grouping, so the
+        // base createNodeData() still calls populateNodeData() while hidden (the ungrouped early-return
+        // does not apply). Unlike bar it has no datum hide animation (no animationResetFns), so any nodes
+        // produced here would re-project the stale box geometry through the current y-scale and render
+        // stretched (CRT-1144). Produce no nodes when hidden so the series renders nothing.
```
iMoses, line 436: nit: concise comments

iMoses (approved):
LGTM, one nit

packages/ag-charts-enterprise/src/series/box-plot/boxPlotSeries.ts
alantreadway: We'll follow-up post release.

PR #7220, opened 2026-06-22
CRT-1129 Preserve x-position when keyboard focus crosses series of differing node density

iMoses:
## Summary

Keyboard focus navigation between cartesian series did not preserve the x-position when the two series had different node densities. In a chart mixing a **range-area** series (two nodeData entries per x-position — low + high markers, only even indices focus-enabled) with a **line** series (one per x-position), arrowing between them landed on the wrong x: range→line jumped to x×2 and line→range jumped to x÷2.

**Root cause:** focus tracks `focus.datumIndex` as a raw *nodeData array index* and `onArrow` carried it unchanged across series boundaries, assuming a 1:1 index↔x mapping that range-area violates.

**Fix:** when focus moves to a different series, translate the position by **category (x) value** rather than carrying the raw node index — reusing the same `getCategoryValue` ⇄ `datumIndexForCategoryValue` mechanism that shared tooltips already use (`chart.ts`).

## Changes

- `series.ts` — new `focusDatumIndexForCategoryValue` (base default = data index; correct for one-node-per-x series).
- `dataModelSeries.ts` — generic override mapping a category value to the first **focus-enabled** nodeData index for that data point. Yields `2k` for range-area and `k` for line automatically via the existing `isDatumEnabled` filter, so no range-area-specific code is required and any future N-per-x series benefits.
- `seriesTypes.ts` — added to the `ISeries` contract.
- `seriesAreaManager.ts` — remap `focus.datumIndex` through the new method when the focused series actually changes; falls back to the previous behaviour when the x-domains don't overlap. The standalone (gauge/hierarchy) and within-series left/right paths are untouched.

## Test plan

- New enterprise test `rangeArea.keyboardNav.test.ts` (line + range-area sharing x categories): navigate to a known x, cross with ArrowDown/ArrowUp in both directions, assert the logical data index / x-value is preserved. Confirmed it fails without the fix (`expected 1 to be 2` — the x÷2 bug) and passes with it.
- `build:types` and `lint` (0 errors) pass for `ag-charts-community` and `ag-charts-enterprise`.
- `yarn nx test ag-charts-community` — all pass.
- `yarn nx test ag-charts-enterprise` — only two pre-existing ErrorBars (AG-16943) snapshot failures, confirmed present on the base branch and unrelated to this change.

Manual: verified the `step-interpolation-combination` example no longer shifts the x-position when arrowing between the line and range-area series.

Fix #CRT-1129

### DIFF
--- a/packages/ag-charts-community/src/chart/series/dataModelSeries.ts
+++ b/packages/ag-charts-community/src/chart/series/dataModelSeries.ts
@@ -256,4 +256,15 @@ export abstract class DataModelSeries<
             if (objectsEqual(categoryValue, xValue)) return datumIndex;
         }
     }
+
+    public override focusDatumIndexForCategoryValue(categoryValue: any): number | undefined {
+        const dataIndex = this.datumIndexForCategoryValue(categoryValue);
+        const nodeData = this.getNodeData();
+        if (dataIndex === undefined || nodeData === undefined) return;
+        for (let i = 0; i < nodeData.length; i += 1) {
+            if (nodeData[i].datumIndex === dataIndex && this.isDatumEnabled(nodeData, i)) {
+                return i;
+            }
+        }
+    }
 }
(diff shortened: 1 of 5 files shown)


olegat (commented):
There's a moderate risk so I would advise that we hold off this release. Left further details on JIRA.

iMoses:
Leaving this to Oli

PR #7221, opened 2026-06-22
CRT-1145 Test partial user padding merges per-side over override

alantreadway:
## What

Follow-up to #7215 (merged). Adds one unit test in `optionsGraph.test.ts` covering a gap in the `$applyPadding` merge coverage.

## Why

#7215 made `$applyPadding` resolve with **per-side** precedence `user > override > default`, but the added unit cases only covered:
- single-number override expanded to every side
- partial override merged over defaults
- full user option preferred over a full override

They did **not** assert a *partial user object* layered over a *partial theme-override* — i.e. that each side resolves independently and a partial user object does not wholesale-replace the override/default sides.

## Test

```
default:  { top: 4, right: 8, bottom: 4, left: 8 }   ($applyPadding)
override: { right: 15, bottom: 15 }
user:     { left: 20 }
=> result: { top: 4, right: 15, bottom: 15, left: 20 }
```

Each side is sourced from a different layer (top→default, right/bottom→override, left→user), so a regression to wholesale object replacement is caught. Passes against the merged implementation; `format`, `build:types`, and `lint` all green.

### DIFF
--- a/packages/ag-charts-community/src/module/optionsGraph.test.ts
+++ b/packages/ag-charts-community/src/module/optionsGraph.test.ts
@@ -920,6 +920,26 @@ describe('OptionsGraph', () => {
                     axes: expect.any(Object),
                 });
             });
+
+            it('should merge a partial user option over a partial theme-override per side', () => {
+                const themeConfig = {
+                    line: {
+                        padding: {
+                            $applyPadding: { top: 4, right: 8, bottom: 4, left: 8 },
+                        },
+                    },
+                };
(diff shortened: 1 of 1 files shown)


iMoses (approved):
LGTM, but a small nit
Maybe we could change the overrides to be from `common` to make sure a more "unique" combination works as expected?

alantreadway:
Good call — switched the override to the `common` namespace (default still from `line`), so the test now also covers the common-namespace fallback in `dangerouslyGetThemeOverride`. Pushed in 0b04b58. Still green.

PR #7224, opened 2026-06-22
CRT-1119 Make Data Source example Reload button show fresh data

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-1119

The Data Source (async-data) example carried a `Reload` button whose only visible
effect was a brief loading overlay, after which the same deterministic dataset
returned — so the button's purpose was unclear.

This keeps the button and makes a reload produce a visible change:

- The fake server now returns a different randomised price series on each reload
  (a new seed per `Database.refresh()`), so clicking Reload visibly changes the
  chart.
- The price random-walk gains a gentle mean-reversion towards the centre so any
  seed stays within the chart's fixed y-axis range (400..1600) and the line never
  clips off the top or bottom.
- The button is relabelled **Reload Data** to make its purpose obvious.
- The docs `### Triggering a Reload` section now references the labelled button
  and explains the loading-overlay-then-refresh behaviour.

## Validation
- `yarn nx lint ag-charts-website` — 0 errors (pre-existing warnings only)
- `yarn nx generate-examples ag-charts-website` — all framework variants generated cleanly
- The data-generation bounds were verified analytically across 40 successive seeds: the series stays within [518, 1489], comfortably inside the fixed [400, 1600] axis.
- e2e (`test:e2e`) status noted in the PR thread.

Fix #CRT-1119

### DIFF
--- a/packages/ag-charts-website/src/content/docs/async-data/_examples/async-data/data.ts
+++ b/packages/ag-charts-website/src/content/docs/async-data/_examples/async-data/data.ts
@@ -5,7 +5,13 @@ import { createSeededRandom } from './seededRandom';
  * frontend developer you can safely ignore this part of the example.
  */
 export const Database = {
-    get: () => (data ??= generate()),
+    get: () => (data ??= generate(seed)),
+    // Re-generate the dataset with a new seed so a reload returns a visibly different price series, mimicking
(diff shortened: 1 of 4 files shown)


AG-DavidG (approved):
LGTM

PR #7229, opened 2026-06-23
CRT-1149 Fix Control key state tracking on Windows

olegat:
https://ag-grid.atlassian.net/browse/CRT-1149

Claude diagnosed that the root cause is that holding down the Control key on Windows send repeated keydown events but only one keyup events when releasing. Whereas on macOS, holding down and releasing the the Control key always sends exactly 1 keydown and 1 keyup event. This causes the candidacyUnion counter to never reach 0 on Windows.

Fortunately, the KeyboardEvents `ctrlKey` and `metaKey` properties are consistent across all platform, so we can reuse the mousemove counting mechanism.


alantreadway (approved):
LGTM

PR #7230, opened 2026-06-23
CRT-1126 Fix gauge line highlighted over empty space

iMoses:
## Summary

Linear and radial gauges highlighted their **line target** when hovering empty space far from the line (e.g. above the bar, under the top label of the simple-bullet example). The line registered as "hovered" almost anywhere in the gauge area.

**Root cause:** the highlight comes from the `NEAREST_NODE` pick fallback, which calls `node.distanceSquared`. Gauge target markers are positioned via `translationX/Y` (their `x/y` stay `0`), but `Marker.distanceSquared` reads `this.x/this.y` — and the `MatrixTransform` mixin inverse-transformed the point for `pickNode`/`pickNodes` but **not** for `distanceSquared`. So the picker measured the target's distance from the scene **origin**, not its rendered position; near the gauge's local origin the distance collapsed toward `0` and the line won the pick. Secondarily, the `line` marker (a thin segment of length `size`) was modelled as a circle of radius `size/2`, giving an oversized hit zone even when positioned correctly.

**Fix:** make `distanceSquared` transform-aware in the `MatrixTransform` mixin (mirroring the existing `pickNode` override), and hit-test the line target against its actual segment instead of a bounding circle.

## Changes

- `transformable.ts` — `MatrixTransform.distanceSquared` now inverse-transforms the point before measuring, matching `pickNode`/`pickNodes`.
- `node.ts` — base `distanceSquared` default (`Infinity`), mirroring `containsPoint`, so the mixin's `super` call is valid for non-shape nodes. `path.ts`/`line.ts` gain the now-required `override` keyword (no behaviour change).
- `marker.ts` — split the circle hit-test into a non-transforming `distanceSquaredLocal` seam called directly by `isPointInPath` (already-local coords) and via the transform-wrapped public `distanceSquared`, so the exact and nearest pick paths can't double-transform.
- `gauge-util/lineMarker.ts` — new `LineMarker` overriding `distanceSquaredLocal` to measure distance to the line segment (thin band) for `line` targets; other target shapes fall back to the circle.
- `linearGaugeSeries.ts` / `radialGaugeSeries.ts` — `markerFactory` builds `LineMarker`.

## Test plan

- New `marker.test.ts` — a translated `Marker` measures distance from its rendered position, not the origin (confirmed failing pre-fix: `19900` vs `0`).
- New `gauge-util/lineMarker.test.ts` — `LineMarker` hits along the line, misses beside/beyond it, is transform-correct, and falls back to the circle for non-line shapes.
- `linearGaugeSeries.test.ts` (CRT-1126) — hovering empty space near the line registers no highlight, hovering the line does. Verified non-vacuous: fails when the `MatrixTransform` fix is reverted.
- `build:types` and `lint` pass for `ag-charts-community` and `ag-charts-enterprise`.
- `yarn nx test ag-charts-community` — all pass (3693). Both gauge snapshot suites pass (no visual regression from the `LineMarker` swap).
- `yarn nx test ag-charts-enterprise` — only the pre-existing crossfilter (integratedChartsExamples) and ErrorBars (AG-16943) snapshot failures remain; confirmed present on the base branch and unrelated to this change.

Shares the root cause with **AG-15859**.

Fix #CRT-1126

### DIFF
--- a/packages/ag-charts-community/src/chart/marker/marker.test.ts
+++ b/packages/ag-charts-community/src/chart/marker/marker.test.ts
@@ -1,0 +1,21 @@
+import { describe, expect, it } from 'vitest';
+
+import { Marker } from './marker';
+
+describe('Marker', () => {
+    describe('distanceSquared', () => {
+        it('measures distance from the translated position, not the untransformed origin', () => {
+            const marker = new Marker();
+            marker.shape = 'square';
+            marker.size = 20;
+            marker.translationX = 100;
+            marker.translationY = 100;
+
+            // Pointer over the marker's rendered centre (its translation) reads as a hit.
+            expect(marker.distanceSquared(100, 100)).toBe(0);
+
+            // Pointer near the scene origin — far from the rendered marker — must not read as a hit.
+            expect(marker.distanceSquared(0, 0)).toBeGreaterThan(0);
+        });
+    });
+});
--- a/packages/ag-charts-community/src/chart/marker/marker.ts
+++ b/packages/ag-charts-community/src/chart/marker/marker.ts
@@ -52,7 +52,7 @@ class InternalMarker<D = any> extends Path<D> {
     }
 
     override isPointInPath(x: number, y: number): boolean {
-        return this.distanceSquared(x, y) <= 0;
+        return this.distanceSquaredLocal(x, y) <= 0;
     }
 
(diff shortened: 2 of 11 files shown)


alantreadway (approved):
LGTM

PR #7231, opened 2026-06-23
CRT-1132 Allow data: in frame-src CSP so Firefox chart PNG download works (#7226)

taktran:
Port of https://github.com/ag-grid/ag-charts/pull/7226


alantreadway (approved):
LGTM

PR #7232, opened 2026-06-23
Gate browser benchmarks on multi-run cross-run variance

alantreadway:
## Why

A `/benchmarks` run on PR #7217 (CRT-1143) reported a `+39.4%` "notable regression" on `axes-1M-time / Zoom`. It wasn't real: CRT-1143 only touches the `chart.update()` options path (the zoom benchmark drives zoom via synthetic wheel events, never `update()`), and the *same example's* Initial-Load/Chart-Create — a path the change can't run — moved `+30%` in lockstep. The "regressions" clustered **by CI shard/VM, not by code path** (per-shard median offset −0.5% … +7.0%).

Root cause: GitHub-hosted runners are shared Azure VMs. Head runs first, base second on the same VM; CPU drift between the two passes is a systematic per-VM bias. CI ran a **single run per side** (`runCount:1`, `crossRunCv:null`), so only the weak intra-run sample CV gated — which is how `+39.4%` passed as "denoised".

## What changed

- **`benchmark.yml`** — run each shard **`AG_BENCHMARK_RUNS` (=3)** times per side, interleaved (head,base,head,base,…); the report job merges per-run and passes every run to the comparison so the **cross-run-CV gate already in `compare-browser-results.js` engages**. Job timeout 45→90 min. A run is excluded from the gate if it is incomplete (missing shard) or degraded (any errored/timed-out example under soft-fail); ≥2 clean runs are required for a variance estimate.
- **`browser-benchmark.ts` + `contention.ts`** — whole-VM **CPU steal + PSI** sampling per example and runner **host identity**, as *logged diagnostics* (preserved per shard via `merge-browser-results.js` → `shardEnvironments`). Linux-only; `null` on macOS/dev. `contention.ts` parsers are unit-tested.
- No change to the comparison's gating maths — the existing cross-run-CV logic is the denoiser; this PR just feeds it enough runs.

## Experiment (the testing you can reproduce)

3× `workflow_dispatch` of this workflow, head vs `latest` (true delta ≈ 0, so all movement is noise):

| Gate | Run 1 | Run 2 | Run 3 | 3-run aggregated |
|---|---|---|---|---|
| False "notable" regressions | 3 | 2 | 1 | **0** |

`axes-1M-time / Zoom` (the false +39.4% on #7217) shows raw +9.1% but **cross-run CV 23.6% → demoted**. ~**52/116** tests exceed a 10% run-to-run noise floor — invisible to single-run gating.

**Rejected on the data (had been prototyped):**
- **CPU steal gating** — `0.0%` on every Azure EPYC shard; no signal.
- **PSI gating** — real (multi-second stalls) but non-discriminating: the biggest %-movers are micro-benchmarks with *low* PSI; high-PSI heavy examples are %-stable.
- **Calibration auto-normalisation** — *widened* dispersion every run (median |Δ| ~2% → ~3.8%) and added false positives; a short instantaneous probe is too noisy a divisor.

Steal/PSI/host are kept as logged diagnostics only — host identity already proved its worth (3 different EPYC models within one run).

## Test plan

- `npx tsx --test tools/benchmark/contention.test.ts` — pass.
- Reverted comparison over the 3 control runs → **0** notable (cross-run-CV gate alone).
- **Validation run of this multi-run workflow** ([run 28010735948](https://github.com/ag-grid/ag-charts/actions/runs/28010735948)) — **green end-to-end**: 5 shards × 3 runs × 2 sides fanned out, per-run merge produced `head.r1..3`/`base.r1..3`, compare emitted `runCount: 3` and **`notable: []`**. The clearest proof the gate works: `multi-series / Datum Highlight / 15x` moved +11.3% / +19ms (not flagged "noisy") but was correctly demoted because its cross-run CV is 8.27% (noise band 2×8.27% = 16.5% > 11.3%) — under the old single-run YAML (`crossRunCv: null`) that gate is inert and this would have surfaced as a false notable.

**Targets `b14.0.0`** so the multi-run gating is active for the current release cycle's `/benchmarks` runs. Cherry-pick of the `latest` PR #7228 — identical commit (`git range-diff` clean), validated by the same run linked above.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -376,10 +376,13 @@ jobs:
                   merge-multiple: true
 
             # Merge each run's shards into one head/base report per run, then hand every run to
-            # the comparison so it can median across runs and gate on cross-run CV. Soft-fail only
-            # covers failures within a shard; a wholly missing shard would silently shrink a run, so
-            # per-run shard counts are enforced. A flaky run is tolerated as long as >=2 survive
-            # (two runs are the minimum for a cross-run variance estimate).
+            # the comparison so it can median across runs and gate on cross-run CV. Two integrity
+            # checks keep a degraded run out of the gate: a wholly missing shard would silently
+            # shrink a run, so per-run shard counts are enforced; and AG_BENCHMARK_SOFT_FAIL lets a
+            # shard finish with errored/timed-out examples (the report is written before the runner
+            # exits non-zero), so each shard's failure counts are inspected too. A flaky run is
+            # tolerated as long as >=2 clean runs survive (the minimum for a cross-run variance
+            # estimate).
             - name: Merge shard results
               env:
                   SHARDS: ${{ needs.resolve.outputs.shards }}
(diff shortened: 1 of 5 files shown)


iMoses (approved):
LGTM

PR #7233, opened 2026-06-23
CRT-1140 Revert arrow-key annotation move fix

alantreadway:
## Summary

Reverts the CRT-1140 fix (PR #7210, commit `f7aaacb263`) on `b14.0.0`.

Following discussion with product, we are backing out the CRT-1140 change for now and shipping **b14.0.0** with the known issue. This restores the prior keyboard-move behaviour:

- Keyboard move step returns to `Math.max(bandwidth, ctrlShift ? 10 : 1)` for both axes.
- Removes the regression test added in PR #7210.

## Known issue (accepted for now)

On band/ordinal x-axes (e.g. the financial preset's ordinal-time axis) an unmodified arrow press moves the annotation by the band width rather than the documented 1px. This is the behaviour we are knowingly shipping.

## Notes

- The snap-aware follow-up (`e02d5495ad`, branch `CRT-1140/annotation-keyboard-step-snap-aware`) was never merged to `b14.0.0`, so no revert is needed for it — it simply won't be shipped.

## Testing

- `yarn nx build:types ag-charts-enterprise` ✅
- `yarn nx lint ag-charts-enterprise` ✅ (0 errors)
- `annotations.test.ts` ✅ 35/35 pass

### DIFF
--- a/packages/ag-charts-enterprise/src/features/annotations/annotations.test.ts
+++ b/packages/ag-charts-enterprise/src/features/annotations/annotations.test.ts
@@ -1,13 +1,7 @@
 import { afterEach, describe, expect, it } from 'vitest';
 
 import { type AgCartesianChartOptions, AgCharts } from 'ag-charts-community';
-import {
-    deproxy,
-    extractImageData,
-    setupMockCanvas,
-    setupMockConsole,
-    waitForChartStability,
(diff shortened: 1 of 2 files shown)


iMoses (approved):
LGTM

PR #7236, opened 2026-06-23
CRT-1120 Refine union-types API reference display

iMoses:
https://ag-grid.atlassian.net/browse/CRT-1120

Follow-up refinements from the AG-14806 review (union-types API display, shipped in #7160) plus related polish on the website API reference page.

### What changed
- **`theme` renders as a plain, non-expandable row.** It has its own dedicated themes-api page, so it links there rather than expanding inline. Added a `noExpand` config (`noExpand={['theme']}` on the options page), threaded through the reference components and the options navigation.
- **Renamed the affordance** "See/Hide available types" → "See/Hide available interfaces" across the UI, e2e tests, and comments.
- **Type text (pink) is clickable only when it has a code block** to toggle — a `code` member's definition or a mixed-union signature. Such rows now show a pointer cursor; child-property rows, union variant lists and plain rows keep the default cursor.
- **Special-type union code blocks show the alias only.** `formatTypeToCode` gained an `expandReferences` flag so series/axes/annotations unions no longer inline their sub-interfaces (each variant has its own navigable page).
- **Scoped include/exclude/prioritise to the top-level interface only** — nested interfaces and union variants render their full member set.
- **Hid `AgThemeColorParam`** from the api-ref, and **constrained the expanded code block width** so it stays within the row bounds.
- **Prevented the docs header link icon from shrinking** (`flex: 0 0 28px`).

### Test plan
- `apiReferenceHelpers.test.ts`: added `formatTypeToCode` coverage for both `expandReferences` modes.
- `e2e/api-ref-page.spec.ts`: added a test asserting `theme` renders as a plain non-expandable row; updated affordance-label assertions.
- `yarn nx test ag-charts-website` (388 passed) and `yarn nx lint ag-charts-website` green locally.

### Notes for reviewers
- SCSS changes live in `external/ag-website-shared` (shared with the AG Grid website via git-subrepo). The cursor additions are inert for Grid; please sanity-check the `flex-basis`/`min-width` constraint on `.expandedUnionTypes` against Grid's reference docs.
- `noExpand` matches by member name (like the existing `keepExpanded`), so it is global to any member named `theme`; currently only `AgChartOptions.theme` exists.

Fix #CRT-1120

### DIFF
--- a/external/ag-website-shared/src/components/link-icon/LinkIcon.module.scss
+++ b/external/ag-website-shared/src/components/link-icon/LinkIcon.module.scss
@@ -9,6 +9,7 @@
 }
 
 .docsHeaderIcon {
+    flex: 0 0 28px;
     width: 28px;
     height: 28px;
     padding: 0;
--- a/external/ag-website-shared/src/components/reference-documentation/ApiReference.module.scss
+++ b/external/ag-website-shared/src/components/reference-documentation/ApiReference.module.scss
@@ -145,6 +145,9 @@
 
     &.expandedUnionTypes > .expandedContent {
         margin-left: $spacing-size-6;
+        // Account for the left margin so the code block stays within the row's bounds.
+        flex-basis: calc(100% - #{$spacing-size-6});
+        min-width: calc(100% - #{$spacing-size-6});
     }
 
     &.expandedChildProps::after,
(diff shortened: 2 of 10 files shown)


alantreadway (approved):
LGTM

PR #7237, opened 2026-06-23
CRT-1143 Skip opaque option keys in removal-sentinel scan

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-1143

Follow-up to the CRT-1143 full-`update()` replace fix, which introduced a stack overflow.

`ChartOptions.containsRemovalSentinel` walks the option diff to decide whether a subtree was removed. It descended into `context` — an opaque user pass-through that AG Grid's cross-filter integration sets to a component instance (`context: this`), and which can equally be a self-referential plain object — following its internal cycle until the stack overflowed.

Skip the opaque pass-through payloads **by name**, reusing the shallow-key set already defined for the options graph (`SHALLOW_OPTION_KEYS`: `context`, `data`, `topology`), and gate the walk on `isPlainObject` so non-plain pass-throughs such as a DOM `container` are not descended either. The removal sentinel is only ever written onto plain-object nodes, so declining to recurse cannot miss one — it only avoids following the cycle.

## Test plan
- New regression tests in `optionsModule.test.ts`: a full `update()` no longer throws when `context` carries circular references — class instance (reference changed / newly introduced) or a self-referential plain object.
- `nx build:types`, `nx lint`, and `nx test ag-charts-community` (full suite) all green.

## Scope
Fixes the CRT-1143 regression in the removal-sentinel scan. The real AG Grid integration (`context: this`, a class instance) is fully handled on every update path. Fully hardening *every* options walk against a self-referential **plain-object** context (e.g. a changed-reference plain-object cycle in `jsonDiff`, or the font/CSS walks on a slow-path update) is a broader, pre-existing concern those walks predate CRT-1143 and lack cycle guards; tracked separately rather than widened into this release-branch fix.

Fix #CRT-1143

### DIFF
--- a/packages/ag-charts-community/src/module/optionsGraph.ts
+++ b/packages/ag-charts-community/src/module/optionsGraph.ts
@@ -98,6 +98,12 @@ export function createOptionsGraph(
     });
 }
 
+// Opaque user pass-through payloads: stored shallow in the options graph and skipped by the
+// removal-sentinel scan in `optionsModule`. Never descended into, so user-supplied cycles in these
+// values (e.g. `context` set to a grid component or a self-referential object) cannot drive an
+// options walk into a stack overflow.
+export const SHALLOW_OPTION_KEYS = new Set<string>(['context', 'data', 'topology']);
+
 /**
  * The OptionsGraph combines the theme config, params, palette, overrides and user options into a graph which can then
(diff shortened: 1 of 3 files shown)


iMoses (approved):
LGTM

PR #7238, opened 2026-06-23
CRT-1120 Resolve computed type aliases in api-ref via TS checker

iMoses:
https://ag-grid.atlassian.net/browse/CRT-1120

Build a `ts.Program`/`TypeChecker` in the docs reference generator so non-structural type aliases (mapped / conditional / indexed-access) that reduce to a literal union are expanded into their concrete values instead of being emitted as unrenderable raw text.

`AgThemeColorParam` (`ColorKeys<Required<AgChartThemeParams>>`) previously resolved to an interface with no members, so its api-ref row expanded empty — hence the stop-gap that hid it. The generator now flattens it to the concrete union of theme-param keys, and that stop-gap (`hiddenInterfaces` entry) is removed.

Changes:
- `type-mapper.ts`: construct a `ts.Program` + `TypeChecker` (compiler options from the nearest tsconfig); iterate inputs in their original order so the resolver's order-sensitive generics handling is unchanged.
- `type-utils.ts`: thread an optional `checker` through `formatNode`; new gated `expandComputedType` emits a literal union only when a computed alias genuinely reduces to one, otherwise falls back to existing behaviour.
- `executor.ts`: pass input file paths (not contents) to `TypeMapper`.
- `apiReferenceHelpers.ts`: drop `AgThemeColorParam` from `hiddenInterfaces`.

Regenerating the resolved interfaces changes exactly three keys — all computed types: `AgThemeColorParam` and `SeriesType` now resolve to concrete literal unions, and the internal `ColorKeys` helper sheds a leaked generic binding. Everything else is byte-identical. The independent `compare-types` path (raw-text, no checker) is unaffected.

Note: the expanded `AgThemeColorParam` includes `focusShadow`/`popupShadow` because `CssColor` and `CssShadow` are both bare `string` aliases, so they are structurally assignable to `AgCssColorOrRef`. This is the true type semantics; tightening it would require distinct branding in `ag-charts-types` (out of scope).

Fix #CRT-1120

### DIFF
--- a/external/ag-website-shared/src/components/reference-documentation/ApiReference.module.scss
+++ b/external/ag-website-shared/src/components/reference-documentation/ApiReference.module.scss
@@ -182,7 +182,11 @@
         margin-top: $spacing-size-3;
         width: 100%;
 
-        .expandedChildProps & {
+        // When the row stacks, the right column drops onto its own line beneath the side-line that
+        // marks the expansion. Indent it past the line so the description and toggle button clear it,
+        // matching the type meta above. Both expanded variants draw the line, so both need the indent.
+        .expandedChildProps &,
+        .expandedUnionTypes & {
             padding-left: $spacing-size-6;
         }
     }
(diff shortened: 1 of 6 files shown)


iMoses:
image

alantreadway (approved):

PR #7242, opened 2026-06-23
CRT-1125 [CLAUDE] Drive focus indicator via native focus({ focusVisible })

olegat:
https://ag-grid.atlassian.net/browse/CRT-1125

Move focus-visibility state tracking out of the library and into the browser by using the experimental `focusVisible` option of `HTMLElement.focus()`.

Previously the focus indicator was shown/hidden via an inline `opacity` override on the swap-chain announcer, with `:focus-visible` only read once (via getComputedStyle) on the first focus event. This was flimsy and never forwarded the desired visibility across announcer swaps.

Now:
  - The first focus-frame defers to the browser's `:focus-visible` decision.
  - A mouse-click on the focused series-area forces the indicator hidden, and a keydown forces it visible, by re-focusing the already-focused announcer with `focus({ focusVisible })`. This is a no-op refocus, so it updates `:focus-visible` without swapping announcers or re-announcing to screenreaders. Re-announcement still happens only when the focused content changes.
  - Blur returns to the "waiting for first focus-frame" state.
  - The desired `focusVisible` is persisted on the swap chain and forwarded on every announcer swap, so the indicator no longer flickers during keyboard navigation.

`FocusIndicator.overrideFocusVisible` (the inline-opacity hack) is replaced by `setDesiredFocusVisible`; the public `chartService.overrideFocusVisible` contract name is retained.

lib.dom (TS 5.8) does not yet type `focusVisible` on `FocusOptions`, so a local `FocusOptionsExperimental` type adds it. A static assertion guards the workaround and fails to compile once the bundled TS lib gains `focusVisible` natively, signalling that the local type can be removed.

Note: this is an experimental branch; the two-latest-major-versions browser support policy is intentionally relaxed (Safari < 18.4 lacks `focusVisible`).

### DIFF
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -636,7 +636,7 @@ export abstract class Chart extends Observable implements ModuleInstance, ChartS
     }
 
     overrideFocusVisible(visible: boolean | undefined): void {
-        this.seriesAreaManager.focusIndicator?.overrideFocusVisible(visible);
+        this.seriesAreaManager.focusIndicator?.setDesiredFocusVisible(visible);
     }
 
     // Use a wrapper to comply with the @typescript-eslint/unbound-method rule.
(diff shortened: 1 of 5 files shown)


olegat:
Closing - e2e playwright test failure suggest that this approach will not work.

PR #7244, opened 2026-06-23
AG-17518 Add e2e tests for axes/series expansion in api-ref panel

iMoses:
https://ag-grid.atlassian.net/browse/AG-17518

Adds CI coverage for the API reference right-hand-side panel, asserting that expanding the `axes` and `series` nodes does not throw the *"Avoid using type-literals in user facing typings…"* error. All tests run under `setupIntrinsicAssertions`, which fails on any `pageerror`/console error, so the type-literals error is caught wherever a node is expanded.

- **Options tab (AC1):** new test expands the top-level `axes` node and asserts its axis-type variants render. The `series` half was already covered by the existing series-expansion tests.
- **Themes tab (AC2):** two new tests drill the `overrides` tree — `overrides → common → axes` (axis-type variants) and `overrides → bar → series` (themeable series properties).

Verified against the running site: all three new tests pass (chromium), and the expansions are non-vacuous (child rows asserted visible).

Fix #AG-17518

### DIFF
--- a/packages/ag-charts-website/e2e/api-ref-page.spec.ts
+++ b/packages/ag-charts-website/e2e/api-ref-page.spec.ts
@@ -34,6 +34,12 @@ function getNavigationText(page: Page, text: string): Locator {
     return getNavigationTree(page).getByText(text, { exact: false });
 }
 
+async function expandNavNode(page: Page, matcher: string | RegExp) {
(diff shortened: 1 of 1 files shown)


alantreadway (approved):
LGTM

PR #7247, opened 2026-06-23
Fix cross-OS snapshot flakiness in error-bar and crossfilter tests

iMoses:
Two enterprise snapshot tests rendered differently on macOS (local) than on Linux (CI) but never failed CI, because CI regenerates baselines with `--update`. Both are deterministic test-harness issues, not product bugs.

## error-bar AG-16943 "via update" tests
`proxy.update({...})` was called with raw options, dropping the harness's `animation: { enabled: false }` default and re-enabling animation. With no animation mock installed, the post-update frame was captured at a wall-clock-dependent point, so the settled state was captured on macOS but an early transition frame on Linux. Routed the update options through `prepareEnterpriseTestOptions` so the settled state is captured deterministically — matching the integrated-charts update tests.

## Crossfilter integrated-charts examples
`prepareTestOptions` set `params: undefined` for `ag-default-dark`, which also dropped the bundled-font (`Verdana` → registered Arimo) normalization. The crossfilter examples are the only ones using the dark theme, so they fell back to the theme's default font stack — which is unregistered and resolves to different system fonts per OS (San Francisco/Helvetica vs DejaVu/Liberation). Pinned `fontFamily` for dark themes while still letting the dark theme supply colours.

## Result
The 7 affected baselines are regenerated and now use the bundled, cross-platform-identical font and deterministic state. Full community (3707) and enterprise (2507) suites pass locally.


alantreadway (approved):
LGTM

PR #7253, opened 2026-06-24
Define explicit Fibonacci retracement ratio sets

alantreadway:
Resolves the SonarCloud quality-gate code smell flagged on `latest` for the Fibonacci retracement annotation utils.

The 4-band ratio set was derived by filtering specific magic-number literals (`23.6`, `78.6`) out of the 6-band array. Each band now maps to its own named constant (`FIBONACCI_RETRACEMENT_RATIOS_4` / `_6`), removing the magic numbers and making the sets robust to future changes of the source array.

The 10-band set concatenates the retracement and extension ratios and is now sorted with a numeric comparator, so band order stays ascending (significant for the rendering loop) without relying on the source arrays being pre-ordered relative to each other.

The resulting ratio arrays are identical to before, so there is no behavioural change.

Verified with `nx format`, `nx build:types ag-charts-enterprise`, and `nx lint ag-charts-enterprise` (0 errors).


packages/ag-charts-enterprise/src/features/annotations/utils/fibonacci.ts
```diff
@@ -2,14 +2,14 @@ import type { Bounds4 } from 'ag-charts-core';
...
+const FIBONACCI_RETRACEMENT_RATIOS_4 = [0, 38.2, 50, 61.8, 100];
+const FIBONACCI_RETRACEMENT_RATIOS_6 = [0, 23.6, 38.2, 50, 61.8, 78.6, 100];
 const FIBONACCI_EXTENSION_RATIOS = [161.8, 261.8, 361.8, 423.6];
-const FIBONACCI_RATIOS = [...FIBONACCI_RETRACEMENT_RATIOS, ...FIBONACCI_EXTENSION_RATIOS];
 
 const FIBONACCI_RATIOS_MAP: Record<FibonacciBands, number[]> = {
-    10: FIBONACCI_RATIOS,
-    6: FIBONACCI_RETRACEMENT_RATIOS,
-    4: FIBONACCI_RETRACEMENT_RATIOS.filter((r) => r !== 78.6 && r !== 23.6),
+    10: [...FIBONACCI_RETRACEMENT_RATIOS_6, ...FIBONACCI_EXTENSION_RATIOS].sort((a, b) => a - b),
```
iMoses: nit: `sort` is redundant

iMoses (approved):
LGTM

packages/ag-charts-enterprise/src/features/annotations/utils/fibonacci.ts
alantreadway:
```suggestion
    10: [...FIBONACCI_RETRACEMENT_RATIOS_6, ...FIBONACCI_EXTENSION_RATIOS],
```

alantreadway: Fixed, ta

PR #7257, opened 2026-06-24
Harden dataSource scroll-zoom tests against timing flake

alantreadway:
## Summary

Fixes intermittent CI failures in `packages/ag-charts-enterprise/src/features/data-source/dataSource.test.ts` (the `Unit & Integration Tests` job). These are a **load-sensitive timing flake**, not a product regression — first observed on an unrelated PR ([#7252](https://github.com/ag-grid/ag-charts/pull/7252), label-collision work that touches no zoom/data-source code).

### Root cause

The suite asserted immediately after `scrollAction(...)`, which only waits a fixed **50 ms**. A wheel-driven zoom commits on a later frame and re-requests data asynchronously, so under CPU load the effect lands *after* that wait. The assertions then read stale state:

- window tests read the pre-zoom window (`expected undefined to deeply equal 'four'`);
- the source test sees no request (`expected [] to include 'user-interaction'`);
- the invalid-response-recovery tests emit `getData`-validation warnings *after* `expectWarningsCalls()` drained them, tripping `setupMockConsole`'s `afterEach`.

With `retry: 0`, any lost race hard-fails the job.

### Fix

Replace the implicit fixed-delay reliance with a bounded `settleUntil(predicate, description)` poll that re-stabilises the chart until the observable effect of each scroll is seen before asserting:

- window tests → the captured window value changed;
- source tests → the `user-interaction` source was captured;
- recovery / financial tests → the `getData` call count incremented.

No assertion was loosened — the value/snapshot/warning checks are unchanged; they now run only once the async effect has actually landed. A genuine failure still surfaces (the poll is bounded and throws a descriptive timeout).

### Validation

- Reproduced the flake on the **pre-change** suite under oversubscribed parallel contention (~7% of runs failed, matching the CI signatures).
- The **hardened** suite passes **80/80** oversubscribed parallel runs (16-way on 14 cores; per-run test time 16–20 s, i.e. genuinely under load).
- `yarn nx format`, `build:types`, and `lint` all pass for `ag-charts-enterprise`.

### DIFF
--- a/packages/ag-charts-enterprise/src/features/data-source/dataSource.test.ts
+++ b/packages/ag-charts-enterprise/src/features/data-source/dataSource.test.ts
@@ -320,8 +320,8 @@ describe('DataSource', () => {
                 { x: 'seven', y: 50 },
             ]);
 
-            let windowStart;
-            let windowEnd;
+            let windowStart: unknown;
+            let windowEnd: unknown;
 
             await prepareChart(
                 {
@@ -359,8 +359,8 @@ describe('DataSource', () => {
                 { x: ['delta', 'seven'], y: 50 },
             ]);
 
-            let windowStart;
-            let windowEnd;
+            let windowStart: unknown;
+            let windowEnd: unknown;
             await prepareChart(
                 {
                     getData: (window) => {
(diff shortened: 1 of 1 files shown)


iMoses (approved):
LGTM

PR #7277, opened 2026-06-29
Split sync/api-state/themes/bar-series -test pages: move already-automated examples to -e2e

alantreadway:
https://ag-grid.atlassian.net/browse/AG-17752

Phase 2a of the `-test` → `-e2e` rationalisation: the mechanical residue.

After the Phase 2 disposition re-analysis on the post-1a/1b surface, 18 examples were found that are **already asserted by a handwritten Playwright spec** but never got migrated off their `-test` page. This PR drains the clean ones — a direct continuation of 1a/1b, **slug-retarget only, no new tests authored**.

Each page is split with `git mv` (history preserved, diff reads as renames): the already-automated examples move to a new `<page>-e2e` page; everything else stays on a slimmed `<page>-test`.

### Pages split (one commit each)

| Page → new `-e2e` | Moved (already-automated) | Residue stays on `-test` |
| --- | --- | --- |
| sync | single-series-sync, multi-series-sync, multi-series-horizontal-sync, multi-series-implicit-key-sync, mixed-series-sync, financial-charts-sync | secondary-axis-sync (manual) |
| api-state | legend-zoom-e2e → legend-zoom | initial-state, initial-state-zoom-category (automatable → later phase) |
| themes | css-variables | css-variables-dark-mode (automatable → later phase) |
| bar-series | animation-with-padding | animation, animation-grouped-category, animation-vertical (manual), data-per-series (automatable → later phase) |

Specs retargeted to the new slugs: `synchronised.spec.ts`, `state.spec.ts`, `css-variables.spec.ts`, `animation-initial-load.spec.ts`.

### Held back (not in this PR)

The two explicit 1a hold-backs remain held: **`accessibility-test`** (slug pending Oli confirmation) and **`security-test`** (one example, `strict-csp-icons`, still needs a new test before the page is fully automated). The automatable residue on api-state / themes / bar-series is left for the later spec-authoring phase.

### Notes

- `legend-zoom-e2e` drops its redundant `-e2e` suffix to `legend-zoom` (Oli #7).
- No slug-embedded snapshots needed renaming — every changed spec names its screenshots explicitly, and none of these pages opt into the slug-derived snapshot spec.

### Verification

`generate-examples`, `validate-examples`, `format:mdoc` and `lint` (0 errors) all pass locally. The full Playwright e2e suite — the real gate for the retargeted slug references — runs in CI on this PR.

PR #7278, opened 2026-06-29
Consolidate styler examples into stylers-e2e + add stylers.spec.ts

alantreadway:
https://ag-grid.atlassian.net/browse/AG-17752

Phase 2b of the `-test` → `-e2e` rationalisation: **styler consolidation** (Oli #8).

Stacked on #7277 (Phase 2a). The disposition re-analysis flagged ~23 `styler` / `styler-and-itemstyler` / `styler-highlight-state` examples scattered across 8 series `-test` pages, none of which had a behavioural spec. This consolidates them onto the existing `stylers-e2e` page and adds `stylers.spec.ts`.

### Consolidation (commit 1)

All 23 examples move into `stylers-e2e` via `git mv` (series-prefixed: `box-plot-styler`, `nightingale-styler-and-itemstyler`, …). The source pages are then:

| Source page | Outcome |
| --- | --- |
| box-plot-series-test | **deleted** (fully drained) |
| nightingale-series-test | **deleted** |
| radar-area-series-test | **deleted** |
| radial-bar-series-test | **deleted** |
| range-area-series-test | **deleted** |
| radar-line-series-test | slimmed → keeps `animation` |
| radial-column-series-test | slimmed → keeps `positive-negative-data` |
| range-bar-series-test | slimmed → keeps 10 animation examples |

Also removes a stray `box-plot-series-test/technical-review-plan.md` (an internal analysis doc, referenced by nothing) orphaned by that page's deletion.

### `stylers.spec.ts` (commit 2, revised by review-feedback commits below)

- **Render screenshot** per consolidated example — visual-regression baseline for the series-level `styler` colour-override path that `item-styler-test.spec.ts` never covered.
- **Branch-coverage test** for each of the 8 `-highlight-state` examples: drives an item highlight and a series highlight via the keyboard, then asserts both the series `styler` and the `itemStyler` were invoked with the expected `highlightState` for that phase, and never `undefined`/empty. Invocations are observed via a read-only `window.agE2E` tracking hook (see below), so coverage is enforced, not assumed.
- **Highlight visual screenshots** (item + series state) per highlight-state example.

### Addressing review feedback (Oli + Codex)

A second round addresses all of Oli's five points plus Codex's P2. Root cause was shared: the spec communicated styler invocations over a `console.log('[styler]', …)` channel and asserted that one `Set` of states *eventually* filled — ordering-blind, unable to tell a series `styler` from a marker/item `itemStyler`, and contradicting the codebase's sanctioned approach.

- **Observe callbacks via the `agE2E` hook, not console logs** (Oli #2). Each `-highlight-state` example now exposes `window.agE2E = { popStylerCalls() }` (precedent: `e2e/state.spec.ts` + `active-e2e/.../line-example`). The spec pulls and drains recorded calls instead of scraping logs.
- **Per-phase pop-and-assert** (Oli #3). The spec pops after the item-highlight phase and again after the series-highlight phase, asserting each phase's expected states — so a broken `highlightSeries()` can no longer be masked by `highlightItem()`.
- **`kind` discriminator** (Oli #5 + Codex P2). Recorded calls carry `kind: 'styler' | 'itemStyler'`, so series-styler coverage can't be satisfied by item-styler invocations (or vice-versa). The 5 examples that previously branched only in their series `styler` (`box-plot`, `nightingale`, `radial-bar`, `radial-column`, `range-bar`) gained a `highlightState`-branching `itemStyler`, completing item-styler highlight coverage. The default/`none` branch returns `{}`, so only the highlight visuals change.
- **Per-example `describe` structure** (Oli #1) replaces the three duplicated global `for`-loops, each with a shared `beforeEach` nav.
- **Pointer/touch hover coverage** (Oli #4): each highlight-state example now has a `mousemove over node1` test. The `node1` coordinates were **measured** against each rendered example in a browser — every hover verified to drive the item styler into `highlighted-item` via the `agE2E` hook — not guessed, and converted to page coordinates at runtime via `canvasToPageTransformer`. The new `-node1` screenshot baselines are generated by CI on `gha/snapshots-<branch>`.

### Prevent recurrence — new e2e rule

Adds `website-e2e-testing.md` (via `.rulesync/`, glob-scoped to `ag-charts-website/e2e/**/*.spec.ts`) capturing the lessons: observe callbacks via the `agE2E` hook (not console logs); pop-and-assert per interaction phase; tag a shared channel with a discriminator; don't guess pointer coords; prefer table-driven per-example `describe`; and let CI regenerate screenshot baselines via the `gha/snapshots-<branch>` convention rather than running `-u` locally. Kept as a Charts-specific in-repo rule, separate from the generic vitest guidance.

### Tooling fix — `validate-examples` nx cache

The `validate-examples` input glob was anchored with `{projectRoot}`, which nx matches only against files owned by the `ag-charts-website` project. Example sources are owned by per-example synthetic projects (task-autogen plugin), so the glob matched nothing and **editing an example never invalidated the cache** — that is why a type error in these examples passed locally on a stale cache hit and only surfaced in CI's Validate Examples job. Re-anchored with `{workspaceRoot}` (as the sibling generate target already does), mirroring the docs/gallery globs `tsconfig.examples.json` compiles. Confirmed by A/B: an example edit now forces a re-run and a revert restores the cache hit, so this class of error is caught locally going forward.

### Baseline regeneration

Exactly 10 baselines change — the 5 enhanced examples × {item, series} highlight visuals — as a direct result of the new `itemStyler` branches. The 3 already-itemStyler examples and all render baselines are byte-identical (the console→hook swap and `highlightState ?? ''` coercion don't affect rendering). Regenerated baselines were taken from CI's `gha/snapshots-<branch>` output; no local `--update`.

### Verification

`generate-examples`, `validate-examples`, `format:mdoc`, and `lint` all pass. (A `highlightState: string` vs `string | undefined` type error in the examples initially slipped past a local `validate-examples` run and was caught by CI — the root cause was the nx cache bug fixed above, now resolved.) The full **32-shard CI E2E matrix is the gate and is green** on the head commit — local Docker was unavailable in this environment, so CI is the authoritative run for both the branch-coverage assertions and the regenerated screenshot baselines.


olegat:
Good start! Here are possible improvements:

## 1. Unnecessary duplication

The PR duplicates the `-styler` examples into `-styler-highlight-state` example,
but the duplicates do not add anything beyond just console-logging the callback
values.

This duplication looks unnecessary, these

```typescript
    for (const example of STYLER_EXAMPLES) {
        test(`renders ${example}`, async ({ page }) => {
            const { url } = toExamplePageUrl('stylers-e2e', example, 'vanilla');
            await gotoExample(page, ...);
            ...
        });
    }

    for (const example of HIGHLIGHT_STATE_EXAMPLES) {
        test(`exercises every highlightState branch for ${example}`, async ({ page }) => {
            const { url } = toExamplePageUrl('stylers-e2e', example, 'vanilla');
            await gotoExample(page, url);
            ...
        });

        test(`highlight visuals for ${example}`, async ({ page }) => {
            const { url } = toExamplePageUrl('stylers-e2e', example, 'vanilla');
            await gotoExample(page, url);
            ...
        });
    }
```

could simply be:
```typescript
    for (const example of EXAMPLES) {
        beforeEach(async ({ page }) => {
            const { url } = toExamplePageUrl('stylers-e2e', example, 'vanilla');
            await gotoExample(page, url);
        });
        test(`renders ${example}`, async ({ page }) => {
            ...
        });

        test(`exercises every highlightState branch for ${example}`, async ({ page }) => {
            ...
        });

        test(`highlight visuals for ${example}`, async ({ page }) => {
            ...
        });
    }
```

## 2. Use e2e "mocks" in the examples

The stylers.spec.ts and example communicate "highlight state" with each other
through console logs. This usually fine for simple tests, but we've also
recently had great success in adding hooks/tracking in the examples which the
spec.ts file can call explicitly:

They work much like Jest Mocks which keep tracking the input args and output
return. Example:
* https://github.com/ag-grid/ag-charts/blob/3c30b42c3c8fe4ac2107d14ed55ea457aa842a13/packages/ag-charts-website/e2e/state.spec.ts#L70
* https://github.com/ag-grid/ag-charts/blob/3c30b42c3c8fe4ac2107d14ed55ea457aa842a13/packages/ag-charts-website/src/content/docs/active-e2e/_examples/line-example/main.ts#L139
* https://github.com/ag-grid/ag-charts/blob/3c30b42c3c8fe4ac2107d14ed55ea457aa842a13/packages/ag-charts-website/src/content/docs/active-e2e/_examples/line-example/main.ts#L60

The main advantage is that it gives the test more fine-grain control on when to
retrieve values and without being too much at the mercy when code is run (or not
run) through Playwright.

## 3. Callback-ordering vulnerability

The `highlightState` tests check that all REQUIRED_HIGHLIGHT_STATES are called,
but it does not check when. For example, if a bug is introduced where
`highlightSeries()` no longer calls the callbacks correctly, then the tests
could still pass if the `highlightItem()` fills the Set completely.

This could be improved by retrieving & clearing the callback in/outs each
time that they are asserted:

```typescript
test(`exercises every highlightState branch for ${example}`, async ({ page }) => {
  await highlightItem(page);
  expect(popStylerCallbacks()).toEqual(EXPECTED_VALUE1)
  await highlightSeries(page);
  expect(popStylerCallbacks()).toEqual(EXPECTED_VALUE2)
}
```

## 4. No mouse/touch hovering tests

The `highlightItem()` / `highlightSeries()` tests keyboard navigation but not
mouse/touch hovering.

It's kind of difficult to get AI to generate good tests for the mouse, because
at the tests need to know the X/Y coords of the mouse, and these obviously
differ for each chart-type. So at some point, someone will need to manually go
through all the examples and decide on X/Y values that test the
styler/itemStyler code.

However, it might be a good idea to get Claude to facilitate this manual
procedure by generating placeholders:

```typescript
const EXAMPLES = [
  {
     name: 'box-plot-styler',
     node1: { x: NaN, y: NaN }, // PLACEHOLDER
  },
  {
     name: 'nightingale-styler',
     node1: { x: NaN, y: NaN }, // PLACEHOLDER
  },
  // and so on...
];

test.describe('stylers', () => {
    for (const example of EXAMPLES) {
        beforeEach(async ({ page }) => {
            const { url } = toExamplePageUrl('stylers-e2e', example.name, 'vanilla');
            await gotoExample(page, url);
        });
        test.skip(`mousemove on node1 for ${example}`, async ({ page }) => {
            await page.mouse.move(example.node1.x, example.node1.y);
            await expect(page).toHaveScreenshot(`${example}-node1.png`);
        });
    }
});
```

## 5. No interaction tests on `itemStyler` examples

The stylers.spec.ts file only tests keyboard interactions on examples ending in
`-styler-highlight-state`, which only include `styler` callbacks.

The `-styler-and-itemstyler` tests only check the initial frame.

There are some exceptions (e.g. range-area-styler-highlight-state also includes
a `marker.itemStyler` in the chart options), but most of the itemStyler
interactions are untested.

alantreadway:
Thanks Oli — all five addressed:

**2 (mocks not logs):** examples now expose `window.agE2E.popStylerCalls()`; the spec pulls via `page.evaluate` instead of scraping console. (Per your `state.spec.ts` / `line-example` precedent.)

**3 (ordering):** pop-and-assert per phase — init/`none`, then `highlightItem`, then `highlightSeries`, each asserted against only its own invocations.

**5 + Codex P2 (itemStyler / conflation):** records carry a `kind: 'styler' | 'itemStyler'` discriminator and coverage is asserted per kind. Added highlight-driven `itemStyler`s to the 5 highlight-state examples that lacked one (box-plot, nightingale, radial-bar, radial-column, range-bar).

**4 (mouse/touch):** scaffolded as `test.skip` with `node1: { x: NaN, y: NaN }` placeholders — coords need a human to measure per chart type.

**1 (structure):** collapsed into one render loop + a per-example `describe` with shared `beforeEach`.

Bonus: tracked down why the example type error slipped past local `validate-examples` — the nx cache key was anchored with `{projectRoot}`, which doesn't see the synthetic per-example projects, so example edits never invalidated it. Fixed by switching to `{workspaceRoot}`.

olegat (approved):
Good improvements. LGTM as a starting point.

The duplication of example like `'box-plot-styler'` & `'box-plot-styler-highlight-state'` still looks unnecessary but that's not a big deal.

PR #7281, opened 2026-06-29
Raise bundle size and tree-shake limits 5% above measured

iMoses:
Re-measured the actual bundle sizes on `latest` and raised every limit to measured x 1.05, giving headroom for incremental code growth without prematurely failing CI.

## What changed
- `packages/ag-charts-community/.size-limit.js` — brotli src/dist limits
- `packages/ag-charts-enterprise/.size-limit.js` — brotli src/dist limits
- `tests/bundle-treeshake-tests/scenarios.mjs` — gzip limits (max across esbuild/vite/webpack)

## How values were derived
- size-limit: ran `size-limit` per package, took the reported brotli size, multiplied by 1.05, rounded up to the nearest whole kB.
- tree-shake: ran the bundle tree-shake suite, took the max gzip size across the three bundlers per scenario, multiplied by 1.05, rounded up to the nearest 1000 bytes.

All three suites pass with the new limits.


alantreadway (approved):
LGTM

PR #7290, opened 2026-06-29
AG-17263 [CLAUDE] Enrich Series.pickNodes() to expose the picked scene-node

olegat:

Hit-testing already resolves the exact scene-node under the pointer, but pickNodes() discarded it and returned only node-data. Because the org-chart card and expander share one datum, the click handler cannot tell them apart.

This enriches the pick result so the scene-node is carried through, without changing any behaviour. Phase 2 will use Node.tag on the exposed node to make the expander the sole expand/collapse target.

- Add `target: Node<unknown>` to `SeriesNodePickMatch`; the field is as accurate as possible, falling back to the series `contentGroup` for hit-test modes (e.g. "closest") that match on datum geometry and cannot resolve the leaf efficiently.
- Replace `PickResult.datums`/`distance` with `picks: SeriesNodePickMatch[]`.
- `pickNodesExactShape()` now returns matches carrying the hit node.
- Populate `target` with the real leaf where it is already in hand: base exact-shape, `pickNodeNearestDistantObject`, cartesian data exact-shape, and the pyramid/map-shape/map-line/flow-proportion node-selection picks. Gauges, sunburst and radial series inherit this via the helpers they delegate to.
- Update the only `PickResult` consumer (the seriesAreaManager aggregator) and the treemap (`path.length`) and dataModel (`datumIndex`) exact-shape sorts to the new shape.

### DIFF
--- a/packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts
+++ b/packages/ag-charts-community/src/chart/series/cartesian/cartesianSeries.ts
@@ -969,7 +969,7 @@ export abstract class CartesianSeries<TTypes extends CartesianSeriesTypes> exten
         return result;
     }
 
-    protected pickNodeDataExactShape(point: Point): SeriesNodeDatum[] | undefined {
+    protected pickNodeDataExactShape(point: Point): SeriesNodePickMatch[] | undefined {
         const { x, y } = point;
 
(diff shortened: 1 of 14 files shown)


alantreadway (approved):
LGTM

PR #7294, opened 2026-06-30
AG-17355 remove Bryntum Gantt link from docs nav

AG-DavidG:
Removes the "Gantt Charts" nav group (single external "See Bryntum Gantt" campaign link) added in #7194.

- File: `packages/ag-charts-website/src/content/docs-nav/nav.json`
- The group contained only an outbound link to `ag-grid.com/campaigns/bryntum-gantt` — no chart-type page.

Reverts the nav addition only.


alantreadway (approved):
LGTM

PR #7298, opened 2026-06-30
Type SeriesModuleDefinition.themeTemplate and replace @ts-expect-error theme suppressions

iMoses:
## Summary

Tightens the typing of series theme templates and removes the `@ts-expect-error`-based escape hatch for undocumented theme defaults.

- Add a typed `themeTemplate?: SeriesThemeTemplate<TOptions>` field to `SeriesModuleDefinition`, so series theme templates are validated against `ExtensibleTheme<S>` for the module's series type.
- Introduce `undocumentedThemeOptions()` in `ag-charts-core` — an identity helper whose return type erases its keys. Spreading it into a theme template passes undocumented defaults through at runtime while keeping every sibling and nested property of the literal validated. This replaces scattered `// @ts-expect-error undocumented option` comments, which suppressed only the first excess-property error and masked validation of everything following them.
- Promote the linear-gauge `scale` from `AgLinearGaugePreset` to `AgLinearGaugeThemeableOptions` (with the matching `gaugePreset.ts` options-def move), making it themeable, and update `themeMap.ts` gauge entries to use the themeable option types.

## Why

The previously untyped theme templates let invalid theme keys slip through, and the `@ts-expect-error` pattern was actively dangerous: it disabled type checking for an entire trailing block of theme properties, not just the intended undocumented key. Typing the field surfaced exactly the places relying on that suppression (e.g. `collisionAvoidance`, `undocumentedGroupFills/Strokes`), which now route through the explicit helper.

## Behaviour

No runtime behaviour change. Every default moved into an `undocumentedThemeOptions({ ... })` block retains its original nesting, and theme merge is key-based.

## Testing

- `yarn nx format` — no changes
- `yarn nx build:types` — core, types, community, enterprise all pass
- `yarn nx lint` — all four packages clean
- `yarn nx test ag-charts-community` — 3730 passed, 0 snapshot diffs
- `yarn nx test ag-charts-enterprise` — 2517 passed, 0 snapshot diffs


alantreadway (approved):
LGTM

PR #7304, opened 2026-06-30
AG-17714 Cross-series label obstacles & collideWith seriesItem opt-in (cartesian MVP)

iMoses:
## Summary

Allows cartesian series labels to route around the rendered geometry of *other* series (bars, histogram bins), not just their own series' markers and labels. Cross-series geometry avoidance is gated behind a new `collideWith.seriesItem` category that defaults **off** (opt-in), so default render output is unchanged.

## Changes

- **New `seriesItem` obstacle category, opt-in.** `LabelCollisionAvoidance.resolveCollideWith()` now defaults `seriesItem` off (`enabled === true`) while `marker`/`label` stay on by default. The category only routes labels around cross-series geometry when explicitly enabled.
- **Bar & histogram contribute obstacles.** Both expose their node geometry via `getLabelObstacles()` using the new shared `rectLabelObstacles()` helper, which skips phantom (stacking/feather) and zero-area rects and returns `undefined` when there's nothing to contribute.
- **`collideWith` threaded through per-render series.** `applyLabelAvoidance()` now also stamps the resolved per-category `collideWith`, and bubble/map-line/map-marker pass it so the `seriesItem` opt-in is honoured. (Line/area already bake it at build time.)
- **Skip work when nothing places labels.** `LabelManager` short-circuits obstacle gathering and placement when no series places labels.

## Why it's a no-op by default

Bar and histogram are the only series that contribute `seriesItem` obstacles, and both are introduced here. Before this change there were no `seriesItem` obstacles in existence, so flipping the default to off combined with the new contributors produces no observable change unless `collideWith.seriesItem` is explicitly enabled.

## Test plan

- `rectLabelObstacles` unit tests (mapping, phantom/zero-area skipping, empty -> undefined).
- `resolveCollideWith` tests (disabled -> undefined; marker/label default on, seriesItem default off; explicit opt-in).
- Bar & histogram `getLabelObstacles()` tests asserting a rect obstacle per node.
- Core `placeLabels` tests: routes around `seriesItem` obstacles only when the category is enabled; disabled-category obstacles do not perturb other categories.
- Visual snapshot tests (bar + line): line labels route around bars when `seriesItems` enabled; left over the bars by default.
- `nx format`, `build:types`, `lint`, and the affected `test` suites pass for core/community/enterprise.

## Out of scope

- Non-rect obstacle geometry (line/area paths, pie sectors, hierarchical tiles).
- The full `allowCollisionWith*` / `collisionPriority` model (deferred to v2).

Fix #AG-17714


alantreadway (approved):
LGTM

PR #7307, opened 2026-07-01
AG-17664 Define collisionAvoidance option types and map legacy placement

iMoses:
## Summary

Establishes a single, canonical type definition for the (still undocumented) `collisionAvoidance` label option and wires the legacy `label.placement` into the collision-avoidance model for bubble/scatter, without changing any rendered output.

- **Canonical types in `ag-charts-types`** — new `chart/collisionAvoidanceOptions.ts` defines `AgChartLabelCollisionAvoidanceOptions` and its parts (`AgChartLabelCollisionStrategy`, `AgChartLabelCollideWithOptions`, `AgChartLabelCollideWithCategoryOptions`, and an 8-value `AgChartLabelCollisionPlacement`). The community label classes now `implement` these interfaces instead of duplicating them locally, so the shape has one source of truth.
- **Legacy placement mapping** — `applyLabelPlacements` (new, in `ag-charts-core`) stamps the ordered fallback placements onto label data per-render, mirroring the existing `applyLabelAvoidance`. Bubble/scatter's `getLabelData` seeds the reposition model from the legacy `label.placement`, so a configured `reposition` strategy is now honoured while the default resolves to the same single placement as before.

## Scope

Per the ticket author's clarification, this is a **partial** against AG-17664 — deliberately scoped:

- **Not linked to the public API.** The interfaces exist in `ag-charts-types` but are **not** attached to `AgChartLabelOptions` or any series label interface. The option stays undocumented (validator remains `undocumented(...)`, theme defaults keep their `@ts-expect-error`).
- **No deprecations.** No legacy-to-new deprecation or conflict warnings.

Deferred to a follow-up (original AC3/AC4): promoting `collisionAvoidance` to a documented public option and adding legacy-to-new deprecation/conflict warnings.

Mapping is scoped to bubble/scatter (cartesian). Map-marker/map-line (topology/enterprise) are intentionally left unchanged.

## Test plan

- Added unit tests for `LabelCollisionAvoidance.placements()` resolution order (strategy placements > legacy seed > fallback).
- Full community and enterprise suites pass with **all image snapshots unchanged** — render output is pixel-identical, which is the acceptance bar for the mapping.
- `nx format`, `build:types` (community + enterprise), and `lint` (types/core/community) are green.

Fix #AG-17664


alantreadway (approved):
LGTM

PR #7314, opened 2026-07-01
AG-17263 [CLAUDE] Restrict org-chart expand/collapse to the expander control

olegat:

Clicking a node's card toggled its collapse state, the same as clicking the expander. Only the expander should do this. Building on the enriched pick result, the clicked scene-node is now threaded to the click handler so it can tell the card and expander apart via Node.tag.

- Tag the expander's scene-nodes (pill, count, chevron) and the card surface with a new `OrganizationNodeTag` selector.
- Carry the hit scene-node through to the click event: `PickedNodes.target` and `SeriesAreaClickEvent.target`, sourced from the topmost match so an expander that overlaps the card bottom still resolves to the expander.
- Gate the toggle behind `shouldToggleCollapseOnClick(target)` in the network click handler. The base allows any click; OrganizationSeries requires the expander tag. A missing target (keyboard Enter/Space on a focused node) still toggles, preserving keyboard accessibility.
- Remove the obsolete commented-out geometric expander hit-test and the now-dead `expanderContainsPoint` helper.

### DIFF
--- a/packages/ag-charts-community/src/chart/series/pickManager.ts
+++ b/packages/ag-charts-community/src/chart/series/pickManager.ts
@@ -1,6 +1,7 @@
 import { objectsEqual } from 'ag-charts-core';
 import type { AgActiveItemState } from 'ag-charts-types';
 
+import type { Node } from '../../scene/node';
 import type { ActiveManager } from '../interaction/activeManager';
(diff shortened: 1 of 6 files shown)


alantreadway (approved):
LGTM

PR #7322, opened 2026-07-02
Add frame-trajectory animation test harness and spike cases

alantreadway:
https://ag-grid.atlassian.net/browse/AG-17755

Spike proving property-style animation testing: a `spyOnAnimationFrames()` harness steps the real `AnimationBatch` frame-by-frame (~30 frames) so tests can assert structural invariants over interpolated scene-node geometry, instead of pixel-comparing frozen intermediate ratios.

- `createSceneGeometrySampler` walks the whole scene (series content/label groups, axis/grid groups) into a map keyed by stable, human-readable node paths (`series[0]/rect[B]`, `axis[left]/text[100]`).
- `expectSceneTrajectory` applies a declarative spec: named nodes assert monotonic/progression/bounded behaviour per property; every unnamed node defaults to "must not move on any frame", so axis wobble, label flicker and sibling disturbance fail by default. Spec keys support `*` globs; entries matching no node fail as a typo guard.
- Failures report all violations together, each with a per-frame sparkline and value table, so the culprit cluster (e.g. one wobbling axis moving many labels) is visible at a glance.
- Five spike cases across bar (data update, add/remove, scale-affecting), pie (initial-load sector sweep) and line (path-bbox — the approach's limit case, as line/area expose no per-vertex geometry).
- Replaces the COLUMN_TIME update-animation 25/50/75% image snapshots with endpoint-only (0%/100%) snapshots, deleting 3 PNGs.

Wider rollout across the remaining per-ratio snapshot blocks is a follow-up.

Specs contain no `any` allowances: every animated node carries a directional expectation (bars/ticks/labels shift left on category add and right on remove, scale-affecting reflows assert one-way gutter/plot shifts, tick swaps assert fade-in/out, pie angles sweep monotonically clockwise from a fixed 12 o'clock anchor).

Trajectories are also time-bounded by animation phase: `captureAnimationFrames` records which `AnimationBatch` phases (remove/update/add/initial/trailing) each inter-frame interval passed through, and spec expectations can be scoped as `{ during: 'add', expect: [...] }` — the property must stay anchored outside its phase window, and violations report the phase the movement actually happened in. `axisReflowSpec()` provides general axis expectations (directional moves only during `update`, endpoint-bounded fades only during remove/update/add, opt-in gutter/plot-edge tracking) for tests that exercise axes incidentally.

**Known limitations / follow-ups** (from self-review):
- The deleted intermediate snapshots exercised a *time* x-axis; CASE 1 asserts the same update shape on a *category* axis, so time-scale interpolation at intermediate frames is no longer pixel-covered.
- The rAF mock does not intercept `cancelAnimationFrame`, so capture flows must drain frames between updates (documented in the harness); phase attribution is per-interval and coarsest at phase boundaries (also documented).
- If the approach lands, the harness (~600 lines) should move out of `chart/test/utils.ts` into a dedicated module before wider rollout, and repeated key-prefix filters in tests could get a small shared helper.


olegat (approved):
LGTM.

Discussed on Slack: Conceptually this spike is a step in the right direction. The only unknown right now whether the new scene-inspection utilities in `test/utils.ts` will be high-maintenance, or whether we'll need some formal way to serialise the scene-graph state.

PR #7327, opened 2026-07-03
AG-17263 [CLAUDE] Add E2E test for org-chart expander clicks

olegat:

The expander-only expand/collapse behaviour can't be covered by a JSDOM unit test: JSDOM's canvas returns false for isPointInPath, so the rounded expander pill can't be hit-tested. A Playwright E2E test drives a real browser instead.

- Add `e2e/org-chart-expander.spec.ts` against the `simple-org-chart` example. It clicks the CEO node's card body (tree stays fully expanded) then its expander pill (tree collapses to just the CEO card), asserting each state via screenshot. The collapsed baseline is visibly distinct, so the test cannot pass vacuously.
- Add the three chromium baseline screenshots.


alantreadway (approved):
LGTM

PR #7346, opened 2026-07-06
AG-17706 Add API tab for Caption context-menu item

AG-DavidG:
https://ag-grid.atlassian.net/browse/AG-17706

Fix #AG-17706


alantreadway (approved):
LGTM

PR #7347, opened 2026-07-06
AG-17780 Harden label placement: test coverage + array-placement guard

iMoses:
https://ag-grid.atlassian.net/browse/AG-17780

Fix #AG-17780


alantreadway (approved):
LGTM

PR #7350, opened 2026-07-06
ci: fix dead staleness check in snapshot-notify (empty run_started_at)

alantreadway:
## Root cause

`${{ github.run_started_at }}` is **not a valid workflow context property** — GitHub Actions expressions resolve unknown context properties to an empty string, silently. So `snapshot-notify`'s `run_started_at` input has been empty on every run since it was added, making the staleness comparison `[[ "${snapshot_commit_time}" < "" ]]` always false: whenever a `gha/snapshots-<run-id>` branch exists, the action reports `updated=true`.

On attempt 1 this is coincidentally correct (the snapshot really was pushed during the run). But it makes **"Re-run failed jobs" after a snapshot failure structurally unable to go green**: the snapshot branch from attempt 1 survives, `updated=true` again, and the `report` job re-fails even when every test passes on the rerun. Observed on [run 28787208508](https://github.com/ag-grid/ag-charts/actions/runs/28787208508) — attempt 2 had all test jobs green, `report` red on `Snapshot branch has new updates (commit: 2026-07-06T11:18:29Z, run started: )` (note the empty timestamp).

## Fix

- Resolve the timestamp inside the action via `gh api repos/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID --jq .run_started_at`. The API's `run_started_at` **resets per attempt**, which is exactly the semantics the staleness check wants: a snapshot pushed during the current attempt is new (fail + notify); one left over from a previous attempt is stale (pass).
- Fail loud (`::error` + exit 1) if the timestamp can't be resolved, so this can't silently regress to always-`updated=true` again.
- Remove the `run_started_at` input entirely (single call site, in `ci.yml`, updated in the same PR).

This also removes a failure class from the mainline CI-failure handler rollout: the handler's rerun of a transient-looking snapshot blip could previously never clear `report`.

### DIFF
--- a/.github/actions/snapshot-notify/action.yml
+++ b/.github/actions/snapshot-notify/action.yml
@@ -4,10 +4,6 @@ inputs:
     branch:
         description: 'Snapshot branch name to use.'
         required: true
-    run_started_at:
-        description: 'DEPRECATED — ignored. The action resolves the current attempt start time from the runs API itself.'
-        required: false
-        default: ''
     github_token:
         description: 'GitHub token to use for PR creation.'
         required: true
--- a/.github/actions/snapshot-notify/action.yml
+++ b/.github/actions/snapshot-notify/action.yml
@@ -5,8 +5,9 @@ inputs:
         description: 'Snapshot branch name to use.'
         required: true
     run_started_at:
-        description: 'ISO timestamp when the workflow run started (github.run_started_at). Used to detect stale snapshot branches from previous attempts.'
-        required: true
+        description: 'DEPRECATED — ignored. The action resolves the current attempt start time from the runs API itself.'
+        required: false
+        default: ''
     github_token:
         description: 'GitHub token to use for PR creation.'
         required: true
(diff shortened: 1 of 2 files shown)


.github/actions/snapshot-notify/action.yml
```diff
@@ -25,11 +22,25 @@ runs:
...
 
               branch=${{ inputs.branch }}
-              run_started_at="${{ inputs.run_started_at }}"
+              # `github.run_started_at` is NOT a valid workflow context property — it
+              # always renders empty, which made the staleness check below dead code
+              # (any existing snapshot branch → updated=true), so a `rerun --failed`
+              # after a snapshot failure could never turn the run green. Resolve the
+              # timestamp from the runs API instead; its run_started_at resets per
+              # attempt, which is exactly the wanted semantics: a snapshot pushed
+              # during THIS attempt is new (fail/notify), one left over from a
```
iMoses: This comment really is useless. Talking about why the previous hallucinations were wrong has no value to people reading this in the future.

iMoses (approved):
LGTM, one nit

.github/actions/snapshot-notify/action.yml
alantreadway: Trimmed to a one-liner stating just the current-attempt semantics.

.github/actions/snapshot-notify/action.yml
```diff
@@ -25,11 +26,25 @@ runs:
...
```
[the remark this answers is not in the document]
alantreadway: Switched to the attempt-scoped endpoint (`/actions/runs/$GITHUB_RUN_ID/attempts/$GITHUB_RUN_ATTEMPT`). Verified against run 28787208508: attempt 1 → 11:09:29Z, attempt 2 → 11:36:08Z.

.github/actions/snapshot-notify/action.yml
alantreadway: Fixed

PR #7358, opened 2026-07-07
Bar animation: trajectory coverage for -test page actions, page retired

alantreadway:
Tracked under [AG-17755](https://ag-grid.atlassian.net/browse/AG-17755). Supersedes #7329 — its harness commits are folded in here (minus the line/area suite additions, which move to the follow-up line and area PRs) so this PR stands alone against `latest`.

## Harness

- Trajectory-harness extensions from #7329: path interior geometry (`top@<i>` stations, `subpaths`, `cutout`), cross-node `frameInvariants`, clip-window sampling, `degenerate`/`settlesAt` expectations, `expectNoAnimation`, plus the axis animation suite (`cartesianAxisAnimation.test.ts`).
- `captureUpdate`: the settle → before-sample → action → capture → after-sample flow with endpoint integrity checks, shared by all action CASEs.
- The path-geometry readers are also consumed by the stacked line/area PRs that follow this one.
- Reviewer note: the default `monotonicTol` for `expectSceneTrajectory` directional checks is loosened from `1e-6` to `0.5` so the crisp-pixel snap on settling frames does not read as a reversal. This applies to the pre-existing trajectory CASEs too — a sub-half-pixel reversal is now tolerated suite-wide.
- `lineSeries.test.ts` changes are limited to adapting spike CASE 3 to the station sampling the new harness performs; the new line/area CASEs arrive in their own PRs.

## Bar `-test` page coverage (`barSeries.test.ts`, "animation -test page actions")

Every control on the `bar-series-test` examples maps to a trajectory case:

- **Randomise** — height-only monotonic tween under a pinned domain; the scale-affecting variant is CASE 5.
- **Remove Data** — leaving bars collapse during the remove phase, survivors widen, bottom axis reflows right. Enter/exit of a single bar is CASE 2.
- **Remove/Add Series** — survivors widen/narrow monotonically; an entering series spawns at final geometry and fades in from opacity 0. Stacked remove + re-add contiguity is CASE 10 (CRT-950).
- **Switch Direction / Grouped ↔ Stacked** — series re-creation replays the initial-load reveal; end-state tiling/partition invariants assert the stacked and grouped layouts.
- **Move legend / Change Theme** — snap contracts (`expectNoAnimation`) in standalone and integrated modes.
- **Integrated mode** — initial-load reveal, `resetAnimations()` replaying the reveal on a data update, and the grouped-category axis example's initial load.
- **Grouped-category Remove Data** — pins observed behaviour: the reflow snaps (no tween runs) and afterwards every leaf label sits on a band centre and every bar inside a band.

Each expectation was negative-tested (deliberately wrong direction/target fails naming the actual behaviour).

## Page removal

`bar-series-test` (three animation examples + the page) is deleted — its manual checklist is now automated. `bar-series-e2e/animation-with-padding` is retained (consumed by the Playwright `animation-initial-load` spec).

## Product findings (asserted around, not fixed)

1. A data update on a grouped-category axis does not animate — the whole reflow snaps before the first frame (accepted as intended behaviour).
2. The removed group's separator line/grid nodes linger in the scene with zero width — ticketed as [AG-17795](https://ag-grid.atlassian.net/browse/AG-17795).

## Verification

Full `ag-charts-community` + `ag-charts-enterprise` suites, `build:types`, lint, format, `generate-examples`/`validate-examples` all green.


alantreadway:
Thanks — walked each finding against the scene serialiser. Key fact: `Path.toSVG` (`extendedPath2D.ts`) emits **only** `M`/`L`/`C`/`Z`, space-joined, from a four-value `Command` enum (Move/Line/Curve/ClosePath). That determines most of these.

**[P1] numeric-arg validation / comma-separated / compact syntax** — the serialiser never emits commas or compact syntax, so that specific input can't occur. But the underlying point (a bare `Number()` accepting `NaN` and surfacing as an opaque non-finite failure downstream) is fair as a diagnostic gap. Fixed in b5fe015ccf: `num()` now validates each coordinate and throws with the offending token + path, matching the existing fail-loud handling of unsupported commands. Negative-tested end-to-end (injected a non-finite coord into `toSVG`; the flattener throws `non-finite coordinate at token N in '…'`).

**[P2] Q/A/H/V hard-fail** — the scene graph has no such commands, so this input is unreachable. The `throw` is a deliberate tripwire: if a future scene-graph change emits a new command, we want it to fail loudly and gain explicit support rather than silently corrupt geometry. Working as intended.

**[P2] station crossing on vertical/degenerate segments** — the `min`-y selection prefers the top edge, so baseline/close edges don't win; a vertical segment is only included when it sits exactly at a station-x; and stations 0 and 4 coincide with the x-extremes by construction. No misattribution in practice.

**[P2] `expectNoAnimation` doesn't check phases** — a *skipped* batch still advances through its phases doing no work, so asserting "no active phase" would false-fail on legitimate snaps (this rationale is in the helper's comment). Full-scene constancy from frame 0 is the correct snap signal, and the snap call-sites carry anti-vacuous guards proving the change actually landed. No change.

Net: one diagnostic hardening applied; the other three rest on inputs the serialiser cannot produce or on deliberate fail-loud design. CI is green.

iMoses (approved):
LGTM

PR #7359, opened 2026-07-07
AG-17798 Resolve generic type parameters in API reference code blocks

iMoses:
https://ag-grid.atlassian.net/browse/AG-17798

Fix #AG-17798


alantreadway (approved):
LGTM

PR #7363, opened 2026-07-08
AG-17637 Fix DevRegression where dragging on axis title does nothing

olegat:

Axes Event Listeners should be using `currentXY` (relative to the `role="region"` proxy-elem that's listened to) instead of `offsetXY` (relative to target, which is the deepest element that responds to pointer-events).

With the new AxisWidgets refactor, the a11y axis title is now a descendant of the axis `role="region"` proxy (the once that listens for wheel/mouse events on an axis). This can meddle with the `offsetXY` values when entering/leaving the a11y axis title element. Whereas the `currentXY` values are always relative to the element that's being listened to.


alantreadway (approved):
LGTM

PR #7364, opened 2026-07-08
AG-17814 Unify label text-fit surface across all series

iMoses:
https://ag-grid.atlassian.net/browse/AG-17814

Fix #AG-17814


alantreadway (approved):
LGTM

PR #7366, opened 2026-07-09
AG-17766 Add ag-charts-demos package for demo app examples

alantreadway:
https://ag-grid.atlassian.net/browse/AG-17766

Fix #AG-17766


iMoses (approved):
LGTM

PR #7369, opened 2026-07-09
Make setup-nx free-disk-space conditional and parallelise the delete

alantreadway:
Adds a `free_disk_space` input (default `true`) to the `setup-nx` composite action so jobs that don't exhaust runner disk can opt out of the cleanup, following the ag-grid precedent (#14371). The delete now runs the disjoint trees concurrently (best-effort), drops the redundant `$AGENT_TOOLSDIRECTORY` entry (it resolves to `/opt/hostedtoolcache`), and is bracketed with `df -h /` to log starting headroom and the amount reclaimed.

Opts out the `lint` and `validate` jobs — read-only checks with no build artifacts or Docker image loads. Further opt-outs will be validated by flipping a candidate job to `free_disk_space: false` and confirming CI stays green (the `df` output helps spot jobs that start with ample space); the default stays protective so no other job silently loses the cleanup.

First step toward reconciling the duplicated `setup-nx` action across ag-grid/ag-charts/ag-studio into `ag-shared`; landing the fix here first so the shared version starts from the corrected code.

### DIFF
--- a/.github/actions/setup-nx/action.yml
+++ b/.github/actions/setup-nx/action.yml
@@ -33,6 +33,13 @@ inputs:
             - 'rw'
             - 'ro'
             - 'off'
+    free_disk_space:
+        description: 'Delete large preinstalled toolchains to reclaim root-partition space (avoids runner ENOSPC). Set false on jobs that do not exhaust disk to skip the slow delete.'
+        required: false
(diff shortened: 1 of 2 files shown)


iMoses (approved):
LGTM

PR #7370, opened 2026-07-09
AG-17814 Hoist label fit resolution out of per-datum loops

iMoses:
https://ag-grid.atlassian.net/browse/AG-17814

Fix #AG-17814


alantreadway (approved):
LGTM

PR #7377, opened 2026-07-10
AG-6439 Add inside label placement centred and fit to markers

iMoses:
https://ag-grid.atlassian.net/browse/AG-6439

Fix #AG-6439


iMoses:
Addressed the Codex review in 04f5a03 (regression tests included):

- **Own-marker collision** — an `inside` candidate now skips only its own marker obstacle (matched by centre), so it still avoids neighbouring markers. (`labelPlacement.ts`)
- **Fit per placement** — the marker fit is applied only when `inside` is the *sole* placement; a mixed fallback list keeps full-size text so directional candidates aren't marker-constrained. (`lineSeries.ts`, `areaSeries.ts`)
- **Disabled markers** — area now gates the inside container on `marker.enabled` (size 0 → hidden), matching line.

Not changed — **shape-aware bounds**: `insideMarkerContainer` is deliberately the inscribed square, which is conservative for any marker shape (text never overflows). A shape-specific bound is an enhancement, not a correctness fix, so it's left as documented.

alantreadway (approved):
LGTM

PR #7382, opened 2026-07-10
AG-17608 Move series actions to legend-item context menu

yinkelly:
- Add a legend-item context menu with the colour submenu and remove series actions
- Use itemStyler to toggle emphasised point labels and add highlight theme overrides in the example
- Update context-menu docs to describe the legend-item colour flow


alantreadway (approved):
LGTM

PR #7396, opened 2026-07-13
Release resolved themes for inline theme option objects

alantreadway:
`getChartTheme` cached resolved `ChartTheme` instances in a `Map` keyed by the input `value`. For object-valued inputs the key was held **strongly**, so a fresh inline theme object per chart pinned its resolved theme — and the deep config tree it owns — for the process lifetime. Workloads that render many charts with distinct inline theme objects (dashboards, SSR, chart thumbnail generation) leaked memory unboundedly.

This was the root cause of the website-build memory exhaustion that PR #7395 mitigated by recycling batch workers; that mitigation stays in place as a safety net.

## Change
- Key object-valued theme inputs in a `WeakMap` so resolved themes become collectable once their options object is unreferenced.
- Keep the bounded stock-name (string) keys in the strong `Map`.
- Reset both caches on module-registry change.

Theme resolution and cache-hit semantics for live references are unchanged.

## Evidence
Repeatedly rendering one example (10 themes × 2 DPI) through a single process:

| | JS heap growth / iteration |
|---|---|
| Before | ~40 MB, linear (unbounded → runner OOM) |
| After | ~0.3 MB, flat |

## Test
Added a `FinalizationRegistry` regression test asserting resolved themes are collected once their inline options object is dropped. Verified it fails (0/20 collected) with the strong-Map behaviour and passes (20/20) with the fix. Full community (3801) and enterprise (2551) suites pass.

### DIFF
--- a/packages/ag-charts-community/src/chart/mapping/themes.test.ts
+++ b/packages/ag-charts-community/src/chart/mapping/themes.test.ts
@@ -318,4 +318,35 @@ describe('themes.ts', () => {
             expect(logs[1][1]).toBe('miss');
         });
     });
+
+    // Heap/finalizer timing can be flaky - retry to keep the signal reliable.
+    describe('theme cache memory', { retry: 5 }, () => {
+        async function collectGarbage() {
+            for (let i = 0; i < 10; i++) {
+                globalThis.gc?.();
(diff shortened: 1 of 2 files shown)


lsjroberts (approved):
LGTM

PR #7397, opened 2026-07-13
Speed up staging example generation

alantreadway:
The Website Build's example batch crawls (873 tasks / 4 threads → ~282 s), dominated by staging example generation. Two independent causes, fixed here:

## 1. Redundant prettier formatting (code)
In prod/staging mode every generated file is prettier-formatted. Each example is emitted as **12 dark-mode × framework variants**, and the shared data/topology source files are byte-identical across them — yet were re-formatted once per variant. On large map examples (~900 KB `topology.ts`) a single prettier pass takes seconds.

A per-example memo cache (created in `generateFiles`, threaded through `getGeneratedContents` → `getOtherScriptFiles`) formats each distinct `(parser, source)` once and reuses it across the variants. Scoped to one example and discarded when it finishes; callers that omit it (website Astro pages) are unaffected. **Output is unchanged.**

| example | before | after |
|---|---|---|
| bubble-map (909 KB topology) | 12,097 ms | 2,329 ms |
| map-lines/lines | 8,293 ms | 1,575 ms |
| 100--stacked-column (simple) | 74 ms | 73 ms |

~5× on heavy map examples; simple examples unaffected.

## 2. Staging examples were never cached (CI)
`website_build` uses a read-only Nx cache and is the only job running `generate-example -c staging`, so staging per-example outputs had no writer and regenerated every build. They're now generated in the read-write `init` job (scoped to `latest`), so `website_build` restores them instead.

Known follow-up (out of scope): the `nx_cache_hash` includes package.json versions, so the daily `beta` version bumps rotate the cache key and cap cross-build reuse to within-a-day windows.

`tsc --noEmit`, build, and the plugin test suite pass.


plugins/ag-charts-generate-example-files/src/executors/generate/generator/utils/formatCache.ts
lsjroberts, line 1: Just curious, why does this appear as a binary file?

lsjroberts (approved):
LGTM

plugins/ag-charts-generate-example-files/src/executors/generate/generator/utils/formatCache.ts
alantreadway: Good catch — it was a raw NUL byte used as the cache-key delimiter on line 22; git treats any file containing a NUL as binary, which is why the whole file rendered as "binary". Fixed in 70ff176 by using the `\u0000` escape sequence instead (same runtime NUL separator that can't collide with parser names or source, but the source stays plain text and now diffs normally).

PR #7399, opened 2026-07-13
AG-6439 Centre inside-placement labels on scatter and bubble markers

iMoses:
## Summary

`placement: 'inside'` labels rendered off-centre on scatter and bubble markers, unlike line/area which centre them. This aligns scatter/bubble with the shared behaviour.

**Root cause:** scatter/bubble carry a bespoke `updateLabelNodes` (`bubbleSeries.ts`) that positioned the label text at the raw placed-box top-left, **without the label-padding inset** the shared `PlacedLabelCartesianSeries` base applies (`text.x = point.x + labelPadding.left`). The placement engine centres the *padded* box on the marker, so the text must be inset by the padding to sit centred. Line/area do this; scatter/bubble didn't, so any boxed inside label (a background box is common for contrast) was pushed toward the top-left. Unboxed labels have zero padding, so the defect only affected boxed inside labels — which is why existing tests (unboxed) never caught it.

## Change

One-line-per-axis padding inset in `BubbleSeries.updateLabelNodes`, mirroring the shared base. Covers scatter via inheritance. `expandLabelPadding` returns 0 for unboxed labels, so the unboxed path is unchanged (no existing-snapshot churn).

## Tests

New cross-series coverage in `labelFit.test.ts` for the boxed inside case — bubble marker, scatter marker, and a non-centred `pin` marker (whose inscribed label rectangle sits away from the marker centre) — each pairing an image snapshot with a non-vacuous text assertion. New baselines were visually confirmed as centred.

## Verification

- `yarn nx test ag-charts-community` — full suite green; no existing snapshots changed.
- `build:types`, `format`, `lint` clean.

## Notes

Follow-up to the merged AG-6439 (#7377). Scope is deliberately narrow: the padding inset only. A separate divergence (bubble applies a marker-shape anchor offset that line/area don't) was investigated and found to be *correct* for non-centred marker shapes — the anchor and inside-offset together map the label onto the marker's visual inscribed rectangle — so it was intentionally left unchanged.


alantreadway (approved):
LGTM

PR #7403, opened 2026-07-14
SE-61 fix privacy redirect

seanlandsman:


examples/server-side-rendering/package.json
```diff
@@ -8,7 +8,7 @@
     "server": "tsx examples/03-express-server.ts"
   },
   "dependencies": {
-    "ag-charts-server-side": "14.0.1-beta.20260713.1437",
+    "ag-charts-server-side": "14.0.0",
```
taktran: I think this isn't right

taktran (approved):

PR #7404, opened 2026-07-14
AG-9227 Rename label orientation values; drop range-area support

iMoses:
https://ag-grid.atlassian.net/browse/AG-9227

Fix #AG-9227


alantreadway (approved):
LGTM

PR #7410, opened 2026-07-14
AG-17261 Add collapsedChange event

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-17261


packages/ag-charts-community/src/chart/interaction/collapsedManager.ts
```diff
@@ -44,34 +57,87 @@ export class CollapsedManager implements MementoOriginator<CollapsedMemento> {
...
+        };
+
+        const getDatum = seriesId ? this.getDatum[seriesId] : undefined;
+
+        this.chartService.callListener({
+            type: 'collapsedChange',
+            source,
+            preventDefault,
+            collapsed: Object.keys(collapsedIds).map((id) => ({
+                itemId: id,
```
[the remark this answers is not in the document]
lsjroberts: Standalone series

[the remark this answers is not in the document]
lsjroberts: This is already the case with `getState()` reporting it as strings. This is a separate issue related to https://ag-grid.atlassian.net/browse/AG-17721

packages/ag-charts-community/src/chart/interaction/collapsedManager.ts
```diff
@@ -22,19 +29,28 @@ export class CollapsedManager implements MementoOriginator<CollapsedMemento> {
 
     restoreMemento(_version: string, _mementoVersion: string, blob: CollapsedMemento | undefined) {
         if (blob) {
-            this.collapse(blob);
+            const defaultGetDatumSeriesId = Object.keys(this.getDatum).at(0);
+            this.collapse(blob, defaultGetDatumSeriesId, 'api-call');
```
lsjroberts, line 33: To get the datums on the event, the `CollapsedManager` has to have a way to retrieve them from the series. The methods other than `restoreMemento()` could just be given the `getDatum()` callback directly, but `restoreMemento()` is called independently of series. So it has to use a "default" one.

Since this is only used by standalone series, then this should be a fine approach for now.

If we ever need to do collapsing on non-standalone series we may need to re-think this, but seems unlikely for any time soon.

packages/ag-charts-enterprise/src/series/organization/organizationSeries.test.ts
alantreadway, line 1: Perhaps we could add an e2e test to exercise an actual click-to-expand/collapse?

packages/ag-charts-angular/tests/zone-callback-inventory.json
```diff
@@ -7,6 +7,10 @@
         "classification": "zone-wrap",
         "reason": "Chart listener; re-entered into the zone by patchChartOptions via patchListeners(options.listeners)."
     },
+    "AgBaseChartListeners.collapsedChange": {
+        "classification": "zone-wrap",
+        "reason": "Chart listener; re-entered into the zone by patchChartOptions via patchListeners(options.listeners)."
+    },
```
alantreadway, line 13: Sanity check: Are we happy that no changes are needed for this new listener to actually be wrapped in an Angular zone-aware wrapper?

packages/ag-charts-community/src/chart/interaction/collapsedManager.ts
```diff
@@ -44,34 +60,87 @@ export class CollapsedManager implements MementoOriginator<CollapsedMemento> {
...
+        if (!changed) return;
+
+        let defaultPrevented = false;
+        const preventDefault = () => {
+            defaultPrevented = true;
+        };
+
+        const getDatum = seriesId ? this.getDatum[seriesId] : undefined;
+
+        this.chartService.callListener({
```
alantreadway, line 125: I'm just wondering if we should instead issue an event on the `eventsHub` and having `chart.ts` subscribe to it and call the user callback if registered to reduce coupling?

lsjroberts: I'm not sure that would reduce coupling really, and more just obfuscate the coupling, since it would still need to return the `defaultPrevented` status.

I think if, like with zoom, there becomes multiple things that can impact/modify the collapse change then it would make sense to have a "request collapse" & "apply collapse" event chain.

As is, I think having a request -> apply -> complete event chain is unnecessary at the moment.

packages/ag-charts-angular/tests/zone-callback-inventory.json
lsjroberts: Confirmed event working as expected with the angular framework example.

packages/ag-charts-website/e2e/org-chart.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+                            },
+                        },
+                    ]),
+                ])
+            );
+            consoleLogs.clear();
+
+            // Expand to reset to initial
+            await page.mouse.click(henry7After.x, henry7After.y);
+            await delay(100); // Small delay to allow focus highlight to update
```
lsjroberts, line 77: There was some flakiness in this test, which this appears to fix. But will keep an eye out for if it flakes again.

alantreadway (approved):
LGTM

PR #7429, opened 2026-07-15
AG-17637 TC2 Fix the `axisId` parameter in contextmenu events.

olegat:


olegat:
Aborting - this approach has confirmed breaking-change. See 87a7e127528db4da6149ce8a79fc6f16eca969a8 for more detail

PR #7432, opened 2026-07-15
AG-17724 Add org expander showAllChildren and showDirectChildren

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-17724


packages/ag-charts-enterprise/src/series/organization/organizationSeries.ts
lsjroberts, line 1: Most of these changes are just re-ordering methods to group by public/protected/private.

alantreadway (approved):
LGTM

PR #7447, opened 2026-07-16
Reduce glob-attached rule context load; add agentic eval harness

alantreadway:
Trims and re-scopes repo-tracked agent rules so irrelevant guidance stops attaching (series test files: ~5.8k → ~3.2k attached words; e2e specs now load only the e2e rule), and adds `tools/agentic-eval/`: a rule-attachment simulator plus an A/B no-degradation eval harness (headless agent runs in worktrees, condition-blind judge, token metrics).

Companion upstream guide changes: ag-grid/ag-dev-prompts#562 (must merge there for the fetched-rule trims to stick past the next install).

Sanity eval (sonnet + opus × 3 tasks × baseline/modified rules): no judge-score regression in any pair; input tokens and cost down 20–60% in most cells. Results in `reports/agentic-eval/sanity-run/summary.json` (not committed).

### DIFF
--- a/.rulesync/rules/cartesian-series-types.md
+++ b/.rulesync/rules/cartesian-series-types.md
@@ -2,198 +2,39 @@
 root: false
 targets: ['*']
 description: 'CartesianSeries consolidated generic types pattern documentation'
-globs: ['**/series/cartesian/**/*.ts', '**/series/**/*Series.ts']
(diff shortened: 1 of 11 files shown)


AG-DavidG (approved):
LGTM

PR #7465, opened 2026-07-20
fix: avoid _previousSplit TypeError when destroy races performUpdate

### LINKED ISSUE #7464: [Bug]: TypeError on `_previousSplit` when destroy() races an in-flight performUpdate
### Link to reproducible scenario

https://codesandbox.io/p/sandbox/yms3fk

(Gist source: https://gist.github.com/alexander7161/fc275f45ee1e53a1dbc789266a7f43cd)

### Describe the bug

**Steps to reproduce**
1. Open the CodeSandbox / gist HTML (ag-charts-community **14.0.1** via CDN).
2. Click **Reproduce**.
3. Open the browser console / on-page log.

**Actual behaviour**

Console logs:

```text
AG Charts - update error TypeError: Cannot assign to read only property '_previousSplit' of object '#<_CartesianChart>'
```

Firefox reports: `"_previousSplit" is read-only`.

**Expected behaviour**

If a chart is destroyed (or otherwise frozen as at the end of `destroy()`) while an async `performUpdate` is in flight, the in-flight update should no-op / cancel cleanly — no TypeError, no `AG Charts - update error` console noise.

**More information**

### Root cause

In `Chart.destroy()` (14.0.1), the instance is still frozen at the end:

```js
this.destroyed = true;
Object.freeze(this);
```

`performUpdate` is async and calls `updateSplits()` between stages. `updateSplits` always does:

```js
this._previousSplit = performance.now();
```

with **no** `if (this.destroyed) return` guard. After `Object.freeze(this)`, that assignment throws. `tryPerformUpdate` catches it and logs `AG Charts - update error`.

The live repro applies the terminal state of `destroy()` (`destroyed = true` + `Object.freeze(this)`) on the first `updateSplits` call during an update, which is the state a completed `destroy()` leaves for any update still mid-flight (e.g. framework unmount calling `destroy()` while `performUpdate` is awaiting between stages). Calling `destroy()` re-entrantly from inside `updateSplits` can abort before `Object.freeze`; the freeze itself is what triggers the TypeError.

Same failure mode on 13.2.0 with a real `chart.destroy()` call from inside `updateSplits` (freeze runs and the error is logged).

### Impact

In React apps using `ag-charts-react`, unmount runs `destroy()` in effect cleanup. If an options `update()` is still in flight, this race surfaces as handled console errors (tab does not crash).

### Suggested fix

Minimal:

```js
updateSplits(splitName) {
  if (this.destroyed) return;
  // ...
}
```

and/or skip logging in `tryPerformUpdate` when `this.destroyed`, and/or await/cancel the update mutex before `Object.freeze(this)` in `destroy()`.

### Version

14.0.1

### Does the issue occur for a specific framework only?

All languages

### Is the issue only observable on a specific browser?

All browsers

alexander7161:
## Summary

Fixes #7464.

When `destroy()` races an in-flight `performUpdate`, `updateSplits` could assign `_previousSplit` after `Object.freeze(this)` in `performTeardown`, producing:

`AG Charts - update error TypeError: Cannot assign to read only property '_previousSplit'`

(Firefox: `"_previousSplit" is read-only`).

`destroy()` already sets `destroyed` early and queues teardown behind the update mutex; this change makes the update path honor that:

- `updateSplits` no-ops when `destroyed`
- `tryPerformUpdate` catch skips logging / further instance writes when `destroyed` (avoids a follow-on read-only error on `runningUpdateType`)

## Test plan

- [x] Extended `Chart destroy() / performUpdate() race condition` with a case that freezes during `updateSplits` (asserts no `console.error` via `setupMockConsole`)
- [ ] Existing destroy/performUpdate race tests still pass

### DIFF
--- a/packages/ag-charts-community/src/chart/chart.test.ts
+++ b/packages/ag-charts-community/src/chart/chart.test.ts
@@ -1607,4 +1607,29 @@ describe('Chart destroy() / performUpdate() race condition', () => {
         'does not throw TypeError when destroy() races with a %s series update',
         testDestroyRace
     );
+
+    it('does not throw when updateSplits runs after the chart instance is frozen', async () => {
+        const proxy = AgCharts.create({
+            container: document.body,
+            data,
+            series: [{ type: 'bar', xKey: 'x', yKey: 'y' }],
+        }) as AgChartProxy;
+        const innerChart = deproxy(proxy);
+        await waitForChartStability(innerChart);
+
+        const chartProto = Object.getPrototypeOf(innerChart);
+        const origUpdateSplits = chartProto.updateSplits;
+        vi.spyOn(chartProto, 'updateSplits').mockImplementationOnce(function (this: any, name: string) {
+            // Mirror performTeardown's terminal state mid-update (Object.freeze(this)).
+            this.destroyed = true;
+            Object.freeze(this);
+            return origUpdateSplits.call(this, name);
+        });
+
+        innerChart.update(ChartUpdateType.FULL);
+        await waitForChartStability(innerChart);
+
+        // setupMockConsole afterEach asserts console.error was not called
+        // (would see: Cannot assign to read only property '_previousSplit').
+    });
 });
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -977,6 +977,10 @@ export abstract class Chart extends Observable implements ModuleInstance, ChartS
     private _previousSplit: number = 0;
 
     private updateSplits(splitName: string) {
+        // destroy() sets `destroyed` before teardown is mutex-queued; skip timing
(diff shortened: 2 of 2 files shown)


AG-DavidG:
Hi @alexander7161 

Thank you for using our product and for offering to contribute to it.

We appreciate your readiness to work on this yourself.

However, the way this feature will be implemented by our team will depend on many factors, such as supporting a wider range of use cases than the specific one you have, ensuring it works well with other features of the products, ensuring doesn't degrade performance, as well as follows coding conventions and patterns used in our codebase. We would also need to comprehensively test this new feature as well as document it.

For this reason, we do not encourage customers to invest time in submitting pull requests, as our implementation of this functionality may significantly differ from the code you may build to support your specific scenario.

If you have found any bugs or require additional functionality, I would encourage you to raise this via our support portal if you do have an active AG Charts Enterprise licence.

Kind regards,

David

PR #7468, opened 2026-07-20
AG-17263 TC2 Auto-expand collasped nodes when pressing ArrowDown

olegat:


alantreadway (approved):
LGTM

PR #7471, opened 2026-07-21
AG-17907 Change zoom to not reset on setState when undefined

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-17907


alantreadway (approved):
LGTM

PR #7481, opened 2026-07-21
AG-17916 [CLAUDE] Split makeGetItemsParams into per-scope methods

olegat:
https://ag-grid.atlassian.net/browse/AG-17916

Pure refactor of the context-menu `getItems` params assembly; no behaviour change.

- Add a `GetItemsParams` alias for the `[AgContextMenuGetItemsParams, Caller[]]` return tuple.
- Reduce `makeGetItemsParams` to a `showOn` dispatcher, moving each case's body into its own method: `makeGetItemsParamsSeriesNode`, `makeGetItemsParamsAxis`, `makeGetItemsParamsCaption` and `makeGetItemsParamsLegendItem`. The trivial `always`/`series-area` case stays inline, and `default` keeps the exhaustiveness assertion.


alantreadway (approved):
LGTM

PR #7488, opened 2026-07-21
Change line/area label placement theme default to 'top'

iMoses:
Changes the line and area series `label.placement` theme default from the fallback list `['top', 'bottom']` to the scalar `'top'`, so labels seat above their point by default rather than cascading to the first placement that fits the bounds.

Regenerated the affected community, enterprise, and website e2e snapshots, and updated the `labelCollisionAvoidance` assertion test that encoded the old fallback-list behaviour.


alantreadway (approved):
LGTM

PR #7489, opened 2026-07-22
AG-17879 Clarify collision-avoidance option docs

AG-DavidG:
https://ag-grid.atlassian.net/browse/AG-17879

Fix #AG-17879


iMoses (approved):
LGTM

PR #7490, opened 2026-07-22
AG-17906 Beside bar label placements and series-area label collision

iMoses:
## AG-17906

Adds "beside" bar-series label placements and makes cascade outside labels respect the series area.

### Beside placements

Six new `beside-{before|after}-{start|center|end}` values on `AgBarSeriesLabelPlacement` (bar series only; histogram/waterfall/range-bar unchanged). A `beside-*` label offsets perpendicular to the value axis, to the side of its segment: `before`/`after` choose the side (a column's left/right, a horizontal bar's above/below), `start`/`center`/`end` position it along the segment's length. Intended for tiny stacked segments with no room to place a label along the value axis. `before`/`after` flip with a reversed category axis so they keep their physical meaning.

Stacked bars also reject an `outside-*` placement that would point into an adjacent stacked segment, so the cascade falls through rather than mislabelling the neighbour.

### Series-area collision for cascade labels

A cascade `outside`/`beside` bar label that overflows the series area (the plot rect grown by `seriesArea.padding`, i.e. the clip rect) is now treated as a collision, so the cascade falls through to the next user-provided placement instead of the label spilling into the axis-label zone below/beside the plot. This is collision detection, not clamping: the label is never slid to fit, and `seriesArea.padding` is honoured as legitimate label space. Implemented via a collision-only region flag (`flushToRegion`) on the placement engine, so behaviour for all other placed-label series is unchanged.

Before this change the effect was asymmetric: labels near the top happened to look fine because the top margin was empty, while `outside-start` labels dropped into the x-axis labels at the bottom.

### Testing

- New community snapshot tests: the six beside placements (both bar directions), category/value axis reversal, stacked outside-rejection fallback, series-area overflow falling back to inside, and an outside label kept within `seriesArea.padding`.
- Full community and enterprise suites pass; core/community/enterprise type-check and lint clean.

Docs unchanged (intentionally out of scope).


alantreadway (approved):
LGTM

PR #7495, opened 2026-07-22
AG-17261 Change collapsedChange event to have only the delta

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-17261

Also fixes issues with string/numeric item ids.


alantreadway (approved):
LGTM

PR #7534, opened 2026-07-24
AG-17776 [CLAUDE] Keep axis resize cursor after releasing drag over axis

olegat:
https://ag-grid.atlassian.net/browse/AG-17776

Releasing an axis drag while the pointer was still over the axis cleared the ew-resize/ns-resize cursor. onAxisDragEnd unconditionally cleared the zoom cursor, and because the pointer never left the axis region no fresh mouseenter fired to restore it.

Extract the hover-cursor computation from onAxisMouseEnter into updateAxisCursor(direction) and track the hovered axis direction alongside hoveredAxisId. On drag end, re-apply the hover cursor when still hovering an axis instead of clearing it. Recomputing (rather than merely skipping the clear) reproduces exactly what a fresh hover would decide, so a drag that ends fully unzoomed still clears correctly.

Add a self-calibrating Playwright e2e regression test that drags the x-axis and asserts the resize cursor remains after release. The JSDOM unit harness cannot cover this: its zero-size getBoundingClientRect makes the drag layer always synthesise a mouseenter that re-sets the cursor, masking the regression.

### DIFF
--- a/packages/ag-charts-enterprise/src/features/zoom/zoom.ts
+++ b/packages/ag-charts-enterprise/src/features/zoom/zoom.ts
@@ -102,6 +102,7 @@ export class Zoom extends AbstractModuleInstance {
     private readonly buttons: ZoomToolbar;
 
     private hoveredAxisId?: AxisID;
+    private hoveredAxisDirection?: ChartAxisDirection;
 
     // State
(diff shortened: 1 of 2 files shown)


alantreadway (approved):
LGTM

PR #7536, opened 2026-07-24
AG-17263 [CLAUDE] Restore org-chart Enter/Space expand/collapse when clickToExpand

olegat:

The Alt+Up/Down keybindings landed alongside a change to OrganizationSeries.hasBuiltinListener (commit 413ab9f2c6, "TC4 Fix clickToExpand on card images/text") that switched the guard from `target == null || ...` to `target != null && ...`. Keyboard Enter/Space activations emit a synthetic click with no pointer target, so the new `target != null` guard silently dropped them, even though the comment still claimed keyboard activations were allowed. Result: Enter/Space no longer expanded/collapsed the focused node — a breaking change.

- hasBuiltinListener now returns `target?.tag === Expander || clickToExpand`. For any non-null (pointer) target this is identical to the previous expression, so click behaviour is unchanged; the only behavioural difference is the null (keyboard) case, which now toggles when clickToExpand is enabled (the default). Alt+Up/Down remain independent of clickToExpand.
- Re-add a screen-reader instruction advertising Enter/Space, shown only when clickToExpand is enabled (matching where the keys actually work); the Alt instruction is still announced unconditionally. New ariaDescriptionToggleNode locale key added to en-US and translated across all locale files to keep key parity.
- Update the Organisation Chart docs Accessibility section to describe arrow-key navigation, Alt+Down/Up, and Enter/Space (when clickToExpand is enabled).
- Add keyboard tests covering Enter/Space toggling with clickToExpand on, the no-op when it is off, and the corresponding announcement instructions.

### DIFF
--- a/packages/ag-charts-enterprise/src/series/organization/organizationSeries.test.ts
+++ b/packages/ag-charts-enterprise/src/series/organization/organizationSeries.test.ts
@@ -1769,6 +1769,24 @@ describe('OrganizationSeries', () => {
             seriesArea.dispatchEvent(new KeyboardEvent('keydown', { key, code: key, bubbles: true }));
         }
 
+        function pressKeyOnSeriesArea(key: string, code: string) {
+            const seriesArea = document.querySelector<HTMLElement>('.ag-charts-series-area');
+            if (!seriesArea) throw new Error('series-area element not found');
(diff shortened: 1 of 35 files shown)


alantreadway (approved):
LGTM

PR #7538, opened 2026-07-24
Fix bar-family label collisions and auto-fit pie sector labels

iMoses:
## Label collision fixes (bar family)

- Align bar label collision footprints with their drawn boxes and fix box reservation, so a hideable label no longer collides a whole box-extent before its drawn box reaches the neighbour.
- Contribute baked bar/histogram/range-bar inside labels as cross-series collision obstacles, and gate series-area label collision behind `collideWith.seriesArea`.
- Wire the placement-array cascade for histogram, waterfall and range-bar, so a label that cannot fit one placement falls back to the next instead of being dropped.
- Range-bar's two end labels share a single bar rect. Treat an already-placed sibling as an obstacle so the second label cascades outside rather than stacking on top of it or being dropped, and fix the range-bar `outsideStyle` default colour so outside labels stay legible.

## Pie/donut sector labels

- Auto-fit pie and donut sector labels to their wedge, with wrapping and truncation inside the sector geometry. The manual `maxWidth`/`maxHeight`/`wrapping`/`truncate` sector-label options (unreleased, never intended to be public) are demoted to undocumented overrides.

## Behaviour change

Bar-family series now default to cross-series label avoidance (`collideWith.seriesItems` / `seriesArea`). Existing charts mixing a bar-family series with another labelled series may relocate or hide labels that previously overlapped. Snapshots have been regenerated accordingly.


iMoses:
### Codex review follow-up

Thanks — worked through all 9 findings. Summary of the resolution:

**Fixed (with regression tests):**
- **[P1] Preserve full text until histogram/waterfall cascades resolve** — histogram and waterfall bound the label text to the first placement's inside container, truncating (or dropping) a label before it could cascade outside. Now only an inside-only placement binds the text. `2d6b12b7`
- **[P1] Respect reversed value axes in range-bar candidates** — the low/high ends were hardcoded to the bottom/top of the rect, so both labels annotated the wrong ends on a reversed axis. The bar direction is now derived from the value axis. Reversed-axis snapshots regenerated. `55f3d245`
- **[P1] Per-orientation/placement anchors and box extents** — `buildBarLabelCandidates` resolved the box extent once from the first placement's style; it now resolves per placement, so an inside/outside cascade with differing `insideStyle`/`outsideStyle` padding sizes and offsets each candidate correctly. `ddacc95e`
- **[P2] Border extents in bar-family offsets** — histogram/waterfall/range-bar offset outside labels by padding only while sizing with the border; they now use the border-inclusive extent, matching bar. `aea14995`

**Not changed, with rationale:**
- **[P1] Public sector-label fit type break** — these options were never released; they leaked into the public type prematurely and are intentionally kept undocumented at this point, so removing them is not a breaking change.
- **[P1] Restrict own-box obstacle exemptions to genuine own geometry** — the obstacle index carries no owner identity to scope by, and the exemption only re-includes labels that genuinely overlap the range-bar's own rect, which is the intended avoidance for a series that opts into label collision. Left as-is.
- **[P1] Series-area containment for inside marker labels** — by design: an inside marker label is fitted to and rides with its marker, and the behaviour is gated behind the (default-disabled) bubble/scatter `collideWith.seriesArea` work. Out of scope for this PR.

alantreadway (approved):
LGTM

PR #7540, opened 2026-07-24
AG-17975 Add preset fast-update path for gauge value/targets updates

alantreadway:
https://ag-grid.atlassian.net/browse/AG-17975

Fix #AG-17975


iMoses (approved):
LGTM

PR #7545, opened 2026-07-24
AG-17975 Reduce financial demo streaming CPU usage

alantreadway:
https://ag-grid.atlassian.net/browse/AG-17975

Fix #AG-17975

### DIFF
--- a/packages/ag-charts-community/src/chart/fonts/fontManager.ts
+++ b/packages/ag-charts-community/src/chart/fonts/fontManager.ts
@@ -6,11 +6,19 @@ export class FontManager {
     private observers: Array<ResizeObserver> = [];
     private destroyed = false;
 
-    // Fonts never unload once available, so specs confirmed by `check()` stay available for the
-    // chart's lifetime. Caching them lets streaming updates skip the per-update native check.
+    // Caches specs confirmed by `check()` so streaming updates skip the per-update native check.
+    // `check()` reports a spec as available even when no matching `@font-face` exists yet, so a
+    // later font declaration or load can make a cached "confirmed" verdict stale; the
+    // `loadingdone`/`loadingerror` listener below clears the cache whenever the document's font set
+    // changes, forcing the next update to re-check.
     private readonly confirmedFontSpecs = new Set<string>();
+    private watchedFontSet?: FontFaceSet;
 
     constructor(private readonly ctx: DynamicContext<ChartRegistry>) {}
+
+    private readonly onFontSetChange = () => {
+        this.confirmedFontSpecs.clear();
+    };
 
     public updateFonts(fonts?: Set<string>) {
         if (!fonts || fonts.size === 0) return;
@@ -30,6 +38,8 @@ export class FontManager {
 
         const fontSet = getDocument('fonts');
         if (fontSet == null) return;
+
+        this.watchFontSet(fontSet);
 
         const pending: Array<Promise<unknown>> = [];
         for (const spec of fontSpecs) {
@@ -53,12 +63,27 @@ export class FontManager {
         });
     }
 
+    // The document `FontFaceSet` outlives individual charts, so a cached "confirmed" spec must be
+    // invalidated whenever the set changes. Attach lazily (never in the constructor: SSR has no font
+    // set, and only the first real `waitForFonts` call proves one exists) and only once per set.
+    private watchFontSet(fontSet: FontFaceSet) {
+        if (this.watchedFontSet === fontSet || !('addEventListener' in fontSet)) return;
+        this.watchedFontSet = fontSet;
+        fontSet.addEventListener('loadingdone', this.onFontSetChange);
+        fontSet.addEventListener('loadingerror', this.onFontSetChange);
+    }
+
     public destroy() {
         this.destroyed = true;
         for (const observer of this.observers) {
             observer.disconnect();
         }
         this.observers = [];
+        if (this.watchedFontSet != null) {
+            this.watchedFontSet.removeEventListener('loadingdone', this.onFontSetChange);
+            this.watchedFontSet.removeEventListener('loadingerror', this.onFontSetChange);
+            this.watchedFontSet = undefined;
+        }
     }
 
     private loadFonts(fonts: Set<string>) {
--- a/packages/ag-charts-community/src/chart/fonts/fontManager.ts
+++ b/packages/ag-charts-community/src/chart/fonts/fontManager.ts
@@ -6,6 +6,10 @@ export class FontManager {
     private observers: Array<ResizeObserver> = [];
     private destroyed = false;
 
+    // Fonts never unload once available, so specs confirmed by `check()` stay available for the
+    // chart's lifetime. Caching them lets streaming updates skip the per-update native check.
+    private readonly confirmedFontSpecs = new Set<string>();
+
     constructor(private readonly ctx: DynamicContext<ChartRegistry>) {}
 
     public updateFonts(fonts?: Set<string>) {
@@ -29,8 +33,11 @@ export class FontManager {
 
         const pending: Array<Promise<unknown>> = [];
         for (const spec of fontSpecs) {
+            if (this.confirmedFontSpecs.has(spec)) continue;
             try {
-                if (!fontSet.check(spec)) {
+                if (fontSet.check(spec)) {
+                    this.confirmedFontSpecs.add(spec);
+                } else {
                     pending.push(fontSet.load(spec));
                 }
             } catch {
--- a/packages/ag-charts-community/src/chart/fonts/fontManager.ts
+++ b/packages/ag-charts-community/src/chart/fonts/fontManager.ts
@@ -6,11 +6,9 @@ export class FontManager {
     private observers: Array<ResizeObserver> = [];
     private destroyed = false;
 
-    // Caches specs confirmed by `check()` so streaming updates skip the per-update native check.
-    // `check()` reports a spec as available even when no matching `@font-face` exists yet, so a
-    // later font declaration or load can make a cached "confirmed" verdict stale; the
-    // `loadingdone`/`loadingerror` listener below clears the cache whenever the document's font set
-    // changes, forcing the next update to re-check.
+    // OPTIMIZATION: skips the native `check()` per streaming update. `check()` reports a spec as
+    // available even with no matching `@font-face`, so a later declaration or load can make a
+    // verdict stale — the `loadingdone`/`loadingerror` listener clears the cache to re-check.
     private readonly confirmedFontSpecs = new Set<string>();
     private watchedFontSet?: FontFaceSet;
 
@@ -63,14 +61,23 @@ export class FontManager {
         });
     }
 
-    // The document `FontFaceSet` outlives individual charts, so a cached "confirmed" spec must be
-    // invalidated whenever the set changes. Attach lazily (never in the constructor: SSR has no font
-    // set, and only the first real `waitForFonts` call proves one exists) and only once per set.
+    // The document `FontFaceSet` outlives individual charts, so a cached verdict must be invalidated
+    // whenever the set changes. Attached lazily: SSR has no font set, and only the first real
+    // `waitForFonts` call proves one exists. Font availability is per-document, so a swapped set
+    // (a chart moved into another document) invalidates every verdict cached against the old one.
     private watchFontSet(fontSet: FontFaceSet) {
         if (this.watchedFontSet === fontSet || !('addEventListener' in fontSet)) return;
+        this.unwatchFontSet();
+        this.confirmedFontSpecs.clear();
         this.watchedFontSet = fontSet;
         fontSet.addEventListener('loadingdone', this.onFontSetChange);
         fontSet.addEventListener('loadingerror', this.onFontSetChange);
+    }
+
+    private unwatchFontSet() {
+        this.watchedFontSet?.removeEventListener('loadingdone', this.onFontSetChange);
+        this.watchedFontSet?.removeEventListener('loadingerror', this.onFontSetChange);
+        this.watchedFontSet = undefined;
     }
 
     public destroy() {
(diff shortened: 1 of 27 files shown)


packages/ag-charts-community/src/chart/fonts/fontManager.ts
```diff
@@ -29,8 +33,11 @@ export class FontManager {
 
         const pending: Array<Promise<unknown>> = [];
         for (const spec of fontSpecs) {
+            if (this.confirmedFontSpecs.has(spec)) continue;
             try {
-                if (!fontSet.check(spec)) {
+                if (fontSet.check(spec)) {
```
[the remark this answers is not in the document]
alantreadway: Fixed in ba615a151a8 — the confirmed-spec cache is now cleared on FontFaceSet `loadingdone`/`loadingerror` events (listener attached lazily, removed in `destroy()`), so later-declared fonts are re-checked and awaited.

packages/ag-charts-community/src/dom/domManager.ts
```diff
@@ -497,6 +504,11 @@ export class DOMManager extends BaseManager {
         this.container = pendingContainer;
         this.pendingContainer = undefined;
         this.agDocument.setContainer(pendingContainer);
+        // Shadow-path watchers live on the previous shadow root and are orphaned by the move; drop
+        // the mirror so they are re-registered against the new root. Normal-path watchers are
+        // children of `this.element`, which moves with its subtree, so their mirror stays valid.
+        this.shadowCssVariableWatchers.clear();
```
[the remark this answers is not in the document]
alantreadway: Fixed in ba615a151a8 — the shadow watcher mirror is now only cleared when the resolved shadow root actually changes; same-root container moves keep the existing watchers and listener.

packages/ag-charts-demos/src/demos/financial/components/grid.ts
```diff
@@ -56,6 +56,22 @@ export const baseColDef = <T>(): ColDef<T> => ({
 
 export const getRowId = <T extends { ticker: string }>({ data }: GetRowIdParams<T>) => data.ticker;
 
+// True when two snapshots of the same row carry identical displayed values. Feeds hand
+// back a fresh object every tick, so identity never matches; compare values instead and
+// treat the rolling `history` window by its length and newest (tail) value.
+export const rowValuesEqual = <T extends object>(a: T, b: T): boolean => {
+    for (const key of Object.keys(a) as (keyof T)[]) {
+        const av = a[key];
+        const bv = b[key];
```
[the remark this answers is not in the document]
alantreadway: Fixed in ba615a151a8 — histories are now compared element-wise (windows are ~24 values, so the cost is trivial).

packages/ag-charts-demos/src/demos/financial/components/SparklineCell.tsx
```diff
@@ -70,20 +50,81 @@ export function SparklineCell({ data: row }: CustomCellRendererProps<SparkRow>)
...
+    const pointsRef = useRef<SparkPoint[]>([]);
+    const seqRef = useRef(0);
+    const baselineRef = useRef(baseline);
+    // Streaming hands a fresh history array every tick, so the effect below fires
+    // even when the drawn line is unchanged. Track a cheap content signature and
+    // skip the redundant chart update when it matches.
+    const contentRef = useRef<string>();
+    const content = `${history.length}:${history[history.length - 1]}:${baseline}`;
 
     useLayoutEffect(() => {
```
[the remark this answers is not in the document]
alantreadway: Fixed in ba615a151a8 — the content signature was redundant as well as lossy: `scrollShift()` already value-compares the window and the effect no-ops when nothing changed, so the signature short-circuit has been removed entirely.

packages/ag-charts-community/src/scale/colorScale.ts
```diff
@@ -110,6 +120,21 @@ export class ColorScale extends AbstractScale<number, string> {
         // Colour derives from the value's position in the Number domain, so finite precision suffices.
         const xn = toNumber(x);
 
+        // Lookup sits after refresh() so a reconfigured scale never serves a stale colour.
+        const cached = this.convertCache.get(xn);
+        if (cached !== undefined) {
+            return cached;
+        }
+
```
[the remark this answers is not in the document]
alantreadway: Fixed in ba615a151a8 — the cap now stops inserting new entries instead of clearing the map, so intra-pass hits are preserved for datasets beyond the cap. The cache is still fully cleared on `update()` (any domain/range/mode change).

packages/ag-charts-enterprise/src/features/shared-toolbar/sharedToolbar.ts
```diff
@@ -155,6 +162,31 @@ export class SharedToolbar extends AbstractModuleInstance {
...
+        this.cachedWidth = width;
+        this.cachedWidthSignature = signature;
+        return width;
+    }
+
+    private widthSignature(): string {
+        return SharedToolbar.SECTION_ORDER.map((section) => {
+            const active = this.activeSections.has(section) ? '1' : '0';
+            const buttons = this.sectionButtons[section]
+                .map((b) => `${b.label ?? ''}~${b.icon ?? ''}~${b.iconPosition ?? ''}`)
```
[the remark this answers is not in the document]
alantreadway: Not changed — this matches the accepted precedent in `zoomToolbar.ts` (`cachedContainerHeight`), which likewise keys on content and ignores late font/theme metric changes. A theme change rebuilds the toolbar options (new button content signature), and a late webfont load altering toolbar width is the same accepted trade-off as the existing zoom toolbar cache. The limitation is documented in a code comment; happy to wire both caches to font-load invalidation in a follow-up if preferred.

packages/ag-charts-enterprise/src/features/ranges/ranges.ts
```diff
@@ -171,6 +182,34 @@ export class Ranges extends AbstractModuleInstance {
...
+        this.cachedToolbarHeight = undefined;
+    }
+
+    private getToolbarHeight(
+        opts: _ModuleSupport.NormalisedRangesOptions,
+        toolbar: _ModuleSupport.BaseToolbar
+    ): number {
+        if (
+            this.cachedToolbarHeight !== undefined &&
+            this.cachedToolbarHeightOpts === opts &&
```
olegat, line 195: There's two potential bugs here:

1. **Stale-height from theme-param change:** chart.ts exits early here when apply new options that only update the theme-params: https://github.com/ag-grid/ag-charts/blob/17ab69c3357a58e2feaf14fa8a06cd80f81cae27/packages/ag-charts-community/src/chart/chart.ts#L1782 Therefore `chartState.setValue` is not called: https://github.com/ag-grid/ag-charts/blob/17ab69c3357a58e2feaf14fa8a06cd80f81cae27/packages/ag-charts-community/src/chart/chart.ts#L1792. For example, if you update `theme.params.rangesMinSize` (which affects the height) then this would return the outdated cached height.

2. **Font loading**: Once the FontManager emits a `'font:load'` event, then the chart will re-render potentially with a new height.

We may need to feed in `chartOptions.themeParameters` somewhere (like we do with `DOMManager.setThemeParameters`) and invalid the cache entirely when `'font:load'` is emitted.

packages/ag-charts-demos/src/demos/financial/components/SparklineCell.tsx
```diff
@@ -70,20 +49,70 @@ export function SparklineCell({ data: row }: CustomCellRendererProps<SparkRow>)
...
+            if (points[shift + i].y !== history[i]) {
+                matches = false;
+                break;
+            }
+        }
+        if (matches) {
+            return { removed: points.slice(0, shift), appended: history.slice(retained) };
+        }
+    }
+    return undefined;
```
olegat: This return statement is unreachable.

The for-loop is inclusive (`shift <= points.length`). Therefore in the final iteration, `retained` is set to `0`. 

In the final iteration, `matches === true` because:
- The first condition (if-statement, `retained > history.length`) is always false because 0 is never greater than the length.
- The second condition (for-statement. `i < retained`) is also always false because `i` is initialised to 0 which is never greater than 0.

Although this isn't a bug, the comment ("otherwise undefined, signalling the caller to reseed") is misleading.

packages/ag-charts-demos/src/demos/financial/components/SparklineCell.tsx
```diff
@@ -70,20 +49,70 @@ export function SparklineCell({ data: row }: CustomCellRendererProps<SparkRow>)
...
     }, []);
 
     useEffect(() => {
-        chartRef.current
-            ?.update(sparklineOptions(containerRef.current!, history, baseline))
-            // eslint-disable-next-line no-console
-            .catch((e) => console.error(e));
+        // The baseline drives segmentation, which a transaction cannot change, so a shift in it —
+        // or any history that isn't a clean scroll of the current one — forces a full reseed.
+        const scroll = baseline === baselineRef.current ? scrollShift(pointsRef.current, history) : undefined;
```
olegat: Claude is telling me that this optimisation will be slow after approximately 2 minutes:

> `baseline` comes from `row.baseline`, which useStreamingMarket sets to `feed.sessionOpen`
> (useStreamingMarket.ts Line 32), and MarketFeed.sessionOpen is `this.bars[0].open`
> (data.ts Lines 363-365). MarketFeed.tick() pushes a bar and then runs
> `if (this.bars.length > MAX_BARS) this.bars.shift()` (data.ts Line 384). seedHistory produces
> HISTORY_BARS = 240 bars and MAX_BARS = 480, so from tick 241 onward the array shifts on EVERY
> tick, `bars[0]` changes on every tick, and `sessionOpen` -- and hence `baseline` -- changes on
> every tick.
> From that point the ternary above always yields `undefined`, so every tick takes the full
> reseed branch: a fresh `history.map(...)` allocation plus
> `chart.update(sparklineOptions(...))`. That is the pre-change behaviour plus the extra
> scrollShift/allocation work, for every sparkline in every visible grid row.
> At the default 500ms tick that is ~2 minutes; at the fastest speed option, sooner. Anyone
> profiling this demo for more than two minutes measures the old code path.
> Note "session open" drifting is arguably a pre-existing data bug (a session open should be
> fixed), but this change makes the sparkline fast path depend on it, so it now matters.

packages/ag-charts-community/src/chart/fonts/fontManager.ts
```diff
@@ -46,12 +61,32 @@ export class FontManager {
...
 
+    // The document `FontFaceSet` outlives individual charts, so a cached verdict must be invalidated
+    // whenever the set changes. Attached lazily: SSR has no font set, and only the first real
+    // `waitForFonts` call proves one exists. Font availability is per-document, so a swapped set
+    // (a chart moved into another document) invalidates every verdict cached against the old one.
+    private watchFontSet(fontSet: FontFaceSet) {
+        if (this.watchedFontSet === fontSet || !('addEventListener' in fontSet)) return;
+        this.unwatchFontSet();
+        this.confirmedFontSpecs.clear();
+        this.watchedFontSet = fontSet;
```
olegat, line 72: Event listeners could potentially leak here. When the `this.watchedFontSet` changes, `removeEventListener` should be called to remove them.

alantreadway: Already fixed in 3f21a1928a8, pushed just before this review landed: `watchFontSet` bails when the set is unchanged, and otherwise calls `unwatchFontSet` (removing both listeners) before attaching to the new set. `destroy()` calls it too. The confirmed-spec cache is cleared on swap as well, since font availability is per-document.

packages/ag-charts-demos/src/demos/financial/components/SparklineCell.tsx
alantreadway: Correct — the inclusive loop always matched on the final iteration, so both the `undefined` return and the comment were wrong. Fixed in a803893945e: `scrollShift` is now total (exclusive loop, degenerate case returned explicitly as a full replacement, which the transaction handles). The reseed is now gated only on the baseline, which is the one thing a transaction genuinely cannot express.

packages/ag-charts-demos/src/demos/financial/components/SparklineCell.tsx
alantreadway: Confirmed and fixed in a803893945e — `openPrice` is now captured once in the constructor, read by both `sessionOpen` and `quote()`. Measured over 600 ticks:

| | before | after |
|---|---|---|
| distinct baselines | 361 | 1 |
| sparkline reseeds | 360 | 0 |

Two notes: I changed `quote()` at the same time, since splitting them would leave the displayed % change sign disagreeing with the sparkline's green/red split. And this does change visible demo behaviour — % change is now measured from a fixed session open rather than a rolling 480-bar window start. That matches `MoverFeed`, whose rows already use a fixed baseline "so it agrees with the trend sparkline", but flagging it as a behaviour change rather than a pure internal fix.

packages/ag-charts-enterprise/src/features/ranges/ranges.ts
alantreadway: Both wired up in a803893945e, though the mechanism differs from the example given.

**Font loading** — fixed; this cache and `sharedToolbar`'s width cache both reset on `font:load`.

**Theme params** — fixed via a new `theme:params-change` event, emitted by `DOMManager.setThemeParameters` after its existing change guard, so it fires exactly when the CSS variables are rewritten and stays silent otherwise. Both toolbar caches subscribe. (A `theme` delta takes `slowSetup` — it is not in `FAST_PATH_OPTIONS` — so `themeParameters` gets a fresh reference and the guard passes.)

On the specific example: `rangesMinSize` is not a theme param. `minSize` is a ranges *option*, written to `--ag-charts-ranges-min-size` via `setModuleCSSVariables` from `opts`, so it lands in `processedOptions` and the existing `opts`-reference key already catches it. Ranges also overrides font-size/family/weight from options (`ranges.css:47-49`), so I could not construct a params-only change that moves the ranges height.

Where I did reproduce it is `sharedToolbar`: `.ag-charts-toolbar` reads `--ag-charts-chrome-font-size-medium`, `-font-family`, `-font-weight` and `--ag-charts-button-border-width` straight from the global params (`toolbar.css:8,43-45`), and its key is button content only. Wired both regardless — same cost, and it closes the class rather than the one param I could reproduce.

Added `domManager` tests for the emit side (fires on change, silent on an identical object, fires when a param is lazily added to the same object).

ag-alantreadway (approved):

PR #7550, opened 2026-07-27
AG-17637 [CLAUDE] Resolve crosshair label formatter index to nearest tick

olegat:

A crosshair sits at an interpolated position rather than on a tick, so `Axis.formatDatum` passed `NaN` as the index to the axis label formatter. Any `label.formatter` reading `params.index` therefore received `NaN` on hover.

For the crosshair source, resolve the value to the nearest tick via `resolveTickIndex`, so the formatter receives the same index it reports for that tick (and that the `showOn: 'axis'` context menu reports at that point). Non-finite conversions fall back to `NaN`, and all other formatter sources (tooltip, series-label, annotation-label, legend-label) are unchanged.

Builds on the reversed-axis picked-index fix, so the resolved crosshair index is consistent with both the axis labels and the context menu.

### DIFF
--- a/packages/ag-charts-community/src/chart/axis/axis.ts
+++ b/packages/ag-charts-community/src/chart/axis/axis.ts
@@ -1062,11 +1062,22 @@ export abstract class Axis<
         );
         const { type, value } = formatParams;
 
+        // A crosshair sits at an interpolated position rather than on a tick, so resolve the nearest tick to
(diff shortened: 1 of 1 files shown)


olegat:
Closing (descoped for this release)

https://ag-grid.atlassian.net/browse/AG-17989

PR #7555, opened 2026-07-27
AG-17844 Onboard range-area labels onto the collision placement engine

iMoses:
https://ag-grid.atlassian.net/browse/AG-17844

Fix #AG-17844


alantreadway (approved):
LGTM

PR #7559, opened 2026-07-27
Fix CSS variable lookup matching Object.prototype member names

iMoses:
`resolveValueOrSymbol` tested every resolved string against the CSS variables map with the `in` operator, which walks the prototype chain of the plain object holding them. Any option string equal to an `Object.prototype` member name was therefore treated as a declared CSS variable and replaced by that member.

Affected strings: `toString`, `valueOf`, `constructor`, `hasOwnProperty`, `isPrototypeOf`, `toLocaleString`, `propertyIsEnumerable` (each becomes a function) and `__proto__` (becomes an object).

This applies to every value passing through option resolution, from theme templates and user options alike — e.g. a chart created with `title: { text: 'constructor' }` resolved `title.text` to a function, which then flows into rendering and fails far from the cause.

Switched the lookup to `Object.hasOwn`, matching the existing pattern in `ag-charts-core/src/utils/data/value.ts`. Real CSS variable keys are `var(--x)`-shaped and can never collide with prototype names, so legitimate variables are unaffected — the existing `var(--brand)` test still passes.

Added a regression test covering both entry paths, verified to fail before the fix.

### DIFF
--- a/packages/ag-charts-community/src/module/optionsGraph.test.ts
+++ b/packages/ag-charts-community/src/module/optionsGraph.test.ts
@@ -1272,6 +1272,27 @@ describe('OptionsGraph', () => {
         });
     });
 
+    describe('css variables', () => {
+        it('should not treat `Object.prototype` member names as CSS variables', () => {
+            const themeConfig = { line: { themeValue: 'valueOf' } };
+            const userOptions = prepareOptions({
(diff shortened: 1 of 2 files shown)


lsjroberts (approved):
LGTM

PR #7566, opened 2026-07-27
AG-17241 Update isCollapsed state to be undefined outside org series

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-17241


alantreadway (approved):
LGTM

PR #7567, opened 2026-07-27
AG-17993 render inside labels the marker cannot hold

iMoses:
https://ag-grid.atlassian.net/browse/AG-17993

Fix #AG-17993


alantreadway (approved):
LGTM

PR #7568, opened 2026-07-27
AG-17273 Fix missing children props in org series item styler type

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-17273


alantreadway (approved):
LGTM

PR #7584, opened 2026-07-28
AG-17947 [CLAUDE] Keep org-chart expander clicks out of node listeners

olegat:
https://ag-grid.atlassian.net/browse/AG-17947

Clicking or double-clicking an Org Chart node's expander toggled expand/collapse and also fired the node's click/double-click events. That conflates two distinct interactions - expanding a node versus activating it - which is unexpected for consumers listening on seriesNodeClick or seriesNodeDoubleClick.

SeriesAreaManager.checkSeriesNodeClick() both fired the user-facing node events and returned the picked node that drives the toggle via the series-area:click event. Add ISeries.firesUserClickListeners(target) so a series can declare that a given scene node owns its clicks outright; OrganizationSeries returns false for the expander pill. The picked node is still returned, so expand/collapse continues to work.

The existing hasBuiltinListener() could not serve as this gate: it is also true for the whole card when node.clickToExpand is enabled, which is what drives the pointer cursor and the keyboard Enter/Space toggle (the latter arrives with an undefined target). Reusing it would have suppressed node events across the entire card.

firesUserClickListeners() is therefore deliberately independent of clickToExpand - a card-body click still fires the node events even when it also toggles. The event surface does not change shape based on an unrelated option, and consumers can call preventDefault() themselves if they want to opt out.

### DIFF
--- a/packages/ag-charts-community/src/chart/series/series.ts
+++ b/packages/ag-charts-community/src/chart/series/series.ts
@@ -1154,6 +1154,10 @@ export abstract class Series<
         return false;
     }
 
+    firesUserClickListeners(_target: Node<unknown> | undefined): boolean {
+        return true;
+    }
+
     protected pickNodesInBBoxPredicate(): PickNodesInBBoxPredicate {
         // By default, pickNodesInBBox just used boxes for hit-testing because it's easier and faster. Series with more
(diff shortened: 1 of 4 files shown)


alantreadway (approved):
LGTM

PR #7586, opened 2026-07-28
AG-17804 Fix horizontal tree layout and refactor to share logic

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-17804

Bug was caused by duplicate code between vertical and horizontal tree layouts that diverged, so refactored to share with a few small fns for the differences.


alantreadway (approved):
LGTM

PR #7589, opened 2026-07-28
Fix crossLineId JSDoc wording

AG-DavidG:
## Summary
- Corrected the JSDoc comment on `AgCommonCrossLineOptions.id` to say `crossLineId` is surfaced in callback and event params, rather than "Context Menu params and events" — `crossLineId` appears more broadly than just the context menu.

## Test plan
- [x] Doc-only change; no functional impact


alantreadway (approved):
LGTM

PR #7592, opened 2026-07-28
AG-17844 Apply the label collision threshold on both axes

iMoses:
https://ag-grid.atlassian.net/browse/AG-17844

Fix #AG-17844


alantreadway (approved):
LGTM

PR #7594, opened 2026-07-28
AG-17951 Fix price-volume preset $applyPadding

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-17951

### DIFF
--- a/packages/ag-charts-enterprise/src/preset/priceVolumePresetModules.ts
+++ b/packages/ag-charts-enterprise/src/preset/priceVolumePresetModules.ts
@@ -98,4 +98,15 @@ export const PriceVolumePresetModule: PresetModuleDefinition<AgPriceVolumePreset
     options: priceVolumeOptionsDef,
 
     create: priceVolume,
+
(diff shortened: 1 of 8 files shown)


packages/ag-charts-enterprise/src/preset/priceVolumePresetModules.ts
```diff
@@ -98,4 +98,15 @@ export const PriceVolumePresetModule: PresetModuleDefinition<AgPriceVolumePreset
...
+    themeTemplate: {
+        padding: {
+            $applyPadding: {
+                top: 6,
+                right: 8,
+                bottom: 6,
+                left: 0,
+            },
+        },
+    },
```
lsjroberts, line 111: Moving the internal `$applyPadding` operation to the theme template means it is resolved as part of the theme, not as a user option. The other operations are not an issue as `$applyPadding` is a special case operator, though ideally all would be in the theme template.

alantreadway (approved):
LGTM

PR #7597, opened 2026-07-28
AG-17417 Always pop InteractionState.ZoomDrag on 'drag-end'

olegat:
https://ag-grid.atlassian.net/browse/AG-17417


alantreadway (approved):
LGTM

PR #7598, opened 2026-07-28
AG-17804 Fix org reversed horizontal layout expander chevron position

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-17804


alantreadway (approved):
LGTM

PR #7600, opened 2026-07-28
Fix org chart spacing example label and doc typo

AG-DavidG:
- Renamed the "Vertical:" spacing control label to "Depth:" in the org-chart-spacing example so it matches what the slider actually controls.
- Removed a duplicated comma in the `clickToExpand` default-value description on the Org Chart docs page.


alantreadway (approved):
LGTM

PR #2226, opened 2024-08-07
CRT-440 / AG-12506 - Fix layouts test page

taktran:
## Changes

* Add support for shadow DOM examples, with `shadowDomSelector` in a `exampleConfig.json` file
* Add support for iframe examples, with `iframeSelector` in a `exampleConfig.json` file
* Add timeout for checking for charts loading in example runner

PR #7611, opened 2026-07-29
AG-17996 - Add consent fields to trial licence and contact forms (b14.1.0)

taktran:
Ports the AG-17996 consent field changes across from ag-grid (ag-grid#14646, merged as `a8e0dc175a9`).

The `ag-website-shared` files were copied manually rather than via `git subrepo`, and are **byte-identical** to the merged ag-grid versions.

Adds three web-to-lead consent checkboxes to the trial licence and contact forms:

| Checkbox | Required | Salesforce field |
| --- | --- | --- |
| Data Processing Consent | Yes — blocks submit | `Data_Processing_Consent__c` |
| Email Marketing Consent | No | `Marketing_Email_Consent__c` |
| Email Tracking Consent | No | `Email_Tracking_Consent__c` |

Email tracking consent only applies to France and Italy. Until IP geolocation is in place, an optional "I live in France or Italy" checkbox reveals it, and is itself submitted so Salesforce can distinguish a French/Italian visitor who declined tracking from one who was never shown the option. Unchecking the declaration also withdraws the tracking consent, so a consent the visitor can no longer see is never submitted.

Also consolidates the trial licence form: the `emailOnly` / `original` variants were never rendered because no call site passed a `type` to the dispatcher, so the `allFields` implementation is promoted to `TrialLicenceForm` and the dispatcher plus `DEFAULT_TRIAL_LICENSE_FORM` / `TrialLicenceFormType` are removed. Includes the trial licence loading state from AG-14065.

## What was checked

- Every `ag-website-shared` file touched here was byte-identical across ag-grid, ag-charts and ag-studio before the change, so overwriting them loses no repo-specific work
- `external/ag-website-shared/src/constants.ts` legitimately differs between repos, so `CONSENT_FIELD_IDS` was inserted as a targeted edit rather than a file copy
- No dangling references to the removed variants or constants remain
- This repo's own Prettier config reports all touched files clean

## Not verified here

This branch has not had a `yarn install`, so the repo's own type-check, lint and dev server were not run against it. The component code is identical to what shipped in ag-grid, where it was verified in a browser across the contact page, the docs trial form and the trial modal, in light and dark mode.

## Needs action before the consent model goes live

1. `dataProcessingConsentId` is the same value in both environments (`00NQ500000IfxTS`). Every other sandbox ID is `00NS9…` and every production ID is `00NQ5…`, so the sandbox entry looks like a production ID in the wrong block — which would mean data processing consent silently fails to land in UAT.
2. The trial licence form's `CreateLeadForTrialLK` Cloud Function must accept the four new payload fields (`dataProcessingConsent`, `marketingEmailConsent`, `emailTrackingConsent`, `franceOrItaly`). If it validates input strictly, the extra fields could break *all* trial licence requests.
---

## AG-17996 — related PRs

| Repo | Branch | PR |
| --- | --- | --- |
| ag-grid | `latest` | ag-grid/ag-grid#14646 (merged) |
| ag-grid | `latest` | ag-grid/ag-grid#14660 — QA TC1 fix |
| ag-grid | `b36.0.2` — production | ag-grid/ag-grid#14658 |
| ag-charts | `b14.0.2` — production | ag-grid/ag-charts#7610 |
| ag-charts | `b14.1.0` — next release | ag-grid/ag-charts#7611 (merged) |
| ag-charts | `b14.1.0` | ag-grid/ag-charts#7619 — QA TC1 fix |
| ag-studio | `latest` | ag-grid/ag-studio#2225 |
| ag-studio | `b2.0.1` — production | ag-grid/ag-studio#2226 |

The QA TC1 fix (contact form errors clearing as soon as a field is corrected) is included as an extra commit on the four still-open port branches — #14658, #7610, #2225 and #2226 — and as its own PR where the port had already merged: #7619 for `b14.1.0` and #14660 for `latest`.

The ag-charts changes are branched off the release branches directly and will reach `latest` after release.


alantreadway (approved):
LGTM

taktran:
Thanks — all three findings were real. Addressed as follows.

**P1 — Studio trial submissions bypass consent / lost privacy disclosure.** Confirmed and fixed. The privacy line was removed from `TrialLicenceFormStudio.tsx` without that form gaining the consent controls that justified removing it from the others, so it was collecting name and email with no disclosure at all. It now gets the same treatment as `TrialLicenceForm`: required data processing consent, optional marketing consent, and a France/Italy declaration revealing optional email tracking consent, with all four sent in its JSON payload.

One correction to the review's framing: this form only renders on **ag-studio** (`packages/ag-studio-docs/src/components/pricing-form/PricingForm.tsx`). In ag-grid and ag-charts it is dead code, so the live regression was limited to the Studio site.

**P2 — consent errors not exposed to assistive technology.** Fixed. The checkbox now gets `aria-invalid` and `aria-describedby` pointing at the message, which carries a matching `id` and `role="alert"`. Verified in a browser: both attributes appear on a failed submit, `aria-describedby` resolves to the real error element, and both clear the instant the box is ticked.

**P2 — stale `submitUrl` / missing `Props` generic.** Both fixed (they were pre-existing rather than introduced here). `submitUrl` is now in the submit callback dependencies, and both trial forms are typed `FunctionComponent<Props>` so consumers can type-check the prop.

Also included: the sandbox consent field IDs now use a consistent `00NS900000Kbw*` set. Previously `dataProcessingConsentId` held a production `00NQ5*` ID, so data processing consent would not have landed in UAT.

Source of these changes is ag-grid/ag-grid#14660; #7611 is already merged, so the equivalent changes land on `b14.1.0` via ag-grid/ag-charts#7619.

Still outstanding and **not** something this PR can fix: `CreateLeadForTrialLK` must be updated to accept the four new payload fields, now including for the Studio form. If it validates input strictly, the extra fields could break trial licence requests.

PR #7615, opened 2026-07-29
CRT-1186 Revert error bar opacity on data selection change

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-1186

Fix #CRT-1186


lsjroberts (approved):
LGTM

PR #7616, opened 2026-07-29
CRT-1167 Keep the overlay content centred after a hide and re-show

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-1167

Fix #CRT-1167

### DIFF
--- a/.rulesync/rules/dom-performance.md
+++ b/.rulesync/rules/dom-performance.md
@@ -31,7 +31,7 @@ Use `DOMElementProxy` (from `ag-charts-community/src/dom/domElementProxy.ts`) as
 -   **`setInnerHTML(html): boolean`** — cached `innerHTML` write. Returns `true` if the write happened. Auto-invalidates `contentStyles` so that the next `setContentStyles` call re-applies styles to the new child elements.
 -   **`setContentStyles(styles)`** — cached style assignment to the first child element (or the container itself if no children). Compared via `JSON.stringify`; skipped if unchanged. Auto-invalidated by `setInnerHTML` — callers no longer need manual `invalidate()` calls.
 -   **`innerHTML`** — getter reading from DOM. Use for read-side checks (e.g., `dom.innerHTML === ''`).
--   **`appendChild(child)`** — delegates to `element.appendChild()`. No caching — structural mutation.
+-   **`replaceChildren(...children)`** — delegates to `element.replaceChildren()`. No caching — structural mutation. Invalidates the `innerHTML` cache key.
 -   **`set innerText(text)`** — delegates to setting `element.innerText`. Invalidates the `innerHTML` cache key since setting innerText changes DOM content.
 -   **`togglePopover(force)`** — delegates to `element.togglePopover()`. Not cached (side effects beyond attribute state).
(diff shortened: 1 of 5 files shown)


alantreadway:
## Summary

The `noVisibleSeries` / `noData` / `loading` overlay text renders centred on its first appearance, but sits ~18px too low on every appearance after that.

`hideOverlay()` writes a no-break space into the overlay container as its cleanup step — the AG-11424 workaround that forces Chromium to re-announce the `aria-live` region. `showOverlay()` then *appended* the content div after that text node. Because `.ag-charts-canvas-overlay > *` is `position: absolute` with no `top` (`src/dom/container.css`), the content resolves to its **static position**, which the preceding in-flow text node pushes down by exactly one line box. The content div's own `height: 100%` + `alignItems: center` then centres the text inside a box that already starts a line too low.

Pre-existing since June 2024 (the NBSP workaround), not a regression in this cycle — the two halves of the mechanism are byte-identical in `b14.0.0`.

## Fix

`showOverlay()` installs the content with `replaceChildren(element)` instead of `appendChild(element)`, so the placeholder is cleared exactly as the new content lands.

`hideOverlay()` is deliberately untouched: the NBSP is still written and still present for the whole hidden interval, which is when the `aria-live` change needs to be detectable. **AG-11424 is preserved.**

The clear has to happen *after* `overlay.getElement()`, not before it — `getElement()` calls `animationManager.animate({ id: 'overlay' })`, whose `checkOverlappingId` synchronously stops any pending remove animation and fires its `onStop`, which writes the NBSP. Clearing first would be undone.

`DOMElementProxy.appendChild` becomes `replaceChildren(...children)` (this was its only call site) and now invalidates the `innerHTML` cache key, since it mutates DOM content.

## Tests

Both were verified anti-vacuous by reverting the fix and watching them fail.

- `overlay.test.ts` → `Overlay > #hide and re-show`: asserts the container holds exactly one child after a hide → re-show cycle, and that the NBSP is still written while hidden. Reverted: 2 children.
- `ag-charts-website/e2e/overlays.spec.ts` → `overlay position`: real-browser bounding-box `y` must match across the cycle. Reverted: fails by exactly 18px, reproducing the reported symptom.

The absence of any hide → re-show coverage in the overlay suite is why this survived from June 2024. jsdom does not implement `innerText`, so the geometry is only observable in the E2E test.

## Verification

`nx format`, `build:types`, `lint` (community + website) clean. `test ag-charts-community` 4098 passed, `test ag-charts-enterprise` 2683 passed, `test:e2e ag-charts-website` (overlay grep) 126 passed.

## Reviewer notes

- Targets `b14.1.0` per the triage verdict (fix in release; narrow scope, low risk).
- AG-11424 has no automated coverage, so a manual Chromium screen-reader pass on the hide-all → show → hide sequence is still worth doing.
- **Adjacent, out of scope:** a custom-`renderer` overlay takes no 'add' animation, so a pending remove animation's `onStop` can write the NBSP *after* new content is installed, wiping it. Unreported; worth its own ticket.

Fixes [CRT-1167](https://ag-grid.atlassian.net/browse/CRT-1167)

ag-alantreadway (approved):

PR #7619, opened 2026-07-29
AG-17996 - Clear contact form errors as soon as a field is corrected (b14.1.0)

taktran:
Fixes QA **TC1** from AG-17996: tabbing past the unchecked data processing consent checkbox showed the required-consent error, and then ticking the box left the error on screen until focus moved away again.

## Cause

`ContactForm` uses react-hook-form with `mode: 'onBlur'`. That mode validates on blur but does **not** re-validate on change before the first submit, so the error state went stale as soon as the value changed — hence "the error goes away after tabbing".

## Fix

One line: `mode: 'onTouched'`. Same first-validation-on-blur behaviour, plus re-validation on change, so a message clears the moment the field is corrected.

## Verified

Reproduced TC1 with real keyboard events, not synthesised ones:

1. Click Message, type `aaa`, TAB TAB → error shown under the unchecked first checkbox (bug state reproduced)
2. SHIFT+TAB, SPACE → checkbox ticked, **error gone immediately** ✅

Checked for regressions, since `mode` affects every field on the form:

- Typing an invalid email **before** any blur still shows no error — errors do not fire while typing into an untouched field
- Blurring an invalid email shows "Enter a valid email"; correcting it now clears immediately rather than waiting for the next blur. This is the same class of bug as TC1, now also fixed for the text fields
- Submit still blocks on every required field, consent included

Format and lint pass.

The trial licence form is not affected — it derives its errors from live state on every render behind the `wasValidated` gate, so its consent error already cleared instantly.

One deviation from the ticket's steps worth noting: the second TAB lands on the "Privacy Policy" link inside the label rather than the second checkbox, because the link is in the tab order. The error state described in TC1 is reached either way.

---

Follow-up to #7611, which is already merged into `b14.1.0` — hence a separate PR rather than an extra commit on that branch. The equivalent fix is added directly to the still-open port PRs on the other branches.
---

## AG-17996 — related PRs

| Repo | Branch | PR |
| --- | --- | --- |
| ag-grid | `latest` | ag-grid/ag-grid#14646 (merged) |
| ag-grid | `latest` | ag-grid/ag-grid#14660 — QA TC1 fix |
| ag-grid | `b36.0.2` — production | ag-grid/ag-grid#14658 |
| ag-charts | `b14.0.2` — production | ag-grid/ag-charts#7610 |
| ag-charts | `b14.1.0` — next release | ag-grid/ag-charts#7611 (merged) |
| ag-charts | `b14.1.0` | ag-grid/ag-charts#7619 — QA TC1 fix |
| ag-studio | `latest` | ag-grid/ag-studio#2225 |
| ag-studio | `b2.0.1` — production | ag-grid/ag-studio#2226 |

The QA TC1 fix (contact form errors clearing as soon as a field is corrected) is included as an extra commit on the four still-open port branches — #14658, #7610, #2225 and #2226 — and as its own PR where the port had already merged: #7619 for `b14.1.0` and #14660 for `latest`.

The ag-charts changes are branched off the release branches directly and will reach `latest` after release.

### DIFF
--- a/external/ag-website-shared/src/components/consent-fields/ConsentCheckbox.tsx
+++ b/external/ag-website-shared/src/components/consent-fields/ConsentCheckbox.tsx
@@ -14,13 +14,27 @@ interface Props {
 }
 
 export const ConsentCheckbox: FunctionComponent<Props> = ({ id, label, error, nested, inputProps }: Props) => {
+    const errorId = `${id}-error`;
+
     return (
         <div className={classnames('input-field', { 'input-error': error }, { [styles.nested]: nested })}>
             <label className={styles.consentLabel} htmlFor={id}>
-                <input type="checkbox" id={id} {...inputProps} />
+                <input
+                    type="checkbox"
+                    id={id}
+                    aria-invalid={error ? true : undefined}
+                    aria-describedby={error ? errorId : undefined}
+                    {...inputProps}
+                />
                 <span>{label}</span>
             </label>
-            <div className={styles.errorContainer}>{error && <p className="error">{error}</p>}</div>
+            <div className={styles.errorContainer}>
+                {error && (
+                    <p className="error" id={errorId} role="alert">
+                        {error}
+                    </p>
+                )}
+            </div>
         </div>
     );
 };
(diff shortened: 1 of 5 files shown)


alantreadway (approved):
LGTM

skewis6 (approved):

PR #7626, opened 2026-07-29
Rulesync: fail on 'paths:' scoping, plus animation-trajectory rule updates

alantreadway:
Agentic tooling only — no product code.

- `verify-rulesync.sh` gains a source-frontmatter check: it **errors** when a `.rulesync/rules/*.md` declares its scope under `paths:` (rulesync accepts `globs:` and *emits* `paths:`, so the scope was silently dropped and the rule loaded in every conversation), and **warns** when a rule declares neither `globs:` nor `alwaysApply: true`. Content verification strips frontmatter before comparing, so this class of defect was structurally invisible to it.
- Fixes the three ag-charts rules that carried the defect: `api-contracts`, `entry-point-hygiene`, `module-support` (~172 lines / ~9 KB of context per conversation, unconditionally).
- Adds SYNC-LOG entry with the migration actions for consuming repos.
- Records animation-trajectory-test learnings in the rule and the migration skill (degenerate x-extents, non-finite values not visible in probe dumps, `monotonicTol` for radian properties, flat sampler keys, whole-scene sampling, the `spyOnAnimationManager`/`spyOnAnimationFrames` scope clash).
- Notes in the project-overview rule and `AGENTS.md` that `/ag-eng:git-conventions` is authoritative for branch/commit/PR format.

`verify-rulesync.sh` passes on this branch.


AG-DavidG (approved):
LGTM

PR #7632, opened 2026-07-29
Fix series-labels example interactivity and layout bugs

AG-DavidG:
Fixes interactivity and layout bugs found in the series-labels documentation examples (packages/ag-charts-website/src/content/docs/series-labels/):

- label-position: switching between bar/bar-horizontal/bubble no longer resets the placement to a hardcoded default when the dropdown already reflects a different choice
- label-orientation: defaults to plain `horizontal` (matching label-position's pattern) instead of a fallback array; data tuned so both fallback directions are demonstrable via the dropdown
- label-showcase: narrow stacked segment now sits between the other two; added a third placement fallback so its label doesn't hide at the container width used on the actual docs page
- label-threshold: widened the collision threshold slider range

https://ag-grid.atlassian.net/browse/CRT-1188

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,10 +43,6 @@ env:
     BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
     AG_LIBRARY: charts
     AG_SKIP_NATIVE_DEP_VERSION_CHECK: 1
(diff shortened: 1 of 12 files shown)


alantreadway (approved):
LGTM

PR #7643, opened 2026-07-30
CRT-1190 Exclude org-chart expander from click-selection

lsjroberts:
https://ag-grid.atlassian.net/browse/CRT-1190


olegat (approved):
LGTM

PR #7653, opened 2026-07-30
Add theme-override provenance test harness

alantreadway:
Enumerates every themeable option from the `Ag*ThemeableOptions` interfaces and asserts that a value supplied through `theme.overrides` resolves identically to the same value supplied in user options, that styling alone does not activate features, and that an explicit `enabled` is always honoured. 23 series types across community and enterprise, ~3,700 comparisons.

Depends on #7649 (targeting `b14.1.0`), whose narrowed `$isUserOption` behaviour these tests assert. **CI here will fail until that fix reaches `latest`** — please merge this only afterwards.

Context: https://ag-grid.atlassian.net/browse/CRT-1195


iMoses (approved):
LGTM

PR #7669, opened 2026-08-03
CRT-1198 Stop the product dropdown being cut off at smaller widths

skewis6:
https://ag-grid.atlassian.net/browse/CRT-1198

Fix #CRT-1198

The Products popup is a fixed `width: 800px` anchored with `left: 0` to the
switcher, which sits in the header's right-hand cluster — so it overhung the
viewport at every width the switcher renders at (75px clipped at 1120, still
5px at 1920), not just small ones.

It now anchors to `.headerInner` and right-aligns to the header's content
column, so its trailing edge lines up with the last nav item and it cannot be
clipped. `top` comes from the header row height because `.headerInner` is
multi-row in the wrapped states below `$docs-search-inline`; the visible effect
is that the popup hangs flush off the header instead of overlapping it by 15px.

Shared component — will need syncing to ag-grid and ag-studio, which both
override `$nav-collapse` / `$docs-search-inline`.


taktran (approved):

markdurrant (approved):
LGTM 👍

PR #7671, opened 2026-08-03
Fix crash when setting legend.pagination.marker.padding

alantreadway:
`legend.pagination.marker.padding` was validated by two separate schemas that had drifted: the copy in `commonChartOptionsDefs` (`ag-charts-core`) used `positiveNumber`, while the `legend` plugin module used `padding`. Both passes run — `validateChart` first, then `validatePluginOptions` — so the stricter copy won and cleared the value, leaving the option `undefined`. `Pagination.updateLabelPosition` then threw `TypeError: Cannot read properties of undefined (reading 'right')` once the legend paginated.

Both documented forms were affected, not just the object form: `OptionsGraph` forces `$applyPadding` on any key named `padding` even when a user option is present, and grafts the expanded object back for re-validation, so a scalar `padding: 6` was rejected too. The option was effectively unusable, despite `ag-charts-types` declaring `padding?: Padding` and documenting both forms.

**Changes**

- Extract `legendOptionsDefs` and `gradientLegendOptionsDefs` as shared consts in `ag-charts-core/src/config/chartDefaults.ts`, referenced by both `commonChartOptionsDefs` and the two plugin modules, so one definition backs both validation passes.
- Resolve the drifted line to `padding` — the form the runtime consumer (`pagination.ts`) requires and the public type documents. The `gradientLegend` half is a pure dedupe (the two copies were byte-identical) with no behaviour change.
- Regression coverage for both the partial-object and scalar forms, verified to fail when the old validator is restored.
- Identity guards on both schema pairs, so re-inlining a copy fails rather than silently drifting again. The shared helper lives in community test space since enterprise can import community but not the reverse.

Net −54 lines.

**Not included**

`ranges` has the same duplicated-schema hazard, but has diverged structurally rather than by one line — the enterprise `rangesOptionsDefs` has `gap`, `buttons` and `...stylesOptions` that core's block lacks. Unifying it means deciding where enterprise-only option shapes should live, so it's left for a follow-up.

### DIFF
--- a/packages/ag-charts-community/src/chart/legend/legendModule.test.ts
+++ b/packages/ag-charts-community/src/chart/legend/legendModule.test.ts
@@ -1,0 +1,12 @@
+import { describe, it } from 'vitest';
+
+import { commonChartOptionsDefs } from 'ag-charts-core';
+
+import { expectSharedOptionsDefs } from '../test/sharedOptionsDefs';
+import { LegendModule } from './legendModule';
+
+describe('LegendModule', () => {
+    it('shares its options schema with commonChartOptionsDefs', () => {
+        expectSharedOptionsDefs('legend', LegendModule.options, commonChartOptionsDefs.legend);
+    });
+});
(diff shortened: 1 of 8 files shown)


iMoses (approved):
LGTM

PR #7672, opened 2026-08-03
Fix update-versions-data silently mislabelling every version

taktran:
`nx update-versions-data ag-charts-website` is currently destructive. Both archive
URLs it read are gone: `charts.ag-grid.com/archive` is a legacy directory listing
that stops at 10.1.0, and `/charts/archive` now redirects to
`/charts/documentation-archive/`. The site also rejects non-browser clients with a
403 regardless of path.

Both failures were swallowed. `getWebsiteVersions()` returned an empty list, and
because `noDocs` is derived from "version absent from the archive", every entry
got `noDocs: true` - which filters it out of `VersionsSelector` and the archive
`MajorTable`. Running the target on `b14.1.0` today empties both.

- Read `/charts/documentation-archive/` and match the absolute hrefs it actually
  serves (`href="https://charts.ag-grid.com/archive/10.0.0/documentation"`); the
  old regex expected an Apache directory listing.
- Collapse the two URLs into that one page. It is a strict superset: it lists 46
  versions (9.0.1 - 14.0.2) against the legacy listing's 11 (9.0.1 - 10.1.0), and
  every version the legacy listing had also appears there.
- Send a browser `User-Agent` so the request is not rejected.
- Throw on a non-OK response or an empty version list rather than writing a file
  derived from nothing.
- Exempt the newest entry from the `noDocs` check: the live release is served
  from the main docs site, so it is never in the archive and must not be flagged.
  Without this, `14.1.0` itself would be dropped from the selector.

`MIN_VERSION` is unchanged at 9.0.1, which still matches the earliest archived
version.

The second commit is the resulting data refresh, adding the nine released
versions missing from the list. Every flag was cross-checked against the live
archive: 14.0.2 and 14.0.1 are archived (no `noDocs`); 13.3.1, 11.2.4, 11.2.3,
10.3.9, 10.3.8, 10.3.7 and 10.3.6 are not (flagged). Nothing lost a flag it
should have kept, and `14.1.0` is untouched.

Same fix as ag-grid and ag-studio, which have the identical script.

### DIFF
--- a/packages/ag-charts-website/scripts/versions-data/updateVersionsData.ts
+++ b/packages/ag-charts-website/scripts/versions-data/updateVersionsData.ts
@@ -4,8 +4,16 @@ import * as fs from 'fs/promises';
 
 // Absolute path from the root directory
 const VERSIONS_DATA_PATH = 'packages/ag-charts-website/src/content/versions/ag-charts-versions.json';
-const WEBSITE_ARCHIVE_URLS = ['https://charts.ag-grid.com/archive', 'https://www.ag-grid.com/charts/archive'];
+// Supersedes the two URLs this previously read. The legacy charts.ag-grid.com/archive
+// listing only went up to 10.1.0, and every version it listed also appears here.
+const WEBSITE_ARCHIVE_URL = 'https://www.ag-grid.com/charts/documentation-archive/';
 const MIN_VERSION = '9.0.1';
+
+// The archive page is served behind a filter that rejects non-browser clients with
+// a 403, so identify as one. Without this every request fails and the script would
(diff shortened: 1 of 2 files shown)


taktran:
Closing in favour of #7674.

This fixed how the archive pages were fetched but kept the underlying mistake: `/charts/documentation-archive/` is generated from the very JSON file the script writes (`MajorTable` filters entries by `!entry.noDocs`), so deriving `noDocs` from it is circular and self-reinforcing - once a version is flagged it drops off the page and stays flagged forever.

#7674 asks the server whether `…/archive/<version>/` exists instead.

PR #7681, opened 2026-08-04
CRT-1200 Set style nonce on dynamically injected style elements

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-1200

Fix #CRT-1200


packages/ag-charts-community/src/chart/overlay/loadingSpinner.ts
```diff
@@ -43,6 +48,9 @@ export function getLoadingSpinner(agDocument: AgDocument, text: string, defaultD
     });
 
     const animationStyles = agDocument.createElement('style');
+    if (styleNonce != null) {
+        animationStyles.nonce = styleNonce;
+    }
```
lsjroberts: It may help prevent this in the future if we have a `createStyleElement(nonce: string | undefined)` function? So we know to pass it through.

lsjroberts (approved):
LGTM, one minor comment.

packages/ag-charts-community/src/chart/overlay/loadingSpinner.ts
alantreadway: Fixed

PR #7706, opened 2026-08-05
AG-14332 Phase 1: Refactor how `Series` creates click events

olegat:

Refactor `seriesNodeClick` / `seriesNodeDoubleClick` in preparation for the `allNodeParams` implementation.

Currently, each concrete/derived series class is responsible for allocating the event-object for clicks. This isn't going to work well on multi-node clicks. You could for example be clicking a marker (line-series) & bar (range-bar series) on a combo chart.

Therefore, each derived series is now only responsible for allocating the datum "parameters", whereas the base `Series` class owns event-object allocations:

- `Series.createNodeParams` - abstract: implemented by concrete/derived series classes.
- `Series.createNodeEvent` - final/non-overridable: allocates event by calling `createNodeParams`.

Phase 2 will call `createNodeParams` in a loop to implement `allNodeParams`.


alantreadway (approved):
LGTM

PR #7711, opened 2026-08-05
Resolve low-risk SonarCloud code smells

alantreadway:
Clears 26 open SonarCloud issues across `ag-charts-core`, `ag-charts-community` and `ag-charts-enterprise`. Every change is behaviour-preserving; no JIRA ticket, this is post-release static-analysis cleanup from the 14.1.0 report.

Rules addressed:

| Rule | Count | Change |
|---|---|---|
| S7719 | 6 | clone dates via `new Date(date)` rather than `.getTime()` |
| S6598 | 4 | call-signature type literals to function types |
| S7780 | 3 | `String.raw` for the mermaid diagram templates |
| S7770 | 3 | pass `Number` / `Boolean` directly instead of wrapping in an arrow |
| S4782 | 3 | drop redundant `?` + `\| undefined` |
| S7752 | 2 | `flatMap` instead of `map().flat()` |
| S7754 | 2 | `some` instead of `filter().length > 0` |
| S7744 | 2 | drop no-op `?? {}` in spreads |
| S7771 | 1 | `at(-1)` instead of `at(length - 1)` |

Deliberately **not** included, having been assessed as false positives or conflicts with existing project decisions:

- **S3863** (58, duplicate imports) — every hit is a value import plus a separate `import type` from the same module, which `eslint.config.mjs` explicitly permits via `no-duplicate-imports: [error, { allowSeparateTypeImports: true }]`. Sonar flags 58 of ~242 identical instances repo-wide, so fixing the flagged subset would be inconsistent churn.
- **S6564** (15, redundant type aliases) — all intentional: semantic aliases over primitives (`ScopeId = string`), public API naming indirection, and function-local shorthand for long generic types.
- **S7741** (3, `typeof` vs `undefined`) — the flagged guards are environment checks on `globalThis.window` / `globalThis.global`. TypeScript types these as always-defined, so `!== undefined` is statically always-true and trades the Sonar issue for a new `sonarjs/different-types-comparison` lint warning.
- **S6679** (1, `Number.isNaN`) — `aggregation.ts` documents `x !== x` as a deliberate hot-loop idiom at lines 442 and 587.
- **S7758** (1, `codePointAt`) — the function parses fixed-width ASCII digits, where `charCodeAt` is correct and faster.
- **S4782** (12 of 15) — the remaining sites inherit `undefined` indirectly through public type aliases; changing them needs either an API-visible `?` removal or `NonNullable<>` noise.

Verification: `nx format`, `build:types` for all three packages, `lint` with no net-new warnings against the 691-warning baseline, and the full community (4164 tests) and enterprise (2729 tests) suites passing.

### DIFF
--- a/packages/ag-charts-community/src/chart/axis/axisLabelUtil.ts
+++ b/packages/ag-charts-community/src/chart/axis/axisLabelUtil.ts
@@ -43,7 +43,7 @@ export function formatAxisLabelValue(
         | (Pick<AgBaseAxisLabelStyleOptions, never> & {
               formatter?: AgAxisLabelFormatterParams extends never
                   ? never
-                  : ((params: AgAxisLabelFormatterParams) => NormalisedTextOrSegments | undefined) | undefined;
+                  : (params: AgAxisLabelFormatterParams) => NormalisedTextOrSegments | undefined;
               format?: string | Record<string, string>;
           })
         | undefined,
--- a/packages/ag-charts-community/src/chart/chart.ts
+++ b/packages/ag-charts-community/src/chart/chart.ts
@@ -150,7 +150,7 @@ function deriveMiniChartInterval(
     intervalOverride: Record<string, unknown> | undefined,
     sourceInterval: object | undefined
 ): Record<string, unknown> {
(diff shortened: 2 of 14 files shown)


lsjroberts (approved):
LGTM

PR #7713, opened 2026-08-05
AG-3390 syncing website-shared

skewis6:


taktran (approved):
Something for later, but I'm wondering if we should move `theme-builder-studio` out of `ag-website-shared`, since it might add noise to future syncing when no other website uses it. Same argument could be made for studio contact form

PR #7715, opened 2026-08-05
AG-18007 Resolve border.enabled independently per label placement

iMoses:
https://ag-grid.atlassian.net/browse/AG-18007

Fix #AG-18007


alantreadway (approved):
LGTM

PR #7719, opened 2026-08-06
Accept space-separated rgb() colours and reject malformed ones

alantreadway:
## The bug

`Color.stringToRgba` split components on commas only, and never checked how many components it ended up with. Two consequences:

1. **The modern CSS syntax does not parse.** `rgb(255 0 0)` splits into a single part, `Number.parseFloat` reads `255` off the front, and the result is a one-element array. The slash-separated alpha form, `rgb(255 0 0 / 0.4)`, fails the same way. The sibling parsers `stringToHsla` and `stringToOklcha` already handle both syntaxes — `rgb()` was left behind.

2. **`validColorString` disagrees with `fromString`.** Because any arity was returned unchecked, `validColorString('rgb(255 0 0)')` returned `true` while `Color.fromString('rgb(255 0 0)')` threw `Malformed rgb/rgba color string`. `validColorString` backs option validation (`optionsDefaults.ts`, `optionsModule.ts`), so a chart configured with `fill: 'rgb(255 0 0)'` passed validation and then threw when the colour was resolved for rendering. Short arities like `rgb(120, 240)` behaved identically.

## The fix

`stringToRgba` now mirrors `stringToHsla` / `stringToOklcha`: split the optional `/ alpha` off first, split the head on whitespace and/or commas, and require exactly three or four components — anything else returns `undefined`, so the validator rejects it instead of leaving it to throw at render time.

Component coercion is factored out into `parseRgbComponent` (percentage relative to 100%, bare number relative to 255, clamped to `[0, 1]`), and alpha reuses the existing shared `parseUnitInterval`. Clamping behaviour for the comma-separated form is unchanged — `rgb(-23, 255, 300)` still yields `0, 1, 1`, and `rgb(100, 25%, 200, 50%)` still yields alpha `0.5`.

## Testing

Two tests added to `color.test.ts`: one covering the space-separated and slash-alpha forms, one asserting `validColorString` and `fromString` agree on both well-formed and malformed `rgb()` strings.

- `yarn nx format`, `yarn nx build:types ag-charts-core`, `yarn nx lint ag-charts-core` — clean
- `yarn nx test ag-charts-core` — 856 passed
- `yarn nx test ag-charts-community` — 4167 passed
- `yarn nx test ag-charts-enterprise` — 2732 passed

No existing test needed changing.

---
_Generated by [Claude Code](https://claude.ai/code/session_012VfCBX5cuzDCMg6utKhQ84)_

### DIFF
--- a/packages/ag-charts-core/src/utils/format/color.test.ts
+++ b/packages/ag-charts-core/src/utils/format/color.test.ts
@@ -180,6 +180,53 @@ describe('Color', () => {
         }
     });
 
+    test('fromRgbaString accepts the space-separated syntax', () => {
+        {
+            const color = Color.fromRgbaString('rgb(120 240 100)');
+            expect(color.r).toBe(120 / 255);
+            expect(color.g).toBe(240 / 255);
+            expect(color.b).toBe(100 / 255);
+            expect(color.a).toBe(1);
+        }
+        {
+            const color = Color.fromRgbaString('rgb(120 240 100 / 0.4)');
+            expect(color.r).toBe(120 / 255);
+            expect(color.g).toBe(240 / 255);
+            expect(color.b).toBe(100 / 255);
+            expect(color.a).toBe(0.4);
+        }
+        {
+            const color = Color.fromRgbaString('rgb(50% 25% 75% / 50%)');
+            expect(color.r).toBe(0.5);
+            expect(color.g).toBe(0.25);
+            expect(color.b).toBe(0.75);
+            expect(color.a).toBe(0.5);
+        }
+        {
+            // The comma-separated syntax also accepts a slash-separated alpha.
+            const color = Color.fromRgbaString('rgb(120, 240, 100 / 0.4)');
+            expect(color.a).toBe(0.4);
+        }
+    });
+
+    test('validColorString agrees with fromString for rgb', () => {
+        for (const str of [
+            'rgb(120, 240, 100)',
+            'rgb(120 240 100)',
+            'rgba(120, 240, 100, 0.4)',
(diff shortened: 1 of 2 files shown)


eeshsaxena (commented):
Nice cleanup, the count check fixes the obvious cases. One malformed shape still slips through though, which is worth catching since rejecting bad input is the point of the PR.

When the alpha is written with a slash but the colour part only has two components, the slashed alpha gets pushed on as a third part and the whole thing is accepted as a valid rgb():

```js
parseColorFunctionParts("rgb(120 240 / 0.4)") // -> ["120", "240", "0.4"]
parseColorFunctionParts("rgb(255 0 / 0.5)")   // -> ["255", "0", "0.5"]
```

So `rgb(120 240 / 0.4)` parses to `rgb(120, 240, 0.4)` (the alpha value becomes the blue channel, ~0.0016 after /255) and `validColorString` returns true, when it should be rejected as missing a channel. `rgb(120, 240 / 0.4)` behaves the same way with commas.

The reason is that `parts.length === 3 || parts.length === 4` can't tell "3 real components, no slash" apart from "2 components plus a slashed alpha". When a slash is present the alpha is always a separate component, so the head has to contain exactly 3 values. Something like:

```js
const parts = head.trim().split(/[\s,]+/);
if (slash !== -1) {
    if (parts.length !== 3) return; // r g b are required before the slash
    parts.push(contents.substring(slash + 1).trim());
}
return parts.length === 3 || parts.length === 4 ? parts : undefined;
```

Might be worth a test case for `rgb(120 240 / 0.4)` in the reject list alongside the existing ones.

PR #7728, opened 2026-08-06
AG-14332 Phase 3: Add `winner` parameter to series event allocation

olegat:

Set `winner` to the current tooltip candidate when applicable.


alantreadway (approved):
LGTM

PR #7732, opened 2026-08-06
e2e: degrade on an unresolvable NX_BASE instead of failing collection

alantreadway:
Two consumer-side fixes from the 2026-08-06 AI Workflow watchdog sweep. The pipeline-side half is in [ag-dev-prompts](https://github.com/ag-grid/ag-dev-prompts) — this PR is only what has to live in this repo.

## 1. `test:e2e` failed at collection on an unresolvable `NX_BASE` (AG-18074)

`e2e/changed-examples.ts`'s header already promises the safe fallback:

> `NX_BASE` names the base ref to diff against. Without it no diff is computable, so callers treat every example as changed — the safe side for both the `affected` flag and framework scoping.

That held for an **absent** `NX_BASE` but not for an **unusable** one. The AI Workflow pipeline exports `NX_BASE=latest-success`, which does not resolve inside the Dockerised Playwright container — so `git diff` exited non-zero, `execSync` threw at module scope, and the whole `test:e2e` invocation died **before a single test ran**, taking `examples-snapshots.spec.ts` and `gallery-examples.spec.ts` with it. AG-18074 lost a complete e2e run to it.

An unusable base is the same situation as an absent one, so it now degrades the same way: `gitChangedFiles` returns `undefined` instead of throwing, and `getChangedExamples` propagates that — every example treated as changed, which is the conservative direction. The rationale is recorded in the function's doc comment so the next reader doesn't "tidy" the try/catch away.

`.rulesync/rules/website-e2e-testing.md` gains the matching `NX_BASE` section. Note that's the **tracked source** — `.claude/rules/` is generated and gitignored here, so editing the generated copy would have looked right locally and been absent from the commit.

## 2. Gitignore `/.ag-dev-prompts/`

Belt-and-braces alongside a `.git/info/exclude` entry landing in the pipeline PR. The pipeline clones its own prompt content into the workspace at `/.ag-dev-prompts/`, which is ignored in **neither** consumer repo — so it shows up as untracked in the agent's `git status` and, more consequentially, is a candidate for the rescue commit's `git add -A` on a run that dies mid-flight. The pipeline PR adds a rescue-path denylist for the same paths; this is the cheap local half.

## Verification

The e2e change was verified from `packages/ag-charts-website` (the pathspecs are package-relative, so running it from the repo root escapes the repo and makes every case look like a failure — worth knowing if you re-check it): a resolvable base returns file lists, an unresolvable one returns `undefined` on all three pathspecs, and an unset `NX_BASE` is unchanged. No behaviour change on CI, where `NX_BASE` resolves normally.


ag-alantreadway (approved):
LGTM

PR #7736, opened 2026-08-06
AG-17709 Discriminate union variants by their literal member

iMoses:
https://ag-grid.atlassian.net/browse/AG-17709

Fix #AG-17709


alantreadway (approved):
LGTM

PR #7742, opened 2026-08-07
AG-14332 Phase 5: Expose `defaultPrevented` in the public event API

olegat:


alantreadway (approved):
LGTM

PR #7743, opened 2026-08-07
AG-3390 Allow the Enzuzo cookie-consent banner in the site CSP

taktran:
Port of [ag-grid#14772](https://github.com/ag-grid/ag-grid/pull/14772) to the charts site.

## What

The Enzuzo banner replacing OneTrust loads from a tag in GTM container `GTM-T7JG534` — the same container the grid and studio sites use (`PUBLIC_GTM_ID` in each `.env.build.production`). Charts has `PRODUCTION_CSP_PHASE = 'enforce'`, so without the origins the loader is hard-blocked:

```
Loading the script 'https://app.enzuzo.com/scripts/cookiebar/061e8460-…'
violates the following Content Security Policy directive: "script-src 'self' …".
The action has been blocked.
```

## Changes

| Directive | Added | Why |
| --- | --- | --- |
| `script-src` | `https://app.enzuzo.com` | Serves the banner bundle `/scripts/cookiebar/<uuid>`. |
| `connect-src` | `https://app.enzuzo.com` | Same origin is the banner's `apiHost`; it XHRs config, cookie list and consent analytics from `/api/public/…`. |
| `connect-src` | `https://gvl.enzuzo.com` | IAB TCF Global Vendor List. Only fetched with TCF on, allowed up front so enabling it later is a console-only change. |

## OneTrust stays for now

`cdn.cookielaw.org` / `*.onetrust.com` are deliberately left in place. One GTM flip switches grid, charts and studio at once while their deploys land at different times, so both banners have to be loadable across that window. A test pins this so the entries aren't tidied away early. Removing OneTrust is a follow-up once Enzuzo is live everywhere.

## What deliberately isn't here

Derived by reading the served bundle rather than from vendor docs:

- **No SHA-256 hash.** GTM injects the banner as an external `<script src>`, not an inline snippet — contrast `GTM_ZOOMINFO_HASH`.
- **No `style-src` change.** The banner's CSS is injected as an inline `<style>`; `'unsafe-inline'` already covers it. It loads no webfonts.
- **No `img-src` / `frame-src` change.** The logo falls under `img-src https:`, and the only iframe it creates is the src-less `__tcfapiLocator`.
- **No `unsafe-eval`.** See below.
- The TCF library itself is a dynamic import from `cdn.jsdelivr.net`, already allowed.

## ⚠️ One thing to watch in the Enzuzo console

The bundle has three `new Function` call sites, all of which throw without `'unsafe-eval'`:

1. `fillTemplate` — template placeholders in banner text. Caught, degrades to empty string.
2. "Display fields" evaluation. Caught, logs an error.
3. String-valued `onConsent` handlers on integrations — **not** caught.

So the Enzuzo config needs to avoid template placeholders in banner copy and string-bodied event handlers. Noted in a comment above the new constants.

## Testing

- `yarn nx run ag-charts-website:test:vitest:update` — 539 tests pass, including four new CSP cases; 2 htaccess snapshots updated.
- `yarn nx lint ag-charts-website` — passes.
- `yarn nx run ag-charts-website:generate-csp` — emitted policy carries `app.enzuzo.com` in both scoped policies and `gvl.enzuzo.com` in their `connect-src`.

## Related

- ag-grid#14772 (`latest`) and ag-grid#14775 (`b36.1.0`)
- The `b14.1.0` port and the two studio PRs follow.


taktran:
~~To be released next week~~ (not necessary)

seanlandsman (approved):

PR #7772, opened 2026-08-11
Serve snapshot-review reports from a plain branch, not gh-pages

alantreadway:
Snapshot-review reports were published to `gh-pages`, so a reviewer could not open one until Pages rebuilt and redeployed the whole site — measured at 163–706s on this repo. Pages branch builds are single-flight, so a later push cancels the in-flight build and a report could miss its own CI run entirely. That is what produced the false "Report publish failed" comments.

CI now pushes reports to a `snapshot-review` branch, which has no Pages build attached: the content is readable via `raw.githubusercontent.com` the moment the push lands (~0.4s measured). The static viewer already published to `gh-pages` under `snapshot-viewer/` renders them, since raw serves `text/plain`.

### What this removes

- `Wait for published reports` and `Check for in-flight Pages deploys` steps
- the per-commit deploy marker they were gated on
- the three-state pending/live/failed link logic they fed — a successful push *is* the live state, so two states remain

### Links name a commit, not a branch

A branch URL is served with a cache TTL and its content moves under a link already shared into a PR comment. `pr-preview-publish` gains a `commit_sha` output, and `report_base_url` carries a `__SHA__` placeholder that `Finalize` substitutes once the push has landed.

### Retention

- `pr-cleanup.yml` gains a job removing `pr-<N>/` from the new branch on PR close (own concurrency group — it races `snapshot_review`, not the gh-pages publisher)
- the janitor sweeps both branches; renamed `gh-pages-janitor.yml` → `preview-branch-janitor.yml` to match
- the janitor now treats a not-yet-created publish branch as a clean no-op rather than a failed run, since a branch is created lazily by its first publish

### gh-pages seams deliberately left alone

- `pr_preview` (UMD bundles) still publishes to `gh-pages` and keeps the `pr-preview-gh-pages-<N>` lock
- `jira-agent-pipeline.yml` bundle URLs are gh-pages preview URLs, unaffected
- `gh-pages-snapshot-viewer.yml` still publishes the viewer to `gh-pages` — it is a static page that benefits from being served as `text/html`

### Notes

- pushes to `snapshot-review` do not trigger CI: `ci.yml`'s push trigger only matches `latest`, `next` and `b*` branches
- `gh-pages-commit.sh`'s sync path writes a zero-byte `.nojekyll` to the new branch too; harmless, and not worth branching the script for

### Verified end-to-end

Tested on this PR by temporarily perturbing four committed baselines across the range the report ranks over (4 px, 249 px, 1,444 px, 263,466 px). That commit has been reverted; the branch now carries only the publishing change.

What the run produced:

- `snapshot-review` created as a true orphan (`parents=0`), holding `pr-7772/snapshot-current.html` and `pr-7772/snapshot-regenerated.html` at ~530 KB each
- both links in the PR comment resolved to the publishing commit rather than the branch — e.g. `…viewer.html?repo=ag-grid/ag-charts&sha=20d4af4b6f…&path=pr-7772/snapshot-current.html`
- both rendered in the viewer, including the regenerated-baselines report

The `pr-7772/` directory is still on the branch and its links still work, so the mechanism can be inspected during review. Close-time cleanup reclaims it.


iMoses (approved):
LGTM

PR #7779, opened 2026-08-11
Fix negative-value currency formatting in series-labels examples

AG-DavidG:
Three series-labels docs examples format negative bar values incorrectly, and none had a tooltip matching the label's `$` formatting:

- `label-orientation`: label formatter interpolated `params.value` after a leading `$`, so a negative value's own minus sign landed after the dollar sign (`$-5m profit` instead of `-$5m profit`).
- `label-position`: same bug in its bar-series label formatter (`$-5m` instead of `-$5m`), visible when switching to the Bar/Horizontal Bar series (Q3/Q5 have negative `profitChange` values).
- `label-fitting` (the "Collision Avoidance" section example): same formatter pattern; fixed for consistency even though its current dataset has no negative values.

Fixed all three formatters to place the sign before the `$`, and added a `tooltip.renderer` to each bar series so the tooltip matches the label's `$`/sign formatting instead of showing the raw unformatted value. The tooltip renderers omit `title` since the chart already supplies the category heading automatically.

Verified in the browser (dev server) that labels and tooltips both correctly show `-$5m` for negative bars, and that the `label-fitting` tooltip reads cleanly without a redundant title row.

### DIFF
--- a/packages/ag-charts-website/src/content/docs/series-labels/_examples/label-fitting/main.ts
+++ b/packages/ag-charts-website/src/content/docs/series-labels/_examples/label-fitting/main.ts
@@ -6,11 +6,6 @@ import { DataType, data } from './data';
 
 ModuleRegistry.registerModules([BarSeriesModule, LegendModule, CategoryAxisModule, NumberAxisModule]);
 
-function formatCurrency(value: number) {
-    const sign = value < 0 ? '-' : '';
-    return `${sign}$${Math.abs(value)}m`;
-}
-
 const options: AgCartesianChartOptions<DataType> = {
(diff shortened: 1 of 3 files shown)


olegat (commented):
Examples `label-fitting` and `label-orientation` do not have negative datums, so that `formatCurrency` is not strictly necessary for them.

Example `label-position` does have two negatives in the bar example, although simply using `format: '$'` seems sufficient to render the negative sign in the correct position.
https://plnkr.co/edit/I4wmcHi1RS3nDTDx?open=main.js

alantreadway (approved):

PR #7786, opened 2026-08-12
AG-18116 Default inside label colour to chartBackgroundColor

iMoses:
https://ag-grid.atlassian.net/browse/AG-18116

Fix #AG-18116


alantreadway (approved):
LGTM

PR #7790, opened 2026-08-12
Fix negative-value currency formatting in series-labels examples (#7779)

alantreadway:
Cherry-pick of #7779 onto the b14.1.0 release branch.

The `label-position` example now uses the built-in `format: '$'` label option so
the negative sign is placed before the currency sign. Tooltips across the
`label-fitting`, `label-orientation` and `label-position` examples get matching
`$` formatting.


alantreadway:
Superseded by #7787, which is the same cherry-pick plus a follow-up preserving the millions suffix.

PR #7791, opened 2026-08-12
AG-9809 Add more axis-click tests

olegat:
https://ag-grid.atlassian.net/browse/AG-9809

Note: Some tests are skipped. I believe that these tests have correct expectations. Fixes for these will come separately.


alantreadway (approved):
LGTM

PR #7801, opened 2026-08-12
Fix nav entry for Development Validation docs page

AG-DavidG:
Fixes the docs nav entry left over from moving the dev-time validation overlay docs to their own page: the item was pointing at the old Overlays location and duplicated, instead of sitting under Development Validation's new location.


alantreadway (approved):
LGTM

PR #7804, opened 2026-08-13
AG-3390 Add staging and report-only configurations to the website preview:csp target

taktran:
Port of ag-grid#14821 to Charts.

## What

`preview:csp` only ever served the **enforced production** policy, so validating the staging policy — or serving either policy in report-only form — meant invoking the script by hand.

`preview-csp.ts` already parsed `--env` and `--mode`; this just exposes them as nx configurations, following the same `{"args": ...}` shape `generate-csp` already uses in the same file.

```bash
yarn nx run ag-charts-website:preview:csp                      # production, enforce (unchanged default)
yarn nx run ag-charts-website:preview:csp:staging              # staging policy, enforced
yarn nx run ag-charts-website:preview:csp:report-only          # production policy, report-only
yarn nx run ag-charts-website:preview:csp:staging-report-only
```

The environments genuinely differ — staging resolves to `us-central1-stripe-testing-19784.cloudfunctions.net` and `test.salesforce.com`, production to the live equivalents.

This is arguably worth more on Charts than on Grid: Charts **staging enforces** the `site` scope (`PRODUCTION_CSP_PHASE`/`htaccessRules.ts`), so there is no report-only grace period and a stale hash or stray inline script breaks staging immediately. Being able to serve the staging policy locally is the only way to catch that first.

## Policy only, not the build

The configurations select the **policy**, deliberately not the build. `dependsOn` still builds the default config, so `PUBLIC_SITE_URL` stays `https://localhost:4601`, `PUBLIC_BASE_URL` stays `/charts`, and the example runner keeps working.

A staging/production *build* served from localhost would be self-defeating: those configurations bake their remote origin into the output **and flip `PUBLIC_BASE_URL` to `/`**, so absolute-URL fetches leave localhost and the examples-scope `connect-src 'self'` blocks them — violations that are artefacts of the origin mismatch rather than of the policy. That stays an explicit two-step, documented in the usage comment along with the matching `--base`.

## Testing

- `yarn nx run ag-charts-website:generate-csp --configuration=staging|production` confirms the two envs emit different origins, so `--env` is meaningful.
- `nx show project ag-charts-website` lists all four configurations.
- `yarn nx lint ag-charts-website` passes; no behaviour change to the default target.

### DIFF
--- a/packages/ag-charts-website/project.json
+++ b/packages/ag-charts-website/project.json
@@ -88,6 +88,12 @@
       "command": "tsx scripts/csp/preview-csp.ts --port=${PORT}",
       "options": {
         "cwd": "packages/ag-charts-website"
+      },
+      "configurations": {
+        "staging": { "args": "--env staging" },
+        "production": { "args": "--env production" },
+        "report-only": { "args": "--mode report-only" },
+        "staging-report-only": { "args": "--env staging --mode report-only" }
       }
     },
     "lint:eslint": {},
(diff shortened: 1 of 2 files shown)


alantreadway (approved):
LGTM

PR #7806, opened 2026-08-13
AG-3390 Report CSP violations instead of failing page verification

taktran:
https://ag-grid.atlassian.net/browse/AG-3390

### DIFF
--- a/.github/workflows/post-deploy-verification.yml
+++ b/.github/workflows/post-deploy-verification.yml
@@ -45,9 +45,6 @@ jobs:
         name: Page Verification (Staging)
         outputs:
             prev_run_id: ${{ steps.prev_run.outputs.id }}
-            # The suite navigates via PUBLIC_SITE_URL rather than a Playwright baseURL, so the
-            # CSP report carries no site URL of its own for the reporting job to quote.
-            site_url: ${{ inputs.staging_url || steps.staging_url.outputs.url }}
         # For manual dispatch: always run.
         # For workflow_run: only proceed if triggered by a push to this repo (not a fork PR),
         # and only if the upstream CI run succeeded. This guards against the untrusted-checkout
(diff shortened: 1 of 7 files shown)


taktran:
Review findings addressed, with one declined.

**Fixed**

- Baselines are now only used when the previous report describes the same site, and a run against a hand-given `staging_url` is no longer uploaded as an artifact, so it cannot become the next run's baseline.
- Only a complete run can report a violation as resolved (`RUN_COMPLETE`), so a run that died part way through cannot announce that a policy was fixed.
- Inline violations are grouped by the hash they need, so two different blocked scripts no longer merge into one entry whose identity is the union of their hashes.
- Hash hints carry their disposition, so a report-only policy's hash is not offered as the fix for an enforced violation.
- A CSP console message with no hash is kept as an annotation instead of being dropped. The violation listener is installed on documents, so a worker's CSP failure reaches the console and nothing else.
- Hints are attributed to the frame that logged them (`msg.location()`), not the top-level page, so an iframe's violation is not answered with another document's hash.
- Slack block text and the resolved list are bounded, since Slack rejects the whole message past its per-block limit.
- The self-check asserts its own injected script's hash and that the script did not execute, so unrelated violations cannot satisfy it; its cleanup is narrowed to what its injected reload provoked and runs before the assertions, so a failure cannot leak the synthetic violation into the report.

**Declined: use `workflow_run.head_sha` for the checkout**

The existing `page-verification` job deliberately checks out `github.sha`, with a comment explaining why: referencing `workflow_run.head_sha` in a privileged `workflow_run` job is the untrusted-input pattern OpenSSF Scorecard flags. The new job follows the same rule for the same reason. The cost is that `COMMIT_SHA` can name the branch tip rather than the deployed commit if two pushes overlap, which mislabels a notification but changes nothing about what was measured.

alantreadway (approved):
LGTM

PR #7815, opened 2026-08-13
AG-3390 Take the CSP baseline from the last successful run

taktran:
https://ag-grid.atlassian.net/browse/AG-3390

### DIFF
--- a/.github/workflows/post-deploy-verification.yml
+++ b/.github/workflows/post-deploy-verification.yml
@@ -43,8 +43,6 @@ jobs:
     page-verification:
         runs-on: ubuntu-latest
         name: Page Verification (Staging)
-        outputs:
-            baseline_run_id: ${{ steps.prev_run.outputs.baseline_id }}
         # For manual dispatch: always run.
         # For workflow_run: only proceed if triggered by a push to this repo (not a fork PR),
         # and only if the upstream CI run succeeded. This guards against the untrusted-checkout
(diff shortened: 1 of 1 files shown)


taktran:
All three findings fixed, and the baseline lookup moved into the reporting job as a result.

**A successful run is not necessarily a run with a report** (raised on all three PRs). Correct, and my own change created the clearest case: a dispatch against a hand-given URL deliberately skips the upload, so it would displace an older valid baseline. The lookup now walks back over the ten most recent successful runs and takes the newest whose artifact listing actually contains a non-expired `csp-violations`, rather than trusting the first hit.

**A failed download was treated as absence.** That distinction is now unnecessary: by the time the download runs, the artifact is known to exist, so any failure is a real error and fails the job instead of silently reporting the whole site as new. The one tolerated case is an unreadable artifact listing for a candidate run, which emits a `::warning` naming the run it skipped rather than passing in silence.

**The lookup's exit status was masked by `echo`.** Right - `echo "baseline_id=$(gh run list ...)"` succeeds even when the query does not, so an auth or rate-limit failure produced an empty baseline and every violation was reported as new. The query is now assigned to a variable under `set -e`, so a failed lookup aborts the step. That also motivated moving the whole lookup out of the `page-verification` job into `csp-report`: a failing baseline query should not be able to redden the gating job. The shared previous-run step goes back to emitting only the `sha` and `outcome` the Slack step needs.

Verified before pushing: the search logic was run against the live API for all three repos and selected a run holding the artifact in each; forced to look for an artifact name that does not exist, it walks all ten candidates and reports no baseline; and pointed at an unreachable repo it exits non-zero rather than yielding an empty baseline. A dispatched dry run of the workflow on each branch is in flight to confirm the same end to end.

seanlandsman (approved):

PR #7830, opened 2026-08-14
ci(jira-agent-pipeline): one ticket, one concurrency group (AG-18177)

alantreadway:
Serialises this repo's AI Workflow pipeline jobs per JIRA ticket, so one ticket means one concurrency group.

Design, rationale and rollout: [AG-18177](https://ag-grid.atlassian.net/browse/AG-18177) (parent: [AG-18171](https://ag-grid.atlassian.net/browse/AG-18171)).

### DIFF
--- a/.github/workflows/jira-agent-pipeline.yml
+++ b/.github/workflows/jira-agent-pipeline.yml
@@ -142,14 +142,13 @@ on:
 permissions:
     contents: read
 
-concurrency:
-    # For workflow_run events the dispatch payloads are empty, so fall back
-    # to the failing run's head branch (which encodes the issue key in the
-    # agent branch shape: `ghabot-ag-<NNNNN>-<slug>`). Same group as the
-    # dispatch-driven jobs so a fresh CI-failure iteration queues behind any
-    # in-flight pr-iterate for the same ticket instead of interleaving.
-    group: claude-${{ github.event.client_payload.issue_key || inputs.issue_key || github.event.workflow_run.head_branch }}
-    cancel-in-progress: false
+# No workflow-level `concurrency:` (AG-18177). It resolved to two different groups
+# for one ticket — `claude-<KEY>` on a dispatch, `claude-ghabot-<proj>-<n>-<slug>`
+# on a workflow_run — so it never serialised the pair it exists to serialise. A GHA
+# expression can't normalise that and workflow-level `concurrency` can't read
+# `needs`; `jobs.<id>.concurrency` can, so the group is resolved by the
+# `concurrency-key` job below and consumed per job. Why, in full:
+# ag-dev-prompts docs/ai-workflow.md -> "Normalising the concurrency group".
 
 env:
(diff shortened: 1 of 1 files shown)


alantreadway:
Amended: `queue: max`, and the promote now shares the agent's group.

**Why this changed after the original push.** The shipped shape was built around a GHA constraint that turns out to be liftable: only **one** run may be *pending* per concurrency group, so a third trigger evicts the waiter. Everything awkward here followed from it — the promote had to live in a `-promote-<head_sha>` sibling group so it could never be evicted, and agent-vs-promote overlap had to be handed to a separate in-flight-agent interlock.

`concurrency.queue` removes the constraint: `queue: max` allows **100 pending, FIFO**, and is live on github.com. Probed empirically on `ag-grid/ag-dev-prompts` before touching this file (an unrecognised key inside `concurrency:` is a *workflow validation* error, which here would stop every pipeline run):

| Question | Result |
|---|---|
| Does `queue:` parse? | yes — 6 runs, no validation error |
| Is pending time charged against `timeout-minutes`? | **no** — a run pending 4m06s ran normally under `timeout-minutes: 2` |
| What status does a pending run report? | `status: pending`, conclusion empty |
| Does cancelling the holder hand over immediately? | yes — holder cancelled 10:44:10, successor started 10:44:14 |

**What changed in this file (4 edits):**
1. `agent-run` — added `queue: max`. `cancel-in-progress: false` stays; the `true` pairing is a validation error.
2. `ci-success-handler` — dropped the `-promote-<head_sha>` suffix, so it shares the plain ticket group, plus `queue: max`.
3. `agent_interlock_wait_minutes: 20` → `0`.
4. `timeout-minutes: 25` → `10` (it was only 25 to outlast the interlock wait; queue time isn't charged).

**The defect that justifies it**, found while stress-testing and not previously known: on the codex-hold path `ci-success-handler` withholds the Needs-Review promotion and emits `dispatch_pr_iterate=true`; that iterate goes `pending` in the ticket group, and under `queue: single` **any** other entrant — a drag, an AG-17918 self-heal, a watchdog re-fire — cancels it. Result: green PR, open P0/P1 findings, promotion held by design, and nothing left to fire (CI has settled). Unlike an evicted stage it is *not* recovered by the next drag.

**The suffix is not coming back.** It existed only to dodge eviction, and it bought that by letting two green commits' handlers overlap on writes that have no compare-and-swap. With no eviction it is strictly worse than the plain ticket group.

A workflow-level lock was designed and **rejected** — workflow-level expressions can't reach `needs`, and GHA has no regex/split/trim, so the group would be the raw `client_payload.issue_key`; ` AG-18155` and `AG-18155\n` would form different groups, which is the AG-18177 defect reintroduced where no code can fix it. Full rationale and the other four objections: `docs/ai-workflow.md` § *Normalising the concurrency group* in `ag-dev-prompts` (merged, #871).

ag-alantreadway (approved):
LGTM

PR #7838, opened 2026-08-17
AG-3390 Allow esm.sh in the CSP for the ES-module example runner

taktran:
Port of the `latest` change (`ffdd62c0fe`, AG-17103) to the release line. `script-src` and
`connect-src` now allow `https://esm.sh`.

Nothing on this branch fetches that origin. The ES-module example runner landed on `latest`
only; the examples here still load through SystemJS, which takes React from jsdelivr. The point
is that the release line emits the same policy as `latest`, so the header does not depend on
which branch built a given deployment — the same reason the Enzuzo, LinkedIn and Make origins
are already on both.

Companion PRs, same change: ag-grid#14867 (`latest`) and ag-grid#14870 (`b36.1.0`).

The `cspRules.ts` comment differs from the one on `latest` (it explains the port rather than the
runner change that motivated it); the two policy lines are identical.

## Verification

- `vitest run --root packages/ag-charts-website htaccess csp` — 59 tests pass (29 `cspRules`,
  28 `htaccessRules`, 2 `redirectsChecker`).
- `.htaccess` snapshots regenerated: the only delta is the added origin in the two directives.
- `nx run-many -t lint,build:types -p ag-charts-website` clean (10 pre-existing warnings in
  `pages.ts`, untouched here).


taktran:
Closing for now — the release-line port will go up once the change is verified on staging. Branch `ag-3390/csp-allow-esm-sh-b14.1.0` is left on the remote, so this can be reopened as-is.

PR #7840, opened 2026-08-17
ci(jira-agent-pipeline): use the shared post-lock guard actions

alantreadway:
Replaces the inline headcheck/route shell in this stub with the `jira-agent-headcheck` and `jira-agent-route` composite actions from `@ag-grid/dev-prompts`, and trims the stub's comments to pointers at the private docs. Behaviour is unchanged apart from two fixes carried by the shared actions: a fork PR's branch tip is now read from the fork, and the head check fails open when it cannot run.

https://ag-grid.atlassian.net/browse/AG-18177

### DIFF
--- a/.github/workflows/jira-agent-pipeline.yml
+++ b/.github/workflows/jira-agent-pipeline.yml
@@ -1,108 +1,18 @@
 name: Jira Agent Pipeline
 
-# ============================================================================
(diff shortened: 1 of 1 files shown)


ag-alantreadway (approved):
LGTM

PR #7843, opened 2026-08-17
AG-17801 Fix background regions category ranges and warnings

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-17801

Fix #AG-17801


alantreadway (approved):
LGTM

PR #7844, opened 2026-08-17
SE-62 Link the API reference pages in the served HTML

taktran:
https://ag-grid.atlassian.net/browse/SE-62


taktran:
Not that good a fix, dropping it for now

PR #7845, opened 2026-08-17
AG-9809 TC7 Avoid axis cursor clash when both click+drags are enabled

olegat:
https://ag-grid.atlassian.net/browse/AG-9809


alantreadway (approved):
LGTM

PR #7852, opened 2026-08-18
AG-3390 Route history writes through a shared helper

taktran:
Follow up to https://github.com/ag-grid/ag-studio/pull/2410 to sync with charts


markdurrant (approved):
LGTM 👍

PR #7861, opened 2026-08-19
AG-17184 Align org chart node text by its backing box

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-17184

Fix #AG-17184


alantreadway (approved):
LGTM

PR #7868, opened 2026-08-20
Handle invalid version strings in changelog transformVersion

taktran:
## Summary
Add validation to the `transformVersion.charts()` function to safely handle invalid or placeholder version strings (e.g. "Next", "TBD", empty strings, or malformed versions) by returning `null` instead of attempting transformation. Update the Pipeline component to gracefully handle `null` return values.

## Key Changes
- **transformVersion.ts**: Added validation to check that grid versions have exactly 3 numeric parts before transformation; returns `null` for invalid versions
- **transformVersion.ts**: Updated return type annotation from `string` to `string | null` to reflect the new behaviour
- **Pipeline.tsx**: Updated the scheduled version display logic to handle `null` return values, falling back to "Scheduled" when transformation fails
- **transformVersion.test.ts**: Added comprehensive test suite covering valid versions and edge cases (placeholder versions, empty strings, malformed versions)

## Implementation Details
The validation uses a regex check (`/^\d+$/`) to ensure each version part is numeric, preventing errors when processing non-release versions. The Pipeline component now checks for both `null` returns and the "NEXT" placeholder before deciding whether to append the version number to the "Scheduled" message.

https://claude.ai/code/session_014t3L58eBLX25TsrwvoeJwE

### DIFF
--- a/external/ag-website-shared/src/components/changelog/Pipeline.tsx
+++ b/external/ag-website-shared/src/components/changelog/Pipeline.tsx
@@ -39,11 +39,11 @@ const getColDefs = (library: Library) => [
             if (hasFixVersion) {
                 const latestFixVersion = fixVersionsArr.length - 1;
                 const fixVersion = fixVersionsArr[latestFixVersion];
-                if (fixVersion.toUpperCase() === 'NEXT') {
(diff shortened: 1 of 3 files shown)


skewis6 (approved):
LGTM

PR #7871, opened 2026-08-20
Encrypt and authenticate the cloud dependency cache

alantreadway:
The baked `node_modules` tree published for Claude Code cloud environment setup is now encrypted with a shared passphrase, and carries an HMAC over the ciphertext that is verified before any decryption. Assets are keyed by lockfile hash and key generation so the passphrase can be rotated without sessions fetching assets they cannot open.

Also drops the temporary branch trigger left in the bake workflow.

Requires a `CLOUD_CACHE_KEY` repository secret, and the same value as `AG_CLOUD_CACHE_KEY` in each cloud environment. Setup and rotation are documented in `external/ag-shared/docs/claude-code-cloud-sessions.md`.


alantreadway:
/pr-review

All three findings addressed in 5cff6eb1 and 64e9d7c7.


---
_Generated by [Claude Code](https://claude.ai/code)_

alantreadway:
/pr-review

Both P2s addressed in 14c8a077: the key check is now on format (64+ hex) rather than length, and the setup step marks AG_CLOUD_CACHE_KEY required for the fast path.


---
_Generated by [Claude Code](https://claude.ai/code)_

alantreadway:
/pr-review

Re-running against 07f55b5e so the verdict matches the head being merged.


---
_Generated by [Claude Code](https://claude.ai/code)_

PR #7884, opened 2026-08-21
Add hot-path detection and review guidance

alantreadway:
Gives PR review a mechanical way to tell when a change lands on code that runs per datum, per frame or per interaction, and what evidence would settle the question. No JIRA ticket — this is tooling for the review flow.

## What it adds

- `tools/hot-paths/hot-path-index.json` — four tiers ordered by call frequency (per-datum, per-frame, per-interaction, bundle), each with its globs and the invariants it must hold, plus the stage-2 signal groups and weights.
- `tools/hot-paths/detect.js` — two-stage detector. Stage 1 matches changed non-test files against the tier globs; stage 2 scores each hit on hot-path signals in the added lines or their context, proximity to an existing hot-path marker comment, and whether an added line executes inside a loop.
- `tools/hot-paths/benchmark-map.js` — maps changed files to the benchmark examples that exercise them and emits a ready-to-run `/benchmarks` command. Each example's series types, axis types and test cases are derived from its `main.ts` on every run, so the map cannot go stale as examples change.
- `.rulesync/skills/hot-paths/SKILL.md` — the review guide, loaded on demand: the tier table, the five regression shapes this repo has actually shipped (each cited to a real commit), how to write a finding, and what the evidence is worth.
- A `## Review guidelines` section in the root rule pointing at the detector. That heading is what `/ag-eng:pr-review` already reads for repo-specific rules, so this is the whole wiring on the repo side.

## Calibration

Measured, not assumed — over the last 48 PR merges (back-merges excluded):

| | Rate |
|---|---|
| Stage 1 (tier globs alone) | 67% |
| Stage 2 (scored, triggers the review pass) | 27% |
| Recall on 13 known performance commits | 10/13 |

The scoring is what makes the pass affordable. The single biggest recall gain was the in-loop check: token matching on added lines alone caught 3 of 13, because a new guard inside an existing per-datum loop mentions no loop machinery of its own.

## The consuming half

The generic review pass lives in the `ag-eng` plugin and is proposed separately in ag-grid/ag-dev-prompts#941. It hardwires no file list and skips silently in a repo that declares no `## Review guidelines` hot-path contract — verified against ag-grid and ag-studio, where neither declares one. This PR is useful without that one: the detector and skill stand alone for manual use.

## Test plan

- [x] `node --test tools/hot-paths/hot-paths.test.js` — 7/7 (glob semantics, tier specificity, benchmark-map derivation, and that an unrecognised path recommends nothing rather than guessing)
- [x] `yarn nx format` clean
- [x] `detect.js --pr <n>` verified end to end, including the merged-PR fallback to the merge commit's first-parent range (a merged PR's merge-base is its own head, which would otherwise report an empty diff)
- [x] No `packages/**` source changed, so `build:types`, `lint` and the chart suites do not apply. `tools/**/*.js` sits outside the eslint project, as does the existing `tools/benchmark/benchmark-examples.js`.


alantreadway:
Round-4 review addressed in 4 of 5 findings. The bullet-series finding is rejected, with evidence, for the second time:

```
$ find packages -ipath "*bullet*" -name "*.ts"
packages/ag-charts-website/src/content/gallery/_examples/simple-bullet/main.ts
packages/ag-charts-website/src/content/docs/linear-gauge/_examples/bullet/main.ts

$ grep -rn "'bullet'" packages/ag-charts-types/src/
(no matches)
```

The cited path `packages/ag-charts-enterprise/src/series/bullet/bulletSeries.ts` does not exist. "Bullet" here is a linear-gauge presentation, not a series type, so a mapping rule for it would tag a type no chart can declare — the opposite of what the mapper is for.

The other four are fixed: the in-loop walk now stops at an arrow that is not an array iteration (a listener registered inside a loop was being charged the loop's frequency); the shared-series rule names the base files rather than matching `chart/series/*.ts` by shape, which had swamped `seriesAreaManager.ts` and `pickManager.ts`; the generic radar rule matches only the shared radar path; and the index comment now names `detect.js`.

Calibration re-derived after the loop-walk change and unchanged: 48 PR merges → stage 1 67%, triggered 27%, recall 6/8. Test suite 24/24.

alantreadway:
Round-5 review: four of five findings fixed, one rejected with a measurement.

**Rejected — "removed lines should score nothing."** Measured both ways over the same 13 known performance commits:

| removed-line weight | trigger rate | recall |
| --- | --- | --- |
| zero | 29% of 48 PR merges | **9**/13 |
| context (half) | 29% of 48 PR merges | **10**/13 |

The commit that drops out is `26d7f84244` — the epoch-column rescan the skill cites as the canonical regression shape, and a deletion. Performance fixes in this repo are frequently deletions, and a change that removes a TypedArray or a fast path from a tier-1 file is precisely what the gate exists to put in front of a reviewer; `passes/hot-path.md` says so explicitly ("if the PR removes a fast path, a cache, a reuse buffer, or a gate, that is a finding"). So removed lines keep the context weight — not as strong as introducing the construct, not silent.

The labelling half of that finding was right and is in: removed matches now report `where: "removed"`, so a reviewer sees what the change displaced instead of reading it as code still present.

**Fixed:**
1. Multiline loop headers. Prettier leaves `) {` at the opener's own indentation, and lowering the walk's bar there skipped the `for (` itself — the strongest signal, lost in exactly the long headers that iterate over data.
2. Multiline deferred callbacks. The walk now checks the call opener above an arrow that landed on its own line.
3. Markers only count inside a comment, so a string literal reading `allocation-free` no longer collects the marker weight.
4. The index's calibration figure now matches the skill's.

29 tests pass. Calibration: 48 PR merges → stage 1 67%, triggered 29%, recall 10/13.

ag-alantreadway (approved):
LGTM

PR #7887, opened 2026-08-21
AG-18117 Fix org series set active show highlight styles and tooltip

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-18117


alantreadway (approved):
LGTM

PR #7890, opened 2026-08-21
Merge b14.1.0 into latest

seanlandsman:
Merge-up of the `b14.1.0` release branch into `latest`.

`b14.1.0` is **12 commits ahead** of `latest` and 801 behind.

## Conflicts

A dry-run merge (`git merge-tree`) reports **34 conflicting files**, so this needs manual resolution before it can merge. The clusters are:

- `external/ag-website-shared/` components — `ContactForm.tsx`, `initCaptcha.ts`, `LinkIcon.tsx`. This is the vendored subrepo, so resolution should match whatever the shared upstream now holds rather than being decided per-repo.
- add/add conflicts where the same file was created independently on both sides, e.g. `buildRoadmapMarkdown.ts`, `apiReferenceRouting.ts`
- `packages/ag-charts-website/` api-documentation components

The add/add cases are worth a closer look than the content conflicts — the same path created twice usually means the two sides diverged on approach, not just on lines.

## Included

Among the 12 commits is the SE-113 blog-URL work (`e194d5c7ca`), which replaced `blog.ag-grid.com` links with `www.ag-grid.com/blog/` across 56 files. That change is already live in production.


seanlandsman:
Superseded by the dedicated merge branch, which resolves the 34 conflicts.

PR #7902, opened 2026-08-25
chore(ci): use the explore driver for PR review

alantreadway:
Switches the PR review workflow to the `explore` driver, which is the default in `@ag-grid/dev-prompts` from 1.6.273. Codex is given the PR intent, a ranked changed-file list and the unresolved review threads, and reads the diff itself instead of being handed a pre-chunked one.

Also drops the `anthropic-api-key` input, which the action accepts and ignores.

See ag-dev-prompts `docs/installation.md` → PR Review action — drivers & secrets.

### DIFF
--- a/.github/workflows/pr-review.yml
+++ b/.github/workflows/pr-review.yml
@@ -25,7 +25,7 @@ on:
                 type: string
                 required: true
             driver:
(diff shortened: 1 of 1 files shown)


ag-alantreadway (approved):
LGTM

PR #7906, opened 2026-08-25
AG-17597 Update $ref operation to allow targeting leaves

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-17597


alantreadway (approved):
LGTM

PR #7908, opened 2026-08-25
AG-18293 Serve the example page's scripts instead of inlining them

sdwvit:
https://ag-grid.atlassian.net/browse/AG-18293

Fix #AG-18293


alantreadway (approved):
LGTM

PR #7917, opened 2026-08-25
Consolidate the bar-family label wiring

iMoses:
Splits the bar-family label geometry out of the placement engine and de-duplicates the wiring the bar, histogram, range-bar, waterfall, funnel and pyramid series each restated.

- `barLabelGeometry.ts` takes the bar-specific helpers (rotation, insets, regions, datum builders, obstacle builders) out of `labelPlacement.ts`, leaving the engine with the generic cascade. The move is byte-identical; the module is re-exported from core's `main.ts`, so the public surface is unchanged.
- `barLabelDataContext` and `barLabelObstaclesFor` derive the six collision/fit values and the obstacle set every bar-family `getLabelData`/`getLabelObstacles` was resolving by hand.
- Two module-level flags the engine threaded through its cascade are gone: `refittingToObstacles` becomes a `fontSearch` argument on `fitLabelToCandidate` (dropping a `try/finally` from a per-candidate path), and the flush-into-region block, previously inlined twice, becomes `flushCandidateBox`. `shrinkCompassCandidate` reads the cascade scratch it was being handed as eleven arguments.
- `LabelFitDescriptor.boundByRegion` is now required rather than defaulting to `true`, so the two construction sites state which behaviour they want.

Also adds the test and benchmark net this refactor leans on: a brute-force oracle for the positioned-candidate path, a `LabelManager` suite covering the placement cache, carry-forward and resolution order, a scaling test that fails if per-label cost picks up a term in the label count, and a `series-labels-collision` benchmark page separating placement cost from draw cost.

No behaviour change: the community and enterprise snapshot suites pass unmodified.

### DIFF
--- a/packages/ag-charts-community/src/chart/labelUtil.ts
+++ b/packages/ag-charts-community/src/chart/labelUtil.ts
@@ -561,9 +561,9 @@ export interface BarLabelDataContext {
     readonly collideWith: CollideWith;
     readonly threshold: number;
     /** Drawn-box footprint of `text` under the configured label font: the measured glyph plus {@link box}. */
-    measureBox(text: NormalisedTextOrSegments): { width: number; height: number };
+    readonly measureBox: (text: NormalisedTextOrSegments) => { width: number; height: number };
     /** Per-candidate fit inputs for `text`, or `undefined` when the label opted out of overflow control. */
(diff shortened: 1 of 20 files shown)


alantreadway (approved):
LGTM

PR #7918, opened 2026-08-25
AG-13301 Let chart captions shrink to fit before truncating

iMoses:
https://ag-grid.atlassian.net/browse/AG-13301

Fix #AG-13301


alantreadway (approved):
LGTM

PR #7928, opened 2026-08-26
Invalidate axis tick-layout cache on font load

alantreadway:
The axis tick-layout cache key covers domain, range and nice bounds only, so a webfont that resolves after the first layout left label thicknesses measured against the fallback font while the labels painted in the loaded one.

Result was a bistable render: the `google-fonts` E2E snapshot has been flipping between two images 1px apart in bottom-axis thickness since March, depending on when the font resolved (four regenerations on 2026-03-03 alternating between the same two blobs; the snapshot CI regenerated in #7925 is bit-identical to the one committed then).

Axes now drop the cache on `font:load`, so the layout that follows re-measures with the loaded font.

Honouring the new metrics exposed a second problem: a relayout while an animation batch is in flight destroys that batch, so a font resolving mid-animation cost the chart its initial animation (caught by the `synchronised` E2E specs). `FontManager` now holds the notification until `animation-stop`, with `update:complete` as a second flush trigger so a skipped batch cannot strand the correction.

Supersedes #7925 — that PR just flips the baseline to the other of the two states. No baseline regeneration is needed: the fonts shard passed against the committed baseline on this branch.

### DIFF
--- a/packages/ag-charts-community/src/chart/axis/axis.ts
+++ b/packages/ag-charts-community/src/chart/axis/axis.ts
@@ -485,8 +485,7 @@ export abstract class Axis<
         this.cleanup.register(
             this.moduleCtx.widgets.containerWidget.addListener('mousemove', (e) => this.onMouseMove(e)),
             this.moduleCtx.widgets.containerWidget.addListener('mouseleave', () => this.endHovering()),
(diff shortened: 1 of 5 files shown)


lsjroberts (approved):
LGTM

PR #7935, opened 2026-08-26
AG-7605 TC2 Consolidate all crossline clicks into 1 event with allClickParams

olegat:
https://ag-grid.atlassian.net/browse/AG-7605


alantreadway (approved):
LGTM

PR #7936, opened 2026-08-26
AG-18234 Skip label processing when there is nothing to draw

alantreadway:
https://ag-grid.atlassian.net/browse/AG-18234

Fix #AG-18234


alantreadway:
/benchmarks simple-sparkline high-perf-line high-perf-area large-dataset

iMoses (approved):
LGTM

PR #7938, opened 2026-08-26
AG-18140 Fix bar label itemStyler resolution and funnel label placement

iMoses:
https://ag-grid.atlassian.net/browse/AG-18140

Fixes for AG-18140:

- Bar-family label `itemStyler` geometry is resolved lazily, so the styler runs once per rendered label, reporting the placement that label was drawn at, instead of once per probed candidate.
- Funnel labels now avoid the neighbouring stage bars, matching every other bar-family series (`collideWith.seriesItems`, which the funnel theme never opted into).
- Cone funnel `start`/`end` now run top to bottom down a vertical divider, per the documented contract, instead of anchoring on the value-axis origin.
- Cone funnel RTL mirrors `before`/`after` where the divider spans the vertical axis, as it already mirrored `start`/`end` for a horizontal one.

Fix #AG-18140


alantreadway (approved):
LGTM

PR #7949, opened 2026-08-27
AG-8901 Let cross line labels reserve space from other labels

iMoses:
Adds `label.reserveSpace` to cartesian cross lines. An opted-in label is offered to label placement as a fixed obstacle, so other labels are placed clear of it.

Stacked on #7923 (AG-7486); review that first. GitHub retargets this to `latest` when #7923 merges.

## What ships

- `LabelSource` seam on `LabelManager`, split into `LabelSource` (identity, invalidation, obstacles) and `PlacedLabelSource` (the placement half). `ISeries` satisfies the latter structurally, so no series code changes.
- `label.reserveSpace` on `AgCartesianCrossLineLabelOptions`, default `false`.

## Notes for review

- The label is an **obstacle**, not a placed label. With no region and no fit, `placeFromPositionedCandidates` can only hand back the box it was given, so routing it through the solve would buy nothing and leave a writeback that cannot fire. Obstacles are also seeded before any label resolves, so the result does not depend on solve order.
- The reserved box is inserted **unrotated**: it is already the rotated footprint, and carrying the rotation through would have `labelObstacleBox` inflate it a second time, so a 90 degree label would reserve a transposed box. Covered by `reserves a rotated label the space it actually occupies`.
- Named `reserveSpace`, not `avoidCollisions`: everywhere else in this codebase `avoidCollisions` means the label itself may move or hide, which is the opposite of this.
- `unregisterSource` takes the instance. A recreated axis registers its replacement before the old one is destroyed, so deleting by id alone would unregister the live source and silently disable the feature for the rest of the chart's life.

## Scope

AG-8901 is **partially** delivered; please do not close it. Product cut angle/polar cross lines and radar (TC3/TC4) to a separate ticket. The truncated-label tooltip acceptance criterion and label wrapping are also not addressed here.

## Not covered

Two pre-existing issues in #7923's code, raised on that PR rather than fixed here: the relaxed `overflows` condition can render an arbitrarily small series area, and `clampPaddingPair`'s `total > 0 ? ... : 0` guard is unreachable.

## Testing

`nx build:types`, `docs-resolved-interfaces`, `build:test`, `lint` (0 errors), `test ag-charts-community` (4573 passed), `test ag-charts-enterprise` (3032 passed). No snapshot churn.


alantreadway (approved):
LGTM

PR #7953, opened 2026-08-27
AG-18348: Fix anchor links using a stale pathname after SPA navigation

skewis6:
https://ag-grid.atlassian.net/browse/AG-18348

Fix #AG-18348

Tab clicks passed a bare hash to `navigate()`, so the `history` package filled the
pathname in from its own cached location, which Astro's `<ClientRouter />` only
refreshes on `popstate`. After a client-side navigation the URL kept the pathname of
the page loaded first. The full location is now passed explicitly - the same shape
#7948 settled on for `SideNavigation.tsx`. `navigate()` is kept rather than
`replaceHistoryUrl` because only a push fires the `browserHistory` listener that keeps
sibling tab groups in sync, and that the scroll spy depends on.

The ticket's second issue asked for the property reference anchor links to migrate to
`replaceHistoryUrl`. That migration is already in place by another route: `LinkIcon`
spreads incoming props before setting its own `onClick`, so the `useScrollToAnchor`
handler passed to it was overridden and never ran, and `LinkIcon`'s handler already
calls `replaceHistoryUrl`. That is why the reported symptoms do not occur on `latest` -
repeated property anchor clicks leave `history.length` unchanged and Astro's
`history.state` keys intact. The dead prop and the now-unreachable hook are removed
instead of migrating code that never executed.

`e2e/docs-page-nav.spec.ts` passes on this branch.

### DIFF
--- a/external/ag-website-shared/src/components/tabs/TabsWithHtmlChildren.astro
+++ b/external/ag-website-shared/src/components/tabs/TabsWithHtmlChildren.astro
@@ -112,7 +112,16 @@ const { tabItemIdPrefix, headerLinks, snippetTabs } = Astro.props as Props;
                     this.updateTabs();
 
                     const tabId = getTabId({ id: this.selected.id, prefix: this.tabItemIdPrefix });
-                    navigate(`#${tabId}`);
+                    // Pass the whole location: a bare hash is resolved against the `history`
+                    // package's cached location, which Astro's `<ClientRouter />` never
+                    // refreshes, leaving the pathname on whichever page loaded first. Keep
+                    // `navigate` rather than `replaceHistoryUrl` - the `browserHistory`
+                    // listener above syncs sibling tab groups, and only a push fires it.
+                    navigate({
+                        pathname: window.location.pathname,
+                        search: window.location.search,
+                        hash: tabId,
+                    });
                 });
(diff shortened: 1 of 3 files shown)


packages/ag-charts-website/src/components/pages-navigation/components/SideNavigation.tsx
```diff
@@ -34,11 +35,7 @@ export function SideNavigation({ headings, delayedScrollSpy }: Props) {
                                     onClick={(event) => {
                                         event.preventDefault();
                                         scrollIntoViewById(slug);
-                                        navigate({
-                                            pathname: window.location.pathname,
-                                            search: window.location.search,
-                                            hash: slug,
-                                        });
+                                        replaceHistoryUrl(`#${slug}`);
```
taktran: So this was reverted here: https://github.com/ag-grid/ag-charts/pull/7948/changes/258a81120827bdd53f8e6ae9417dc965e8fe9a06 because it was failing CI - maybe you found something else though?

skewis6: Nope, I've reverted the change

taktran (approved):

PR #7960, opened 2026-08-27
AG-17743: hook documentation example dark mode into website CSS

skewis6:
Documentation example charts now adopt the site's dark mode through CSS instead of the load-then-swap workaround.

Three separate concerns:

- `example-chart-theme.css` (new, folded into every example's `ag-example-styles.css`) declares the nine colour parameters `ag-default-dark` retunes, keyed off `:root[data-dark-mode='true']`.
- The page shell owns `data-dark-mode` on the example document via a render-blocking `<head>` script, and keeps it in step with the site toggle.
- The injected snippet points the default theme's params at those variables. AG Charts resolves `var()` colours against the container and watches them, so first paint is correct and a toggle re-resolves them with no `chart.update()`.

The snippet is now prepended rather than appended, so the theme is settled before the chart is created.

**Palette-sensitive series keep the theme-name swap.** `ag-default-dark` also retunes the palette (stroke set, hierarchy ramp, second sequential colours), and none of it can be driven from CSS: `theme.palette` takes no option for the hierarchy or sequential colours (validation rejects them), and supplying `strokes`/`up`/`down`/`neutral` would flip `paletteType` and change how ~8 series types render in light mode too. Scatter, bubble, candlestick, ohlc, treemap, cone-funnel, the gauges and the map backgrounds therefore stay on the swap, which — being prepended — still gets their first paint right; only their toggle costs a re-render.

Examples that set their own `backgroundColor`/`foregroundColor` keep today's behaviour, and Plunker/CodeSandbox exports get none of this plumbing.

### Acceptance criteria

- **AC1** (correct dark colours on first paint) — met for every default-theme example, both paths.
- **AC2** (toggle without `chart.update()`) — met except for the palette-sensitive series above, which re-render. Closing that fully needs a product change: extending `AgChartThemePalette` with the hierarchy/sequential colours, or reworking `paletteType` so injected palette values don't trip the `user-indexed`/`user-full` branches.

### Verification

Browser-tested on the docs site with `optionsMutationFn` instrumented to record what each example resolved:

| example | dark first paint | toggle light→dark |
| --- | --- | --- |
| bar, pie | `ag-default` + var params, `#192232` | `rgb(25,34,50)`, 0 extra mutation calls |
| scatter, treemap, candlestick | `ag-default-dark`, `#192232` | `#192232`, re-rendered |

Light first paint correct in all five. `cssVariableParams.test.ts` pins both halves of the design: that the nine params reproduce `ag-default-dark`'s parameters, and the exact set of palette entries the dark theme retunes — so a change to either theme fails here rather than silently mis-colouring examples.

`nx run-many -t lint test -p ag-charts-generate-example-files ag-charts-community` passes.

Out of scope: annotation colour persistence (AG-11798 / AG-12484).

### DIFF
--- a/packages/ag-charts-website/src/components/example-runner/getExampleDarkModeInitScript.ts
+++ b/packages/ag-charts-website/src/components/example-runner/getExampleDarkModeInitScript.ts
@@ -1,0 +1,38 @@
+/**
+ * Source of the example document's `data-dark-mode`, which `example-chart-theme.css` turns into the
+ * chart theme's colour variables. Runs render-blocking in the example's `<head>` so the attribute is
+ * set before the chart is created, and keeps it in step with the site's toggle afterwards.
+ */
+export const getExampleDarkModeInitFragment = ({ ignoreDarkMode }: { ignoreDarkMode?: boolean }) =>
+    ignoreDarkMode
+        ? ''
+        : `<script nonce="123123">
+(function () {
+    const htmlEl = document.documentElement;
+
+    let storedDarkmode;
+    try {
+        storedDarkmode = localStorage['documentation:darkmode'];
+    } catch (e) {
+        /* Storage can be unavailable, in which case the OS preference stands in. */
+    }
+
+    const setDarkmode = (darkmode) => {
+        htmlEl.dataset.darkMode = String(darkmode === true);
+    };
+
+    setDarkmode((storedDarkmode ?? String(matchMedia('(prefers-color-scheme: dark)').matches)) === 'true');
+
+    // Two delivery channels, each read from the property its own channel provides: a real
+    // postMessage when this example runs inside the example-runner iframe, and a same-page
+    // CustomEvent when it is embedded directly in the docs page.
+    const onColorSchemeChange = (data) => {
+        if (data?.type === 'color-scheme-change') {
+            setDarkmode(data.darkmode);
+        }
+    };
+
+    window.addEventListener('message', (event) => onColorSchemeChange(event.data));
+    window.addEventListener('ag-color-scheme-change', (event) => onColorSchemeChange(event.detail));
+})();
+</script>`;
--- a/plugins/ag-charts-generate-example-files/src/executors/generate/generator/utils/getDarkModeSnippet.ts
+++ b/plugins/ag-charts-generate-example-files/src/executors/generate/generator/utils/getDarkModeSnippet.ts
@@ -1,26 +1,82 @@
-// NOTE: This is duplicated from `external/ag-website-shared/src/utils/getDarkModeSnippet.ts`
+// NOTE: The homepage gallery keeps its own copy of this plumbing in
+// `packages/ag-charts-website/src/utils/getDarkModeSnippet.ts`.
 export const DARK_MODE_START = '/** DARK MODE START **/';
 export const DARK_MODE_END = '/** DARK MODE END **/';
 
+/**
+ * Points the default theme at the colour variables `example-chart-theme.css` flips with
+ * `data-dark-mode`. AG Charts resolves `var()` colours against the chart container and watches them
+ * for changes, so an example paints in the right scheme first time and repaints itself on a toggle.
+ *
+ * Only the parameters `ag-default-dark` retunes are listed; the rest are `$ref`s onto these and
+ * follow. Any other theme - a preset's included - also differs by palette, which no parameter can
+ * carry, so it keeps resolving the `-dark` theme name and still needs an explicit re-render.
+ */
 const THEME_MUTATION = `
-const getDarkmodeTheme = (theme = 'ag-default', preset) => {
-    const baseTheme = preset === 'price-volume' ? 'ag-financial' : theme.replace(/-dark$/, '');
-    return darkmode ? baseTheme + '-dark' : baseTheme;
+const DEFAULT_THEME_PARAMS = {
+    axisLineColor: 'var(--ag-example-chart-axis-line-color)',
+    backgroundColor: 'var(--ag-example-chart-background-color)',
+    borderColor: 'var(--ag-example-chart-border-color)',
+    chromeBackgroundColor: 'var(--ag-example-chart-chrome-background-color)',
+    crosshairLabelBackgroundColor: 'var(--ag-example-chart-crosshair-label-background-color)',
+    foregroundColor: 'var(--ag-example-chart-foreground-color)',
+    gridLineColor: 'var(--ag-example-chart-grid-line-color)',
+    groupedCategoryLineColor: 'var(--ag-example-chart-grouped-category-line-color)',
+    subtleTextColor: 'var(--ag-example-chart-subtle-text-color)',
 };
 
+const isDarkmode = () => document.documentElement.dataset.darkMode === 'true';
+
+const getDarkmodeTheme = (theme = 'ag-default', preset) => {
+    const baseTheme = preset === 'price-volume' ? 'ag-financial' : theme.replace(/-dark$/, '');
+    return isDarkmode() ? baseTheme + '-dark' : baseTheme;
+};
+
+let hasSwappedTheme = false;
+
 __chartAPI.optionsMutationFn = function update(options, preset) {
-    const nextOptions = { ...options };
     const theme = options.theme;
-    if (isAgThemeOrUndefined(theme)) {
-        nextOptions.theme = getDarkmodeTheme(theme, preset);
-    } else if (typeof theme === 'object' && isAgThemeOrUndefined(theme.baseTheme)) {
-        nextOptions.theme = {
-            ...options.theme,
-            baseTheme: getDarkmodeTheme(theme.baseTheme, preset),
+    const isThemeObject = typeof theme === 'object' && theme != null;
+    const baseTheme = isThemeObject ? theme.baseTheme : theme;
+    const params = isThemeObject ? theme.params : undefined;
+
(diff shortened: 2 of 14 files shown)


packages/ag-charts-website/src/components/example-runner/getExampleDarkModeInitScript.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * Source of the example document's `data-dark-mode`, which `example-chart-theme.css` turns into the
+ * chart theme's colour variables. Runs render-blocking in the example's `<head>` so the attribute is
+ * set before the chart is created, and keeps it in step with the site's toggle afterwards.
+ */
+export const getExampleDarkModeInitFragment = ({ ignoreDarkMode }: { ignoreDarkMode?: boolean }) =>
+    ignoreDarkMode
+        ? ''
+        : `<script nonce="123123">
```
taktran, line 9: This `nonce` doesn't look right - it should be generated on the server side no?

plugins/ag-charts-generate-example-files/src/executors/generate/generator/utils/getDarkModeSnippet.ts
```diff
@@ -1,26 +1,122 @@
-// NOTE: This is duplicated from `external/ag-website-shared/src/utils/getDarkModeSnippet.ts`
+// NOTE: The homepage gallery keeps its own copy of this plumbing in
+// `packages/ag-charts-website/src/utils/getDarkModeSnippet.ts`.
 export const DARK_MODE_START = '/** DARK MODE START **/';
```
taktran: Can we delete `DARK_MODE_START` and `DARK_MODE_END` now?

plugins/ag-charts-generate-example-files/src/executors/generate/generator/utils/getDarkModeSnippet.ts
```diff
@@ -1,26 +1,122 @@
...
+
 const getDarkmodeTheme = (theme = 'ag-default', preset) => {
     const baseTheme = preset === 'price-volume' ? 'ag-financial' : theme.replace(/-dark$/, '');
-    return darkmode ? baseTheme + '-dark' : baseTheme;
+    return isDarkmode() ? baseTheme + '-dark' : baseTheme;
 };
 
+let hasSwappedTheme = false;
+
 __chartAPI.optionsMutationFn = function update(options, preset) {
```
taktran, line 74: Do we still need to monkey patch `optionsMutationFn`?

packages/ag-charts-website/src/components/example-runner/getExampleDarkModeInitScript.ts
taktran: Maybe needs to be an external script in `public`

plugins/ag-charts-generate-example-files/src/executors/generate/generator/utils/getDarkModeSnippet.ts
```diff
@@ -1,26 +1,108 @@
...
 };
+
+// The page shell owns the example document's \`data-dark-mode\`, setting it before this runs and
+// keeping it in step with the site's toggle.
+new MutationObserver(() => {
+    if (!hasSwappedTheme) return;
+    document.querySelectorAll('[data-ag-charts]').forEach((element) => {
+        const chart = __chartAPI.getInstance(element.parentElement);
+        if (chart == null) return;
+        chart.update(chart.getOptions());
```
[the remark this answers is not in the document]
skewis6: Good catch, fixed in 9e35296305.

`hasSwappedTheme` is replaced by a `WeakSet` of the containers that actually took the swap, and the observer checks each chart against it — so a CSS-driven chart is no longer dragged into an update by a palette-sensitive sibling.

Verified in the browser by wrapping `chart.update` per instance and toggling dark mode:

| example | path | `chart.update` calls |
| --- | --- | --- |
| `simple-bar` | CSS variables | 0 |
| `simple-scatter` | theme-name swap | 1 (from the observer) |

plugins/ag-charts-generate-example-files/src/executors/generate/generator/utils/getDarkModeSnippet.ts
skewis6: Half of them, yes — done in 9e35296305.

The *values* have to stay: `stripOutExampleGeneratorCode` strips this snippet out of the code viewer and the Plunker/CodeSandbox exports using `DARK_MODE_REGEX`, which is built from the same two delimiters in `external/ag-website-shared/src/utils/extraCodeSnippets.ts`. Drop the markers and the harness code shows up in every example's published source.

The `export`s, though, were dead — nothing imports them, `examplesGenerator.ts` only takes `getDarkModeSnippet`. They're now plain `const`s, and the comment points at `extraCodeSnippets.ts` as the file that owns the values.

Worth noting this duplication predates the PR (same two exported consts on `latest`, with a NOTE saying as much). The plugin can't import `@ag-website-shared` — no path mapping, separate SWC build — so a single shared constant would mean either adding that mapping or a relative hop into `external/`. Happy to do it in a follow-up if you'd like it properly de-duplicated.

plugins/ag-charts-generate-example-files/src/executors/generate/generator/utils/getDarkModeSnippet.ts
skewis6: Yes, for now — it's doing two jobs neither of which CSS can do on its own, and it isn't new here (`latest` already assigns `optionsMutationFn` at line 11 of this same file; the PR changes what it returns, not the fact of the hook).

1. **Pointing the default theme's params at the variables.** The examples don't set a theme, so something has to inject `params: { backgroundColor: 'var(--ag-example-chart-background-color)', … }` into their options before the chart is built. Only the library or the example itself could avoid that, and we don't want ~900 examples each declaring a theme.

2. **The fallback swap.** Palette-sensitive series, non-default themes, presets, and containers the variables can't reach still need `ag-default-dark` resolved by name — a `var()` can't carry `theme.palette`.

Job 1 is the one that could go away: if `ag-default` declared its colour params as `var(--ag-charts-*, <current default>)` in the library, examples would pick the site's scheme up with no injection at all, and this hook would shrink to just the palette fallback. That's a library-side change well beyond this ticket, so I've left it — happy to raise it if it's worth pursuing.

packages/ag-charts-website/src/components/example-runner/getExampleDarkModeInitScript.ts
skewis6: Normally yes, but a server-generated nonce isn't available here — and for the examples scope it would actively make things worse. Taking both of your comments together:

**Why not a real nonce.** The site's CSP lives in `src/utils/htaccess/cspRules.ts` and ships as `.htaccess` headers on a static build; that file states outright that static hosting rules out per-request nonces. For the `examples` scope, `script-src` is `… 'unsafe-eval' 'unsafe-inline'` (`cspRules.ts`, the `scope === 'examples'` branch), and the same file warns that per CSP2+ **any** hash or nonce in a policy makes the browser ignore the `'unsafe-inline'` that scope still relies on. So introducing a genuine nonce for this one script would void `'unsafe-inline'` and break every other inline script the examples depend on.

**So what is `123123` doing?** It's the repo's existing fixture nonce, not a real secret and not part of the site policy:

- `FrameworkTemplate.astro:190` already emits `<script nonce="123123">` for the E2E style block, and sets the same value on the style element it injects.
- The four `security-e2e/_examples/*/head.html` fixtures declare `'nonce-123123'` in their own `<meta>` CSP.

It's also not load-bearing for this script: the fragment is injected *ahead of* the example's own `head.html` (deliberately, so it stays render-blocking even for an example that pins its own CSP), so it precedes any example-declared `<meta>` policy — and a meta-delivered CSP only governs what follows it. The nonce is there for consistency with the sibling script above it.

**On moving it to `public/`** — that would work: `'self'` is allowed by all four CSP fixtures, and a classic `<script src>` in `<head>` is still render-blocking, so `data-dark-mode` would still be set before the chart is created. It costs a request per example iframe and needs base-URL plumbing in both `FrameworkTemplate.astro` and `GalleryFrameworkTemplate.astro`, and it doesn't buy any CSP headroom given inline is already permitted for this scope.

Three options, your call and I'll make the change either way:

1. Leave as-is — consistent with `FrameworkTemplate.astro:190`.
2. Drop the `nonce` attribute entirely, since nothing enforces it here.
3. Move it to `public/` as an external script.

I'd lean to (1) or (2); (2) removes the magic value that prompted your comment without adding a request.

taktran: Ok leaving as 1 is fine.

plugins/ag-charts-generate-example-files/src/executors/generate/generator/utils/getDarkModeSnippet.ts
taktran: It's fine as is

plugins/ag-charts-generate-example-files/src/executors/generate/generator/utils/getDarkModeSnippet.ts
taktran: Fine for now

taktran (approved):
LGTM

taktran:
This will need to go into `b14.2.0`

PR #7965, opened 2026-08-27
AG-13301 Shrink text segments proportionally and let layout shrink captions

iMoses:
https://ag-grid.atlassian.net/browse/AG-13301

Fix #AG-13301


alantreadway (approved):
LGTM

PR #7966, opened 2026-08-28
AG-18340 Support label fitting options on Pie and Donut labels

iMoses:
https://ag-grid.atlassian.net/browse/AG-18340

Fix #AG-18340


alantreadway (approved):
LGTM

PR #7972, opened 2026-08-28
AG-18312 Apply themed padding to background region labels

lsjroberts:

Fix #AG-18312


packages/ag-charts-enterprise/src/features/background-regions/backgroundRegions.test.ts
```diff
@@ -358,7 +358,7 @@ const THEMED: AgCartesianChartOptions = {
                             },
                             fontSize: 14,
                             fontWeight: 'bold' as const,
-                            padding: { top: 12, right: 20, bottom: 12, left: 20 },
+                            padding: { top: 4, right: 40, bottom: 4, left: 8 },
```
lsjroberts, line 361: these were the same as the defaults before so showed no change 🤦‍♂️

alantreadway (approved):
LGTM

PR #7976, opened 2026-08-28
AG-18360 Centre the org chart when the held view becomes unreachable

lsjroberts:
https://ag-grid.atlassian.net/browse/AG-18360

Fix #AG-18360


alantreadway (approved):
LGTM

PR #7978, opened 2026-08-28
SE-141 Rewrite the gallery hub's copy and family headings

skewis6:
https://ag-grid.atlassian.net/browse/SE-141

The hub's title, H1, meta description and intro now resolve from a single `GALLERY_HUB_COPY` that the page and its `.md` twin both read, so the two describe the gallery identically. Each chart family renders as an H2 carrying the family's anchor id — the grid had no headings at all — and each card is anchored on the H1 its example page serves.

The rewritten H1 does not fit the fixed-height sticky toolbar, so it sits with the intro above the first card and scrolls with the page; the toolbar keeps "AG Charts Gallery" as a plain label. Two points worth a look in review: that label now reads as a near-duplicate of the H1 below it, and visible family headings reverse the AG-10044 decision to remove them.

The thumbnail alt-text guard is widened to any label that already names a chart. Without it "Bar Chart Example" would be given an alt of "Bar Chart Example Chart"; it also fixes a pre-existing case, where "Line Chart With Gaps" was already producing "Line Chart Chart with Gaps".

Fix #SE-141

### DIFF
--- a/packages/ag-charts-website/src/components/gallery/components/GalleryExampleImage.tsx
+++ b/packages/ag-charts-website/src/components/gallery/components/GalleryExampleImage.tsx
@@ -15,7 +15,7 @@ interface Props {
 }
 
 const optimizeAltTextForSeo = (label: string): string => {
-    if (label.toLowerCase().endsWith('chart')) {
+    if (label.toLowerCase().includes('chart')) {
         return label;
     }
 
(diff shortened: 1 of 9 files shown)


taktran:
This will need to go into `b14.2.0`

skewis6:
Closing as already merged to RC

PR #7979, opened 2026-08-28
AG-14332 Prevent data selection when click preventDefault() is called

olegat:


alantreadway (approved):
LGTM

PR #7980, opened 2026-08-28
AG-18312 Apply a single-number padding from a theme override

lsjroberts:

Fix #AG-18312


alantreadway (approved):
LGTM

PR #7983, opened 2026-08-28
AG-17613 Omit an axis with no visible series from event coordinates

olegat:
https://ag-grid.atlassian.net/browse/AG-17613

Toggling off the only series bound to a secondary axis left `coordinates` still reporting that axis, with `value: NaN`, `domain: [NaN, NaN]` and `index: -1`. The same action on a primary axis omitted it entirely, so the two directions disagreed.

An axis whose every series is hidden has an empty domain. A primary axis keeps that empty domain, so `scale.invert` returns null and `Axis.pickValue` bails out. A secondary axis additionally runs `calculateNiceSecondaryAxis` to align its ticks to the primary tick count, and there `findMinMax([])` returns `[]`: the destructured bounds are undefined and the arithmetic fabricates a [NaN, NaN] domain. That domain is not null, so it passes `pickValue`'s guard and reaches the listener.

Skip the alignment when the domain is empty. A secondary axis in this state then behaves exactly like a primary one - no ticks, and its key absent from `coordinates` rather than present with NaN values.

The tests hide each y axis in turn and assert the two behave identically. They check `axisId in coordinates` is false rather than merely undefined, since consumers branch on key presence, and that the axis still holding a visible series continues to resolve.

### DIFF
--- a/packages/ag-charts-community/src/chart/axis/generateTicks.ts
+++ b/packages/ag-charts-community/src/chart/axis/generateTicks.ts
@@ -282,6 +282,7 @@ function calculateRawTicks<TScale extends Scale<TDatum, number, TickInterval<TSc
     if (
         tickGenerationType === TickGenerationType.CREATE_SECONDARY &&
         primaryTickCount != null &&
+        domain.length > 0 &&
         ContinuousScale.is(scale)
     ) {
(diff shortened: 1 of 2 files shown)


alantreadway (approved):
LGTM

PR #7987, opened 2026-08-28
AG-18366 Report a hidden axis as undefined in event coordinates

olegat:
https://ag-grid.atlassian.net/browse/AG-18366


iMoses (approved):
LGTM

PR #7991, opened 2026-09-01
AG-15902 Improve callout labels collision with variable radius pie sectors

iMoses:
https://ag-grid.atlassian.net/browse/AG-15902

Fix #AG-15902


packages/ag-charts-community/src/chart/series/polar/donutSeries.ts
```diff
@@ -1471,16 +1621,225 @@ export class DonutSeries extends PolarSeries<
...
+                if (limit > 0) {
+                    for (const side of sides) {
+                        const offset = smallestClearingPush(d, placementOf(d, side), limit);
+                        if (offset != null) consider(side, offset);
+                    }
+                }
+
+                if (best.sectors > 0) {
+                    // Nothing clears the sectors, so settle for the side that at least moves off their mid-angle.
+                    place(sides.length > 1 ? sides[1] : undefined, 0);
```
olegat, line 1820: > **sector fallback discards `best` and can force a placement that was rejected as hidden**
>
> `consider` (Line 1781) refuses any candidate whose box `isBoxHidden` returns true, so such a placement can never become `best`. This fallback then ignores `best` entirely and hard-codes sides[1] - which may be exactly one of the placements that was skipped for being hidden. The label is placed where the algorithm already decided it cannot be seen, and the final pass at Line 2086 then hides it. It also discards a strictly better-ranked candidate: when no candidate clears the sectors, `best` still holds the placement with the fewest crossed callout lines (and the smallest offset), and that ranking is thrown away. Suggested: when best.sectors > 0, still honour best.side/best.offset, and fall back to sides[1] only when `best` is the untouched seed.

olegat (approved):
LGTM.

I strongly suspect that we'll see some unexpected side effects. But these should be limited to just pie/donut. Also, this change won't be going into this release so we've got time to test and address the side effects.

PR #7992, opened 2026-09-01
AG-18121: fix keyboard tabbing in the API reference

skewis6:
[AG-18121](https://ag-grid.atlassian.net/browse/AG-18121)

Charts equivalent of ag-grid/ag-grid#15024. The charts API reference is a different component, so this is a port of the same fix rather than a file copy.

Safari omits links and buttons from the tab order unless they carry an explicit `tabindex`. This declares it on the reference's disclosure buttons (`ChildPropertiesButton`, `UnionTypesButton`, `CodeCollapsibleButton`) and its links, and adds `aria-expanded` with state-aware labels — the labels previously read "See…" even when expanded, while the visible text already said "Hide".

Those three buttons use `button-as-link` / `button-style-none`, both of which strip the design system's focus ring, so they would otherwise become tabbable with no visible focus. The shared stylesheet restores an indicator, scoped to those two classes so the grid and studio reference buttons (`button-tertiary`) keep their existing ring.

**Not covered:** `OptionsNavigation.tsx` builds its nav tree from `<span onClick>` elements, and `Properties.tsx` has a span-based property-name expander. Those are keyboard-inaccessible in every browser and need a markup change rather than a tabindex, which is more than belongs on a release branch — worth a follow-up ticket.

**Testing:** ESLint, Prettier, `sass` compile, esbuild parse, `api-documentation` suite 34/34. Not browser-verified.


packages/ag-charts-website/src/components/api-documentation/components/ApiReference.tsx
```diff
@@ -469,7 +477,12 @@ function ApiReferenceRow({
                                     (child) => isValidElement(child) && child.type === 'code'
                                 );
                                 return (
-                                    <a href={href} className={hasCode ? 'meta-link' : undefined} {...props}>
+                                    <a
+                                        tabIndex={0}
+                                        href={href}
+                                        className={hasCode ? 'meta-link' : undefined}
+                                        {...props}
```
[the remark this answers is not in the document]
skewis6: Fixed in 4e7bf7784b. Extracted the anchor renderer into a shared `docsMarkdownComponents` and passed it to both Markdown instances, so union-variant descriptions get the same keyboard-reachable links as member descriptions.

packages/ag-charts-website/tsconfig.lint.tsbuildinfo
```diff
@@ -0,0 +1 @@
...,2451,2615],[65,71,1197,2451,2617,2618],[65,71,1197,2451,2620],[65,71,1197,2451,2621,2622,2623,2624,2625],[65,71,1197,2451,2626,2674,2675],[65,71,1070,1197,2451,2634,2644,2651,2654,2662,2668,2673],[65,71,1197,2451,2674],[65,71,1197,2451,2687,2688,2689],[65,71,1070,1197,2451,2681,2682],[65,71,1197,2451,2679,2683,2684,2685],[65,71,1197,2451,2476],[65,71,1197,2451,2699],[65,71,1197,2451,2607],[65,71,1197,2451,2701],[65,71,1070,1197,2451,2478],[65,71,1197,2451,2677],[65,71,1197,2451,2594,2678],[65,71,1197,2451,2680],[65,71,1197,2451,2692],[65,71,1197,2451,2680,2681,2703,2704,2705,2706,2707,2708,2709,2710,2711,2712,2713],[65,71,1197,2451,2680,2705],[65,71,1070,1197,2451,2679,2714],[65,71,1197,2680,2703,2705],[65,71,1197,2451,2547,2680,2705],[65,71,1070,1197,2451,2716],[65,71,1070,1197,2451,2717],[65,71,1197,2451,2718],[65,71,2455],[65,71,843,1070,1197,2451,2453,2454,2456,2457,2461,2464,2466,2469,2471,2473,2475,2592,2594,2596,2600,2602,2606,2610,2612,2616,2619,2634,2644,2651,2654,2662,2668,2673,2676,2686,2690,2698,2700,2702,2715,2719,2731,2743,2747,2759,2775,2779,2786,2791,2795,2804,2810,2818,2821,2827,2832,2837,2856,2861,2865,2869,2870,2871,2872,2876,2877,2878,2882,2883,2884,2886],[65,7
```
taktran: Is this file intentional?

skewis6: No — that was an accidental commit of a local build artefact. `tsconfig.lint.json` sets `composite: true`, so the type-check I ran to verify the previous fix emitted a `.tsbuildinfo` even under `--noEmit`, and it got swept into that commit. Removed in d402c6bfd2, so the diff is back to just the three source files.

Worth noting separately: nothing in `.gitignore` covers `*.tsbuildinfo`, and this was the only one tracked in the repo — so the same trap is there for anyone who runs that config by hand. I have left that out of this PR to keep a release-branch change minimal, but happy to raise it as a follow-up.

taktran (approved):

PR #7994, opened 2026-09-01
SE-143 Fail the build on a retired gallery title, meta or H1

skewis6:
https://ag-grid.atlassian.net/browse/SE-143

Fix #SE-143

### DIFF
--- a/packages/ag-charts-website/astro.config.mjs
+++ b/packages/ag-charts-website/astro.config.mjs
@@ -23,10 +23,12 @@ import agCssAsString from './plugins/agCssAsString';
 import agDemosStatic from './plugins/agDemosStatic';
 import agDevCsp from './plugins/agDevCsp';
 import agDevMarkdownNegotiation from './plugins/agDevMarkdownNegotiation';
+import agGallerySeoChecker from './plugins/agGallerySeoChecker';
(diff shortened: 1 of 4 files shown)


skewis6:
Addressed in bb9e437e2d, with one deviation from the suggested fix.

The gap was real: the retired description (`Example of a {title} Chart built with AG Charts JavaScript Charting Library. Edit source code with CodeSandbox & Plunker. ...`) contains "Example" and is non-empty, so it passed. The checker now rejects it and a test covers it.

It is matched on `CodeSandbox & Plunker`, not on the brand phrasing the review pointed at. `AG Charts JavaScript Charting Library` is still live copy in `src/content/homepage/homepage.json`, `src/content/metadata/metadata.json` and `src/utils/markdown-pages/apiReferencePageContent.ts`, so matching it would fail the build on legitimate pages. The dead editor names appear nowhere else in the site and identify the retired string uniquely. A test asserts the live brand phrasing still passes.

For the record on scope: AC1 asks for a title matching the retired pattern, a title without "Example", or "an empty title or meta", so retired-description detection was not a stated requirement of the ticket. It is a fair gap to close regardless, hence the fix.

taktran (approved):

PR #8001, opened 2026-09-02
AG-18353 Refactor: clickedOn→type, allClickParams→allMatchedParams

olegat:
https://ag-grid.atlassian.net/browse/AG-18353
* Rename `allClickParams` to a more neural name `allMatchedParams`
* Brand hit objects using `event.type` instead of context-menu `showOn`


iMoses (approved):
LGTM

PR #8002, opened 2026-09-02
Fix the failing security workflow and the failing OSSF Scorecard check

seanlandsman:
Resolves the failing `security.yml` workflow and the failing OSSF Scorecard check.

Both Snyk gating jobs now pass locally against the workflow's own invocations, and the Scorecard score is projected to clear its threshold.

## Dependency scan — was 15 high/critical findings, now 0

- **All 15 cleared.** Most were a stale lockfile against ranges `package.json` already declared, so relocking was enough — no range widening.
- **Two production-runtime findings, both fixed at source rather than suppressed:**
  - `ws` 8.18.2 → 8.21.3 (CVE-2026-48779, and CVE-2026-45736 which an existing suppression had been hiding). Shared with the published `ag-charts-server-side` via `jsdom`.
  - `follow-redirects` 1.15.11 → 1.16.0 (CVE-2026-40895), reached through `skia-canvas`. The existing `**/http-proxy/follow-redirects` resolution never applied here because it is scoped to the `http-proxy` subtree — worth knowing, as that pattern reads broader than it is.
- **`@vitest/mocker` time-boxed.** Advisory published mid-work; `vitest` is pinned to exactly `2.1.9` in five manifests, so there is no range to relock within and the fix needs vitest 4.x. Dev-only, expiring 2027-03-02. **The vitest 4.x upgrade is worth scheduling** — this repo is the only one still on 2.1.9 and it attracted a finding within hours.
- Verified: **zero** suppressed findings inside the published production runtime closure (3 third-party roots, 72 packages, computed by traversal from the packages that actually publish).

## SAST — was 9 findings, now 0

`snyk-viewer/index.mjs` excluded from Snyk Code. It is a manually-run, localhost-only developer tool, never built or shipped, and every flagged sink is already guarded before the read or write. Rule-level `ignore:` entries are inert for Snyk Code (verified), and Snyk Code reads only the scan-root policy, so this cannot live in the `ag-shared` subrepo and is duplicated in each repo root.

## Ignore-policy hygiene

- All expiries extended to **2027-03-02**. The previous cohort expired at exactly the timestamp of the weekly cron.
- **Ignore placement corrected.** `snyk test --yarn-workspaces` evaluates each workspace project against *its own* `.snyk`, not the root — so an entry in the root policy silently covered nothing. Four existing entries were also pinned to a beta version no longer in the tree and had stopped matching. Both classes are now fixed, and new entries use version-agnostic paths so they survive the next bump.

## OSSF Scorecard — was 7.1 against a 7.3 threshold

- **Token-Permissions 0 → 10.** `ci-failure-handler.yml` was the only workflow without a top-level `permissions:` block, which Scorecard treats as write-all; `cloud-cache-bake.yml` had top-level `contents: write`, now scoped to the single job that needs it for `gh release`. The other 22 job-level writes are load-bearing and unchanged — they sit under declared `read` top levels and carry no penalty.
- **122 GitHub Action references pinned** to commit SHAs with accurate version comments, resolved via the API. Four resolutions independently reproduced pins already in the repo, byte for byte.
- **`sonarsource/sonarqube-scan-action` was tracking `master` with no tag** — running whatever that branch head happened to be on every invocation. Now pinned with a dated comment, so it needs deliberate bumps rather than silent drift.
- **4 of 6 OSV advisories cleared** (3 by the dependency work above, 1 by relocking `postcss-selector-parser`). The two remaining are held by exact pins outside this repo's control and ride the Angular and bundler majors.
- Projected: Token-Permissions 0 → 10, Pinned-Dependencies 0 → 5, Vulnerabilities 4 → 5, aggregate **7.1 → ~8.2** (worst modelled case 7.82).

## Verification

- Snyk dependency scan: **exit 0**, 0 findings, 19/19 projects, 4,537 dependencies
- Snyk Code: **exit 0**, 0 results
- `build:types` and `lint`: clean on all eight packages
- Tests: green except `quadrantPreset > ITEM_STYLERS_NUMERIC`, a pre-existing 3-pixel snapshot diff that reproduces identically on unmodified `latest`


taktran (approved):
LGTM

PR #8003, opened 2026-09-02
AG-18378 Fire chart-level click on axis areas without a listener

olegat:
https://ag-grid.atlassian.net/browse/AG-18378

An axis proxy element covers the axis area whenever any axis feature asks for one, and the context menu is enabled by default, so every enterprise cartesian chart has a proxy over each axis. Chart-level click/doubleClick is dispatched from the chart container and requires an exact target match, so a click landing on a proxy was dropped - the listener fired in community but not in enterprise.

-   Hand a click that no axis listener wants back to the chart-level
    click/doubleClick listener, checked per axis so an axis without a listener
    still falls back when a sibling axis has one. Skipped when zoom is enabled,
    which keeps zoom's axes suppressed as they were before.
-   Drop the two chart-wide hasClickListeners() gates on the axis click and
    dblclick handlers, so the per-axis check is reached when no axis has a
    listener at all.

Test harness (findTarget.ts) previously routed a click to an axis proxy only when zoom or an axis click listener existed, so it modelled a DOM that no longer exists and could not observe this behaviour:

-   Route to a proxy whenever it is visible and pointer events are enabled,
    matching `.ag-charts-proxy-elem { pointer-events: auto }`.
-   Treat proxy bounds as the half-open interval a DOM box occupies, so a proxy
    no longer claims the edge pixel of the series area it abuts. boxContains is
    inclusive at both edges, which mis-routed clicks on cone-funnel nodes that
    sit exactly on that boundary.
-   Move the suppressed-overflow-label click from x=800 to x=799: the axis spans
    x 25..800 of an 800px canvas, so 800 was off the element and unreachable by
    a real click. Its `index` expectation is unchanged; `value` is the
    interpolated 998.71 rather than a clamped 1000.

### DIFF
--- a/packages/ag-charts-enterprise/src/features/axis-dom-proxy/axisDomProxy.test.ts
+++ b/packages/ag-charts-enterprise/src/features/axis-dom-proxy/axisDomProxy.test.ts
@@ -916,6 +916,34 @@ describe('AxisDOMProxy', () => {
             expect(chartClick).toHaveBeenCalledTimes(1);
         });
 
+        // A listener for one event type must not suppress the fallback for the other: the dispatch below
+        // picks its listener by event type, so the unmatched event would otherwise reach nothing at all.
+        test('a click-only axis listener leaves doubleClick to the chart', async () => {
+            chart = await createChart({ axes: { x: { listeners: { click } }, y: {} } });
+
+            await doubleClickAction(...X_AXIS_AREA)(chart);
+            await waitForChartStability(chart);
+
+            expect(click).toHaveBeenCalledTimes(2);
+            expect(chartClick).not.toHaveBeenCalled();
+            expect(chartDoubleClick.mock.calls).toMatchObject([[expect.objectContaining({ type: 'doubleClick' })]]);
+        });
+
+        test('a doubleClick-only axis listener leaves click to the chart', async () => {
+            const doubleClick = vi.fn();
+            chart = await createChart({ axes: { x: { listeners: { doubleClick } }, y: {} } });
+
+            await doubleClickAction(...X_AXIS_AREA)(chart);
+            await waitForChartStability(chart);
+
+            expect(doubleClick).toHaveBeenCalledTimes(1);
+            expect(chartDoubleClick).not.toHaveBeenCalled();
+            expect(chartClick.mock.calls).toMatchObject([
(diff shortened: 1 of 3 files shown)


packages/ag-charts-enterprise/src/features/axis-dom-proxy/axisDomProxy.ts
```diff
@@ -323,6 +323,16 @@ export class AxisDOMProxy extends AbstractModuleInstance {
         if (!div) return;
 
         const axisCtx = this.ctx.axisManager.getAxisIdContext(axisId);
+        // AG-18378: An axis proxy covers the axis area even when only the context menu needs it, so a click no axis
+        // listener wants is handed back to the chart-level listener - unless zoom owns the axis interaction.
+        if (axisCtx != null && !hasAxisClickListener(this.ctx.chartService, axisCtx)) {
```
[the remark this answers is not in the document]
olegat: Thanks. Resolved.

packages/ag-charts-enterprise/src/features/axis-dom-proxy/axisDomProxy.test.ts
```diff
@@ -341,12 +348,13 @@ describe('AxisDOMProxy', () => {
...
         test('click - value/index match', async () => {
-            await clickAction(800, 560)(chart); // far right, where '1000' hid itself
+            // 799, not 800: the axis spans x 25..800 of an 800px canvas, so 800 is off the element.
+            await clickAction(799, 560)(chart); // far right, where '1000' hid itself
             await clickAction(25, 560)(chart); // far left, where '0' hid itself
             await waitForChartStability(chart);
 
             expect(click.mock.calls).toMatchObject([
-                [expect.objectContaining({ value: 1000, index: 5 })],
+                [expect.objectContaining({ value: expect.closeTo(998.71), index: 5 })],
```
olegat, line 357: This is very suss. I'll verify this manually.

olegat: 👍🏻 Seems legit. X starts at 0, so 800 is indeed off-target. The chart does not draw the x=1000 gridline, which means the visible domain actually stops right before 1000.
Image

iMoses (approved):
LGTM

PR #8006, opened 2026-09-02
SE-166 Emit internal links that resolve without a 301 (b14.1.0)

seanlandsman:
Backport of the SE-166 builder fix to b14.1.0, with one deliberate divergence.

b14.1.0 predates the ES-modules example runner and still uses SystemJS. `SystemJs.tsx` injects `var appLocation = '${appLocation}'` as a raw string, consumed by `systemjs.config.js` as `map: { app: appLocation }` with `main: './main.ts'`. Slashing `getExampleUrl` — safe where every consumer re-joins through `pathJoin` — would have changed `appLocation` here. So on this branch `getExampleUrl` stays unslashed and documented as a base, and a new `getExampleLinkUrl` serves the places the browser actually navigates or fetches (example-runner iframe, "open in new tab", gallery markdown, markdoc live URL, debug listing). `appLocation` is byte-identical to what ships today; two new tests assert the base stays unslashed.

### Shared URL canonicaliser

Review on the sibling PRs found that the first cut skipped any URL containing `?` or `#`, so `chartsUrlWithPrefix({ url: '/gallery?utm_source=x' })` still emitted the 301ing `/charts/gallery?utm_source=x` (confirmed live). `external/ag-website-shared/src/utils/addTrailingSlash.ts` now exports `addTrailingSlashToPath`: split the URL into pathname and suffix at the first `?`/`#`, insert the slash into the pathname, and leave the URL alone only when the pathname is empty (`#section`, `?x=1`, the builders' default `''`), already ends in `/`, or names a file (`getFileExtension`, with a purely numeric "extension" exempted so `/archive/13.3.1` — a live 301 to `/archive/13.3.1/` — keeps its slash). `chartsUrlWithPrefix`, `gridUrlWithPrefix`, `getArchiveUrl` and `getDocumentationArchiveUrl` all route through it. The legacy `addTrailingSlash` stays in the same file, unchanged, for the shared search components (`SearchModal.jsx`, `SearchHits.jsx`) that import it. The table-driven test in `addTrailingSlash.test.ts` fails 21 cases against the previous helper. The shared file is byte-identical to the ag-grid and ag-studio copies of `external/ag-website-shared`.

### History

The branch was rebased onto the current `b14.1.0` tip (force-pushed once, `--force-with-lease`) to resolve a conflict with SE-145 (#7968): a single import block in `buildGalleryExampleMarkdown.ts`, where SE-145's `resolveGallerySeo` import sits above the `urlPaths` import this PR changes. Nothing else moved. An intermediate commit (`9eb1b22633`) moved the helper into its own file and deleted `addTrailingSlash.ts`, which broke the website build via the two `.jsx` importers above; `3af391b771` restores the layout and is the version to review.

### Link checker

The build's link checker now fails on any internal link that would redirect: missing trailing slash, doubled slash, non-canonical host (`ag-grid.com`, `charts.ag-grid.com`, `studio.ag-grid.com`) or `http://` scheme, via `external/ag-website-shared/src/utils/internalLinkShape.ts` (`cf5de05976`, shared with the ag-grid and ag-studio copies). Every `<a href>` is checked before the existing ignore/resolution logic, so absolute `https://www.ag-grid.com/...` links and `?`/`#` links are in scope too. First run over the built site flagged exactly 10 links, all `https://charts.ag-grid.com/archive/<pre-10.1>/documentation/` from `getArchiveUrl` — live those return 200 only on `charts.ag-grid.com` (the `www` form is a 404), so the rule was corrected to treat `/archive/` on that host as canonical; the re-run is clean ("Link checker completed with no issues."). `plugins/agLinkChecker.ts` is aligned to the current shared revision, which also brings this branch the `#manage_cookies` footer-anchor ignore entry it lacked.

### E2E shard 6

The first two CI runs failed `api-ref-page.spec.ts` with `504 (Outdated Optimize Dep)` / `[astro-island] Error hydrating ApiReferencePage.tsx` / `Execution context was destroyed`. Reproduced from a cold dev server on `origin/b14.1.0` itself (7 failed / 21 passed, four runs, branch and base identical): Vite discovers `astro/virtual-modules/transitions-*.js`, `algoliasearch/lite` and `react-instantsearch` after its initial crawl, re-optimises and reloads the page ~8 s in. `latest`/b14.2.0 already pre-declare them in `optimizeDeps.include` (AG-17715, `781e02b9f7`); that hunk is back-ported here (`126cef53b8`), and the same cold-server harness then passes 28/28.

With that gone, two tests (`:175`, `:192`) still lost all three attempts with only `page.evaluate: Execution context was destroyed` at `fixture.ts:47`. Pressing Enter on a property-search result calls Astro's `navigate()`; b14.1.0's `Layout.astro` has no `<ClientRouter />` (b14.2.0 gained it in AG-17715), so that is a hard `location.href` load, and the fixture's post-action `waitForCharts` evaluate races it. Pre-existing on the base: run 33617370276 (shard 17) failed the same test with the same error and passed only on retry, and all ten other recent b14.1.0 PRs show 2–4 flaky on that shard. `e3b57a8e95` makes the settle evaluate tolerate a navigation (wait for `load`, settle again, up to three rounds); tests unchanged. Reproduced locally with CDP CPU throttling at 6x, `--repeat-each=20 --retries=0`: before 7 passed / 33 failed, after 40 passed (also 40/40 at 8x; whole spec 28/28 at 6x and unthrottled). `79ef3e6891` adds the `E2E_CPU_THROTTLE` knob used for that reproduction; unset, it does nothing.

### Verification

- `nx run-many -t test:vitest lint -p ag-charts-website ag-website-shared --skip-nx-cache`: ag-website-shared 240 tests pass; ag-charts-website 717 / 718 pass, 0 lint errors. **The one failure is pre-existing on `b14.1.0` since SE-163 (#8000)**: `markdownPages.test.ts › leaves the contact result pages out of the sitemap` asserts `/contact/` is in the sitemap, which SE-163 now canonicalises to the grid site. Same stale assertion as on `latest`; needs its own fix.
- `nx build ag-charts-website` (with `CHECK_LINKS=true`): exit 0, "Link checker completed with no issues."
- `nx format:check --base origin/b14.1.0`: clean.

`external/ag-website-shared` is a git-subrepo; these changes need a subrepo push or they will be rebased away. The same shared fix is in the ag-grid and ag-studio PRs and should be reconciled upstream once.

Jira: SE-166


packages/ag-charts-website/src/components/docs/utils/urlPaths.ts
```diff
@@ -29,6 +29,10 @@ export const getExamplePageUrl = ({ framework, path }: { framework?: Framework;
 
 /**
  * Dynamic path where examples are
+ *
+ * Deliberately unslashed: this is a base, not a link. It becomes `appLocation`, which
+ * `SystemJs` injects verbatim as the SystemJS `app` module-map target, so its exact spelling is
```
taktran, line 34: We don't use systemjs anymore?

taktran (approved):
There are a couple of references to systemjs, which don't seem relevant anymore. Otherwise LGTM

packages/ag-charts-website/src/components/docs/utils/urlPaths.ts
taktran: Nevermind, just read the PR description

packages/ag-charts-website/astro.config.mjs
```diff
@@ -130,6 +130,19 @@ export default defineConfig({
                 'ag-charts-react',
                 'ag-charts-vue3',
             ],
+            // Dev-server only. These are reachable exclusively through a lazily hydrated island
+            // or (for the transitions runtime) a virtual module the dep crawler cannot follow, so
+            // Vite discovers them after its initial crawl, re-optimises, and reloads the page —
+            // 504-ing whichever island requests were in flight. Pre-declaring them keeps the
+            // first load stable.
+            include: [
```
taktran, line 138: This seems like a weird inclusion

seanlandsman: This is the `optimizeDeps.include` block AG-17715 added in `781e02b9f7` (10 Aug); it's on `b14.2.0` and `latest` but never reached `b14.1.0`. The block here is byte-identical to b14.2.0's.

Without it, E2E shard 6 fails 7 of 29 `api-ref-page` tests on every PR-shaped run of this branch: these six deps sit behind a lazily-hydrated island / a virtual module, so Vite only finds them after its initial crawl, re-optimises ~8s in and reloads the page, and the in-flight `ApiReferencePage` hydration gets `504 (Outdated Optimize Dep)`. Cold-dev-server repro on this branch: 7 failed / 21 passed → 28 passed / 0 failed with the block (twice), and the five affected tests went green in CI on the next run.

seanlandsman:
**Why a links PR touches `astro.config.mjs` and the e2e fixture: E2E shard 6 (`api-ref-page.spec.ts`) had two pre-existing failures on b14.1.0, both fixed here.**

The first two CI runs of this branch failed shard 6 with `504 (Outdated Optimize Dep)`, `[astro-island] Error hydrating … ApiReferencePage.tsx` and `page.evaluate: Execution context was destroyed`. Nothing in this PR touches `src/components/api-documentation/`, `e2e/`, `urlWithPrefix.ts` or `urlWithBaseUrl.ts`, and the legacy `addTrailingSlash` used by the site search is byte-identical to the base — so the failures were investigated rather than rerun.

**Root cause 1 — Vite dep-optimiser reload (`126cef53b8`).** Cold-dev-server repro, `node_modules/.vite` cleared, `astro dev` started fresh, `playwright test e2e/api-ref-page.spec.ts` (29 tests):

| run | head | result |
|---|---|---|
| branch ×2 | `3af391b771` | 7 failed / 21 passed / 1 skipped |
| base ×2 | `9e4a851330` (`origin/b14.1.0`) | 7 failed / 21 passed / 1 skipped |

Same seven console-assertion tests every time on both heads. Astro log ~8 s after start: `[vite] ✨ new dependencies optimized: algoliasearch/lite, react-instantsearch, astro/virtual-modules/transitions-router.js, …` → `optimized dependencies changed. reloading`. `latest`/b14.2.0 already pre-declare exactly those modules in `optimizeDeps.include` (AG-17715, `781e02b9f7`); that 13-line hunk is back-ported. Cold repro with it: 28 passed / 0 failed / 1 skipped, twice, no reload line. Provenance, scope (dev server only) and the option of splitting it into its own PR are on the review thread: https://github.com/ag-grid/ag-charts/pull/8006#discussion_r3923188711

**Root cause 2 — hard-navigation race in the e2e fixture (`e3b57a8e95`).** With the reload gone, `:175 navigates to bar series via search keyboard flow` and `:192 clears search query…` still lost all three attempts with only `Execution context was destroyed` at `fixture.ts:47`. Pressing Enter on a property-search result calls Astro's `navigate()` (`apiReferenceRouting.ts`, ported to b14.1.0 in `d6ad59b056`); b14.1.0's `Layout.astro` has no `<ClientRouter />` (b14.2.0 gained it in AG-17715), so that is a hard `location.href` load, and the fixture's post-action `waitForCharts` evaluate races it. This is pre-existing: the base run 33617370276 (shard 17) failed the same test with the identical error on attempt 0 and passed only on Playwright's retry, and all ten other recent b14.1.0 PRs report 2–4 flaky in shard 6, with these two tests retrying in 8 of 10. The fixture now treats a settle evaluate destroyed by a navigation as the navigation: wait for `load`, settle again (bounded to three rounds, since `load` can resolve for the outgoing document). Tests unchanged. Reproduced locally with CDP CPU throttling, `--repeat-each=20 --retries=0` on the two tests:

| throttle | before | after |
|---|---|---|
| none | 20 / 20 pass | — |
| 6x | 7 passed / 33 failed | 40 passed / 0 failed |
| 8x | — | 40 passed / 0 failed |

Whole spec after: 28 passed / 1 skipped at 6x and unthrottled. `79ef3e6891` adds the `E2E_CPU_THROTTLE` knob used for that reproduction (no effect when unset) — reviewers can drop it independently.

Why nobody saw either before: the base runs are pushes, planned push-shaped (six example variants, 32 shards); PR runs are PR-shaped (vanilla only, 16 shards) — `ci.yml`'s shard-planning comment, identical on base and branch. With 16 shards `api-ref-page.spec.ts` runs first in shard 6 against a cold server.

taktran (approved):
LGTM

PR #8007, opened 2026-09-02
AG-18379 Add per-instance module registration to AgCharts.create

iMoses:
https://ag-grid.atlassian.net/browse/AG-18379

Fix #AG-18379


packages/ag-charts-community/src/chart/chartProxy.ts
```diff
@@ -247,6 +248,13 @@ export class AgChartInstanceProxy implements AgChartProxy {
         return this.chart?.setSelection(items);
     }
 
+    isModuleRegistered(moduleId: string): boolean {
+        if (!this.chart) throw new Error(DESTROYED_ERROR);
+
+        const moduleName = findExpectedModuleName(moduleId) ?? moduleId;
+        return this.chart.chartOptions.moduleRegistry.hasModule(moduleName);
```
olegat, line 256: This is always returning `false` on some modules like `ZoomInteractionModule` and `AxisInteractionModule`.
https://plnkr.co/edit/KUBIG6s98hM9w3MJ?open=main.ts

I suspect that `isModuleRegistered('QuadrantPresetModule')` might also be affected, but I've not verified this.

I don't know whether it matters much that `ZoomInteractionModule` and `QuadrantPresetModule` aren't found. But the fact that `AxisInteractionModule` isn't found is probably worth fixing, because right now as things stand that module is explicitly required for the new upcoming axis-click feature (https://ag-grid.atlassian.net/browse/AG-9809).

packages/ag-charts-types/src/chartBuilderOptions.ts
```diff
@@ -271,6 +271,15 @@ export interface AgTypedChartInstance<TDatum, TContext, O extends AgChartInstanc
      */
     clearSelection(): void;
 
+    /**
+     * Returns whether a module is available to this chart, either registered globally or passed to it
+     * via `AgCharts.create(options, { modules })`.
+     *
+     * @param moduleId the exported name of an individual module, such as `'LineSeriesModule'`; bundles such as
+     * `AllCommunityModule` are not module names.
```
olegat: The `@param moduleId` comment is misleading. It claims that bundle-names are not valid input, but the  `'QuadrantChartModule'` input (which is a bundle) returns `true`.
https://plnkr.co/edit/PssNGjrNvARjDEym?open=main.ts

libraries/ag-charts-eslint-rules/rules/validate-module-registration.mjs
olegat, line 1: I'm not familiar with this custom linter, but Claude Code found some error in here. I've not verified these.

libraries/ag-charts-eslint-rules/rules/validate-module-registration.mjs
```diff
@@ -769,22 +869,49 @@ export default {
...
+                        for (const moduleId of modules) {
+                            if (bundleContents.has(moduleId) || intrinsicDefaultSet.has(moduleId)) continue;
+                            const usedByOwner = [...requiredModules.entries()].some(
+                                ([reqMod, info]) =>
+                                    (info.owners.has(owner) || info.owners.has(null)) &&
+                                    (reqMod === moduleId || bundleContents.get(reqMod)?.includes(moduleId))
+                            );
+                            if (usedByOwner) continue;
+                            context.report({
+                                node: arrayNode.elements.find((e) => e?.type === 'Identifier' && e.name === moduleId),
```
olegat:
```
libraries/ag-charts-eslint-rules/rules/validate-module-registration.mjs:906:39: error: 'rule crashes ESLint when two AgCharts.create calls share one options variable' [-Werror-lint-crash]
  906 |                                 node: arrayNode.elements.find((e) => e?.type === 'Identifier' && e.name === moduleId),
      |                                       ^
      | `extractInstanceModules` keys by the options variable name and accumulates every create call's
      | module ids into one entry, but keeps only the FIRST call's `arrayNode`
      | (`instanceModules.get(owner) ?? { modules: new Set(), arrayNode }`). A module id contributed by a
      | later call is therefore not present in the stored array, `find` returns undefined, and
      | `context.report({ node: undefined })` aborts the whole file.
      |
      | REPRODUCED against this branch's rule file with eslint 9.37.0:
      |   const shared = { series: [{ type: 'line', ... }], axes: { ... } };
      |   AgCharts.create(shared, { modules: [LineSeriesModule, NumberAxisModule, OrdinalTimeAxisModule] });
      |   AgCharts.create(shared, { modules: [ContextMenuModule] });
      | =>  Error: Node must be provided when reporting error if location is not provided
      |     Occurred while linting the probe file, reported against its first line
      |     Rule: "aglint/validate-module-registration"
      |     at Program:exit - the context.report call in the rule file (Line 905)
      | This is a lint abort, not a lint error: the file's other violations go unreported too. Track the
      | arrayNode per module id, or per create call rather than per owner.

```

libraries/ag-charts-eslint-rules/rules/validate-module-registration.mjs
```diff
@@ -95,9 +98,46 @@ export default {
...
+         * `AgCharts.create*` call they are inlined into, or null when not attributable.
+         */
+        function ownerOf(node) {
+            if (!node) return null;
+            const ancestors = context.sourceCode.getAncestors(node);
+            const createCall = ancestors.find((a) => isCreateCall(a));
+            if (createCall) return createCall;
+            const declarator = ancestors.find(
+                (a) =>
+                    a.type === 'VariableDeclarator' && a.id.type === 'Identifier' && a.parent.parent.type === 'Program'
```
olegat:
```
libraries/ag-charts-eslint-rules/rules/validate-module-registration.mjs:123:86: error: 'exported options defeat owner attribution, so one chart's per-chart modules silently satisfy another chart's requirements' [-Werror-lint-false-negative]
  123 |                     a.type === 'VariableDeclarator' && a.id.type === 'Identifier' && a.parent.parent.type === 'Program'
      |                                                                                      ^
      | For `export const options = {...}` the ESTree shape is
      | ExportNamedDeclaration > VariableDeclaration > VariableDeclarator, so `a.parent.parent.type` is
      | 'ExportNamedDeclaration', not 'Program'. `ownerOf` falls through and returns null, and
      | `isRequirementSatisfied` then treats a null owner as satisfiable by `expandedAll` - the union of
      | the global registrations and EVERY chart's per-chart modules (see Line 135). A module supplied to
      | chart B silently satisfies chart A's requirement, which is precisely the cross-chart leak this
      | rule was extended to catch.
      |
      | REPRODUCED against this branch's rule file with eslint 9.37.0. Same file, two charts, each
      | passing its own `modules`; chart A needs AreaSeriesModule but is given only LineSeriesModule,
      | chart B is given AreaSeriesModule:
      |   export const exportedOpts = { series: [{ type: 'area', ... }], ... };   -> NO error reported
      |   const       plainOpts    = { series: [{ type: 'area', ... }], ... };   -> error reported:
      |        "Module 'AreaSeriesModule' is required for series type 'area' but not registered"
      | The `export` keyword is the only difference between the two runs. The new fixture cases 17 and 18
      | both use a plain top-level `const`, so nothing exercises this.
      |
      | Two related attribution gaps in the same function, worth fixing together: options passed as a
      | spread (`AgCharts.create({ ...options, container }, { modules })`) make the owner the call node
      | while the requirements inside the top-level `options` are owned by the variable name, giving the
      | opposite result - a false positive; and the `info.owners.has(null)` escape in the
      | over-registration loop (Line 901) inherits the same weakness.
```

packages/ag-charts-community/src/api/agCharts.ts
```diff
@@ -90,6 +93,19 @@ function takeOptionsArgumentIssue<O>(options: O, methodName: string): { options:
...
+function usesEnterpriseModules(moduleScope: ModuleScope): boolean {
+    for (const module of moduleScope.listModules()) {
+        if (module.enterprise) return true;
+    }
+    return false;
+}
+
+// Public module types describe only the identifying fields; the runtime objects are full definitions.
+function instanceModules(params?: AgChartParams): Array<ModuleDefinition | ModuleDefinition[]> | undefined {
+    return params?.modules as Array<ModuleDefinition | ModuleDefinition[]> | undefined;
```
olegat: This `as` has the potential to cause unexpected errors elsewhere in the codebase. I wonder whether there's a better approach we can take.

Perhaps `ChartInternalOptionMetadata.modules` should just be of type `AgChartParams` so type assumptions can be checked instead of asserted?

packages/ag-charts-community/src/chart/chartProxy.ts
iMoses: Good catch. The lookup maps the export name to the registry name through `ExpectedModules`, which only listed modules that own an options key, so plugins like `AxisInteractionModule` (`axis-dom-proxy`) never resolved. Added the entry there and in the lint mappings, plus an enterprise test that checks `AxisInteractionModule`, `ZoomModule` and `QuadrantChartModule` by name with the real modules registered. `ZoomInteractionModule` and `QuadrantPresetModule` aren't exported from the enterprise entry point, so `false` is correct for those. Fixed in d470b54285.

packages/ag-charts-types/src/chartBuilderOptions.ts
iMoses: Fair. `QuadrantChartModule` is the odd one out because the quadrant preset is only exported as a bundle, so there's no individual module name to check. Reworded the JSDoc to say bundles aren't accepted except for that case, and covered it in the enterprise test.

libraries/ag-charts-eslint-rules/rules/validate-module-registration.mjs
iMoses: Both were real, thanks. Fixed in 7f027ad511 with fixture cases 19 to 21 covering them, replies on each thread below.

libraries/ag-charts-eslint-rules/rules/validate-module-registration.mjs
iMoses: Reproduced. The entry kept only the first call's array, so a module from the second call had no node to report against. Each module id now remembers the array it came from, and missing-module reports are deduped per fix target so a per-chart gap isn't swallowed by an unrelated global one. Fixture case 19 is the repro.

libraries/ag-charts-eslint-rules/rules/validate-module-registration.mjs
iMoses: Both fixed. `export const options` now counts as a top-level declarator, and a spread `{ ...options, ... }` registers the modules for the call and for every spread variable, with over-registration judged per array element across all its owners. Cases 20 and 21 cover the two shapes, and all 890 website examples still lint clean under the changed rule.

packages/ag-charts-community/src/api/agCharts.ts
iMoses: Agreed, the cast was hiding the check. `ChartInternalOptionMetadata.modules` is now the public `AgChartModule[]` type, and a small `resolveInstanceModuleScope` helper narrows each entry (known `type`, string `name` and `version`) before it reaches the registry, throwing a clear error for anything else. No cast left, and there's a community test for the rejection path.

olegat (approved):
LGTM

PR #8008, opened 2026-09-02
AG-18387 Show crosslines on axes min/max

olegat:
https://ag-grid.atlassian.net/browse/AG-18387


packages/ag-charts-community/src/chart/crossline/cartesianCrossLine.ts
olegat, line 1: > [!IMPORTANT]
> This is the fix. The rest is just tests.

iMoses (approved):
LGTM

PR #8011, opened 2026-09-02
AG-17886 Fix financial demo chart range, toolbar and demo notice

yinkelly:
https://ag-grid.atlassian.net/browse/AG-17886

Fix #AG-17886


packages/ag-charts-demos/src/demos/financial/components/FinancialChart.tsx
```diff
@@ -44,72 +96,156 @@ function createFinancialOptions(
...
+            const zoomedOut = event.ratioX.start <= 0 && event.ratioX.end >= 1;
+            const start = zoomedOut ? undefined : rangeStartTime(event.rangeX);
+            propsRef.current.onRetainFrom(start == null ? undefined : start - RETAIN_MARGIN_BARS * BAR_INTERVAL_MS);
+
+            viewStartRef.current = start;
+            const wasLive = liveRef.current;
+            liveRef.current = zoomedOut || event.ratioX.end >= LIVE_EDGE_RATIO;
+            // Reset means "back to the live range" here: rendering the whole pinned history instead
+            // would snap back to the baseline window as soon as the feed evicted it.
+            const reset = zoomedOut && dataRef.current.length > propsRef.current.rangeMinutes;
```
olegat: If I zoom out fully (ie. `ratioX.start===0 && ratio.end===1`) then the zoom resets. This means that I can seemling zoom out forever with wheel-down.

https://github.com/user-attachments/assets/8aea9e5c-98c8-4715-86dd-17b6eae3c53c

Is this intentional? This isn't what I would expect. I would expect the zoom to do stop once you've zoomed out to 100%.

packages/ag-charts-demos/src/demos/financial/components/FinancialChart.tsx
```diff
@@ -44,72 +96,156 @@ function createFinancialOptions(
...
+        scheduleFlush(true);
+    }, [isRangeView, scheduleFlush]);
+
+    // Swap the built-in reset for one that lands on the range view, keeping its label and its habit
+    // of greying out when the view it restores is already on screen.
+    const getContextMenuItems = useCallback<AgContextMenuGetItemsCallback<ChartDatum, never>>(
+        ({ defaultItems }) =>
+            defaultItems.map((item) =>
+                item === 'reset-zoom'
+                    ? { label: RESET_ZOOM_LABEL, enabled: !isRangeView(), action: resetToRange }
```
olegat: The default reset behaviour resets the zoom to `initialState.zoom`.

So I wonder whether this custom "reset to selected range" behaviour can be achieved by calling `chart.update({ initialState: { zoom: newRange }})` when a range button is pressed in `FinancialApp.tsx`.

If that works, then it would simplify the demo a lot. We wouldn't need options like `zoom.enableDoubleClickToReset: false`, `listeners.doubleClick: onDoubleClick`, `contextMenu.getItems`

packages/ag-charts-demos/src/demos/financial/components/FinancialChart.tsx
yinkelly: Good catch, not intentional at all. Agree zoom should stop

olegat (approved):
LGTM – Tested it out locally. Looks great! Well done 😃

PR #8016, opened 2026-09-03
Merge b14.2.0 into latest

iMoses:
Merge-back of `origin/b14.2.0` into `latest`.

One conflict in `donutSeries.ts` (imports only) resolved as the union of both sides. The `expandLabelPadding` import carried by b14.2.0 was dropped: AG-15902 had already replaced its only call site with `expandLabelBoxExtent`, so it was dead after the merge.


taktran (approved):
Website bits look good 👍🏼

olegat (approved):
LGTM.

`imoses/b14.2.0-into-latest` is the resolved merge of `b14.2.0` into `latest`. Looks like it was just a conflict in the imports AFAICT.

PR #8026, opened 2026-09-03
AG-3390 - Allow the Google Ads hosts the shared GTM container now needs

taktran:
Ports ag-grid/ag-grid#15057 and ag-grid/ag-grid#15060 to this site's CSP. The shared GTM container gained a Google Ads "Google tag" destination (`AW-873243008`) plus an internal promo-tracking GA4 event tag, and the policy here was missing what they need.

## What the tag does, by consent state

Under the **default consent-denied** state gtag.js sends one non-personalised beacon (`npa=1`) and nothing else. With **all cookies accepted** in the Enzuzo banner it does considerably more, so the second half of this only shows up once you accept cookies. The constant records that, so the next person doesn't read a consent-denied page load as proof the policy is complete.

| Origin | Directive | What it is | Without it |
| --- | --- | --- | --- |
| `www.googleadservices.com`, `pagead2.googlesyndication.com` | `connect-src` | `/pagead/conversion`, `/ccm/conversion` beacons, tried with `fetch()` before an `` fallback | Degrades silently to the less reliable pixel |
| `googleads.g.doubleclick.net` | `script-src` | `/pagead/viewthroughconversion/<id>`, injected as an external `<script>` | Hard-blocked |
| `ad.doubleclick.net` | `connect-src` | `/ccm/s/collect`, sent with the Fetch API (`stats.g.doubleclick.net` was already allowed; this host was not) | Hard-blocked |

Also pins the promo-tracking tag's inline `<script>` by SHA-256 hash alongside the other GTM-injected inline scripts in `SITE_SCRIPT_HASHES`.

Deliberately still not allowed, pending an actual violation: `adservice.google.com/pagead/regclk` and `ade.googlesyndication.com/ddm/activity` (dead fallbacks present in every gtag.js payload), with a test pinning that.

All of this was validated on grid staging with all cookies accepted; the console is clean there with the equivalent policy.

## Test plan

- [x] `htaccess` vitest suite passing, including new coverage for every origin above, a test that each stays out of the directive that doesn't need it, and the not-allowed list.
- [ ] Validate on staging once deployed: accept all cookies, then confirm the console is clean.


taktran:
Superseded: retargeted at b14.2.0 in the follow-up PR: https://github.com/ag-grid/ag-charts/pull/8027

PR #8038, opened 2026-09-04
AG-18382 Align funnel and pyramid label before/after with stageLabel sides

iMoses:
https://ag-grid.atlassian.net/browse/AG-18382

Fix #AG-18382


olegat (approved):
LGTM

PR #8037, opened 2026-09-04
Merge b14.2.0 back to latest

AG-DavidG:
Merge the b14.2.0 release branch back into latest.

PR #8040, opened 2026-09-04
Merge b14.2.0 into latest

iMoses:
Merges b14.2.0 back into latest, replacing #8037.

Also includes a fix for the Framework Package Tests (angular) job that blocked #8037. Inside the `mcr.microsoft.com/playwright:v1.52.0` container, `ng new` runs `npm install` with npm 10.9.2, whose resolver crashes with `Cannot read properties of null (reading 'edgesOut')` on the `vitest@^4.0.8` peer set since the vitest 5.0.0 family was published on 3 Sept. `tests/shared/run.sh` now runs `npm i -g npm@11` after configuring the npm cache, so every framework and SSR package test picks it up. npm 11 rather than 12 because the image ships Node 22.14 and npm 12 requires 22.22+.

Verified locally with `yarn nx run angular-package-tests:test:package:modules:local-latest`, which now passes end to end.


AG-DavidG:
LGTM

AG-DavidG (approved):

PR #8049, opened 2026-09-04
Harden dev-prompts npm install against registry stalls

AG-DavidG:
npm.pkg.github.com stalled the `@ag-grid/dev-prompts` install with no output for 5+ minutes (no 4xx/5xx) instead of failing fast, so the composite action's install ran until GitHub Actions' blunt 5-minute job timeout killed the whole job — cancelling `preflight` and/or `concurrency-key` before the pipeline ever reached the agent. Observed on AG-18413: 6+ consecutive dispatch failures over roughly 3 hours (09:02-11:55 UTC 2026-09-04).

**Root cause:** the runner's bundled npm 10.9.8 has an Arborist bug (`Cannot read properties of null (reading 'edgesOut')`) resolving certain peer-dependency graphs, which started biting once vitest 5.0.0 hit the npm registry on 3 Sept — the same underlying bug Ido's `e51fe6603f9` already upgraded npm to work around in the package-test container. It manifested there as a fast crash; here, under `--silent`, the same failure appears to leave the process hanging instead of exiting, which is why it read as a network stall at first.

**This PR:**
1. Upgrades the runner's npm to 11 before the install, which should prevent the underlying crash/hang rather than just recovering from it.
2. Wraps the install in a per-attempt `timeout` plus npm's own `--fetch-timeout`/`--fetch-retries` tuning as a bounded fallback (worst case ~3 min across 3 attempts instead of the full 5-minute job timeout), in case the hang has some other cause too.

Not included here: caching the install (e.g. via `actions/cache`) so a registry blip costs at most one cache-miss instead of blocking every job of every run identically — a reasonable follow-up but a separate change.

### DIFF
--- a/external/ag-shared/github/actions/install-ag-dev-prompts/action.yml
+++ b/external/ag-shared/github/actions/install-ag-dev-prompts/action.yml
@@ -49,8 +49,10 @@ runs:
                       "@ag-grid/dev-prompts@${DEV_PROMPTS_CHANNEL}"; then
                       exit 0
                   fi
-                  echo "::warning::npm install of @ag-grid/dev-prompts@${DEV_PROMPTS_CHANNEL} attempt ${attempt} timed out or failed; retrying"
-                  sleep 5
+                  if [ "$attempt" -lt 3 ]; then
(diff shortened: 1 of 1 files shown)


AG-DavidG:
Unecessary at the moment - the timeout extension worked

PR #8060, opened 2026-09-07
AG-18426 Validate the licence key for community-only charts

iMoses:
https://ag-grid.atlassian.net/browse/AG-18426

Fix #AG-18426


olegat (approved):
LGTM

PR #8074, opened 2026-09-07
AG-18420 Expand unions of type aliases in the API reference into their values

iMoses:
https://ag-grid.atlassian.net/browse/AG-18420

Fix #AG-18420


olegat (approved):
LGTM

PR #8077, opened 2026-09-07
AG-18448 Update dev-validation docs for renamed validation options

AG-DavidG:
https://ag-grid.atlassian.net/browse/AG-18448

Fix #AG-18448


packages/ag-charts-website/src/content/docs/dev-validation-test/_examples/validation-throw/main.ts
```diff
@@ -0,0 +1,62 @@
...
+
+// Invalid on purpose: opacity must be between 0 and 1, so this raises a validation warning, which
+// `throwOn` can turn into a thrown error instead of a console warning.
+function applyInvalidOptions() {
+    const isWarningSelected = (document.getElementById('throw-on-warning') as HTMLInputElement).checked;
+    const throwOn: ('error' | 'warning' | 'deprecation')[] = isWarningSelected ? ['warning'] : [];
+    chart?.destroy();
+    chart = undefined;
+    try {
+        chart = AgCharts.create({
```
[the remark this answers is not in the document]
AG-DavidG: Moved to the -test page and removed from the docs

packages/ag-charts-types/src/chart/chartOptions.ts
```diff
@@ -369,57 +369,37 @@ export interface AgBaseChartOptions<
...
      * Default: `[]`
      */
     throwOn?: AgChartValidationSeverity[];
-    /**
-     * Called for each validation problem the chart raises — an invalid option value or a runtime error
-     * caught during a chart update. The reported problems are the same set the validation overlay
-     * shows, not every diagnostic the chart can write to the console. Never gated by
-     * `consoleOn` or `showOverlayOn`.
-     *
-     * Default: `undefined`
```
olegat, line 422: Optional: Maybe worth keeping the default? All the neighbouring options document their defaults.

packages/ag-charts-types/src/chart/chartOptions.ts
```diff
@@ -369,57 +369,37 @@ export interface AgBaseChartOptions<
...
      */
     showOverlayOn?: AgChartValidationSeverity[];
     /**
-     * The severities of validation problem that cause the chart to throw instead of warning and
-     * falling back to a default. Each severity is handled independently of the others, so
-     * `['warning']` throws on a warning-severity problem but not on an error-severity one. An empty
-     * array never throws, matching the behaviour of charts that do not set this option.
-     *
-     * Console output is never suppressed by this option — the console record of a problem is written
-     * before the throw.
```
olegat, line 411: When both `consoleOn:['error']` and `throwOn:['error']` are set, two things happen when an error is reported: (1) error is logged to the console and (2) the error is thrown.

However, we no longer document this behaviour. Is that intentional?

packages/ag-charts-website/e2e/example-options.ts
```diff
@@ -57,6 +57,12 @@ export const EXAMPLE_OPTIONS: Record<string, Record<string, ExampleOverrides>> =
         'validation-overlay': { ignoreConsoleWarnings: true },
```
olegat: nit: `validation-overlay` and `validation-overlay-dark` are being removed in this PR. They can be removed from here too.

packages/ag-charts-website/src/content/docs/dev-validation/index.mdoc
```diff
@@ -1,45 +1,89 @@
...
 ---
 
-Option misconfiguration and caught runtime errors are normally reported only to the browser console, where they are easy to miss. The validation overlay surfaces them on the chart instead. Issues are grouped by severity — errors, then warnings, then deprecations — each group with its own count and accent colour, under a header summarising the totals.
+AG Charts reports option misconfiguration and caught runtime errors. Reports go to the browser console or a development overlay, the chart can throw to halt execution, and an event can be raised for each issue.
+
+## Severity Levels
+
+Each validation issue is reported as one of three severities: `error`, `warning`, or `deprecation`.
+
+All validation options take an array of one or more of these severities, and apply only to the ones listed. An empty array disables that option entirely.
```
olegat: nit: The `issueRaised` validation option is an exception here, it takes a function not a array of string.

packages/ag-charts-website/src/content/docs/dev-validation/index.mdoc
```diff
@@ -1,45 +1,89 @@
...
 ```
 
-`overlaySeverity` is an inclusive threshold: each severity also shows every louder one.
+In this example:
 
-- `'error'` — errors only.
-- `'warning'` — errors and warnings.
-- `'deprecation'` — errors, warnings and deprecations.
-- `'none'` — nothing. This is the default, so the overlay never appears unless you opt in.
+- Errors and warnings are shown in the overlay, but deprecations are not.
```
olegat: nit: The "but deprecations are not" wording here is a bit confusing here. The sentence suggests the example uses deprecated option which are being suppressed. But in reality the `validation-overlay-multi` example doesn't use any deprecated options at all AFAICT.

Perhaps there is a deprecated options like `type: 'column'` that we can use in the example?

packages/ag-charts-website/src/content/docs/dev-validation/index.mdoc
```diff
@@ -1,45 +1,89 @@
...
+}
+```
+
+In the above example:
+
+- Applying the invalid option with `consoleOn: ['warning']` selected logs a warning to the console.
+- Applying it with `consoleOn: []` selected logs nothing.
 
-{% chartExampleRunner title="Multiple Validation Issues" name="validation-overlay-multi" type="generated" /%}
+## Throwing on Validation Issues
```
olegat, line 57: Optional: It might be worth including a `throwOn` example.

packages/ag-charts-website/src/content/docs/dev-validation-test/_examples/validation-throw/main.ts
```diff
@@ -0,0 +1,63 @@
...
+        y: { type: 'number' },
+    },
+    validations: {
+        // Disabled so the only console output is the explicit log below, not also the default warning.
+        consoleOn: [],
+        throwOn: ['warning'],
+    },
+};
+
+let chart: ReturnType<typeof AgCharts.create> | undefined = AgCharts.create(options);
```
olegat: If the example uses the `throwOn: ['warning']` option, then I think the example should demonstration the best-practice by using a try-catch to create a chart:

```ts
let chart: ReturnType<typeof AgCharts.create> | undefined = undefined;
try {
  chart = AgCharts.create(options);
} catch (e) {
  console.log(`exception thrown when creating chart`, e);
}
```

olegat (approved):
LGTM - some minor non-blocking nits/feedback.

packages/ag-charts-types/src/chart/chartOptions.ts
AG-DavidG: Restored — added `Default: undefined`, matching the pattern used elsewhere for callback options that default to unset (e.g. `getItems` on the context menu).

packages/ag-charts-types/src/chart/chartOptions.ts
AG-DavidG: Good catch, not intentional — restored a version of that line: "Console output is never suppressed by this option." on both the `throwOn` JSDoc and the "Throwing on Validation Issues" prose section.

packages/ag-charts-website/e2e/example-options.ts
AG-DavidG: Removed both.

packages/ag-charts-website/src/content/docs/dev-validation/index.mdoc
AG-DavidG: Fixed — reworded to name the three array-typed options explicitly (`consoleOn`, `showOverlayOn`, `throwOn`) rather than claiming it of "all validation options".

packages/ag-charts-website/src/content/docs/dev-validation/index.mdoc
AG-DavidG: `type: 'column'` doesn't exist as a deprecated option in the current codebase, so I've reworded the bullet instead to describe what `showOverlayOn` would do generally, rather than implying this specific example demonstrates every severity: "Any errors or warnings present would be shown in the overlay, but deprecations would not, since `showOverlayOn` only lists `'error'` and `'warning'`."

packages/ag-charts-website/src/content/docs/dev-validation/index.mdoc
AG-DavidG: Left as prose + snippet only (optional, and no live example added) — an interactive `throwOn` example on this page would need to avoid the imperative `create`/`destroy` pattern that isn't framework-generator compatible, which is exactly why the existing throw example was moved to the internal `dev-validation-test` page rather than kept here.

packages/ag-charts-website/src/content/docs/dev-validation-test/_examples/validation-throw/main.ts
AG-DavidG: Good point — the initial `AgCharts.create(options)` call also has `throwOn: ['warning']` in effect, so it should follow the same try/catch pattern. Wrapped it accordingly; the button handler already did this.

PR #8082, opened 2026-09-08
AG-17959 Web analytics demo enhancements

yinkelly:
https://ag-grid.atlassian.net/browse/AG-17959

Fix #AG-17959


packages/ag-charts-demos/src/demos/web-analytics/components/TrafficChart.tsx
```diff
@@ -199,17 +281,64 @@ export function TrafficChart({
...
             listeners: {
                 selectionChange: ({ source }) => {
                     // Ignore our own api-call echoes; only user interaction should push a new selection upward.
                     if (source === 'api-call') return;
                     onSelectionChange(selectionToDays(chartRef.current?.getSelection() ?? []));
                 },
+                crossLineClick: ({ crossLineId }) => onAnnotationSelect(crossLineId),
+                // A cross-line click never reaches here: the engine fires its cross-line
+                // callbacks and returns before the chart-level listeners.
+                click: () => onAnnotationSelect(null),
```
olegat, line 321: The user can end up in a state where both an annotation and a marker(s) is selected. Step 1: Click "Spring launch", Step 2: Click a marker. It's unclear to me whether this mixed-type selection state is desirable. If not, then we may want to call `onAnnotationSelect(null)` when in the `selectionChange` callback too.

packages/ag-charts-demos/src/demos/web-analytics/components/OverviewView.tsx
```diff
@@ -1,122 +1,194 @@
...
     const [filterModel, setFilterModel] = useState<Record<string, unknown>>({});
+    const [displayedRowCount, setDisplayedRowCount] = useState(0);
+    const onGridStateChange = useCallback((model: Record<string, unknown>, rowCount: number) => {
+        setFilterModel(model);
+        setDisplayedRowCount(rowCount);
+    }, []);
     const gridRef = useRef<SessionsGridHandle>(null);
-    const metricLabel = METRIC_BY_KEY[metric].axisTitle;
+    // Clicked annotation, and the day the add-event form is open on (null = closed).
+    const [selectedAnnotationId, setSelectedAnnotationId] = useState<string | null>(null);
```
olegat, line 53: The selected annotation can disappear and restore itself when the range changes. E.g. the "API outage" annotation is only visible in the "Last 90 days". We may want to call `setSelectedAnnotationId(null)` when the selected annotation disappears from the visible range

olegat (approved):
LGTM - just some minor feedback (non-blocking). Repro-steps & videos left on JIRA.

PR #8083, opened 2026-09-08
AG-17886 Highlight the active ticker by row selection

yinkelly:
https://ag-grid.atlassian.net/browse/AG-17886

Fix #AG-17886

### DIFF
--- a/packages/ag-charts-demos/src/demos/financial/components/TickerGrid.tsx
+++ b/packages/ag-charts-demos/src/demos/financial/components/TickerGrid.tsx
@@ -2,10 +2,10 @@ import {
     type CellKeyDownEvent,
     type ColDef,
     type FullWidthCellKeyDownEvent,
-    type RowClassRules,
+    type RowSelectionOptions,
 } from 'ag-grid-community';
 import { AgGridReact } from 'ag-grid-react';
-import { useEffect, useMemo, useRef } from 'react';
+import { useCallback, useEffect, useMemo, useRef } from 'react';
 
 import { baseColDef, getRowId, gridTheme, rowValuesEqual } from './grid';
 
@@ -17,6 +17,9 @@ interface TickerGridProps<T extends { ticker: string }> {
     activeTicker: string;
     onSelect: (ticker: string) => void;
 }
+
+// Selection carries the active highlight, so moving it must never tear down a row's cell renderers.
+const rowSelection: RowSelectionOptions = { mode: 'singleRow', checkboxes: false };
 
 // A titled watchlist-style grid: a fixed row set whose values stream in place,
 // with the active ticker highlighted and rows selectable.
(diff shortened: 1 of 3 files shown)


packages/ag-charts-demos/src/demos/financial/components/TickerGrid.tsx
```diff
@@ -48,15 +51,17 @@ export function TickerGrid<T extends { ticker: string }>({
...
+        const api = gridRef.current?.api;
+        if (!api) return;
+        const node = api.getRowNode(activeTicker);
+        // A ticker belongs to one list, so the lists without it clear theirs and one row stays selected.
+        if (node) api.setNodesSelected({ nodes: [node], newValue: true });
+        else api.deselectAll();
     }, [activeTicker]);
+    useEffect(() => {
+        syncSelection();
+    }, [syncSelection]);
```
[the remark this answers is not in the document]
yinkelly: Confirmed and fixed in 46cedcad2e. Traced it in ag-grid-react 35.0.1: `AgGridReact.setGridApi` runs from `ctrlsSvc.whenReady`, and the grid's controllers register from children mounted in a later render pass, so `gridRef.current.api` is indeed undefined in the mount effect. `syncSelection` now takes an optional `GridApi` and is called from `onFirstDataRendered` with the event's own api, alongside the `activeTicker` effect.

packages/ag-charts-demos/src/demos/financial/components/grid.ts
```diff
@@ -61,6 +62,8 @@ export const baseColDef = <T>(): ColDef<T> => ({
     sortable: true,
     resizable: false,
     suppressMovable: true,
+    // The grid's own Space handler toggles the row, which would drop a highlight the app owns.
+    suppressKeyboardEvent: ({ event }) => event.key === ' ',
```
olegat: This line seems to add a regression where the page scrolls down when I press Space

yinkelly: Good catch — fixed in 0be904dc5d. The cause: `onSpaceKeyDown` ends with an unconditional `event.preventDefault()`, and suppressing the grid's handling skipped that along with the selection toggle, so nothing cancelled Space's default scroll. The callback now cancels it itself before returning true.

On whether the suppression is necessary: it guards a narrow case — focus a cell in the row that is *already* active and press Space. `inferNodeSelections` returns `newValue: !currentSelection` for `source: 'spaceKey'`, so the grid deselects the row; `onCellKeyDown` then calls `onSelect` with the same ticker, `activeTicker` does not change, and the effect never re-runs — that list ends up with no highlight while the chart still shows the ticker, unrepairable except by picking a different ticker. Space on any other row looks fine, which is why it passes casual testing.

One thing I could not settle and would value your eyes on, since you have the repro: `onCellKeyDown` already calls `event.preventDefault()` for Space, and by my reading of `processCellKeyboardEvent` the `cellKeyDown` dispatch happens outside the suppression block, so that call should already have cancelled the scroll — making the one in the ColDef redundant. If you drop a `console.log(event.key, document.activeElement)` beside it and it fires while the page still scrolls, my model is wrong. If it fires and there is no scroll, I will revert grid.ts to the pure predicate and keep the single `preventDefault` in the component.

olegat (approved):
LGTM

PR #8086, opened 2026-09-08
AG-18452 Route validation issues from the Logger through the eventsHub

iMoses:
https://ag-grid.atlassian.net/browse/AG-18452
https://ag-grid.atlassian.net/browse/AG-18435
https://ag-grid.atlassian.net/browse/AG-18436
https://ag-grid.atlassian.net/browse/AG-18437
https://ag-grid.atlassian.net/browse/AG-18438
https://ag-grid.atlassian.net/browse/AG-18439
https://ag-grid.atlassian.net/browse/AG-18440
https://ag-grid.atlassian.net/browse/AG-18446

Fix #AG-18452
Fix #AG-18435
Fix #AG-18436
Fix #AG-18437
Fix #AG-18438
Fix #AG-18439
Fix #AG-18440
Fix #AG-18446


packages/ag-charts-core/src/logging/logger.ts
```diff
@@ -10,6 +11,25 @@ export function isLogLevel(value: unknown): value is LogLevel {
...
+    message: string;
+    /** A logged Error's stack, then any further console arguments, one per line. */
+    detail?: string;
+    /** A logged Error, so a subscriber that throws can chain it. */
+    cause?: unknown;
+}
+
+function stringifyLogContent(value: unknown): string {
+    if (value instanceof Error || typeof value !== 'object' || value == null) return String(value);
+    return JSON.stringify(value);
```
olegat: This `JSON.stringify` can throw an error if `value` is cyclical. This means that a logger could potentially end up unintentionally throwing instead of logging.

packages/ag-charts-community/src/chart/validation/chartValidations.ts
```diff
@@ -0,0 +1,270 @@
...
+    /**
+     * Replaces the collection with what a re-validating pass raised. `handledBy` is the instance the pass
+     * reported through; what it or this instance already told the same listener is not told again.
+     */
+    beginCycle(issues: readonly LogIssue[], handledBy: ChartValidations = this) {
+        const previous = new Set(this.collection.keys());
+        if (handledBy !== this && handledBy.listener === this.listener) {
+            for (const key of handledBy.told) this.told.add(key);
+        }
+        this.collection.clear();
```
olegat, line 168: It looks like this line clears one-off errors when calling `chart.update()`.

For example:
1. Right-Click chart and Click "Download".
2. Download fails: "Unable to download chart" is drawn on the overlay.
3. Call `chart.update(...newOpts)`
4. Error message disappears from overlay.

The test `'a cycle drops a runtime error the previous cycle collected'` suggests that this is intentional. Can you confirm that this is intentional?

packages/ag-charts-core/src/logging/logger.ts
```diff
@@ -43,40 +76,55 @@ export class Logger {
...
+    }
+
+    // Console first, then subscribers: a subscriber that throws must not lose the console record.
+    private emitIssue(severity: LogLevel, message: unknown, logContent: unknown[]) {
+        if (!this.hasIssueListeners) return;
+        const details = logContent.map(stringifyLogContent);
+        const issue: LogIssue =
+            message instanceof Error
+                ? { severity, message: message.message, cause: message }
+                : { severity, message: String(message) };
```
olegat: If `message` is an object then `String(message)` outputs the constant `"[object Object]"`; so two distinct entries like `errorOnce({a:1})` and `errorOnce({a:2})` could be considered the same message.

packages/ag-charts-community/src/util/listeners.ts
```diff
@@ -40,6 +42,7 @@ export class Listeners<EventType extends string, EventHandler extends Handler> {
             try {
                 listener.handler(...params);
             } catch (e) {
+                rethrowFailFast(e);
```
olegat, line 45: 1. Re-throwing exits the for-loop early, aborting following listeners. Do we want `throwOn` to throw an exception for all listeners? If so, then this needs to be restructured.
2. Nit (_non-blocking_): One of the key advantages of using exceptions is that the developer can retrieve the stack-trace of where the error originated to understand the root cause better. Catch & rethrowing cancels out that advantage, so it would be nice to avoid a re-throw.

packages/ag-charts-core/src/logging/logger.ts
iMoses: Fixed: `stringifyLogContent` now falls back to `String()` when `JSON.stringify` throws (circular structures, bigint), so a logging call can never throw on serialisation. Covered by a Logger test with both cases.

packages/ag-charts-core/src/logging/logger.ts
iMoses: Good catch. `emitIssue` now serialises the message the same way the `*Once` cache key does: an object with its own `toString` (the validator's issue objects) uses it, a plain object is JSON-stringified, and anything unserialisable falls back to `String()`. So `errorOnce({a:1})` and `errorOnce({a:2})` are distinct issues. Test added.

packages/ag-charts-community/src/chart/validation/chartValidations.ts
iMoses: Yes, intentional, and unchanged from before this PR: the previous collector's `setIssues` also reset the runtime feed on every options pass (`this.runtimeIssues = []`), on the basis that the update following the pass re-reports anything that still fails. A failed download is a one-off action that no update re-evaluates, so it goes with the cycle, as it did before. Keeping one-off action errors across `update()` calls would be a behaviour change; happy to pick it up as a follow-up if you think it is wanted.

packages/ag-charts-community/src/util/listeners.ts
iMoses: 1. Yes, that is the intended semantics of `throwOn`: it aborts the operation in progress, the same way it aborts the rest of an update pass, so the remaining listeners do not run. Only a `FailFastError` is re-thrown here; an ordinary listener exception is still logged and the loop carries on.
2. `rethrowFailFast` re-throws the same `Error` instance unchanged, so its `stack` still points at the original throw site in `ChartValidations.onIssue`; nothing is lost by the catch and re-throw.

packages/ag-charts-community/src/chart/chart.ts
```diff
@@ -1777,23 +1752,10 @@ export abstract class Chart implements ModuleInstance, ChartService {
...
-    }
-
     applyOptions(newChartOptions: ChartOptions) {
-        // Registered from the same options object in the same statement pair, so this pass's issues
-        // reach the listener this pass declared without depending on when chartState observers flush.
-        this.setIssueListener(newChartOptions.processedOptions.validations?.issueRaised);
-        this.validationCollector.setIssues(newChartOptions.validationIssues);
+        if (newChartOptions.revalidated) {
+            this.ctx.validations.beginCycle(newChartOptions.issues, newChartOptions.validations);
+        }
```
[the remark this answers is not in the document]
iMoses: Fixed. Data-key, invalid-value and callback warnings are now owned by the pass that raised them (`beginPass`/`endPass` on the data and update phases). A completed pass that does not raise an issue again retires it, which covers the same-length fast path where `beginCycle` is skipped. Option issues have no owning pass and survive until the next validated cycle. Retirement only runs when the pass could have re-raised the issue: a full data reprocess for data warnings, and a node-data-regenerating update with a cleared callback cache for callback warnings; an incremental `applyTransaction` or a legend toggle keeps them. Tests: `drops a data warning once a same-length data update supplies the key`, `drops an invalid data value warning…`, `drops a callback warning…`, plus the three `keeps …` cases.

packages/ag-charts-core/src/logging/logger.ts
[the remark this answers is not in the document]
iMoses: Fixed. `stringifyLogContent` falls back to `String()` when `JSON.stringify` throws (circular structures, bigint), so a logging call cannot throw on serialisation. Covered by a Logger test with both cases.

packages/ag-charts-community/src/module/optionsModule.ts
```diff
@@ -468,17 +406,24 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
...
         } catch (error) {
-            throw this.decorateOptionsProcessingFailure(error);
+            // A rejected pass must not leave its `validations` settings on the chart that keeps its old options.
+            if (baseChartOptions != null) {
+                this.validations.configure(getValidations(baseChartOptions.processedOptions));
+            }
+            // An error raised while processing (a throwing datum getter, a callback invoked during
+            // validation) escapes ahead of the update loop's catch, so it is reported here instead.
+            rethrowFailFast(error);
+            this.logger.error(error);
```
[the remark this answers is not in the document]
iMoses: Fixed. The error is now reported inside the `catch`, under the failing pass's `validations` settings (so a new `issueRaised` listener or `throwOn` is honoured), and the previous chart's settings are restored in `finally` so restoration also happens when reporting throws. Tests under `a processing failure on a warm update` in optionsModule.test.ts.

packages/ag-charts-core/src/logging/logger.ts
```diff
@@ -10,13 +11,39 @@ export function isLogLevel(value: unknown): value is LogLevel {
...
+}
+
+// A logged value can be anything a user callback threw, so serialisation must not be the thing that throws.
+function stringifyLogContent(value: unknown): string {
+    if (value instanceof Error || typeof value !== 'object' || value == null) return String(value);
+    if (value.toString !== Object.prototype.toString) return String(value);
+    try {
+        return JSON.stringify(value);
+    } catch {
+        return String(value);
```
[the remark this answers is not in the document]
iMoses: Fixed. Only a callable, non-default `toString` is treated as custom; everything else goes through `JSON.stringify`, and the fallback is a literal `[object Object]` rather than another coercion, so a null-prototype argument cannot throw out of the logging call. Test: `serialises a null-prototype object, which has no toString, rather than throwing`.

packages/ag-charts-community/src/chart/validation/chartValidations.ts
```diff
@@ -0,0 +1,311 @@
...
+
+        const key = keyOf(issue);
+        if (!this.collection.has(key)) {
+            this.collection.set(key, issue);
+            // An issue raised by interaction has no pass to re-raise it, so no pass retires it.
+            const pass = this.openPasses.at(-1);
+            if (pass != null) this.passOf.set(key, pass);
+            this.dismissed = false;
+            if ((this.showOverlayMask & SEVERITY_BIT[issue.severity]) !== 0) {
+                this.eventsHub.emit('validation:change', null);
```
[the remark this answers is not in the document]
iMoses: Fixed by preserving dismissal for unselected issues: `dismissed` is reset only when the arriving issue is one `showOverlayOn` shows, which is also the only case that emits `validation:change`, so the subscribers and `hasVisibleIssues()` can no longer disagree. Test: `keeps a dismissed overlay dismissed when an issue the selection does not show arrives`.

olegat (approved):
LGTM

PR #8088, opened 2026-09-08
Restructure module-registry docs, fix RTL numeric formatting text

AG-DavidG:
## Module Registry

- Collapse the four registration subsections (importing, global, per-chart, checking-registered) under one `## Registering AG Charts Modules` heading, instead of splitting "Registering Modules per Chart" out as a sibling H2.
- Fix a self-contradicting timing clause: the intro previously said modules must be registered "before calling `AgCharts.create`" for both the global and per-chart paths, but per-chart modules are passed *as an argument to* that call.
- Conditionalise the per-chart registration snippet per framework (javascript/react/angular/vue), matching the pattern used on the financial-charts page.
- Align the inline "Registering Globally" import snippet with the example source and the page's own Module Selector output (single import, not two from the same package).
- Add the missing `$framework` placeholder to the frontmatter description.

## RTL

Tighten the numeric-formatting bullets under "In this example" to name the two series shown (Sales Growth, Net Profit) directly, rather than referring to a separate "Numbers" section.

### DIFF
--- a/packages/ag-charts-website/src/content/docs/module-registry/index.mdoc
+++ b/packages/ag-charts-website/src/content/docs/module-registry/index.mdoc
@@ -1,6 +1,6 @@
 ---
 title: 'Module Registry'
-description: 'Register AG Charts modules to enable chart types and features before creating charts.'
+description: 'Register $framework AG Charts modules to enable chart types and features before creating $framework charts.'
 ---
 
 Reduce bundle size by importing only the AG Chart modules required by your application, with treeshaking support.
@@ -20,9 +20,13 @@ Alternatively, if you are not concerned about bundle size and you want access to
 
 ## Registering AG Charts Modules
 
-Every module must be registered with the global `ModuleRegistry` **_*before*_** calling any AG Charts APIs such as `AgCharts.create`.
+Every module must be registered before a chart can use it, either [globally](#registering-globally) via `ModuleRegistry` or [per chart](#registering-per-chart) when that chart is created.
 
-Add your registration call near the bootstrap of your application so it runs exactly once.
+Global registration must run **before** any chart is created.
+
+{% note %}
+`ModuleRegistry.registerModules` is idempotent: registering the same module again is ignored unless a different version is detected.
+{% /note %}
 
 ### Importing the Module Registry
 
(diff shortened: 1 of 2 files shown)


packages/ag-charts-website/src/content/docs/module-registry/index.mdoc
```diff
@@ -20,9 +20,13 @@ Alternatively, if you are not concerned about bundle size and you want access to
 
 ## Registering AG Charts Modules
 
-Every module must be registered with the global `ModuleRegistry` **_*before*_** calling any AG Charts APIs such as `AgCharts.create`.
+Every module must be registered before a chart can use it, either [globally](#registering-globally) via `ModuleRegistry` or [per chart](#registering-per-chart) when that chart is created.
 
-Add your registration call near the bootstrap of your application so it runs exactly once.
+Global registration must run **before** any chart is created.
```
iMoses: That's not really correct. We actually have tests to validate that registering global module still work and picks up on that when the chart updates. (`AgChart.update()`)

packages/ag-charts-website/src/content/docs/module-registry/index.mdoc
```diff
@@ -62,26 +66,33 @@ import { AnimationModule, HeatmapSeriesModule } from 'ag-charts-enterprise';
 ModuleRegistry.registerModules([AllCommunityModule, HeatmapSeriesModule, AnimationModule]);
 ```
 
-### Registering Modules
+### Registering Globally
+
+Add the `ModuleRegistry.registerModules` call near the bootstrap of your application so it runs exactly once.
```
iMoses: "so it runs exactly once" - again, not really a limitation, just the simplest approach.

packages/ag-charts-website/src/content/docs/module-registry/index.mdoc
```diff
@@ -20,9 +20,13 @@ Alternatively, if you are not concerned about bundle size and you want access to
 
 ## Registering AG Charts Modules
 
-Every module must be registered with the global `ModuleRegistry` **_*before*_** calling any AG Charts APIs such as `AgCharts.create`.
+Every module must be registered before a chart can use it, either [globally](#registering-globally) via `ModuleRegistry` or [per chart](#registering-per-chart) when that chart is created.
```
iMoses: nit: maybe mention here at the top that you can mix, setting globals for all charts + per chart for specific modules.

packages/ag-charts-website/src/content/docs/rtl/index.mdoc
```diff
@@ -36,14 +36,9 @@ In this example:
...
 - The tooltip and context menu layout are mirrored.
-- Both series have negative months, and those values keep their sign on the left in every language.
-- Tooltips use `mode: 'single'`. Sales growth uses the default tooltip, net profit a custom `renderer` that wraps its own number.
-
-## Numbers
-
-Numbers keep their left-to-right reading order in RTL mode, so the negative labels in the example above render as `-30` rather than `30-`. This covers the sign, the decimal and group separators, and any unit attached to the value, such as `-12.5 kg`. Text surrounding a number still reads right-to-left.
-
-The same ordering applies to axis and series labels, tooltips and crosshair labels, including any text a `formatter` returns. HTML returned from a `renderer` is inserted exactly as supplied and is never modified, so wrap numbers in it with `U+202A` and `U+202C` as the example does.
+- Both series have negative values in some months, and those values keep their sign on the left, rendering as `-30` rather than `30-`. This applies to the sign, the decimal and group separators, and any unit attached to the value, such as `-12.5 kg`.
```
iMoses: "such as `-12.5 kg`" - the unit should literally be attached. It should be `-12.5kg`

iMoses (approved):
LGTM - a few minor notes

PR #8104, opened 2026-09-08
AG-18426 Inject the chart stylesheets into every hosting document

iMoses:
https://ag-grid.atlassian.net/browse/AG-18426

Fix #AG-18426


olegat (approved):
LGTM

PR #8112, opened 2026-09-09
AG-18474 Stop theme template defaults reporting unregistered modules

iMoses:
https://ag-grid.atlassian.net/browse/AG-18474

Fix #AG-18474


olegat (approved):
LGTM

PR #8115, opened 2026-09-09
Fix ESM bundle rewriting className statics to temp identifiers

iMoses:
The `pure-toplevel-side-effects` esbuild plugin renamed a class variable inside its trailing static assignments with a text regex, which also matched string literals. Any class whose variable name equals its `className` string ended up in `main.esm.mjs` as `_c$91.className = "_c$91"`, so `createId()` yields series ids like `_c$91-1` for ESM consumers. The bug has been latent since the plugin landed (#6589) and was surfaced on the docs site when AG-17103 switched the TypeScript and framework example runners from the CJS entry to `main.esm.mjs`: clicking a bar node in the `seriesNodeClick` example logs `Series: _c$91-1`.

This renames references via acorn identifier positions, skipping property names and string contents, and fails the build if a `className` literal is ever rewritten again.

Verified: rebuilt community and enterprise bundles have zero corrupted `className` statics (previously 25 and 38), `test:validate-size` passes, community and enterprise tests pass apart from a quadrant preset snapshot that also fails on unmodified `b14.2.0`.

### DIFF
--- a/esbuild.config.cjs
+++ b/esbuild.config.cjs
@@ -178,11 +178,16 @@ function renameReferences(code, node, from, to) {
             return;
         }
         if (!n.type) return;
+        if (n.type === 'Property' && n.shorthand && n.value.type === 'Identifier') {
+            // `{ Foo }` shares one range between key and value, so expand it to `Foo: _c$0`.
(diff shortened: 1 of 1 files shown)


olegat (approved):
LGTM.

PR #8117, opened 2026-09-09
AG-18458 Add an inner circle total to the sunburst gallery example

yinkelly:
https://ag-grid.atlassian.net/browse/AG-18458

Fix #AG-18458


olegat (approved):
LGTM

PR #8125, opened 2026-09-10
Render an empty chart when the lead series type has no registered module

iMoses:
An enterprise series type in the community package (e.g. `heatmap`) now renders a blank canvas instead of the empty chart an empty `series` array gives.

**Cause.** The theme only builds a defaults entry per registered series type, so a lead type with no module left the chart with no chart-level defaults and the first update crashed on `touch.dragAction`. AG-18240 worked around that by skipping the update entirely whenever the lead type had no module, and its check read the raw user options, so a dropped enterprise series short-circuited the whole chart.

**Fix.** `OptionsGraph` resolves its lead type to a registered theme entry when the requested one has none, so every unusable series ends up as the same empty chart `series: []` gives: missing-module error reported, no series, captions and defaults intact. The AG-18240 skip machinery (`unusableLeadSeriesType`, `queueSkippedUpdate`, the guard in `Chart.update`, the fast-path exclusion) is removed, and its tests are rewritten to assert the consistent empty-chart outcome.

**Known limitations, unchanged from before AG-18240**
- A registry with no series module at all still has no chart-level defaults and crashes, for any options including `series: []`.
- When the fallback engages (only when `line` itself is unregistered and no series survive), theme overrides keyed by the fallback type apply to the empty chart.
- A data-only delta update takes the fast path and does not re-validate a dropped series after its module is registered; a refresh request does.

### DIFF
--- a/packages/ag-charts-community/src/api/agCharts.test.ts
+++ b/packages/ag-charts-community/src/api/agCharts.test.ts
@@ -581,55 +581,84 @@ describe('AgCharts', () => {
             return messages;
         }
 
-        it('reports the implicit `line` default and renders nothing instead of crashing', async () => {
-            await withOnlyBarRegistered(() => {
-                expect(() => (chart = AgCharts.create({ container } as AgChartOptions))).not.toThrow();
+        // No series survive, yet the captions and chart-level defaults are resolved and applied.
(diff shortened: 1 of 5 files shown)


olegat (approved):
LGTM

PR #8126, opened 2026-09-10
AG-18457 `aglint/validate-module-registration` improvements

olegat:
https://ag-grid.atlassian.net/browse/AG-18457


olegat:
Closing. Not worth the effort right now. May revisit in the future.

PR #8128, opened 2026-09-10
AG-18485 Centre donut inner labels on the hole

iMoses:
https://ag-grid.atlassian.net/browse/AG-18485

Fix #AG-18485


AG-DavidG (approved):
LGTM

PR #8129, opened 2026-09-10
AG-18489 Revise label placement, axes and background region docs

AG-DavidG:
https://ag-grid.atlassian.net/browse/AG-18489

Fix #AG-18489


olegat (approved):
LGTM

PR #8137, opened 2026-09-10
Forward overrideDevicePixelRatio through the quadrant preset

iMoses:
The quadrant preset did not forward `overrideDevicePixelRatio` to the underlying cartesian chart, and its options defs did not register it, so the option was stripped as unknown. The website thumbnail generator relies on it for the `@2x` gallery thumbnails, which rendered the quadrant chart at 1x scale inside a 2x canvas (visible on staging and the 14.2.0 archive on retina screens).

Adds the option to the preset's forwarded chart options and registers it as an undocumented positive number, matching the gauge and price-volume presets. Two tests cover the forwarding and the validation registration.

Once merged, the enterprise bundle hash changes, so all gallery thumbnails regenerate on the next deploy.

### DIFF
--- a/packages/ag-charts-enterprise/src/preset/quadrant/quadrantPreset.test.ts
+++ b/packages/ag-charts-enterprise/src/preset/quadrant/quadrantPreset.test.ts
@@ -331,6 +331,29 @@ describe('Quadrant Preset', () => {
         }
     );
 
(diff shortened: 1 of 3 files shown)


AG-DavidG (approved):
LGTM

taktran (approved):
LGTM

PR #8142, opened 2026-09-10
AG-18198 Add corner radius to gallery examples

yinkelly:
https://ag-grid.atlassian.net/browse/AG-18198

Fix #AG-18198


AG-DavidG (approved):
LGTM

PR #8145, opened 2026-09-11
AG-18496 Register axis interaction with enterprise axis modules

iMoses:
https://ag-grid.atlassian.net/browse/AG-18496

Fix #AG-18496


packages/ag-charts-community/src/module/optionsModule.ts
```diff
@@ -989,6 +994,32 @@ export class ChartOptions<T extends AgChartOptions = AgChartOptions> {
...
+        const placeholder = ExpectedModules.get('axis-interaction');
+        if (placeholder == null || this.moduleRegistry.hasModule(placeholder.name)) return;
+        if (placeholder.chartType != null && placeholder.chartType !== this.chartDef?.name) return;
+
+        let missing = false;
+        const strip = (listeners: PlainObject | undefined, ...events: string[]) => {
+            if (!isObject(listeners)) return;
+            for (const event of events) {
+                if (listeners[event] == null) continue;
+                delete listeners[event];
```
olegat, line 1008: nit: I'm not sure this removal is necessary. It doesn't really matter much, but these listeners are only ever called by `AxisInteraction`, so if that module is missing then these callbacks are unreachable anyway.

olegat (approved):
LTGM

packages/ag-charts-community/src/module/optionsModule.ts
iMoses: True, but it is consistent so in case we change the validation side in the future I'd rather leave it in.

PR #8146, opened 2026-09-11
AG-18493 Memoise options-graph `resolvePartial`

olegat:
https://ag-grid.atlassian.net/browse/AG-18493

This is the main bottleneck for the poor highlight/tooltip performance on quadrant charts.

### DIFF
--- a/packages/ag-charts-community/src/module/optionsPartialCache.ts
+++ b/packages/ag-charts-community/src/module/optionsPartialCache.ts
@@ -1,0 +1,129 @@
+import { LRUCache, type PlainObject, isPlainObject } from 'ag-charts-core';
+
+// Sized for the handful of distinct styles a styler yields; one returning a distinct style per datum evicts
+// rather than growing without bound.
+const PARTIAL_CACHE_MAX = 512;
+
+// Depth cap doubles as the cycle guard: a self-referential payload bails out as uncacheable.
+const CACHE_KEY_MAX_DEPTH = 6;
+
+export interface PartialCacheKeyOptions {
+    permissivePath?: boolean;
+    pick?: boolean;
+    proxyPaths?: Record<string, Array<string>>;
+}
+
+interface PartialCacheEntry {
+    value: PlainObject | undefined;
+}
+
+/**
+ * Resolved partials, keyed by request. A styler resolves one per datum and resolving is not a read — operations
+ * graft theme branches onto the graph, resolve them, then roll them back — so repeating an identical request
+ * rebuilds and discards the same sub-graph once per datum.
+ */
+export class OptionsPartialCache {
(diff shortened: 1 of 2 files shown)


packages/ag-charts-community/src/module/optionsPartialCache.ts
```diff
@@ -0,0 +1,143 @@
+import { LRUCache, type PlainObject, isPlainObject } from 'ag-charts-core';
+
+// Sized for the handful of distinct styles a styler yields; one returning a distinct style per datum evicts
+// rather than growing without bound.
+const PARTIAL_CACHE_MAX = 512;
+
+// Depth cap doubles as the cycle guard: a self-referential payload bails out as uncacheable.
+const CACHE_KEY_MAX_DEPTH = 6;
+
+export interface PartialCacheKeyOptions {
```
iMoses, line 10: nit: this is the same shape as `OptionsGraphAccessorResolvePartialOptions` which is already exported from optionsGraph.ts, and the two private signatures in there inline it again. can we just reuse the one type in all three places so they don't drift

packages/ag-charts-community/src/module/optionsPartialCache.ts
```diff
@@ -0,0 +1,143 @@
...
+    proxyPaths?: Record<string, Array<string>>;
+}
+
+interface PartialCacheEntry {
+    value: PlainObject | undefined;
+}
+
+/**
+ * Resolved partials, keyed by request. A styler resolves one per datum and resolving is not a read — operations
+ * graft theme branches onto the graph, resolve them, then roll them back — so repeating an identical request
```
iMoses, line 22: "resolving is not a read" isn't really true. `resolveVertexDependencies` calls `resolveVertex(dependency)` with no object so it defaults to `this.resolved`, and rollback doesn't undo those writes. it still works out because the operations re-resolve their deps before reading, so I don't think it's a problem in practice, but worth rewording so the next person touching the invalidation doesn't rely on it

packages/ag-charts-community/src/module/optionsPartialCache.ts
```diff
@@ -0,0 +1,143 @@
...
+ * rebuilds and discards the same sub-graph once per datum.
+ */
+export class OptionsPartialCache {
```
iMoses, line 25: we should have a test file for this. the cache is invisible to the snapshots so a broken key or a missed invalidation wouldn't fail anything. `keyFor` / `appendCacheKey` / `hasUnmergedCssVariables` are all pure so it's cheap - key distinctness, cached `undefined` vs a miss, and the three invalidation triggers would cover it. fine as a follow up

iMoses (approved):
LGTM - a few minor notes, nothing blocking. happy for these to go in a follow up if you want to merge now

PR #8149, opened 2026-09-11
Promote strict-boolean-expressions to error and fix violations

olegat:
A `string` or `number` in a boolean position is almost always a nullish check written as a truthiness check, which then also rejects the legitimate values `''`, `0` and `NaN`. The bubble/scatter crash on `sizeKey: ''` was exactly this shape.

Enable `@typescript-eslint/strict-boolean-expressions` as an error with `allowString`, `allowNumber`, `allowNullableString` and `allowNullableNumber` all off. Nullable objects, booleans, enums and `any` keep the ordinary truthiness shorthand, so the rule targets the hazard without churning every guard in the codebase.

Fix the resulting violations across core, community, enterprise, website, demos, angular and the test library. Most become `x != null`, `x == null || x === ''`, `.length > 0`, or an explicit `=== 0` / `Number.isNaN` where 0 and NaN previously fell through to the same branch as an absent value. Behaviour is preserved throughout: where the original truthiness deliberately skipped `''` or `0`, both checks are kept rather than collapsing to a nullish test.

Some conversions needed more than a local edit:

- Resolve `title.showInLegend` before use in the donut legend, so the value is the title text or nothing rather than `false`.
- Alias `userContent.title` in the sparkline tooltip so the truthy branch narrows it.
- Type `useRef` as `HTMLDivElement` in `ImageCarousel`, `lastHeading` as `Node | undefined`, and `toTitle` as accepting `undefined` — all contracts the surrounding guards and tests already relied on but the types denied.
- Read indexed positions through `.at()` or an explicit bounds check where the guard is load-bearing at runtime but the index type is not optional.

Drop guards the rule proved dead: `getBBox()` and
`Transformable.toCanvas()` never return null, `Group.childNodes` is a readonly Set, and `CodeViewer`'s loading branch and `Collapsible`'s `versions` guard could never render.


packages/ag-charts-website/src/components/homepage/components/gallery-examples/galleryExamplesData.ts
```diff
@@ -48,12 +48,12 @@ async function getUpdateExampleFunction({
...
     let dataJs = generatedFiles['data.js']?.trim();
 
-    if (replaceNewlineToken) {
+    if (replaceNewlineToken != null && replaceNewlineToken !== '') {
         mainJs = mainJs.replaceAll('\n', replaceNewlineToken);
         dataJs = dataJs?.replaceAll('\n', replaceNewlineToken);
     }
 
-    const exampleFunction = `() => {${dataJs ? dataJs : ''}${mainJs}}`;
+    const exampleFunction = `() => {${dataJs}${mainJs}}`;
```
alantreadway: `dataJs` is `undefined` at runtime for examples without a `data.ts` (the index type hides it). The homepage `sankey`, `simple-radial-gauge` and `simple-linear-gauge` examples have none, so this emits `() => {undefinedconst {...} = agCharts; ...}` – a SyntaxError in the homepage examples bundle. Suggest `${dataJs ?? ''}`.

packages/ag-charts-community/src/scene/node.ts
```diff
@@ -98,8 +98,7 @@ export abstract class Node<TDatum = unknown> {
     static *extractBBoxes(nodes: Iterable<Node>, skipInvisible?: boolean) {
         for (const n of nodes) {
             if (!skipInvisible || (n.visible && !n.transitionOut)) {
-                const bbox = n.getBBox();
-                if (bbox) yield bbox;
+                yield n.getBBox();
```
alantreadway: This guard wasn't dead: `getBBox()` hides `BBox | undefined` behind `as BBox`, and the base `Node.computeBBox()` returns `undefined`. `Image` doesn't override it, so an `Image` in a group whose bbox is computed will now make `BBox.merge` throw. Same root cause for the guard removed in `sceneDebug.ts` (~L285). Suggest keeping the guard, or making `computeBBox` return a real `BBox` and dropping the cast.

packages/ag-charts-enterprise/src/axes/angle-number/angleNumberAxis.ts
```diff
@@ -81,8 +82,8 @@ export class AngleNumberAxis extends AngleAxis<AgNumericValue, LinearAngleScale,
         let rawTicks: number[];
         if (values == null) {
             const { arcLength } = scale;
-            const minTickCount = maxSpacing ? Math.floor(arcLength / maxSpacing) : 1;
-            const maxTickCount = minSpacing ? Math.floor(arcLength / minSpacing) : Infinity;
+            const minTickCount = maxSpacing == null ? 1 : Math.floor(arcLength / maxSpacing);
```
alantreadway: `maxSpacing: 0` passes validation (`positiveNumber` is `numberMin(0)`). Previously this gave `minTickCount = 1`; now it's `Infinity`, which drives the tick step to 0 and renders no ticks. `radialGaugeSeries.ts` handles the same case with `=== 0` – suggest `maxSpacing == null || maxSpacing === 0 ? 1 : …`.

packages/ag-charts-community/src/chart/legend/legend.ts
```diff
@@ -1433,8 +1433,8 @@ export class Legend {
                     aspectRatio < 1
                         ? Math.min(maxCoefficient, minHeightCoefficient * (1 / aspectRatio))
                         : minHeightCoefficient;
-                legendWidth = maxWidth ? Math.min(maxWidth, width) : width;
-                legendHeight = maxHeight ? Math.min(maxHeight, height) : Math.round(height * heightCoefficient);
+                legendWidth = maxWidth == null ? width : Math.min(maxWidth, width);
```
alantreadway: Behaviour change: `maxWidth: 0` / `maxHeight: 0` (valid per `positiveNumber`) were previously ignored and now produce a zero-size legend. Consistent with L633, but worth confirming it's intended. Same at L1451-1452.

packages/ag-charts-community/src/chart/legend/legendManager.ts
```diff
@@ -57,11 +57,11 @@ export class LegendManager implements MementoOriginator<LegendDataMemento> {
     private getRestoredData(datum: AgInitialStateLegendOptions) {
         const { seriesId, itemId, legendItemName, visible } = datum;
 
-        if (seriesId) {
+        if (seriesId != null) {
             const legendData = this.legendDataRecord[seriesId] ?? [];
 
             const data = legendData.map((d) => {
-                const match = d.seriesId === seriesId && (!itemId || d.itemId === itemId);
+                const match = d.seriesId === seriesId && (itemId == null || d.itemId === itemId);
```
alantreadway: Behaviour change: `itemId` is typed `string` but can be the number `0` at runtime (pie/donut use `datumIndex`). Previously `initialState.legend: [{ itemId: 0, ... }]` matched every item in the series; now only item 0. Looks like a latent bug fix – worth a test.

packages/ag-charts-community/src/api/preset/sparkline.ts
```diff
@@ -133,14 +133,15 @@ const tooltipRendererFn = simpleMemorize((tooltip?: AgSparklineTooltip<any>, dat
         }
 
         // Absent renderer or one returning `undefined` falls through to the default content.
-        const hasUserTitle = userContent?.title != null;
+        const userTitle = userContent?.title;
+        const hasUserTitle = userTitle != null;
         const content = userContent?.content ?? defaultTooltipContent(xValue, yValue, datumKey, hasUserTitle);
 
-        return userContent?.title
+        return hasUserTitle
```
alantreadway: Behaviour change: a tooltip renderer returning `{ title: '' }` previously took the `title: content, data: []` branch; it now takes the `data: [{ label: '', value: content }]` branch. Arguably more consistent – intended?

packages/ag-charts-website/src/components/release-notes/Collapsible.tsx
```diff
@@ -61,19 +61,17 @@ const Collapsible: FunctionComponent<Props> = ({
                         Version:
                     </label>
                     <select
-                        value={fixVersion || versions[0]}
+                        value={fixVersion === '' ? versions[0] : fixVersion}
```
alantreadway: `fixVersion` is typed `string`, but `Changelog.tsx` (`useFixVersion`) calls `setFixVersion(version!)` with `undefined` when there's no `?fixVersion` param – the default landing. The old `||` fell back to `versions[0]`; now `value={undefined}` flips the select to uncontrolled (React warning). Suggest `fixVersion == null || fixVersion === ''`.

packages/ag-charts-enterprise/src/axes/angle-number/linearAngleScale.ts
```diff
@@ -32,10 +32,10 @@ export class LinearAngleScale extends LinearScale {
             return { ticks: [], count: 0 };
         }
 
-        const { nice, interval } = ticks;
+        const { interval } = ticks;
         const [d0, d1] = numericDomain;
 
-        if (interval) {
+        if (interval != null) {
```
alantreadway: Gauge `scale.interval.step` is validated with `numericValue`, so `0` is reachable. Previously 0 skipped this branch; now it hits `isDenseInterval(Infinity)` and logs a spurious density warning (and yields no ticks when `min === max`). Suggest `interval != null && interval !== 0`.

alantreadway (commented):
Mostly mechanical and equivalent, but a few conversions change behaviour – three look like bugs (gallery bundle, `getBBox` guard, angle-axis `maxSpacing: 0`), the rest are `0`/`''` edge cases worth confirming as intended.

Also: the description says enums keep truthiness, but the second commit sets `allowNullableEnum: false` – worth updating.

olegat:
> Mostly mechanical and equivalent, but a few conversions change behaviour – three look like bugs (gallery bundle, `getBBox` guard, angle-axis `maxSpacing: 0`), the rest are `0`/`''` edge cases worth confirming as intended.
> 
> Also: the description says enums keep truthiness, but the second commit sets `allowNullableEnum: false` – worth updating.

Thanks. Feedback addressed.

alantreadway (approved):
LGTM

PR #8163, opened 2026-09-14
CRT-1207: Write the clicked gallery family to the URL hash

skewis6:
https://ag-grid.atlassian.net/browse/CRT-1207

## Problem

Clicking a chart family in the gallery sidebar scrolled to that section but left the URL unchanged, so the address bar kept naming whichever family the visitor arrived on. A link shared after clicking through the nav pointed at the wrong section — the edge case called out on the ticket: arrive on `/gallery/#chord` via *View All Examples*, click **Bar**, share, and the recipient still lands on Chord.

## Cause

The nav scrolls with a sticky-header offset rather than letting the browser jump, so its click handler calls `preventDefault()`. That also discards the hash the browser would otherwise have set.

## Fix

Write the hash back with `replaceHistoryUrl`, which is how this codebase already solves this:

- `ab3e30ba88` (AG-18315) fixed the identical bug on the docs side navigation.
- The heading `LinkIcon`s on this very page already do the same.

`replace` rather than `push`, matching those, so clicking through several families doesn't bury the previous page under back-stack entries. A fragment-only URL leaves the path and query untouched, so the archive paths on the ticket and the theme query param are unaffected.

## Testing

New `e2e/gallery-page-nav.spec.ts` clicks every sidebar link and asserts the hash follows, modelled on the existing `docs-page-nav.spec.ts`.

Verified against a local dev server:

- **With the fix — passes.**
- **With the fix reverted — fails with the ticket's exact symptom** (`Expected "#bar", Received ""`), confirming the test is a genuine regression guard rather than a vacuous one.
- ESLint 0 errors; Prettier clean.

No registration needed for the new spec — `testDir` auto-discovers it, and the CI shard balancer assigns unknown files the median duration.

## Note (not addressed here)

Scrolling still leaves the hash stale; only clicks update it. That matches the ticket as written, and the scrollspy already tracks the active family if we want to extend it later.


alantreadway (approved):
LGTM

PR #8166, opened 2026-09-14
CRT-1218 Serve the gallery get-started line as its own paragraph

skewis6:
Closes [CRT-1218](https://ag-grid.atlassian.net/browse/CRT-1218).

The get-started sentence every gallery example page closes with was held on the end of each intro behind a markdown hard line break, which renders as a `<br>` inside the one paragraph — a new line with no gap above it.

It is boilerplate rather than per-example copy, so it moves out of all 117 intros into a single shared constant and is rendered as a paragraph of its own, on the page and in the example's `.md` twin alike. The gap is the design system's paragraph rhythm rather than a new spacing value, and the intro block keeps its existing margin down to the example.


alantreadway (approved):
LGTM

PR #8167, opened 2026-09-14
Show the gallery example before its intro and align the related strip

alantreadway:
Follow-up to the gallery example page redesign in 14.2.0, from the review of the shipped page. None of the SEO-checked surface changes: title, H1, meta description, intro copy and deep links, related links and the `.md` twins are byte-identical, and `gallerySeoChecker` passes unchanged.

**What changed**

- Less top padding on the page: 64px to 40px on desktop, 40px to 24px on mobile.
- The example is shown first and the intro after it. The markup still has the intro directly under the H1; the page container is now a flex column and CSS `order` moves the intro and the related section visually, so crawled order and the markdown twin are unchanged.
- Carousel slides are sized on the hub grid's column count (1, 2 or 3 per page by breakpoint) so a page of the strip is exactly a hub row, flush with the example runner's edges. This replaces the fixed 240px slide that left the strip mis-aligned with the content above it.
- The shared `Carousel` arrows are overlaid on the track edges and rendered only once the slides overflow, so a strip of three shows no controls. The gallery is the only consumer of this component.

**Stacked on #8166** (CRT-1218, two intro paragraphs). Once that merges this PR shrinks to its own commit.

**Not in scope:** the hub page's "rule of 3" is broken by the visible per-family H2s added for SEO; that needs a design call rather than a CSS fix.


markdurrant (approved):
LGTM 👍

PR #8168, opened 2026-09-14
RTI-3451: discard persist placeholders that outlive their swap

skewis6:
Port of ag-grid/ag-grid#15209 — these two scripts were byte-identical to the ag-grid copies and carry the same defect on the same Astro version (6.1.9).

A persist placeholder orphaned into the live document — its partner detached between `astro:before-swap` and the swap, e.g. the Enzuzo banner tearing itself down — leaves one `data-astro-transition-persist` id on two live elements once the partner returns. `swapBodyElement` cannot pair an id twice and throws `TypeError: Cannot read properties of null (reading 'moveBefore')`, aborting the swap after `swapRootAttributes` has already stripped `<html>`, so `astro:after-swap` never restores dark mode.

Fix: mark each placeholder and drop any still in the live document at the next `astro:before-swap` — one that is still around has missed its own swap.

The change is byte-identical to the ag-grid fix, which was verified against the live 36.2.0 build. No equivalent live run here — this environment cannot reach the deployed sites.


alantreadway (approved):
LGTM

PR #8171, opened 2026-09-14
CRT-1240 Give the block-caption test examples room for the website font

iMoses:
https://ag-grid.atlassian.net/browse/CRT-1240

Fix #CRT-1240


alantreadway (approved):
LGTM

PR #8172, opened 2026-09-14
CRT-1246 Report an empty licence key as missing

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-1246

Fix #CRT-1246


iMoses (approved):
LGTM

PR #8173, opened 2026-09-14
CRT-1228 Fix series label example control state in framework variants

iMoses:
https://ag-grid.atlassian.net/browse/CRT-1228

Fix #CRT-1228


alantreadway (approved):
LGTM

PR #8182, opened 2026-09-14
CRT-1231 Show fixed width state in the bars example controls

iMoses:
https://ag-grid.atlassian.net/browse/CRT-1231

Fix #CRT-1231


alantreadway (approved):
LGTM

PR #8183, opened 2026-09-14
CRT-1234 Hide crosshair label while dragging a labelled cross-line

alantreadway:
https://ag-grid.atlassian.net/browse/CRT-1234

Fix #CRT-1234


iMoses (approved):
LGTM - not sure about the changed E2E snapshot though

alantreadway:
> LGTM - not sure about the changed E2E snapshot though

Verified, that test includes dragging the annotation, and the snapshot is taken mid-drag, so the exact case this PR targets.

PR #8191, opened 2026-09-15
Merge b14.2.0 into latest

alantreadway:
Merge-back of `origin/b14.2.0` into `latest`.

Conflict resolution:
- `donutSeries.test.ts`: both sides added an import on the same line. Kept both: `DonutNodeTag` from CRT-1205 (b14.2.0) and `polarExamples` from AG-15902 (latest). Both are used in the file.

Version stamps auto-merged to `b14.2.0`'s `14.2.0-beta.20260914.1507`, which is newer than `latest`'s `14.2.0-beta.20260911`.


AG-DavidG (approved):
LGTM

PR #8195, opened 2026-09-15
AG-18361 Use pointer-capture when dragging

olegat:
https://ag-grid.atlassian.net/browse/AG-18361

* Remove separate mouse / touch drag implementations.
* Add new device-agnostic pointer drag implementation.
  * Use `setPointerCapture()`


olegat:
Blocked: The use of `HTMLElement.setPointerCapture` is preventing `click` events from firing on descendants. The `seriesWidget` hierarchy needs to be restructured to work around this.

PR #8197, opened 2026-09-15
Fix flaky quadrant item-styler pattern-fill snapshot test

iMoses:
Fixes a flaky image-snapshot test, no ticket. No runtime code changes.

`quadrantPreset.test.ts > for ITEM_STYLERS_NUMERIC it should render to canvas as expected` intermittently fails on CI with 3-4 differing pixels, all inside the pattern-filled marker.

**Why it flakes**

- The marker centre lands on fractional device coordinates, so the crisp 20x20 pattern tile is resampled at a fractional offset when painted. The blended edge pixels of the tile's stroke are far more sensitive to rasteriser rounding than an ordinary anti-aliased edge.
- Measured with pixelmatch's YIQ metric, those pixels score a delta of 88 against the default `threshold: 0.05` limit of 88.04. Any ±1 colour-level variance between runners (visible on every anti-aliased marker edge, but harmless there) tips them over.
- The scene JSON snapshot is unchanged, so this is not a layout or geometry change.

**Fix**

Use `PATTERN_SNAPSHOT_DEFAULTS` (per-pixel threshold 0.075) for that one case, the tolerance already used by the bubble, area and series-label pattern-fill snapshot suites since the skia-canvas upgrade. The per-case `imageSnapshotDefaults` is threaded through the render `compare()`; every other case keeps the default tolerance. No baselines were regenerated.

Verified locally: the case failed deterministically before the change (macOS output vs the Linux baseline) and passes after; the full file passes 69/69.


alantreadway (approved):
LGTM

PR #8199, opened 2026-09-15
CRT-1251 Correct the showcase page titles and descriptions

yinkelly:
https://ag-grid.atlassian.net/browse/CRT-1251

Fix #CRT-1251


alantreadway (approved):
LGTM

PR #8225, opened 2026-09-17
SE-169 Update the remaining v39 licence copy missed in #8221

taktran:
Follow-up to #8221 (ported in #8224), which replaced `ag-charts-enterprise/LICENSE.html` with the v40 Commercial EULA but missed `packages/ag-charts-server-side/LICENSE.html`, still the v39 Word export.

It is now the same generated v40 document (source: ag-grid/ag-grid#15190). The same fix has been added to #8224 for `b14.2.0`.


markdurrant (approved):
LGTM 👍

PR #8234, opened 2026-09-18
AG-18076 charts theme builder

skewis6:
Fix AG-18076


tools/createAndDeployDocsToS3.sh
```diff
@@ -8,7 +8,7 @@ fi
 
 TARGET_DIRECTORY=$1
 
-cd dist/packages/ag-charts-website
+cd dist/packages/
 
-aws s3 cp dist s3://testing.ag-grid.com/$TARGET_DIRECTORY --recursive
+aws s3 cp ag-charts-website s3://testing.ag-grid.com/$TARGET_DIRECTORY --recursive
```
taktran, line 13: Is this change intentional?

taktran (commented):
Mostly LGTM, just a small q.

Also in general, do we need `theme-builder-charts` to live in `ag-website-shared`? It doesn't seem like charts theme builder changes need to update the other sites

tools/createAndDeployDocsToS3.sh
skewis6: No idea, this is @seanlandsman when he fixed the branch builds i think? https://github.com/ag-grid/ag-charts/commit/cbc4420d1667e642b5dc557a51e3140a099685c6

taktran (approved):
LGTM

PR #8235, opened 2026-09-18
AG-18578 Re-pin test-case repros to staging when an agent PR merges

alantreadway:
Adds a `pull_request: closed` job to the AI Workflow stub that, when an agent PR merges, re-points the ticket's PR-pinned repro Plunkers from the per-PR gh-pages preview to the staging CDN. The job is gated on a merged `ghabot-*` head, writes only the test-case register's pin keys, and always exits 0.

https://ag-grid.atlassian.net/browse/AG-18578


alantreadway:
/pr-review

The checkout is now covered by job-level continue-on-error. The re-pin does not depend on the preview surviving cleanup: the action works from the pin record on each test case and only falls back to the preview manifest for pins that predate that record.

PR #8238, opened 2026-09-18
AG-17392 Migrate series base and line series onto plain options

iMoses:

Fix #AG-17392


alantreadway (approved):
LGTM

PR #8242, opened 2026-09-18
AG-17392 Migrate range-area and radar series onto plain options

iMoses:

Fix #AG-17392


alantreadway (approved):
LGTM

iMoses:
Two behaviour notes for reviewers.

**Radar item markers are now bound by datum id on animatable data.** `latest` used a positional `update(data)` because the marker holder was never marked clean, so the selection was recreated on every update. Once the dirty flag is honoured the nodes persist, and keyed binding (mirroring line series) keeps a marker attached to its category across data swaps. Removed categories are garbage-binned and destroyed on the next update: radar data changes go through the polar animation state machine, whose `resetAllAnimation` calls `itemSelection.cleanup()`. The "data swap (remove)" case in `radarLine.test.ts` covers it.

**Two range-area snapshots change (mixed dashed/undashed segmentation).** The series-level `lineDash: [0]` theme default means every segment now carries a dash value. Previously `SegmentedPath` reused one path and `setProperties` only assigned keys present on the segment, so an undashed segment following a dashed one inherited the dash. Line series never showed this because its theme already had the default. The regenerated baselines are the corrected rendering.

PR #8243, opened 2026-09-18
AG-17392 Migrate polar and gauge series onto plain options

iMoses:

Fix #AG-17392


alantreadway (approved):
LGTM

iMoses:
Review note on the dropped `scale.interval.minSpacing` / `maxSpacing` defaults (0 / 1000).

These were undocumented and, for the linear gauge, had no observable effect, so they were not carried over as theme values. Evidence:

- A fixed `interval.step` is accepted or rejected by `LinearScale.ticks` via `isDenseInterval`, which is pixel based; the tick-count bounds are not consulted on that path.
- On the auto path the ceiling is now `floor(range / (range / 6)) = 6` instead of the pixel extent. With `defaultTickCount = 5`, `tickStep` never prefers a candidate above 6 ticks: any multiplier giving `c > 6` has a neighbour giving `ceil(c / 2)`, which is closer to 5.
- Probed nine domains (0..6, 12, 30, 45, 60, 70, 120, 350, 600) at the stack head and at `origin/latest`; the rendered tick labels are identical in every case.

The radial gauge keeps both as undocumented theme options because its arc-length tick logic reads them directly.

PR #8259, opened 2026-09-21
Deny agent reads of local secret files

alantreadway:
Adds `Read` deny rules to `.claude/settings.json` for `.env.local` and the `credentials/` directory, so agentic tooling cannot read local secret files.


lsjroberts (approved):
LGTM

PR #8263, opened 2026-09-21
Make CI status-change detection robust to stale run listings

alantreadway:
The `report` job posts the Slack CI status alert only when the workflow result differs from the previous run on the branch. That previous result came from the first row of `gh run list`, but the workflow-runs API intermittently returns a stale, unordered page (observed returning runs from March to May 2026 first). On 2026-09-21 the first green run after a red streak (35577788731) therefore logged `status is still success` and no green alert was posted; on 2026-09-18 run 35387386449 logged a change from success to failure when the branch was already red, producing a duplicate red alert.

Changes to the `Check last job status` step:

- constrain the listing to the last 14 days and fetch JSON rather than parsing table output
- sort locally by creation time and exclude this run, unfinished runs and cancelled runs before taking the head
- when no usable previous run is found, treat it as a state change so a real transition is never silently swallowed


iMoses (approved):
LGTM

PR #8265, opened 2026-09-21
Tolerate pattern-fill rasteriser variance in highlight screenshots

lsjroberts:
Raises the per-pixel screenshot threshold for the two `highlight-e2e` pattern-fill examples, whose hatch tiles resample at a different sub-pixel phase between CI runners and so fail a comparison that allows no differing pixels.


alantreadway (approved):
LGTM

